PR #605, opened 2023-11-07
feat(formgen): default to sandbox identifier when branch or stack is not passed to formgen CLI

sdstolworthy:
*Issue #, if available:*

*Description of changes:*

When an app name or stack identifier is not passed to the formgen command, the stack identifier will now be used by default. This will allow users to run the command:
```
amplify generate forms
```
without being required to pass in the stack name.


sobolk:
This got closed as side-effect of renaming repo. Please reopen new PR.

PR #625, opened 2023-11-10
fix: add conditions to auth & unauth roles

bombguy:
*Issue #, if available:*

*Description of changes:*
The default authenticated roles that were being created were overly permissive and was not tied to the specific identity pool. This will leave the created app to be vulnerable as it can be exploited.

This change will apply condition to these roles so its tied down to specific identity pools created when auth construct is deployed.

### DIFF
--- a/packages/auth-construct/src/construct.ts
+++ b/packages/auth-construct/src/construct.ts
@@ -130,12 +130,12 @@ export class AmplifyAuth
       }
     );
 
-    // Auth / UnAuth Roles
-    const { auth, unAuth } = this.setupAuthAndUnAuthRoles();
-
     // Identity Pool
-    const { identityPool, identityPoolRoleAttachment } = this.setupIdentityPool(
-      { auth, unAuth },
+    const {
+      identityPool,
+      identityPoolRoleAttachment,
+      roles: { auth, unAuth },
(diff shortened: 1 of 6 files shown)


packages/auth-construct/src/construct.ts
```diff
@@ -169,13 +169,27 @@ export class AmplifyAuth
...
+        }),
       }),
       unAuth: new Role(this, 'unauthenticatedUserRole', {
-        assumedBy: new FederatedPrincipal('cognito-identity.amazonaws.com'),
+        assumedBy: new FederatedPrincipal('cognito-identity.amazonaws.com', {
+          StringEquals: {
+            'cognito-identity.amazonaws.com:aud': identityPoolId,
+          },
+          'ForAnyValue:StringLike': {
+            'cognito-identity.amazonaws.com:amr': 'authenticated',
```
awsluja, line 190: this one should be 'unauthenticated'

edwardfoyle (approved):

sobolk (approved):

bombguy:
> Can we get a little more context in the PR description? What's the problem and how does adding these conditions solve it?

for more context, here is similar issue from gen 1 CLI: https://github.com/aws-amplify/amplify-cli/pull/1974

PR #629, opened 2023-11-10
feat(auth): OTP via Email

Equartey:
*Description of changes:*
Adds OTP via Email 

### DIFF
--- a/packages/passwordless-auth-construct/package.json
+++ b/packages/passwordless-auth-construct/package.json
@@ -18,7 +18,8 @@
   },
   "dependencies": {
     "@aws-amplify/auth-construct-alpha": "^0.4.2",
-    "@aws-sdk/client-sns": "^3.445.0"
+    "@aws-sdk/client-sns": "^3.445.0",
+    "@aws-sdk/client-ses": "^3.445.0"
   },
   "devDependencies": {
     "@types/aws-lambda": "^8.10.119",
--- a/packages/passwordless-auth-construct/src/otp/construct.ts
+++ b/packages/passwordless-auth-construct/src/otp/construct.ts
@@ -1,6 +1,6 @@
 import { Construct } from 'constructs';
 import { CustomAuthTriggers, OtpAuthOptions } from '../types.js';
-import { PolicyStatement } from 'aws-cdk-lib/aws-iam';
+import { Effect, PolicyStatement } from 'aws-cdk-lib/aws-iam';
 
 /**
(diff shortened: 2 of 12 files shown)


packages/passwordless-auth-construct/src/services/ses_service.ts
```diff
@@ -0,0 +1,72 @@
...
+      },
+      Source: this.config.fromAddress,
+    });
+
+    // Send Email via SES
+    const output = await this.sesClient.send(emailCommand);
+    logger.debug(`SMS sent: ${JSON.stringify(output, null, 2)}`);
+  };
+
+  public mask = (destination: string): string => {
```
Jordan-Nelson, line 47: *note[non-blocking]:* I am noticing that the logic here is not quite the same as Cognito. I don't think that is necessarily an issue but ideally they would be the same. I am going to create a task to investigate and align them if it makes sense.

Jordan-Nelson (approved):

packages/passwordless-auth-construct/package.json
```diff
@@ -18,7 +18,8 @@
   },
   "dependencies": {
     "@aws-amplify/auth-construct-alpha": "^0.2.0-alpha.6",
-    "@aws-sdk/client-sns": "^3.429.0"
+    "@aws-sdk/client-sns": "^3.429.0",
+    "@aws-sdk/client-ses": "^3.445.0"
```
Jordan-Nelson, line 22: *Question:* Does`@aws-sdk/client-sns` need to be a higher version than `@aws-sdk/client-ses`? Should we align the two versions?

sobolk (approved):

packages/passwordless-auth-construct/package.json
Equartey: They can be aligned. Will update 👍

packages/passwordless-auth-construct/src/otp/construct.ts
```diff
@@ -30,6 +30,11 @@ export class AmplifyOtpAuth extends Construct {
         // see: https://docs.aws.amazon.com/sns/latest/dg/sns-using-identity-based-policies.html
         notResources: ['arn:aws:sns:*:*:*'],
       }),
+      // SES IAM policy
```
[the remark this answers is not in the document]
Jordan-Nelson: I am going to mark this as resolved. We will need to refactor this a bit in the magic link PR.

PR #638, opened 2023-11-13
fix: Expand auth types

awsluja:
*Issue #, if available:*

*Description of changes:*
Provides type expansion to improve UX.
Also adds docs for the samsara layer (some props were not inheriting docs, or have different docs due to BackendSecret).

UX:

'defineAuth':
image

'loginWith':
image

'externalProviders':
image

oauth/oidc/saml jsdocs:
image

'triggers':
image


packages/backend-auth/API.md
```diff
@@ -26,9 +27,9 @@ export type AmazonProviderFactoryProps = Omit<AmazonProviderProps, 'clientId' |
 };
 
 // @public (undocumented)
-export type AmplifyAuthFactoryProps = Omit<AuthProps, 'outputStorageStrategy' | 'loginWith'> & TriggerConfig & {
-    loginWith: AuthLoginWithFactoryProps;
-};
+export type AmplifyAuthFactoryProps = Expand<Omit<AuthProps, 'outputStorageStrategy' | 'loginWith'> & TriggerConfig & {
+    loginWith: Expand<AuthLoginWithFactoryProps>;
+}>;
```
sobolk, line 32: From https://github.com/aws-amplify/amplify-backend/pull/606#discussion_r1387325155


Instead of adding more and more type algebra - should we do what data factory did and just define types here and map them to construct types in implementation?

Types without type algebra like this are friendlier to users that don't rely on IDE tool tips to understand what's inside.

awsluja: I don't think data construct has the same issue; using Backend secret as a wrapper for client secrets (Auth has two interfaces, one for cdk users - without BackendSecret support, and one for front end devs - with BackendSecret)

awsluja: I guess we could 'flip' and do the algebra for CDK users

awsluja: trick with that is that construct shouldn't know about its wrapper, and it would if we decided to do algebra in construct

packages/plugin-types/API.md
```diff
@@ -118,6 +118,11 @@ export type ConstructFactoryGetInstanceProps = {
 // @public
 export type DeploymentType = 'branch' | 'sandbox';
 
+// @public
+export type Expand<T> = T extends infer O ? {
```
edwardfoyle, line 122: I'm not sure this belongs in `plugin-types`. Probably best to just keep this in `backend-auth` until we know we need it in several places

awsluja: didn't seem specific to auth so I was agreeing with this comment https://github.com/aws-amplify/amplify-backend/pull/606#discussion_r1387323873 from the old PR

packages/backend-auth/API.md
sobolk: > well i guess that we can avoid this by transforming before passing into construct

Yes, that's what I'm thinking, so that types are easy to understand for both populations of users.
I'd say that in this case DX is more important.

awsluja: maybe we should split this out to its own pr, this would impact Hoang's work with https://github.com/aws-amplify/amplify-backend/blob/main/packages/backend-auth/src/translate_auth_props.ts which translate from factory props to construct props and I haven't spent much time with it

awsluja: image
on the other hand, I can eliminate some of the type algebra ^ because we don't need a "TriggerConfig" type

sobolk: @edwardfoyle @Amplifiyer what do you think?

The https://github.com/aws-amplify/amplify-backend/blob/main/packages/backend-auth/src/translate_auth_props.ts is actually a step forward, it implements some part of the translation we're talking about, so it's not re-work but ~re-use. Btw. `translate_auth_props.ts` is how we'd name a file that has `translate(props: FactoryProps): ConstructProps`.

awsluja: fwiw the type algebra is quite nested; I don't know if non-IDE users should will find this easy either:
image

packages/plugin-types/API.md
edwardfoyle: I suppose I can disagree and commit on this one, but we need to be careful that `plugin-types` doesn't become a dumping ground for all sorts of utility types. Perhaps we need a `utility-types` package?

packages/backend-auth/API.md
edwardfoyle: I sympathize with Kamil's concern, but I also think there's a balance to be struct between type complexity and translation complexity. In this case, I think the `Expand` utility type is less complex than having two separate types and translation between them. The type algebra from the secrets work was significantly trickier hence breaking out into separate types with translation.

packages/plugin-types/API.md
sobolk: Creating separate set of types for factory and construct removes this problem.
Which btw makes me thing that if we keep current amount of type algebra for now then maybe it's better to move this type back to auth so it's easier to remove it when preview customers say that they can't understand our auth types.

packages/backend-auth/API.md
sobolk: There's one more thing here:
Sharing types between construct and factory means that construct break is inherently factory's break.

I think we may go for preview with this as is and log release blocker to revisit this. Perhaps users will tell us feedback by then. (if we do this, we should move `Expand` back to auth until we know more).

packages/plugin-types/API.md
awsluja: I'll move this back to auth

packages/backend-auth/API.md
awsluja: ticket created for this

sobolk: Can we get this in form of github issue ?

edwardfoyle (dismissed):

packages/backend-auth/API.md
awsluja: done https://github.com/aws-amplify/amplify-backend/issues/649

sobolk (approved):

hoangnbn (approved):

PR #642, opened 2023-11-14
feat: Add usage data metrics for sandbox

Amplifiyer:
*Description of changes:* Add usage data metrics for sandbox


packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -161,23 +173,54 @@
...
       // swallow execa error which is not really helpful, rather throw stderr
       throw new Error(aggregatedStderr);
     }
   };
+
+  private getDeploymentTimes = (cdkOutput: string) => {
+    const deploymentTimes: DeploymentTimes = {};
+    // the time can be in fractional or whole seconds. 24.3, 24, 24.22 etc.
+    const regexTotalTime = /✨ {2}Total time: (\d*\.*\d*)s.*/;
+    const totalTime = cdkOutput.match(regexTotalTime);
```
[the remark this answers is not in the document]
Amplifiyer: This input is not from customer and generated by the CLI and is not uncontrolled data.

packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -161,23 +173,54 @@
...
+    if (totalTime && totalTime.length > 1 && !isNaN(+totalTime[1])) {
+      deploymentTimes.totalTime = +totalTime[1];
+    }
+
+    const regexSynthTime = /✨ {2}Synthesis time: (\d*\.*\d*)s/;
+    const synthTime = cdkOutput.match(regexSynthTime);
```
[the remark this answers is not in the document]
Amplifiyer: This input is not from customer and generated by the CLI and is not uncontrolled data.

packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -161,23 +173,54 @@
...
+    if (synthTime && synthTime.length > 1 && !isNaN(+synthTime[1])) {
+      deploymentTimes.synthesisTime = +synthTime[1];
+    }
+
+    return deploymentTimes;
+  };
+
+  private deploymentHotswapped = (cdkOutput: string) => {
+    const regexTotalTime = /✨.*hotswapped!/;
+    return regexTotalTime.test(cdkOutput);
```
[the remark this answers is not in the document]
Amplifiyer: This input is not from customer and generated by the CLI and is not uncontrolled data.

packages/platform-core/src/telemetry/usage_data_emitter.ts
```diff
@@ -0,0 +1,108 @@
...
+      const req = https.request({
+        hostname: this.url.hostname,
+        port: this.url.port,
+        path: this.url.path,
+        method: 'POST',
+        headers: {
+          'content-type': 'application/json',
+          'content-length': payload.length,
+        },
+      });
```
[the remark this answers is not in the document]
Amplifiyer: This is known that we read version information from package.json file on local system.

packages/platform-core/API.md
```diff
@@ -29,13 +43,52 @@ export enum CDKContextKey {
...
     constructor(stackTraceLine: string);
     // (undocumented)
     extract: () => string | undefined;
 }
 
+// @public
+export class LibraryVersionFetcher {
+    // (undocumented)
+    fetch: (absolutePackageJsonPath: string) => string;
+}
```
sobolk, line 69: I'd create `PackageJsonReader` instead. If we have to read whole thing to get version we can as well serve all properties we care about and caller will select what they want.

`CwdPackageJsonLoader` could then use `PackageJsonReader` internally (assuming it does some trick with process.cwd`.

packages/platform-core/API.md
```diff
@@ -29,13 +43,52 @@ export enum CDKContextKey {
...
+    name: string;
+}>;
+
+// @public
+export class UsageDataEmitter {
+    constructor(libraryVersion: string, sessionUuid?: string, url?: UrlWithStringQuery, accountIdFetcher?: AccountIdFetcher);
+    // (undocumented)
+    emitFailure: (command: string, error: Error) => Promise<void>;
+    // (undocumented)
+    emitSuccess: (command: string, deploymentTimes?: DeploymentTimes, hotswapped?: boolean) => Promise<void>;
```
sobolk, line 89: 1. if this has `command` in param names then should this type go to `cli-core` instead?
Will we ever emit metrics from constructs or constructs factories?
2. I think that `emitSuccess` should accept `metrics: Record<string, number>` and `dimensions: Record<string, string>` instead of coupling this interface to deployer types.

packages/platform-core/src/telemetry/account_id_fetcher.ts
```diff
@@ -0,0 +1,22 @@
...
+   * constructor for AccountIdFetcher
+   */
+  constructor(private readonly stsClient = new STSClient()) {}
+  fetch = async () => {
+    const stsResponse = await this.stsClient.send(
+      new GetCallerIdentityCommand({})
+    );
+    if (stsResponse && stsResponse.Account) {
+      return stsResponse.Account;
+    }
```
sobolk, line 17: If this type is to remain public then I'd rather just use stsClient directly in usage data. It doesn't seem to absorb a lot of complexity.
If this becomes internal it's not a problem.

packages/platform-core/src/telemetry/get_installation_id.ts
```diff
@@ -0,0 +1,20 @@
...
+export const getInstallationUuid = async (
+  namespace: string = AMPLIFY_CLI_UUID_NAMESPACE
+) => {
+  const packageJsonName = (
+    await new CwdPackageJsonLoader().loadCwdPackageJson()
+  ).name;
+  const userName = userInfo().username;
+  const cwdPath = process.cwd();
+
+  return uuidV5(cwdPath + packageJsonName + userName, namespace);
```
sobolk, line 19: should this mix some "machine identifier" like mac address ?

packages/platform-core/src/telemetry/get_usage_data_url.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  return parsedUrl;
+};
+
+const getParsedUrl = (): UrlWithStringQuery => {
+  if (!useBetaUrl()) {
+    return url.parse(prodUrl);
+  }
+
+  return url.parse(process.env.AMPLIFY_CLI_BETA_USAGE_TRACKING_URL || '');
```
sobolk, line 24:
```suggestion
  return url.parse(process.env.AMPLIFY_BACKEND_USAGE_TRACKING_ENDPOINT || '');
```

packages/platform-core/src/telemetry/serializable_error.ts
```diff
@@ -0,0 +1,78 @@
+import path from 'path';
+
+/**
+ * Wrapper around Error for easy serialization for usage metrics
```
sobolk, line 4:
```suggestion
 * Wrapper around Error for serialization for usage metrics
```

packages/platform-core/src/telemetry/usage_data_emitter.ts
```diff
@@ -0,0 +1,110 @@
...
+      req.on('error', () => {
+        /* noop */
```
sobolk, line 97: If this goes to cli-core we should `Printer.verbose(error)`.

packages/platform-core/src/telemetry/version_manager.ts
sobolk, line 1: these constants seem to have 1 usage each, we should move them to relevant files.
Alternatively we should name this file `telemetry/constants.ts`

Amplifiyer: These two are coupled with each other. Renaming the file.

packages/platform-core/src/telemetry/get_installation_id.ts
Amplifiyer: the network interfaces could be named differently and some have just all 0s in it. Would this be a better idea https://www.npmjs.com/package/node-machine-id?

packages/platform-core/API.md
Amplifiyer: Refactored `PackageJsonReader`

packages/platform-core/API.md
Amplifiyer: 1. Should we move to cli-core? that's where I had it originally but thought it might grow coupled with CLI and there is no reason for other components to emit metrics as well in the future. What do you think?
2. Good idea.

packages/platform-core/src/telemetry/account_id_fetcher.ts
Amplifiyer: this is not exported outside of platform-core. Is it considered public? Right now it helps with making testing easier.

Amplifiyer: Added the factory to hide this

packages/platform-core/API.md
```diff
@@ -4,7 +4,21 @@
...
+import { STSClient } from '@aws-sdk/client-sts';
+import { UrlWithStringQuery } from 'url';
+import { z } from 'zod';
+
+// @public
+export class AccountIdFetcher {
+    constructor(stsClient?: STSClient);
+    // (undocumented)
+    fetch: () => Promise<string>;
+}
```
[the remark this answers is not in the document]
Amplifiyer: Added factory

packages/platform-core/API.md
sobolk: Re 1.
If we want to keep it in platform-core we should somehow drop `command` wording. 

I'm fine with moving this to `cli-core` if that's where it belongs now. Even if it somehow gets coupled to CLI then recovery from that would be to extract non-coupled part first and have coupled part + telemetry core.

Amplifiyer: I'm removing the command wording regardless. For now let's keep it in platform-core! Moving to cli-core is much easy than the other way around.

packages/platform-core/src/telemetry/get_installation_id.ts
sobolk: The network interface was an example. That library looks promising if it's maintained. So feel free to find best solution for this. We need some machine id.

packages/backend-deployer/API.md
```diff
@@ -5,12 +5,13 @@
 ```ts
 
 import { BackendIdentifier } from '@aws-amplify/plugin-types';
+import { DeploymentTimes } from '@aws-amplify/platform-core';
```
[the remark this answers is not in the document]
Amplifiyer: Moved `DeploymentTimes` in the backend-deployer and made `UsageDataMetrics` API more generic using `metrics` and `dimensions`

packages/platform-core/src/telemetry/usage_data_emitter.ts
Amplifiyer: For now staying in platform-core.

packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -161,23 +173,54 @@ export class CDKDeployer implements BackendDeployer {
...
       stderr: 'pipe',
     });
-    childProcess.stderr?.pipe(aggregatorStream);
-
+    childProcess.stderr?.pipe(aggregatorStderrStream);
+    childProcess.stdout?.pipe(aggregatorStdoutStream);
+    childProcess.stdout?.pipe(process.stdout);
     try {
       await childProcess;
+      return aggregatedStdout;
```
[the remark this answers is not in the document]
Amplifiyer: I experimented something here https://github.com/aws-amplify/amplify-backend/pull/642/commits/96414ec49f8e7cfc0e51c8f91ba72d78c086b965 looks a bit clunky, wdyt?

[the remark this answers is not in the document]
sobolk: Why not register `stdout.on('data',...)` instead of piping?
Regardless of that there's one problem to solve, i.e. `data` or `chunk` might not be coming in a form of lines (could be partial line or many lines).

Can you try this approach to listen to stdout?
https://github.com/aws-amplify/amplify-backend/blob/79ac139975278c462be2a8196a87d47ed4e46f77/packages/integration-tests/src/process-controller/process_controller.ts#L66-L69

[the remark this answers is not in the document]
Amplifiyer: Tried here https://github.com/aws-amplify/amplify-backend/pull/642/commits/ae1a81fdc2eb1f02d2c8a18d49d0ca5f23f2bd79

`on` listener does send data line by line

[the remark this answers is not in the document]
sobolk: > Tried here https://github.com/aws-amplify/amplify-backend/commit/ae1a81fdc2eb1f02d2c8a18d49d0ca5f23f2bd79

Looks good I think. We might want `new CDKOutputListener.listen(childProcess.stdout)`.

> on listener does send data line by line

is this documented somewhere or do we get this behavior because child process happens to emit line by line that fits pipe buffer?

packages/platform-core/src/telemetry/get_installation_id.ts
Amplifiyer: This doesn't look well supported or a good alternative that exists without requiring a local state.

packages/backend-deployer/src/cdk_deployer.ts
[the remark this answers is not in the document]
Amplifiyer: I think it's happening because the data is coming line by line, so there is a chance that a big chunk gets broken down by node. I'll look into using `readline`

packages/platform-core/src/telemetry/usage_data_emitter.ts
```diff
@@ -0,0 +1,131 @@
...
+import { getInstallationUuid } from './get_installation_id.js';
+import { latestPayloadVersion } from './constants.js';
+import { getUrl } from './get_usage_data_url.js';
+import isCI from 'is-ci';
+import { SerializableError } from './serializable_error.js';
+
+/**
+ * Entry point for sending usage data metrics
+ */
+export class DefaultUsageDataEmitter {
```
edwardfoyle, line 15: this should implement the type defined in the factory file

packages/sandbox/package.json
```diff
@@ -28,14 +28,15 @@
...
-    "open": "^9.1.0"
+    "glob": "^10.2.7",
+    "open": "^9.1.0",
+    "parse-gitignore": "^2.0.0"
   },
   "devDependencies": {
     "@types/debounce-promise": "^3.1.6",
-    "@types/parse-gitignore": "^1.0.0"
+    "@types/parse-gitignore": "^1.0.0",
+    "@types/uuid": "^9.0.7"
```
edwardfoyle, line 39: we can use crypto.randomUUID to avoid pulling in this dep

packages/platform-core/src/telemetry/get_installation_id.ts
edwardfoyle: We shouldn't include cwd in this seed. You can run `npx amplify` from anywhere in your directory. It should be the path of the root package.json or something like that

packages/platform-core/src/telemetry/usage_data_emitter.ts
```diff
@@ -0,0 +1,131 @@
...
+   */
+  constructor(
+    private readonly libraryVersion: string,
+    private readonly sessionUuid = uuid(),
+    private readonly url = getUrl(),
+    private readonly accountIdFetcher = new AccountIdFetcher()
+  ) {}
+
+  emitSuccess = async (
+    metrics?: Record<string, number>,
```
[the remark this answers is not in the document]
sobolk: See https://github.com/aws-amplify/amplify-backend/pull/642#discussion_r1393493174 .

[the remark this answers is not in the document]
sobolk: and https://github.com/aws-amplify/amplify-backend/pull/642#discussion_r1393501107 .
above.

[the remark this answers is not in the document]
sobolk: The reasoning behind is that "metrics collector" should expose api to collect metrics (i.e. abstract mechanism of how data is pushed to the backend) but it should be up to caller to know what exactly is being reported.

packages/backend-deployer/src/cdk_deployer.ts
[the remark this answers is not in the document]
Amplifiyer: Updated here https://github.com/aws-amplify/amplify-backend/pull/642/commits/66a7ca8e24f4076499ea54465d332727c4b3e798

packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -155,29 +164,58 @@ export class CDKDeployer implements BackendDeployer {
...
+      stdout: 'pipe',
       stderr: 'pipe',
     });
-    childProcess.stderr?.pipe(aggregatorStream);
-
+    childProcess.stderr?.pipe(aggregatorStderrStream);
+    if (stdoutListener) {
+      childProcess.stdout?.on('data', stdoutListener);
+    }
+    childProcess.stdout?.pipe(process.stdout);
```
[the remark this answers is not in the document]
Amplifiyer: Removed `on` listener, now both are using pipes.

packages/platform-core/src/telemetry/usage_data_emitter.ts
```diff
@@ -0,0 +1,131 @@
...
+    dimensions?: Record<string, string>
```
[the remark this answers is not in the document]
Amplifiyer: Same as what Kamil said above

packages/platform-core/src/telemetry/get_usage_data_url.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  return url.parse(process.env.AMPLIFY_BACKEND_USAGE_TRACKING_ENDPOINT || '');
+};
+
+const useBetaUrl = (): boolean =>
+  !!(
+    process.env.AMPLIFY_BACKEND_USAGE_TRACKING_ENDPOINT &&
+    typeof process.env.AMPLIFY_BACKEND_USAGE_TRACKING_ENDPOINT === 'string'
+  );
```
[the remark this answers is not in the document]
Amplifiyer: yup, copied over from classic CLI and cut few things. It can be simplified as you suggested.

packages/platform-core/src/telemetry/get_usage_data_url.test.ts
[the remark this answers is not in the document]
Amplifiyer: yup, changing it

packages/platform-core/src/telemetry/usage_data_emitter.ts
Amplifiyer: updated

packages/platform-core/src/telemetry/get_usage_data_url.ts
```diff
@@ -0,0 +1,31 @@
...
+
+let parsedUrl: UrlWithStringQuery;
+
+const prodUrl = `https://api.cli.amplify.aws/${latestApiVersion}/metrics`;
+
+/**
+ *  Usage data tracking service URL
+ */
+export const getUrl = (): UrlWithStringQuery => {
+  if (!parsedUrl) {
```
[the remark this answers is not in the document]
Amplifiyer: updated

packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -185,37 +175,41 @@ export class CDKDeployer implements BackendDeployer {
...
+    output: DeployResult | DestroyResult,
+    stdout: stream.Readable
   ) => {
     const regexTotalTime = /✨ {2}Total time: (\d*\.*\d*)s.*/;
     const regexSynthTime = /✨ {2}Synthesis time: (\d*\.*\d*)s/;
-    const listener = (chunk: string) => {
-      const data = Buffer.from(chunk).toString();
-      if (data.includes('✨')) {
+    const reader = readline.createInterface(stdout);
+    for await (const line of reader) {
```
sobolk, line 201: very nice.

packages/platform-core/src/telemetry/usage_data_emitter.ts
```diff
@@ -0,0 +1,131 @@
...
+    const data = await this.getUsageData({
+      state: 'FAILED',
+      error,
+      dimensions,
+    });
+    await this.send(data);
+  };
+
+  private getUsageData = async (options: {
+    state: string;
```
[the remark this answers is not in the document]
Amplifiyer: updated.

packages/sandbox/package.json
Amplifiyer: this is actually needed in platform-core (moved there) and since it's core, we should use the browser compatible libraries and I'm not sure if node's crypto is.

packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -161,23 +164,52 @@ export class CDKDeployer implements BackendDeployer {
     // actionable errors being hidden among the stdout. Moreover execa errors are
     // useless when calling CLIs unless you made execa calling error.
     let aggregatedStderr = '';
-    const aggregatorStream = new stream.Writable();
-    aggregatorStream._write = function (chunk, encoding, done) {
+    const aggregatorStderrStream = new stream.Writable();
+    aggregatorStderrStream._write = function (chunk, encoding, done) {
```
sobolk, line 168: I didn't realize that aggregating output is existing thing.
Mind logging GH issue to make this part "streaming and parsing errors" ?

packages/cli/src/backend-identifier/local_namespace_resolver.ts
```diff
@@ -18,6 +18,8 @@ export class LocalNamespaceResolver implements NamespaceResolver {
    * Returns the value of package.json#name from the current working directory
    */
   resolve = async () => {
-    return (await this.packageJsonLoader.loadCwdPackageJson()).name;
+    const name = this.packageJsonLoader.read().name;
```
sobolk, line 21: if we call it `Loader` then `load` is fine. We can also name it reader.

packages/platform-core/API.md
```diff
@@ -29,13 +30,57 @@ export enum CDKContextKey {
     DEPLOYMENT_TYPE = "amplify-backend-type"
 }
 
+// @public
+export class CwdPackageJsonLoader {
+    read(): {
+        name?: string | undefined;
+        version?: string | undefined;
+        type?: "module" | "commonjs" | undefined;
+    };
```
sobolk, line 39: Should this use `type PackageJson` just defined below?
Or is it some api extractor behavior with zod?

packages/platform-core/src/cwd_package_json_loader.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Throws if no package.json is present
+ */
+export class CwdPackageJsonLoader {
+  /**
+   * Returns the contents of the package.json file in process.cwd()
+   *
+   * If no package.json file exists, or the content does not pass validation, an error is thrown
+   */
+  read() {
```
sobolk, line 15:
```suggestion
  read(): PackageJson {
```

this might help

packages/platform-core/src/telemetry/get_installation_id.ts
Amplifiyer: Good point, I don't think it's possible with `npx` to figure out the "root" app since customer can have many nested workspaces. The only other reliable option I can think of is persisting a UUID along with consent information.

sobolk (commented):
LGTM.

packages/platform-core/src/telemetry/get_installation_id.ts
sobolk: How about rendering a file next to `package.json` in `node_modules` where this package gets installed ?

packages/backend-deployer/src/cdk_deployer.ts
Amplifiyer: Will log it, this would be a little harder since we are actually throwing and processing elsewhere.

packages/platform-core/src/telemetry/get_installation_id.ts
sobolk: One more idea.
We could hash absolute path derived from `import.meta.url` , that should be stable.

packages/platform-core/src/telemetry/usage_data_emitter.ts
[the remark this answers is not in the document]
edwardfoyle: That's true, but we also have logic in here where we are looking for specific keys to perform certain transformations. Seems like we could get the type system to help us.

Regardless, I'm not going to block on it

packages/platform-core/src/telemetry/get_installation_id.ts
Amplifiyer: `import.meta.url` is a good idea, but platform-core is built into `commonJs` and doesn't like `import.meta` :(

edwardfoyle: `npx which amplify` would probably get what we want

packages/cli/src/backend-identifier/local_namespace_resolver.ts
Amplifiyer: updated

packages/platform-core/API.md
Amplifiyer: updated

packages/platform-core/src/cwd_package_json_loader.ts
Amplifiyer: updated

packages/platform-core/src/telemetry/usage_data_emitter.ts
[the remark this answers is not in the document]
sobolk: That's a good point. But I'd do that by casting these Records to something typed internally in data collector as we do this only because we want to keep old schema.

packages/platform-core/src/telemetry/get_installation_id.ts
sobolk: But that's ipc call.
how about `__dirname` ?

Amplifiyer: Isn't `__dirname` same as `process.cwd`?

sobolk: No. it's dir of current file afaik https://nodejs.org/api/modules.html#__dirname

packages/platform-core/src/cwd_package_json_loader.ts
sobolk: I might have forgotten that this needs `Promise<PackageJson>` if this is async.

packages/platform-core/src/telemetry/get_installation_id.ts
Amplifiyer: I'll give it a shot

edwardfoyle: No __dirname could work. That's the path of the actual source code file that is executing which should be stable for a given installation

packages/platform-core/src/cwd_package_json_loader.ts
sobolk: nvm I can see the comment in code.

packages/backend-deployer/src/cdk_deployer.ts
Amplifiyer: https://github.com/aws-amplify/amplify-backend/issues/662

packages/platform-core/src/telemetry/get_installation_id.ts
edwardfoyle: Now I'm wondering if we should _only_ seed with `__dirname`? The packageJsonName and userName could both change without actually being a different installation. `__dirname` should be stable for a given installation in all normal scenarios I can think of

sobolk: `__dirname` + some "machine identifier". (if we punt machine id we should create gh issue)

Amplifiyer: userName changing is unlikely though it's usually is included in the path (for unix at least), however for cases where users are installing an "example" app in a /tmp or a root directory, we will get same installation ids across different users.

There is no reliable way to get machine identifier as per my quick research. We will have to punt that.

Amplifiyer: Updated to use `__dirname` for now

packages/cli/src/backend-identifier/local_namespace_resolver.ts
```diff
@@ -12,12 +12,14 @@ export class LocalNamespaceResolver implements NamespaceResolver {
    * packageJsonLoader is assigned to an instance member for testing.
    * resolve is bound to this so that it can be passed as a function reference
    */
-  constructor(private readonly packageJsonLoader: CwdPackageJsonLoader) {}
+  constructor(private readonly packageJsonLoader: CwdPackageJsonReader) {}
```
sobolk, line 15:
```suggestion
  constructor(private readonly packageJsonReader: CwdPackageJsonReader) {}
```

sobolk (approved):

packages/platform-core/src/usage-data/get_installation_id.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Generates a consistent installation Uuid from the cwd + packageJsonName + userName
+ */
+export const getInstallationUuid = (
+  namespace: string = AMPLIFY_CLI_UUID_NAMESPACE
+) => {
+  const packageJsonName = new CwdPackageJsonReader().read().name;
+  const userName = userInfo().username;
+  const modulePath = __dirname;
+
+  return uuidV5(modulePath + packageJsonName + userName, namespace);
```
edwardfoyle, line 17: My vote would still be for just `modulePath + userName`. While packageJsonName should be pretty stable, it is subject to change (we have changed our package.json names on occasion)

packages/platform-core/src/usage-data/get_installation_id.ts
```diff
@@ -0,0 +1,18 @@
+import { userInfo } from 'os';
+import { v5 as uuidV5 } from 'uuid';
+import { CwdPackageJsonReader } from '../cwd_package_json_reader.js';
+
+// eslint-disable-next-line spellcheck/spell-checker
+const AMPLIFY_CLI_UUID_NAMESPACE = 'e7368840-2eb6-4042-99b4-9d6c2a9370e6'; // A random v4 UUID
+/**
+ * Generates a consistent installation Uuid from the cwd + packageJsonName + userName
```
edwardfoyle, line 8: nit: update comment

packages/cli/src/backend-identifier/local_namespace_resolver.ts
Amplifiyer: updated

packages/platform-core/src/cwd_package_json_reader.ts
```diff
@@ -0,0 +1,20 @@
+import * as path from 'path';
+import { PackageJson, PackageJsonReader } from '.';
+
+/**
+ * Loads the contents of package.json from process.cwd().
+ *
+ * Throws if no package.json is present
+ */
+export class CwdPackageJsonReader {
```
edwardfoyle, line 9: I wonder if this should just be another method on PackageJsonReader. ie `.readFromCwd` or such?

edwardfoyle (commented):
A few comments but I don't think they are blockers

packages/platform-core/src/telemetry/get_installation_id.ts
Amplifiyer: Long term - https://github.com/aws-amplify/amplify-backend/issues/665

packages/platform-core/src/usage-data/get_installation_id.ts
Amplifiyer: updated.

packages/platform-core/src/usage-data/get_installation_id.ts
Amplifiyer: I think in that case it's acceptable for installation id to change. I think it's worse to have two customers sharing the installation id vs one customer to have many. Nevertheless I'm not strong on that so changed it to remove package.json name

sobolk: Maybe for machine id we could use `os.hostname()` for now ? https://nodejs.org/api/os.html#oshostname

sobolk (approved):

edwardfoyle (approved):

PR #650, opened 2023-11-14
feat(formgen): change output directory of model documents in formgen

sdstolworthy:
*Issue #, if available:*

*Description of changes:*

The option to customize the output directory of graphql documents used by forms is removed. Graphql documents consumed by generated forms are always generated in a nested graphql directory inside the outDir of the forms.

Additionally, --ui-out-dir has been renamed to --out-dir to maintain consistency with the other commands.


packages/form-generator/src/local_codegen_graphql_form_generator.ts
```diff
@@ -98,20 +98,23 @@ export class LocalGraphqlFormGenerator implements GraphqlFormGenerator {
       renderTypeDeclarations: true,
       apiConfiguration: {
         dataApi: 'GraphQL',
-        fragmentsFilePath: path.join(
+        fragmentsFilePath: `./${path.join(
```
alharris-at, line 101: Does this work on windows?

sdstolworthy: This string is generated straight into the UI Form. If it is left out, the string is sent as `graphql/fragments`, which node attempts to resolve as a node module. I can get a windows vm set up to verify this.

sobolk: Should we get some integration(better)/e2e(worse) test for this where we attempt to use generated code?

sdstolworthy: This is definitely a missing test vector. I'll get a test written for this.

packages/form-generator/src/local_codegen_graphql_form_generator.test.ts
```diff
@@ -251,4 +254,36 @@ void describe('LocalCodegenGraphqlFormGenerator', () => {
     const forms = await l.generateForms();
     assert('writeToDirectory' in forms);
   });
+  const graphqlDirectories = [
```
sdstolworthy, line 257: @alharris-at @sobolk I added these tests which exercise the path generation logic. This is integration-level; it is calling the real codegen package and making assertions against the output.

packages/form-generator/src/local_codegen_graphql_form_generator.ts
sdstolworthy: https://github.com/aws-amplify/amplify-backend/pull/650/files#r1393530504

packages/form-generator/src/local_codegen_graphql_form_generator.test.ts
sdstolworthy: image
These tests are passing on windows

sobolk (approved):

alharris-at (approved):

PR #659, opened 2023-11-15
test: restrict live canary run time

sobolk:
*Issue #, if available:*

*Description of changes:*

Makes live canaries timeout in 15 minutes.

To mitigate this behavior:
image

### DIFF
--- a/.github/workflows/canary_checks.yml
+++ b/.github/workflows/canary_checks.yml
@@ -30,6 +30,7 @@ jobs:
           npm run test
   live_dependency_health_checks:
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     permissions:
       # these permissions are required for the configure-aws-credentials action to get a JWT from GitHub
       id-token: write


Amplifiyer (approved):

.github/workflows/canary_checks.yml
```diff
@@ -30,6 +30,7 @@ jobs:
           npm run test
   live_dependency_health_checks:
     runs-on: ubuntu-latest
+    timeout-minutes: 15
```
sobolk, line 33: this should be plenty enough but not too agressive
image

alharris-at (approved):

srquinn21 (approved):
Hopefully we never need to bump this... but we should probably run book it anyway

sobolk:
> Hopefully we never need to bump this... but we should probably run book it anyway

We should never bump it. 
If we get there parallelization is the answer ,
https://github.com/aws-amplify/amplify-backend/pull/659#pullrequestreview-1732951208

I'll draft some runbook, seems we have enough content to start it now.

PR #664, opened 2023-11-16
fix: `--version` for amplify and disable subcommands version

0618:
*Issue #, if available:*

*Description of changes:*

**Issue**: `amplify --version` and `amplify <subcommand> --version` don't return correct versions. Always return `1.0.0`.
**Reason**: It's because `yargs` always gets the package.json version from the root dir, which means it gets custemer's project version.
**Changes in this PR**:
- Disable subcommand's version
- Use absolute path to get the version from `backend-cli` package
- Add test


packages/cli/src/main_parser_factory.ts
```diff
@@ -8,7 +10,12 @@ import { createConfigureCommand } from './commands/configure/configure_command_f
  * Creates main parser.
  */
 export const createMainParser = (): Argv => {
+  const absolutePath = './node_modules/@aws-amplify/backend-cli';
```
sobolk, line 13: we should use `import.meta.url` to determine this.
But I suggest to wait until @Amplifiyer 's PR goes in that adds components for that to platform core.

.changeset/wet-items-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+'@aws-amplify/backend-cli': patch
+---
+
+`amplify sandbox --version` returns version of the sandbox package
```
edwardfoyle, line 5: Seems like this message needs to be updated

packages/cli/src/main_parser_factory.test.ts
```diff
@@ -16,6 +17,11 @@ void describe('main parser', { concurrency: false }, () => {
     assert.match(output, /generate\s+Generates post deployment artifacts/);
   });
 
+  void it('shows version', async () => {
+    const output = await commandRunner.runCommand('--version');
+    assert.match(output, RegExp(version));
```
edwardfoyle, line 22: Can this just be `assert.equal(output, version)`?

packages/cli/src/commands/configure/configure_command.ts
```diff
@@ -36,6 +36,7 @@ export class ConfigureCommand implements CommandModule<object> {
    */
   builder = (yargs: Argv): Argv => {
     return yargs
+      .version(false)
```
edwardfoyle, line 39: Is there any way we can set this globally on the yargs parser?

.changeset/wet-items-raise.md
0618: Good catch!

packages/cli/src/main_parser_factory.ts
Amplifiyer: You can now use 

https://github.com/aws-amplify/amplify-backend/blob/70685f36bdd089711fab3ba6eab0c375dad35827/packages/platform-core/src/package_json_reader.ts#L8

See example 

https://github.com/aws-amplify/amplify-backend/blob/70685f36bdd089711fab3ba6eab0c375dad35827/packages/cli/src/commands/sandbox/sandbox_command_factory.ts#L57-L59

packages/cli/src/commands/configure/configure_command.ts
Amplifiyer: We decided it's ok to let the `--version` works for every command right? so just setting on the main_parser_factory is sufficient?

0618: We mentioned about having them returning the same version or disable the versions for sub-commands, but didn't have a hard conclusion.
IMO, having `amplify sandbox --version` return the same version as `amplify --version` might be confusing. It makes more sense to just disable `--version` for the sub-commands.

sobolk: I think `--version` doesn't make much sense on subcommands, but I don't think it matters much, so I'm fine either way, with preference of disabling this on subcommands.

packages/cli/src/main_parser_factory.test.ts
0618: Unfortunately, no. It'd have the following error

image

0618: Done 33af02788

.changeset/wet-items-raise.md
0618: Done 24038c422

packages/cli/src/commands/configure/configure_command.ts
edwardfoyle: My preference is also disabling on subcommands. It's just noise in the help printouts

packages/cli/src/main_parser_factory.ts
0618: Done! 4c50d21d4

sobolk (dismissed):

packages/cli/src/commands/configure/configure_command.ts
edwardfoyle: I'm assuming the verdict on this is that it has to be set on each subcommand?

edwardfoyle (approved):

packages/cli/src/commands/configure/configure_command.ts
Amplifiyer: If we want to hide it from subcommands then I believe yes.

0618: Yea, the `--version` option is hidden from `--help` now

0618: > Is there any way we can set this globally on the yargs parser?

Unfortunately, I was not able to find a way to set it globally from yargs. If we want, I guess we can customize a `amplifyYargs`. What do you think? @edwardfoyle

edwardfoyle: Eh, probably not worth it at this point

sobolk (approved):

Amplifiyer (approved):

PR #668, opened 2023-11-16
fix: dependabot alert for tough cookie

0618:
*Issue #, if available:*

*Description of changes:*

Fix https://github.com/aws-amplify/amplify-backend/security/dependabot/13


package.json
```diff
@@ -98,5 +98,8 @@
     "*.json": "prettier --write",
     "*.yml": "prettier --write",
     "*.md": "prettier --write"
+  },
+  "resolutions": {
+    "tough-cookie": "4.1.3"
```
sobolk, line 103: if we do this we should create GH issue to clean this up when verdaccio ships patch.
Or, maybe better, since this is medium severity and only used in our build time we could ignore this for now.
Can you check if this problem affects us in any meaningful way ?

0618: Right, we only use it in verdaccio, which doesn't affact anything other than build/dev. 

```
└─┬ verdaccio@5.27.0
  └─┬ request@2.88.2
    └── tough-cookie@2.5.0
```

I think it makes sense to dismiss the alert.

PR #677, opened 2023-11-16
fix: Force colors to not be stripped off when piping child process stdout

Amplifiyer:
*Description of changes:* Piping the child process output by default strips off the color for consuming streams. This is a workaround to preserve the color being piped to parent process. See https://github.com/Marak/colors.js/issues/127#issuecomment-392249836

### DIFF
--- a/packages/backend-deployer/src/cdk_deployer.ts
+++ b/packages/backend-deployer/src/cdk_deployer.ts
@@ -170,14 +170,15 @@ export class CDKDeployer implements BackendDeployer {
       done();
     };
 
-    // Piping the output by default strips off the color. This is a workaround to
-    // preserve the color being piped to parent process. Later we reset it to whatever it was before.
-    const oldForceColorValue = process.env['FORCE_COLOR'];
-    process.env['FORCE_COLOR'] = '1';
     const childProcess = execa(command, cdkCommandArgs, {
       stdin: 'inherit',
       stdout: 'pipe',
       stderr: 'pipe',
+
+      // Piping the output by default strips off the color. This is a workaround to
+      // preserve the color being piped to parent process.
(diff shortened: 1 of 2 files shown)


packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -169,6 +169,11 @@ export class CDKDeployer implements BackendDeployer {
       aggregatedStderr += chunk;
       done();
     };
+
+    // Piping the output by default strips off the color. This is a workaround to
+    // preserve the color being piped to parent process. Later we reset it to whatever it was before.
+    const oldForceColorValue = process.env['FORCE_COLOR'];
+    process.env['FORCE_COLOR'] = '1';
```
sobolk, line 176: Can we clone `process.env`, append this to a copy and pass to child process ?

packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -184,6 +189,7 @@ export class CDKDeployer implements BackendDeployer {
 
     try {
       await childProcess;
+      process.env['FORCE_COLOR'] = oldForceColorValue;
```
sobolk, line 192: this should be in finally block.

packages/backend-deployer/src/cdk_deployer.ts
Amplifiyer: Good idea, only setting in the child process now.

packages/backend-deployer/src/cdk_deployer.ts
Amplifiyer: no longer needed.

edwardfoyle (approved):

sobolk (approved):

PR #682, opened 2023-11-17
test: add e2e testing for config generation

### LINKED ISSUE #403: e2e test for config generation
We need e2e coverage for `amplify generate config`

See https://github.com/aws-amplify/samsara-cli/pull/402#issuecomment-1758283888

rtpascual:
*Issue #, if available:*

Resolves #403

*Description of changes:*

Add e2e testing for config generation.


packages/integration-tests/src/test-e2e/deployment.test.ts
```diff
@@ -124,6 +126,37 @@ void describe(
...
+                  configOutDir,
+                  '--config-format',
+                  format,
+                ],
+                testProject.projectDirPath
+              )
+                .do(waitForSandboxDeploymentToPrintTotalTime())
+                .do(interruptSandbox())
+                .do(rejectCleanupSandbox())
+                .run();
```
sobolk, line 155: the ask in the issue was to add tests fro `amplify generate config` not sandbox options.
this should likely become part of `pipeline-deploy` test somewhere in this file since it needs some deployment first.

(which would be faster at runtime btw).

packages/integration-tests/src/test-project-setup/minimal_with_typescript_idioms.ts
```diff
@@ -66,10 +64,29 @@ class MinimalWithTypescriptIdiomTestProject extends TestProjectBase {
...
-    const clientConfigStats = await fs.stat(
-      path.join(this.projectDirPath, 'amplifyconfiguration.json')
+  /**
+   * Creates a test project directory and instance.
+   */
+  static createProject = async (
+    e2eProjectDir: string,
+    cfnClient: CloudFormationClient
+  ): Promise<MinimalWithTypescriptIdiomTestProject> => {
+    const { projectName, projectRoot, projectAmplifyDir } =
```
sobolk, line 74: this looks like outcome of bad merge.
This got replaced by `MinimalWithTypescriptIdiomTestProjectCreator.createProject`

packages/integration-tests/src/test-e2e/deployment.test.ts
rtpascual: Would it be useful to keep one test case to make sure `--config-out-dir` and `--config-format` are working as expected?

sobolk: I'd say no, we've got pretty solid coverage in unit tests for that and likely tests for `generate config` are going to catch integration problems. We should hold until we get signal that we need this e2e test.

As for `generate config`, it's presumably one of our core commands that has zero e2e test coverage. Hence the need to run it at least in one of tests.

packages/integration-tests/src/test-project-setup/minimal_with_typescript_idioms.ts
rtpascual: Done in [8fa9444](https://github.com/aws-amplify/amplify-backend/pull/682/commits/8fa9444d2a1dd65c1f980e6ebbb143464eecdcd5)

packages/integration-tests/src/test-e2e/deployment.test.ts
rtpascual: Done in [019bc64](https://github.com/aws-amplify/amplify-backend/pull/682/commits/019bc647c3d54dcf6cc472b2bef215de2661c4fc)

packages/integration-tests/src/test-project-setup/test_project_base.ts
```diff
@@ -90,4 +95,18 @@ export abstract class TestProjectBase {
   async getUpdates(): Promise<TestProjectUpdate[]> {
     return [];
   }
+
+  /**
+   * Verify client config file is generated with the provided directory and format
+   */
+  async assertPostDeployment(dir?: string, format?: ClientConfigFormat) {
+    const clientConfigStats = await fsp.stat(
+      await getClientConfigPath(dir ?? this.projectDirPath, format)
```
sobolk, line 104: do you have latest `main` in this PR?
I'd expect this fails on `json-mobile` https://github.com/aws-amplify/amplify-backend/blob/14274f43776d8233faaa69d870dcf4fa9e8e0a3b/packages/client-config/src/client-config-types/client_config.ts#L19 .

My other thought is that I'd keep `assertPostDeployment` as it was before because it's part of deploy-assert mechanism (and assumes default flags deployment runs with, has nothing to do with extra config generation).

For probing client config configurations I'd just create separate set of assertions.

rtpascual: Yup I saw it failed on `json-mobile` before getting latest `main`, now it is passing locally.

I see, I was thinking to reuse `assertPostDeployment` since it was checking for existence of `amplifyconfiguration.json` in project directory. With no parameters this acts the same way as before with an added benefit of not having to change the default format if we decide to change it in the future.

sobolk: That looks like accidental duplication per https://gtramontina.com/posts/do-repeat-yourself/ .
these assertions are expected to evolve independently, let's not couple them.

sobolk: Ahh.. I totally missed `getClientConfigPath` . (that's re `json-mobile`).

As for accidental duplication I think that:
1. It would be fine to define `assertClientConfigExists` which has logic in current form.
2. Use `assertClientConfigExists` in `assertPostDeployment`

This would address both problems, i.e. it would eliminate duplication AND didn't bend the "post deployent assertion" abstraction to different purpose.

sobolk (approved):

packages/integration-tests/src/test-project-setup/test_project_base.ts
sobolk: looks good, now thanks!

rtpascual: Yeah used `getClientConfigPath` so everything is aligned (case of `json-mobile` -> `json` and defaults) and we would only need to change one spot.

Done in [971d26e](https://github.com/aws-amplify/amplify-backend/pull/682/commits/971d26e3eeb7332a00291e527bf4232db2f7ac09)

bombguy (approved):

PR #690, opened 2023-11-17
chore: pin data construct

alharris-at:
*Issue #, if available:*
N/A

*Description of changes:*
Pin data construct dependency during launch window.

Followup tracked pre-full-release https://github.com/aws-amplify/amplify-backend/issues/691

### DIFF
--- a/packages/backend-data/package.json
+++ b/packages/backend-data/package.json
@@ -29,7 +29,7 @@
   "dependencies": {
     "@aws-amplify/backend-output-storage": "0.2.4",
     "@aws-amplify/backend-output-schemas": "^0.4.0",
-    "@aws-amplify/data-construct": "^1.4.1",
+    "@aws-amplify/data-construct": "1.4.1",
     "@aws-amplify/plugin-types": "^0.4.1",
     "@aws-amplify/data-schema-types": "^0.6.3"
   }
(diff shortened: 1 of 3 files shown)


iartemiev (approved):

packages/backend-data/package.json
```diff
@@ -29,7 +29,7 @@
   "dependencies": {
     "@aws-amplify/backend-output-storage": "0.2.4",
     "@aws-amplify/backend-output-schemas": "^0.4.0",
-    "@aws-amplify/data-construct": "^1.4.1",
+    "@aws-amplify/data-construct": "1.4.1",
```
sobolk, line 32: what motivates this change?
wouldn't customers have option to pin it with package manager if necessary ?

alharris-at: Was thinking about this after posting, and we can also do things like update `latest` tag if a regression is found urgently (we need to get into that habit anyway). We should probably build those muscles now, not at GA.

This was discussed in standup today to bring stability during preview launch window, circling back w/ the team.

alharris-at: Yeah, after reflecting on it a bit. I think not doing this is probably better for long-term health.

PR #697, opened 2023-11-17
fix(@aws-amplify/backend-data) - Align authorization modes with client

stocaaro:
*Description of changes:*
> Align the terminology used in defineData and mark UPPER_CASE_FORMAT as deprecated
> In the client it's `authMode: 'userPool', 'apiKey'`
>
> Acceptance criteria:
> Support both camel case and UPPER_SNAKE_CASE format but mark all UPPER_CASE_FORMAT as deprecated
> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

In this review, with product input, we have decided not to support the UPPER_SNAKE_CASE on the schema definition surface, keeping only the camel case values consistent with the `aws-amplify` client library.

This change does the following:
- Introduces `'iam' | 'userPool' | 'oidc' | 'apiKey' | 'lambda'` as DefualtAuthorizationMode options
- Translates these to the corresponding CDK modes 'AWS_IAM' | 'AMAZON_COGNITO_USER_POOLS' | 'OPENID_CONNECT' | 'API_KEY' | 'AWS_LAMBDA'
- Updates tests and templates to use the new modes

*Testing*
Package tests have been update / extended to cover this behavior. Working on manual testing against a sample app now.

### DIFF
--- a/.changeset/dull-keys-punch.md
+++ b/.changeset/dull-keys-punch.md
@@ -1,6 +1,6 @@
 ---
 'create-amplify': patch
-'@aws-amplify/backend-data': patch
+'@aws-amplify/backend-data': minor
 ---
 
 fix(@aws-amplify/backend-data) - Align with authorization modes used in the aws-amplify client library
--- a/.changeset/dull-keys-punch.md
+++ b/.changeset/dull-keys-punch.md
@@ -3,4 +3,4 @@
 '@aws-amplify/backend-data': patch
 ---
 
-fix(@aws-amplify/backend-data) - Align authorization modes with client
+fix(@aws-amplify/backend-data) - Align authorization modes used in the aws-amplify client library
--- a/.changeset/dull-keys-punch.md
+++ b/.changeset/dull-keys-punch.md
@@ -1,0 +1,6 @@
+---
+'create-amplify': patch
+'@aws-amplify/backend-data': patch
+---
+
+fix(@aws-amplify/backend-data) - Align authorization modes with client
--- a/.changeset/dull-keys-punch.md
+++ b/.changeset/dull-keys-punch.md
@@ -3,4 +3,4 @@
 '@aws-amplify/backend-data': patch
 ---
 
-fix(@aws-amplify/backend-data) - Align authorization modes used in the aws-amplify client library
+fix(@aws-amplify/backend-data) - Align with authorization modes used in the aws-amplify client library
--- a/packages/backend-data/API.md
+++ b/packages/backend-data/API.md
@@ -35,6 +35,11 @@ export type DataProps = {
 // @public
 export type DataSchema = string | DerivedModelSchema;
 
+// Warning: (ae-forgotten-export) The symbol "AuthorizationModeMapping" needs to be exported by the entry point index.d.ts
+//
+// @public
+export type DefaultAuthorizationMode = keyof typeof AuthorizationModeMapping;
+
 // @public
 export const defineData: (props: DataProps) => ConstructFactory<AmplifyData>;
 
@@ -52,10 +57,6 @@ export type OIDCAuthorizationModeProps = {
     tokenExpiryFromAuthInSeconds: number;
     tokenExpireFromIssueInSeconds: number;
 };
-
-// Warnings were encountered during analysis:
-//
-// src/types.ts:91:3 - (ae-forgotten-export) The symbol "DefaultAuthorizationMode" needs to be exported by the entry point index.d.ts
 
 // (No @packageDocumentation comment for this package)
 
--- a/packages/backend-data/API.md
+++ b/packages/backend-data/API.md
@@ -36,7 +36,7 @@ export type DataProps = {
 export type DataSchema = string | DerivedModelSchema;
 
 // @public
-export type DefaultAuthorizationMode = 'AWS_IAM' | 'AMAZON_COGNITO_USER_POOLS' | 'OPENID_CONNECT' | 'API_KEY' | 'AWS_LAMBDA';
+export type DefaultAuthorizationMode = 'iam' | 'userPool' | 'oidc' | 'apiKey' | 'lambda';
 
 // @public
 export const defineData: (props: DataProps) => ConstructFactory<AmplifyData>;
--- a/packages/backend-data/API.md
+++ b/packages/backend-data/API.md
@@ -35,10 +35,8 @@ export type DataProps = {
 // @public
 export type DataSchema = string | DerivedModelSchema;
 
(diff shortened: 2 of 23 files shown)


.changeset/dull-keys-punch.md
```diff
@@ -0,0 +1,6 @@
+---
+'create-amplify': patch
+'@aws-amplify/backend-data': patch
+---
+
+fix(@aws-amplify/backend-data) - Align authorization modes with client
```
sobolk, line 6: who is "client"?

packages/backend-data/API.md
```diff
@@ -15,14 +15,25 @@ export type ApiKeyAuthorizationModeProps = {
...
 };
 
+// @public
+export const AuthorizationModeMapping: {
+    readonly iam: "AWS_IAM";
+    readonly userPool: "AMAZON_COGNITO_USER_POOLS";
+    readonly oidc: "OPENID_CONNECT";
+    readonly apiKey: "API_KEY";
+    readonly lambda: "AWS_LAMBDA";
+};
```
sobolk, line 25: can you define this as `enum AuthorizationMode` ? or `DataAuthorizationMode` ?

Btw. regarding "Continues to honor the CDK modes while adding a deprecation warning when they are used" from PR description.
We're in preview mode, we should just drop them and map this internally in implementation and make types clean.

packages/backend-data/API.md
```diff
@@ -15,14 +15,25 @@ export type ApiKeyAuthorizationModeProps = {
...
-    defaultAuthorizationMode?: DefaultAuthorizationMode;
     apiKeyAuthorizationMode?: ApiKeyAuthorizationModeProps;
     lambdaAuthorizationMode?: LambdaAuthorizationModeProps;
     oidcAuthorizationMode?: OIDCAuthorizationModeProps;
     allowListedRoleNames?: string[];
-};
+} & DefineDataAuthConfig;
+
+// @public (undocumented)
+export type CdkDefaultAuthorizationMode = (typeof AuthorizationModeMapping)[DefaultAuthorizationMode];
```
sobolk, line 36: we should not use `CDK` in types in this layer.

sobolk (changes requested):

packages/backend-data/API.md
stocaaro: I've reached out to Rene re: "We're in preview mode, we should just drop them"

I'm not quite following the thought to express this mapping as an enum. I think for customers to use string literals, the enum would have to be mapped to the new names, but then I'm unclear. Can you pose an example of what this would look like?

.changeset/dull-keys-punch.md
stocaaro: Adding clarification which I will post in a moment.

packages/backend-data/API.md
stocaaro: Sure. I had these named `DeprecatedDefaultAuthorizationMode`, which didn't seem right when I was building the translation logic to map these into the names used in the `CDKAuthorizationModes` object. The name of this destination object informed this name. If we are removing support for these capital snake case types, then I should be able to remove this type entirely.  Hopefully this is what we can move to. Will circle back after conferring with Rene.

packages/backend-data/API.md
sobolk: That type which was here before looks like this enum https://github.com/aws-amplify/amplify-backend/blob/c28d536f5def56aff6b0f7effe2cf9dc7485dae8/packages/client-config/src/client-config-types/client_config.ts#L16-L22 .

packages/backend-data/API.md
```diff
@@ -56,6 +53,10 @@ export type OIDCAuthorizationModeProps = {
     tokenExpireFromIssueInSeconds: number;
 };
 
+// Warnings were encountered during analysis:
+//
+// src/types.ts:91:3 - (ae-forgotten-export) The symbol "DefaultAuthorizationMode" needs to be exported by the entry point index.d.ts
```
sobolk, line 58: check will complain about this, you need to export type.

.changeset/dull-keys-punch.md
```diff
@@ -0,0 +1,6 @@
+---
+'create-amplify': patch
+'@aws-amplify/backend-data': patch
```
sobolk, line 3: this should be `minor` given break in api.
(while we're at 0.x minor bump should be used to signal break)

packages/backend-data/API.md
stocaaro: Yeah. Just fixed this.

packages/backend-data/API.md
```diff
@@ -35,8 +35,10 @@ export type DataProps = {
 // @public
 export type DataSchema = string | DerivedModelSchema;
 
+// Warning: (ae-forgotten-export) The symbol "AuthorizationModeMapping" needs to be exported by the entry point index.d.ts
+//
 // @public
-export type DefaultAuthorizationMode = 'AWS_IAM' | 'AMAZON_COGNITO_USER_POOLS' | 'OPENID_CONNECT' | 'API_KEY' | 'AWS_LAMBDA';
+export type DefaultAuthorizationMode = keyof typeof AuthorizationModeMapping;
```
sobolk, line 41: We should do something in implementation to not drag `AuthorizationModeMapping` into public types.

stocaaro: Got it. I think splitting this string literal from the mapping and moving the mapping directly into where its used to translate from one auth mode structure to another speaks to both this comment and the above comment. I have made this change, will be up in a minute.

sobolk: Whatever it takes :)
Internal mechanism to map this can be changed with patch version release if types don't leak into public surface.

sobolk (approved):

packages/backend-data/src/convert_authorization_modes.ts
```diff
@@ -150,6 +150,20 @@ const computeApiKeyAuthFromResource = (
   };
 };
 
+const authorizationModeMapping = {
```
alharris-at, line 153: Could you type this as a Record<DefaultAuthorizationMode, CDKDefaultAuthorizationMode>?

alharris-at (approved):

iartemiev (approved):

packages/backend-data/src/convert_authorization_modes.ts
stocaaro: I had some of this in an earlier version. My making the mapping `const` we get this strictness for free without any additional public exporting. Would recommend we leave it and add this on a refactor.

PR #706, opened 2023-11-18
fix: turn off type checking if amplify/tsconfig.json is not found

Amplifiyer:
*Description of changes:* fix: turn off type checking if amplify/tsconfig.json is not found

If the amplify/tsconfig.json is not found, TSC doesn't go to the root or parent director looking for tsconfig rather errors out with `error TS5057: Cannot find a tsconfig.json file at the specified directory: 'amplify'.`
This PR checks if we can load the config before executing TSC. If we cannot, we turn off ts compiler.

### DIFF
--- a/.changeset/few-readers-brush.md
+++ b/.changeset/few-readers-brush.md
@@ -1,0 +1,5 @@
+---
+'@aws-amplify/backend-deployer': patch
+---
+
+turn off type checking if amplify/tsconfig.json is not found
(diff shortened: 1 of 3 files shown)


edwardfoyle (approved):
I'm okay if this is the behavior we want. The other option here is to say that you _must_ have a tsconfig in the amplify dir and the way to turn off type checking is to explicitly exclude the directory

Amplifiyer:
> I'm okay if this is the behavior we want. The other option here is to say that you _must_ have a tsconfig in the amplify dir and the way to turn off type checking is to explicitly exclude the directory

The only way to "exclude" the directory is to have a tsconfig (since customers don't have access to tsc command line which we execute), thereby making it more prohibitive, especially for JS/CJS projects.

PR #756, opened 2023-12-01
Cognito domain creation and expose OAuth to frontend config

awsluja:
*Issue #, if available:*

*Description of changes:*
This change enables the creation of a Cognito domain given the specified 'domainPrefix'.
It allows users to create, update, or delete the domain after initial deployments (with a slight workaround).

Also, exposes the OAuth settings to the frontend config.
- social providers 
- oauth (domain, scope, signin/out urls, response type)

UX:
image


packages/auth-construct/src/construct.ts
```diff
@@ -712,18 +756,41 @@ export class AmplifyAuth
...
+          this.oAuthSettings.scopes?.map((s) => s.scopeName) ?? []
+        );
+        output.oauthRedirectSignIn = this.oAuthSettings.callbackUrls
+          ? this.oAuthSettings.callbackUrls.join(',')
+          : '';
+        output.oauthRedirectSignOut = this.oAuthSettings.logoutUrls
+          ? this.oAuthSettings.logoutUrls.join(',')
+          : '';
+        output.webClientId = this.resources.userPoolClient.userPoolClientId;
+        output.oauthResponseType = 'code';
```
awsluja, line 792: note: 'code' seems to be the default for hosted ui (old cli), when would this not be the case?

packages/auth-construct/src/types.ts
```diff
@@ -199,11 +199,19 @@ export type ExternalProviderOptions = {
   /**
    * List of allowed redirect URLs for the identity providers.
    */
-  callbackUrls?: string[];
+  callbackUrls: string[];
```
edwardfoyle, line 202: not strictly part of this PR, but this should either be `redirectUrls` to match the comment, or change the comment to say "callback URLs"

packages/auth-construct/src/construct.ts
edwardfoyle: is this still an open question?

edwardfoyle (dismissed):

packages/auth-construct/src/construct.ts
awsluja: not anymore; 'token' is considered legacy and less secure than 'code' so we don't have a need for 'token' at the moment

packages/auth-construct/src/types.ts
awsluja: image
I'm not sure why but CDK uses the names interchangeably here; I'm not sure if this is intentional

packages/auth-construct/src/construct.ts
```diff
@@ -108,32 +121,56 @@ export class AmplifyAuth
...
     // UserPool - Identity Providers
-    const providerSetupResult = this.setupIdentityProviders(
+    this.providerSetupResult = this.setupIdentityProviders(
       this.userPool,
       props.loginWith
     );
-    this.oauthMappings = providerSetupResult.oauthMappings;
 
-    // UserPool Client
+    this.domainPrefix = props.loginWith.externalProviders?.domainPrefix ?? '';
```
sobolk, line 129: why do we need `?? ''` if we check it in `if` below ?
is `this.domainPrefix` type correct?

packages/auth-construct/src/construct.ts
```diff
@@ -108,32 +121,56 @@ export class AmplifyAuth
...
+    if (
+      this.providerSetupResult.providersList.length > 0 &&
+      this.domainPrefix
```
sobolk, line 132: should we check providers list here?
if I  explicitly provide only `domainPrefix` in inputs I'd be surprised to see it not created.
alternatively we should validate and throw if providing domain prefix doesn't make sense if providers list is empty.

awsluja: I went with the latter, throwing an error if domainPrefix is specified with an empty providers list

packages/auth-construct/src/construct.ts
awsluja: good catch, i updated the type

edwardfoyle (approved):

0618 (approved):

PR #761, opened 2023-12-02
chore: update passwordless API

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- Update public API to be consistent with auth category

### DIFF
--- a/packages/passwordless-auth-construct/src/constants.ts
+++ b/packages/passwordless-auth-construct/src/constants.ts
@@ -1,0 +1,4 @@
+/**
+ * A placeholder for the code or link in messages.
+ */
+export const codeOrLinkPlaceholder = '##codeOrLink##';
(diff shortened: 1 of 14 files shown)


packages/passwordless-auth-construct/src/constants.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * A placeholder for the code or link in messages.
+ */
+export const codeOrLinkPlaceholder = '##codeOrLink##';
```
Equartey, line 4: Question: do we want to allow this to be customized?

Jordan-Nelson: It is only used internally. This isn't exposed to the customer. The API that the customer uses is

```ts
message: (code: string) => `Your code is ${code}`
```

Equartey (approved):

sobolk (approved):

PR #767, opened 2023-12-04
chore: add prefix to metadata keys

Jordan-Nelson:
*Issue #, if available:*

*Description of changes:*
- prefix metadata keys to prevent collisions with customer provided metadata

### DIFF
--- a/packages/passwordless-auth-construct/src/constants.ts
+++ b/packages/passwordless-auth-construct/src/constants.ts
@@ -2,3 +2,10 @@
  * A placeholder for the code or link in messages.
  */
 export const codeOrLinkPlaceholder = '##codeOrLink##';
+
+// Metadata keys for passwordless auth are prefixed to prevent collisions with
+// keys provided by the customer
+export const actionMetadataKey = 'Amplify.Passwordless.action';
+export const deliveryMediumMetadataKey = 'Amplify.Passwordless.deliveryMedium';
+export const redirectUriMetadataKey = 'Amplify.Passwordless.redirectUri';
+export const signInMethodMetadataKey = 'Amplify.Passwordless.signInMethod';
--- a/packages/passwordless-auth-construct/src/constants.ts
+++ b/packages/passwordless-auth-construct/src/constants.ts
@@ -3,9 +3,20 @@
  */
 export const codeOrLinkPlaceholder = '##codeOrLink##';
 
-// Metadata keys for passwordless auth are prefixed to prevent collisions with
-// keys provided by the customer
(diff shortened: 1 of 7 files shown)


packages/passwordless-auth-construct/src/constants.ts
```diff
@@ -2,3 +2,10 @@
  * A placeholder for the code or link in messages.
  */
 export const codeOrLinkPlaceholder = '##codeOrLink##';
+
+// Metadata keys for passwordless auth are prefixed to prevent collisions with
+// keys provided by the customer
+export const actionMetadataKey = 'Amplify.Passwordless.action';
+export const deliveryMediumMetadataKey = 'Amplify.Passwordless.deliveryMedium';
+export const redirectUriMetadataKey = 'Amplify.Passwordless.redirectUri';
+export const signInMethodMetadataKey = 'Amplify.Passwordless.signInMethod';
```
sobolk, line 11: Who's "metadata" this is ? I.e. which service uses it?
(I'm asking because we use cfn stack metadata in couple of other places)

Equartey (approved):
Looks good. Just address @sobolk's question first

packages/passwordless-auth-construct/src/constants.ts
Jordan-Nelson: This is metadata provided in the Cognito RespondToAuthChallenge API (see: https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_RespondToAuthChallenge.html#CognitoUserPools-RespondToAuthChallenge-request-ClientMetadata).

The client invokes RespondToAuthChallenge and provides metadata. Cognito provides that metadata to several lambda functions, including the lambdas that are used for passwordless auth.

The passwordless auth lambdas expect specific metadata, which the Amplify client libraries will provide. However, an Amplify customer could provide their own metadata on top of the metadata that the Amplify client libraries will provide. The prefix "Amplify.Passwordless" reduces the chance that the two sets of metadata collide.

sobolk: Awesome.
Can you please mention that in code comment?
Also consider introducing some kind of grouping, for example:
```
enum CognitoClientMetadataKeys {
 ACTION = '...',
 // and so on
}
```
(like here https://github.com/aws-amplify/amplify-backend/blob/8facb329ed5c263846d1a648375905fc1b7e451e/packages/client-config/src/client-config-types/client_config.ts#L16-L22)

Jordan-Nelson: Enum makes sense 👍 . Updated.

sobolk (approved):

PR #792, opened 2023-12-07
feat(passwordless): Allow OTP Retry

Equartey:
*Description of changes:*
Allow users to make up to 3 attempts when trying to confirm their OTP code before the auth session fails. This is accomplished by: 

1. Saving the OTP and delivery method under the `challengeMetadata` key in the response.
2. On failed attempts, restart the flow and reuse the saved OTP code and delivery method.
3. Fail auth session after 3 attempts 

### DIFF
--- a/packages/passwordless-auth-construct/src/custom-auth/custom_auth_service.ts
+++ b/packages/passwordless-auth-construct/src/custom-auth/custom_auth_service.ts
@@ -126,11 +126,9 @@ export class CustomAuthService {
 
     const clientMetadata = event.request.clientMetadata || {};
     const signInMethod = clientMetadata[CognitoMetadataKeys.SIGN_IN_METHOD];
+    const signInAction = clientMetadata[CognitoMetadataKeys.ACTION];
     const method = this.validateSignInMethod(signInMethod);
-    const action = this.validateAction(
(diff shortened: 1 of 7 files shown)


packages/passwordless-auth-construct/src/custom-auth/custom_auth_service.ts
```diff
@@ -213,11 +251,20 @@ export class CustomAuthService {
   private validateDeliveryCodeDetails = (
     event: CreateAuthChallengeTriggerEvent
   ): CodeDeliveryDetails => {
+    const previousChallenge = event.request.session.slice(-1)[0];
+    const previousDeliveryMedium: string | undefined =
+      previousChallenge.challengeMetadata?.includes('deliveryMedium')
+        ? JSON.parse(previousChallenge.challengeMetadata).deliveryMedium
+        : undefined;
```
israx, line 258:
```suggestion
JSON.parse(previousChallenge.challengeMetadata).deliveryMedium
```
It will return `undefined` if the key is not found.

israx (commented):
LGTM. Just a couple of comments

sobolk (approved):

PR #795, opened 2023-12-08
Initial implementation SignUp passwordless

elorzafe:
*Issue #, if available:*

*Description of changes:*
Adds passwordless sign up
- works for phone_number and email including verification during sign in

Pending items:
- fix linter issues
- export endpoint (currently only visible)

Tested end to end and works as expected (only issue I noticed is that is very slow maybe needs to enable prevent user existance errors)


packages/passwordless-auth-construct/src/sign-up/createUserService.ts
```diff
@@ -0,0 +1,43 @@
...
+    if (params.phone_number) {
+        userAttributes.push({
+            Name: "phone_number",
+            Value: params.phone_number
+        })
+    }
+
+    const client = new CognitoIdentityProviderClient({ region: params.region });
+    const command = new AdminCreateUserCommand({
+        Username: params.email || params.phone_number, // what happens when there is both??
```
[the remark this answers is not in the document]
elorzafe: this is a good question, the payload for sign up a user from API GW (not Cognito) is defined by us not the Cognito trigger as other examples

packages/passwordless-auth-construct/src/sign-up/construct.ts
```diff
@@ -0,0 +1,72 @@
...
+        const defaultMemorySize = 128;
+
+        const commonOptions: NodejsFunctionProps = {
+            entry: path.join(dirname, '.', 'createUserService.js'),
+            runtime: Runtime.NODEJS_20_X,
+            architecture: Architecture.ARM_64,
+            bundling: {
+                format: OutputFormat.ESM,
+            },
+        };
```
Equartey, line 33: suggestion: we should have these settings defined in one spot if they are going to be similar to https://github.com/aws-amplify/amplify-backend/blob/feat/passwordless-auth/packages/passwordless-auth-construct/src/construct.ts#L46-L53

Equartey (commented):
Looking good!

packages/passwordless-auth-construct/src/custom-auth/custom_auth_service.ts
```diff
@@ -187,9 +192,38 @@ export class CustomAuthService {
...
+
+    const answerCorrect = (await verifyResult).response.answerCorrect;
+    logger.debug(`Answer is correct: ${answerCorrect ? 'true' : 'false'}`);
+
+    const userCreatedByPasswordlessSignUp =
+      event.request.userAttributes[PASSWORDLESS_SIGNUP_ATTR_NAME] === 'true';
+    const attributeName =
+      event.request.privateChallengeParameters.deliveryMedium === 'SMS'
+        ? 'phone_number_verified'
+        : 'email_verified';
```
Jordan-Nelson, line 207: The value from `privateChallengeParameters` cannot be used. It will not be there in the magic link flow since a new custom auth session is started for link verification.

The delivery medium (or attribute name) should be stored as the value of the custom attribute.

packages/passwordless-auth-construct/src/custom-auth/custom_auth_service.ts
```diff
@@ -12,7 +12,12 @@ import {
   SignInMethod,
 } from '../types.js';
 import { CognitoMetadataKeys } from '../constants.js';
+import {
+  AdminUpdateUserAttributesCommand,
+  CognitoIdentityProviderClient,
+} from '@aws-sdk/client-cognito-identity-provider';
 
+const PASSWORDLESS_SIGNUP_ATTR_NAME = 'custom:_passwordless_signup';
```
Jordan-Nelson, line 20: *nit:* Sign up is either two words as a verb or hyphenated (sign-up) as a noun/adjective. Either way it should be split up.

```suggestion
const PASSWORDLESS_SIGNUP_ATTR_NAME = 'custom:_passwordless_sign_up';
```

packages/passwordless-auth-construct/src/sign-up/construct.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    api.root.addMethod('PUT', putCreateUserIntegration);
+
+    const createUserPolicy = new Policy(this, `CreateUserPolicy${id}`, {});
+    createUserFunction.role!.attachInlinePolicy(createUserPolicy);
+    const createUserPolicyStatement = new PolicyStatement();
+    createUserPolicyStatement.addActions('cognito-idp:AdminCreateUser');
+    createUserPolicyStatement.addResources(userPool.userPoolArn);
+    createUserPolicyStatement.effect = Effect.ALLOW;
+    createUserPolicy.addStatements(createUserPolicyStatement);
```
Jordan-Nelson, line 73: Question: Is there an advantage to using the more imperative API compared to:

```ts
createUserFunction.addToRolePolicy(
  new PolicyStatement({...})
);
```

For example, see:
https://github.com/aws-amplify/amplify-backend/blob/ca2f170f2e43b67a25178293bee3ecbc3b6ac432/packages/passwordless-auth-construct/src/construct.ts#L94-L115

packages/auth-construct/src/construct.ts
```diff
@@ -134,6 +144,8 @@ export class AmplifyAuth
             ? { scopes: this.getOAuthScopes(externalProviders.scopes) }
             : {}),
         },
+        readAttributes: clientReadAttributes,
```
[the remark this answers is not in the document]
elorzafe: yes correct

packages/passwordless-auth-construct/src/sign-up/construct.ts
elorzafe: But I updated it to make it more readable, thanks

PR #800, opened 2023-12-08
build: add api test

sobolk:
*Issue #, if available:*

**Description of changes:**

This is attempt 3 of building API tests.
Previous attempts:
https://github.com/aws-amplify/amplify-backend/pull/788
https://github.com/aws-amplify/amplify-backend/pull/794

The algorithm is:
1. Checkout latest and baseline repo (i.e. PR content and PR base branches)
2. Publish packages to verdaccio (local registry)
3. Generate test projects
    1. Use baseline API.md to generate usage snippets
    2. Use latest as dependency to test project
4. Compile generated projects.
    1. Passing compilation indicates no breaking change detected (as good as our coverage).
    2. Failing compilation indicates breaking change.

Implemented checks include:
1. Enum validation
2. Type assignability validation

**Out of scope**

This is initial version of this mechanism and is focused on establishing algorithm and check.
1. Further expansion: exported classes and functions/constants/vars are out of scope for this PR and will be additive expansion on the top of it.
2. Further improvements of debug-ability: current version was developed using development loop on local machine that was working quite easy and fast.  Debugging procedure is documented in script jsdocs. This will be improved separately if needed.
3. Integration with semver diffs/changeset. This check can be accompanied by semver diff check. I.e. a version bump coming from changeset can be validated against API diff detected (e.g. no API change - patch, non-breaking API change - minor, breaking API change - major, detection of no API change or intentional major version increment should skip this change for given package).
4. Mechanism to suppress the check for intentional breaks (for now check will be optional).

**Testing**

1. Enabled this check for all packages in repo
2. Added integration tests that assert this mechanism can catch expected breaking changes
3. Unit tests for snippet generators (so that it's easier to work with them locally)


edwardfoyle (commented):
Overall this is looking _wayyy_ better than the original version. It's still a bit dense in places (like generating code from the ast) and there's a bit of "wait, which directory is this operating in?" when reading through the api changes validator but I think with some additional comments throughout this can be mostly fixed.

I also left a comment about seeing if we can have a local script to do the setup for this check which I think would be really helpful for being able to run / debug this locally.

.github/workflows/health_checks.yml
```diff
@@ -43,6 +43,51 @@ jobs:
...
+        run: |
+          mkdir api-validation-projects
+          cd api-validation-projects
+          workingDirectory=$(pwd)
+          cd ..
+          cd base-branch-content
+          baselinePath=$(pwd)
+          cd ..
+          cd pull-request-content
+          latestPath=$(pwd)
```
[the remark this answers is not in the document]
sobolk: That's a scrappy solution to get absolute paths. Will be refactored when I get algorithm ironed out.

scripts/components/api-changes-validator/api_changes_validator.ts
```diff
@@ -0,0 +1,93 @@
...
+      cwd: this.projectPath,
+      stdio: 'inherit',
+    });
+  };
+
+  private createTestProject = async () => {
+    const dependencies: Record<string, string> = {};
+    dependencies[
+      `${this.packageName}-latest`
+    ] = `file://${this.latestPackagePath}`;
```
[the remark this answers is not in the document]
sobolk: I'm planing to try verdaccio instead. Since it turned out we have to do local publish for different reasons. 
If that won't work well I'll try npm link.

.github/workflows/health_checks.yml
```diff
@@ -43,6 +43,51 @@ jobs:
       - uses: ./.github/actions/restore_build_cache
       - run: npm run set-script-shell
       - run: npm run test:coverage:threshold
+  check_api_changes:
```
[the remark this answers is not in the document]
sobolk: I'll look into that.
Other than setup the script is quite nice to use locally.
I.e one can choose where new old and working dir is. And it leaves artifacts in working dir so one can open them in IDE and inspect.

scripts/components/api-changes-validator/api_changes_validator.ts
```diff
@@ -0,0 +1,93 @@
...
+  /**
+   * creates api changes validator
+   */
+  constructor(
+    private readonly packageName: string,
+    private readonly baselinePackageApiReportPath: string,
+    private readonly latestPackagePath: string,
+    private readonly workingDirectory: string
+  ) {
+    this.projectPath = path.join(workingDirectory, this.packageName);
```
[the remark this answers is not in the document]
sobolk: package name is really `<repoRoot>/packages/<packageName>` . Should be already fs friendly.
I'll work on naming.

.github/workflows/health_checks.yml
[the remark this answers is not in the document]
sobolk: This got refactored.

scripts/check_api_changes.ts
```diff
@@ -0,0 +1,48 @@
...
+import path from 'path';
+import { existsSync } from 'fs';
+import { ApiChangesValidator } from './components/api-changes-validator/api_changes_validator.js';
+
+// extract the command args that should be run in each package
+const runArgs = process.argv.slice(2);
+
+const latestRepositoryPath = runArgs[0];
+const baselineRepositoryPath = runArgs[1];
+const workingDirectory = runArgs[2];
```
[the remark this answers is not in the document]
sobolk: done

scripts/components/api-changes-validator/api_changes_validator.ts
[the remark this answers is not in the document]
sobolk: I reworked naming and data that is going around should be better now.

scripts/components/api-changes-validator/api_changes_validator.ts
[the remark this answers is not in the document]
sobolk: I ended up offering two options: using remote and using npm link.
remote is used by script, link is used in integration tests.
should be easy to modify this if needed.

.github/workflows/health_checks.yml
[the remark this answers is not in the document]
sobolk: I made the script assume that it's repo root is "latest".
So debugging this locally is as complex as:
1. checkout/be in your PR branch
2. checkout base of your PR to separate dir (one liner command)
3. create "working directory"
4. call script.

(I've been running this from WebStorm using play button)

Therefore I suggest we park this part until this becomes painful.

scripts/components/api-changes-validator/api_changes_validator.ts
```diff
@@ -0,0 +1,93 @@
...
+    dependencies['typescript'] = '^5.0.0';
+
+    // Add latest dependencies and peer dependencies as public API might use them
+    const latestPackageJsonContent = await fsp.readFile(
+      path.join(this.latestPackagePath, 'package.json'),
+      'utf-8'
+    );
+    const latestPackageJson = JSON.parse(latestPackageJsonContent);
+    if (latestPackageJson.dependencies) {
+      for (const dependencyKey of Object.keys(latestPackageJson.dependencies)) {
```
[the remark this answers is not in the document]
sobolk: I'm gonna punt on this...
image

.github/workflows/health_checks.yml
```diff
@@ -43,6 +43,37 @@ jobs:
       - uses: ./.github/actions/restore_build_cache
       - run: npm run set-script-shell
       - run: npm run test:coverage:threshold
+  test_scripts:
```
sobolk, line 46: I have separated these tests.
They are not very useful when working on `packages/*` and start growing.

.github/workflows/health_checks.yml
```diff
@@ -43,6 +43,37 @@ jobs:
...
+      - name: Publish packages locally
+        run: npm run vend
+      - name: Checkout base branch
+        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # version 3.6.0
+        with:
+          path: base-branch-content
+          ref: ${{ github.event.pull_request.base.sha }}
+      - name: Check API changes
+        run: |
+          mkdir api-validation-projects
```
edwardfoyle, line 75: could the script create a temp dir for validation projects internally? This would remove one more setup step (albiet a simple one)

scripts/components/api-changes-validator/api_usage_statements_generators.ts
```diff
@@ -0,0 +1,249 @@
...
+        this.typeAliasDeclaration.typeParameters
+      ).generate().usageStatement ?? '';
+    const genericTypeParameters =
+      new GenericTypeParameterUsageStatementsGenerator(
+        this.typeAliasDeclaration.typeParameters
+      ).generate().usageStatement ?? '';
+    const baselineTypeName = `${typeName}Baseline`;
+    const functionParameterName = `${constName}FunctionParameter`;
+    // declare type with same content under different name.
+    let usageStatement = `type ${baselineTypeName}${genericTypeParametersDeclaration} = ${this.typeAliasDeclaration.type.getText()}${EOL}`;
```
edwardfoyle, line 148: instead of doing all of this string building ourselves, I wonder if we could use the [typescript printer](https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API#creating-and-printing-a-typescript-ast) to construct this from AST nodes?

scripts/components/api-changes-validator/api_changes_validator.test.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  for (const testProject of testProjectsWithBreaks) {
+    void it(`detects break in project ${testProject}`, async () => {
+      const latestPackagePath = path.join(
+        testProjectsWithBreaksPath,
+        testProject
+      );
+      const expectedErrorMessage = (
+        await fsp.readFile(
+          path.join(latestPackagePath, 'expected-error-message.txt'),
```
edwardfoyle, line 51: could this just be an inline string?

scripts/components/api-changes-validator/api_usage_generator.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    expectedApiUsage: `
+import { SomeClass } from 'samplePackageName';
+    `,
+  },
+];
+
+void describe('Api usage generator', () => {
+  for (const testCase of testCases) {
+    void it(testCase.description, () => {
+      const apiReportAST = ApiReportParser.parse(testCase.apiReport);
```
edwardfoyle, line 92: All of the escaped backticks in the test cases makes them a bit difficult to read. Perhaps instead we could do something like
```suggestion
ApiReportParser.parse(nestInMarkdownCodeblock(testCase.apiReport))
```
where `nestInMarkdownCodeblock` does the pre and postfix of "\```ts" and "```"

.github/workflows/health_checks.yml
sobolk: discussed offline.
while working on this PR I discovered that it's useful to force user to know where working dir is by making parameter mandatory.
The script intentionally leaves generated projects in place and does not delete them after the run. This gives opportunity to inspect generated code and play with this to understand what broke (e.g. one can open this dir in IDE to further understand the break).

scripts/components/api-changes-validator/api_changes_validator.test.ts
sobolk: this is intentional. the test is parameterized and error message is part of project definition.

scripts/components/api-changes-validator/api_usage_generator.test.ts
sobolk: this is done

scripts/components/api-changes-validator/api_usage_statements_generators.ts
sobolk: I did an experiment here https://github.com/sobolk/amplify-backend/commit/c31f73b8937c3f08a4c1395cd812ca09576b0128 .

Which led me to conclusion that using `ts.printer` and `ts.factory` for this purpose is going to be overkill.
I had to spin up extra workspace where I was using expected output and debugger to figure out how to construct desired AST. I covered couple of edge cases around type usage snippet generation but gave up at some point given complexity (the experiment is missing function parameter type mandatory template params).
Given expected low complexity of "usage samples" we generate and high complexity of figuring out how to generate them I think we should keep template approach for now until we discover it doesn't scale well.

Since this is internal detail of the tool and usage snippet generation is under "unit-ish" test control it should be straightforward (but dense) effort to migrate later.

edwardfoyle (approved):
LGTM, need to merge main to appease the PR size check

.github/workflows/health_checks.yml
```diff
@@ -43,6 +43,37 @@ jobs:
...
+  test_scripts:
+    needs:
+      - build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # version 3.6.0
+      - uses: ./.github/actions/setup_node
+      - uses: ./.github/actions/restore_build_cache
+      - run: npm run set-script-shell
+      - run: npm run test:scripts
```
0618, line 55: nit
```suggestion
      - run: |
         npm run set-script-shell
         npm run test:scripts
```

scripts/components/api-changes-validator/api_report_parser.ts
```diff
@@ -0,0 +1,23 @@
+import ts from 'typescript';
+
```
0618, line 2: no test for `ApiReportParser`?

scripts/components/api-changes-validator/api_usage_generator.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    return `${importStatements.join(EOL)}${EOL}${EOL}${usageStatements.join(
+      EOL
+    )}${EOL}`;
+  };
+
+  private generateStatementsForNode = (
+    node: ts.Node
+  ): UsageStatements | undefined => {
+    if (node.kind === ts.SyntaxKind.ImportDeclaration) {
```
0618, line 49: nit: should we use `switch` instead?

scripts/components/api-changes-validator/test-resources/test-projects/with-breaks/enum-break-remove-enum-member/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "enum-break-remove-enum-member",
```
0618, line 2: nit: let's add a `@aws-amplify` prefix to all the internal/external package, just in case

scripts/components/api-changes-validator/test-resources/test-projects/with-breaks/enum-break-remove-enum-member/package.json
```diff
@@ -0,0 +1,12 @@
...
+  "type": "module",
+  "exports": {
+    ".": {
+      "types": "./lib/index.d.ts",
```
0618, line 6: nit: plz sort line 6-9

scripts/components/api-changes-validator/test-resources/test-projects/with-breaks/type-break-make-property-mandatory/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "type-break-make-property-mandatory",
```
0618, line 2: same here

scripts/components/api-changes-validator/test-resources/test-projects/with-breaks/type-break-make-property-mandatory/package.json
```diff
@@ -0,0 +1,12 @@
...
+  "type": "module",
+  "exports": {
+    ".": {
+      "types": "./lib/index.d.ts",
```
0618, line 6: same here for sorting

scripts/components/api-changes-validator/test-resources/test-projects/without-breaks/project-without-breaks/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "project-without-breaks",
```
0618, line 2: same here

scripts/components/api-changes-validator/test-resources/test-projects/without-breaks/project-without-breaks/package.json
```diff
@@ -0,0 +1,12 @@
...
+  "type": "module",
+  "exports": {
+    ".": {
+      "types": "./lib/index.d.ts",
```
0618, line 6: and here for the sorting

scripts/components/api-changes-validator/test-resources/.eslintrc.json
```diff
@@ -0,0 +1,8 @@
+{
+  "rules": {
+    "@typescript-eslint/no-empty-function": "off",
+    "@typescript-eslint/no-unused-vars": "off",
+    "jsdoc/require-jsdoc": "off",
+    "no-duplicate-imports": "off"
```
0618, line 6: why to turn them off?

scripts/components/api-changes-validator/api_report_parser.ts
sobolk: It's covered by `scripts/components/api-changes-validator/api_usage_generator.test.ts` below.

It also runs as part of `scripts/components/api-changes-validator/api_changes_validator.ts` integration test and also runs as part of GH Action check.

While working on this PR I was trying to find balance between providing adequate coverage and local debugability vs writing tests for the sake of having them. This particular component is a this wrapper on top of 3rd party TS API, if it doesn't work 3 other places will go red.

In general this PR relies heavy on typescript compiler (both api and command line) hence testing effort is integration heavy (but still fast, all scripts tests complete <5s ).

scripts/components/api-changes-validator/api_usage_generator.ts
sobolk: we could do that, the reason `if` remained here is that these blocks were much heavier at some point and linter does not allow to define new `const/var` inside switch blocks.

0618 (dismissed):
LGTM. Just added some nit-picking

scripts/components/api-changes-validator/test-resources/.eslintrc.json
sobolk: The typescript idioms we use in integration test and those generated by validator are often "hanging in the void", i.e. they're not expected to be part of a program that does something useful. Their purpose is just to trigger compiler checks.

0618 (approved):
Thanks for updating the PR! LGTM!

edwardfoyle (approved):

sobolk:
A follow up fix https://github.com/aws-amplify/amplify-backend/pull/812 .

Issue #796, opened 2023-12-08
Improperly Constructed ARNs in auth-construct lead to deployment failure


srquinn21:
### Environment information

```plain text
System:
    OS: macOS 13.5.2
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 1.95 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.18.2 - ~/.nvm/versions/node/v18.18.2/bin/node
    Yarn: Not Found
    npm: 9.8.1 - ~/.nvm/versions/node/v18.18.2/bin/npm
    pnpm: Not Found
    bun: Not Found
    Watchman: Not Found
  npmPackages:
    @aws-amplify/backend: ^0.5.5 => 0.5.5
    @aws-amplify/backend-cli: ^0.9.1 => 0.9.1
    aws-amplify: ^6.0.5 => 6.0.5
    aws-cdk: Not Found
    aws-cdk-lib: Not Found
    typescript: ^5.3.2 => 5.3.2
```

### Description

Receiving a deployment failure when trying to configure a SAML provider with the auth-construct due to improperly formatted  ARNS. These two lines are the culprits:

https://github.com/aws-amplify/amplify-backend/blob/main/packages/auth-construct/src/construct.ts#L259
https://github.com/aws-amplify/amplify-backend/blob/main/packages/auth-construct/src/construct.ts#L268

Which produce ARNs with region in the wrong slot: `arn:us-east-1:iam::814763596509:saml-provider/MicrosoftEntraIDSAML`

Two thoughts:
* Is there no other choice but to hand construct these ARNs? Can we not get them programmatically from the resource itself?
* If we must hand construct, we should have a common method that does so and a test that ensure's we serialized the string correctly

srquinn21:
We should use the official parser from the SDK https://github.com/aws/aws-sdk-js-v3/blob/main/packages/util-arn-parser/src/index.ts

PR #802, opened 2023-12-08
bug(796): Replaces arn construction logic with SDK parser

srquinn21:
*Issue 796*
https://github.com/aws-amplify/amplify-backend/issues/796

*Description of changes:*
Replaces manual creation of ARN strings with official builder from AWS SDK

### DIFF
--- a/.changeset/nice-socks-complain.md
+++ b/.changeset/nice-socks-complain.md
@@ -1,0 +1,5 @@
+---
+'@aws-amplify/auth-construct-alpha': patch
+---
+
+Fixes bug with ARN construction when configuring SAML/OIDC
--- a/package-lock.json
+++ b/package-lock.json
@@ -21319,11 +21319,23 @@
       "dependencies": {
         "@aws-amplify/backend-output-schemas": "^0.4.0",
         "@aws-amplify/backend-output-storage": "^0.2.6",
-        "@aws-amplify/plugin-types": "^0.5.0"
+        "@aws-amplify/plugin-types": "^0.5.0",
+        "@aws-sdk/util-arn-parser": "^3.465.0"
       },
       "peerDependencies": {
         "aws-cdk-lib": "^2.110.1",
         "constructs": "^10.0.0"
+      }
+    },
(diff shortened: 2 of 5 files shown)


awsluja (dismissed):
Looks good, nice catch

edwardfoyle (commented):
LGTM, can we get a test for this?

srquinn21:
> LGTM, can we get a test for this?

The `util-arn-parser` functions are well unit tested. Deserves an e2e for OIDC/SAML, though.

sobolk:
> > LGTM, can we get a test for this?
> 
> The `util-arn-parser` functions are well unit tested. Deserves an e2e for OIDC/SAML, though.

We're still passing some inputs to it ? so we should assert that output matches expectation.
(fwiw I wouldn't mock arn parser given it's stability but we should assert that what's coming back matches expectations.).

sobolk (approved):

srquinn21:
> > > LGTM, can we get a test for this?
> > 
> > 
> > The `util-arn-parser` functions are well unit tested. Deserves an e2e for OIDC/SAML, though.
> 
> We're still passing some inputs to it ? so we should assert that output matches expectation. (fwiw I wouldn't mock arn parser given it's stability but we should assert that what's coming back matches expectations.).

I added units in this style. I didn't originally because these appear to be pretty fragile since a change to the input will require an update to the test.

awsluja (approved):

sobolk:
> > > > LGTM, can we get a test for this?
> > > 
> > > 
> > > The `util-arn-parser` functions are well unit tested. Deserves an e2e for OIDC/SAML, though.
> > 
> > 
> > We're still passing some inputs to it ? so we should assert that output matches expectation. (fwiw I wouldn't mock arn parser given it's stability but we should assert that what's coming back matches expectations.).
> 
> I added units in this style. I didn't originally because these appear to be pretty fragile since a change to the input will require an update to the test assertion.

the part with hash looks potentially problematic, but we can replace this with some kind of partial match if this becomes a headache.

PR #811, opened 2023-12-12
test: fix api check - retain version information after local publish

sobolk:
*Issue #, if available:*

*Description of changes:*

Fixes https://github.com/aws-amplify/amplify-backend/actions/runs/7177124185/job/19543184647?pr=810 .

Add option to retain git diff after local publish. I.e. retain information about versions that got published to verdaccio.
Test projects in api check use pinned versions to resolve package under test to make sure they don't accidentally pull wrong version.

I was debating if we should use `latest` in test project's package json but I figured using precise version is going to remove some doubts about what exactly is resolved.

### DIFF
--- a/.github/workflows/health_checks.yml
+++ b/.github/workflows/health_checks.yml
@@ -64,16 +64,17 @@ jobs:
         uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # version 3.6.0
       - uses: ./.github/actions/setup_node
       - uses: ./.github/actions/restore_build_cache
+      - name: Publish packages locally
+        run: npm run vend
       - name: Checkout base branch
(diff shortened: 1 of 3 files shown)


edwardfoyle (approved):
I think I like this option better, but I'm good with either one. Having a `--keepGitDiff` flag on publish:local can also be useful for checking version changes locally

.github/workflows/health_checks.yml
```diff
@@ -65,7 +65,10 @@ jobs:
       - uses: ./.github/actions/setup_node
       - uses: ./.github/actions/restore_build_cache
       - name: Publish packages locally
-        run: npm run vend
+        run: |
+          npm run start:npm-proxy
+          # keep git diff with version increment to make sure test projects resolve right version
+          npm run publish:local -- --keepGitDiff
```
0618, line 71: TIL: `--keepGitDiff`

0618 (approved):

PR #812, opened 2023-12-12
test: fix api check - check api before reseting git after publish

sobolk:
*Issue #, if available:*

**Description of changes:**

This PR is alternative to https://github.com/aws-amplify/amplify-backend/pull/811 .

Fixes https://github.com/aws-amplify/amplify-backend/actions/runs/7177124185/job/19543184647?pr=810 .

It refactors publish local to create programatic interface that can be used to execute logic between publish and resetting git.
Which is important for api check as it needs to retrieve correct version of published artifact.

Note: `.gitignore` entry was added because in GH action we have to checkout baseline and use working dir inside PR checkout dir and local publish needs clean tree to work. Otherwise we could not benefit from build cache (cache actions are not flexible enough to restore to different root :( )

**Out of scope**
It seems that throwing that `.gitignore` entry is going to help script the whole validation procedure, but I'd like to explore that separately. This PR is focused on fixing the bug.


.gitignore
```diff
@@ -15,3 +15,4 @@ verdaccio-logs.txt
 cdk.out/
 e2e-tests
 concurrent_workspace_script_cache.json
+api-validation-temp-dir/
```
sobolk, line 18: See PR description.

scripts/components/publish_runner.ts
0618, line 1: Are you using `components` as a new `helper`/`utils`? 😂

sobolk:
closing in favor of https://github.com/aws-amplify/amplify-backend/pull/811

scripts/components/publish_runner.ts
sobolk: `components` imply some structure there. `helper/utils` don't :P

PR #829, opened 2023-12-18
feat: Passwordless Sign Up

elorzafe:
Adds support for passwordless sign up

## Changes

- Creates APIGW endpoint + lambda to create a user without a password
- Updates custom-auth service to verify user was created by passwordless sign up and also verifies the user after successful sign in
- Adds custom attribute to keep track of a user created by passwordless sign up 

Increase unit tests coverage

## Checklist


packages/auth-construct/src/construct.ts
```diff
@@ -350,6 +350,9 @@ export class AmplifyAuth
         phoneNumber: DEFAULTS.IS_REQUIRED_ATTRIBUTE.phoneNumber(phoneEnabled),
         ...(props.userAttributes ? props.userAttributes : {}),
       },
+      customAttributes: {
+        passwordless_sign_up: new cognito.StringAttribute({ mutable: true }), // amplify-backend/packages/passwordless-auth-construct/src/constants.ts
```
Equartey, line 354: nit: remove comment

packages/passwordless-auth-construct/src/custom-auth/custom_auth_service.ts
```diff
@@ -151,32 +157,35 @@ export class CustomAuthService {
     // If the user is not found or if the attribute requested for challenge
     // delivery is not verified, return a fake successful response to prevent
     // user enumeration
```
Equartey, line 159: request: remove or update this comment

packages/passwordless-auth-construct/src/custom-auth/custom_auth_service.ts
```diff
@@ -151,32 +157,35 @@ export class CustomAuthService {
...
+    const validUser =
+      !event.request.userNotFound && (isVerified || isFirstSignInAttempt);
+
+    // If the user is found and if the attribute requested for challenge
+    // delivery is verified or if the user was created via passwordless sign up (first sign in attempt)
+    if (validUser) {
+      return challengeService.createChallenge(
+        { deliveryMedium, attributeName },
+        destination,
+        event
```
Equartey, line 169: nit/suggestion: IMO this function is easier to reason about when the happy case is at the bottom. Consider inversing the conditional and placing the user not found logic inside the if branch.

packages/passwordless-auth-construct/src/custom-auth/custom_auth_service.ts
```diff
@@ -280,31 +325,57 @@ export class CustomAuthService {
...
+    try {
+      const passwordlessConfiguration = JSON.parse(passwordlessSignUp);
+      if (
+        passwordlessConfiguration.allowSignInAttempt &&
+        passwordlessConfiguration.deliveryMedium === deliveryMedium
+      ) {
+        isFirstSignInAttempt = true;
+      }
+    } catch (_err) {
+      // best effort to parse passwordless_sign_up attribute if the user was created via passwordless sign up
```
Equartey, line 356: question: How are you planning on handling when the value can not be parsed? Might be good to return an error to to client.

packages/passwordless-auth-construct/src/sign-up/create_user_service.ts
Equartey, line 1: request: move this file to the services folder and add unit tests.

packages/passwordless-auth-construct/src/custom-auth/custom_auth_service.ts
```diff
@@ -397,4 +468,41 @@ export class CustomAuthService {
...
   }
+
+  /**
+   * Update user on Cognito UserPool by mark attribute as verified
+   * @param username - The username to be verify the attribute
+   * @param attributeName - The attribute that is going to be mark as verified
+   * @param region - The UserPool region
+   * @param userPoolId - The UserPool ID
+   */
+  private async markAsVerifiedAndDeletePasswordlessAttribute(
```
Equartey, line 479: request: 
1. move this function to `create_user_service.ts`
2.`CognitoIdentityProviderClient` can be a property passed to the class constructor. This should allow dependency injection and resuse for the create user logic.

packages/passwordless-auth-construct/src/sign-up/create_user_service.ts
Equartey: follow up on previous comment: make this a class that has `createUser` and `markAsVerifiedAndDeletePasswordlessAttribute`. Pass all service clients in as constructor params to follow Di and resue where you can.

packages/passwordless-auth-construct/src/sign-up/create_user_service.ts
```diff
@@ -0,0 +1,103 @@
...
+        body: 'Missing parameters',
+      };
+    }
+
+    // this is the format for Cognito UserPool API, eslint doesnt like it otherwise
+    // eslint-disable-next-line
+    const userAttributes: Array<{ Name: string; Value: string }> = [];
+    const passwordlessConfiguration: {
+      allowSignInAttempt: boolean;
+      deliveryMedium?: 'SMS' | 'EMAIL';
```
Equartey, line 39:
```suggestion
      deliveryMedium?: DeliveryMedium;
```
request: use the common enum

packages/passwordless-auth-construct/src/sign-up/create_user_service.ts
```diff
@@ -0,0 +1,103 @@
...
+      headers: {
+        'Access-Control-Allow-Origin': '*',
+      },
+    };
+  } catch (err) {
+    return {
+      statusCode: 400,
+      body: 'Invalid parameters',
+    };
+  }
```
Equartey, line 102: question: is this try/catch needed? Looks like the params are validated above and an error is already thrown if they don't exist.

packages/passwordless-auth-construct/src/sign-up/construct.ts
```diff
@@ -0,0 +1,99 @@
...
+          new PolicyStatement({
+            actions: ['cognito-idp:AdminCreateUser'],
+            resources: [userPool.userPoolArn],
+            effect: Effect.ALLOW,
+          }),
+        ],
+      })
+    );
+
+    verifyAuthChallengeResponse.role!.attachInlinePolicy(
```
Equartey, line 84: question: force unwrapping `verifyAuthChallengeResponse.role!` makes me a little nervous. If there is a chance this won't have a value, can we assert there is a value before this statement?

packages/passwordless-auth-construct/src/sign-up/create_user_service.ts
```diff
@@ -0,0 +1,103 @@
...
+      deliveryMedium: undefined,
+    };
+
+    if (params.email) {
+      userAttributes.push({
+        Name: 'email',
+        Value: params.email,
+      });
+      passwordlessConfiguration.allowSignInAttempt = true;
+      passwordlessConfiguration.deliveryMedium = 'EMAIL';
```
Equartey, line 51: request: these could also use the common enum values

packages/passwordless-auth-construct/src/sign-up/create_user_service.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    if (params.phone_number) {
+      userAttributes.push({
+        Name: 'phone_number',
+        Value: params.phone_number,
+      });
+
+      passwordlessConfiguration.allowSignInAttempt = true;
+      passwordlessConfiguration.deliveryMedium = 'SMS';
```
Equartey, line 61: request: common enum value

Equartey:
Can you also address the code quality comments?

packages/passwordless-auth-construct/src/custom-auth/custom_auth_service.ts
```diff
@@ -210,9 +216,49 @@ export class CustomAuthService {
...
+
+        const attributeVerified =
+          event.request.userAttributes[attributeName] === 'true';
+        // Only update verified attribute the first time
+        if (answerCorrect && !attributeVerified) {
+          const cognitoUserService = new CognitoUserService(
+            new CognitoIdentityProviderClient({ region: event.region })
+          );
+          logger.debug(`Updating user attribute to verified: ${attributeName}`);
+          await cognitoUserService.markAsVerifiedAndDeletePasswordlessAttribute(
```
Equartey, line 246: request: All dependencies should be defined in this [index](https://github.com/aws-amplify/amplify-backend/blob/feat/passwordless-auth/packages/passwordless-auth-construct/src/custom-auth/index.ts). This will need to become a constructor prop on `CustomAuthService`.

packages/passwordless-auth-construct/src/services/cognito_user_service.ts
```diff
@@ -0,0 +1,141 @@
...
+    params: MarkVerifiedAndDeletePasswordlessParams
+  ) {
+    const { username, attributeName, userPoolId, region } = params;
+    const attributeVerified = {
+      Name: attributeName,
+      Value: 'true',
+    };
+
+    const client =
+      this.cognitoClient || new CognitoIdentityProviderClient({ region });
```
Equartey, line 51: request: We've been asked to avoid default definitions for services inside the service, can you make this client required in this class and hoist the default to the `custom-auth/index.ts` file?

packages/passwordless-auth-construct/src/sign-up/construct.ts
```diff
@@ -0,0 +1,103 @@
...
+          statements: [
+            new PolicyStatement({
+              actions: ['cognito-idp:AdminCreateUser'],
+              resources: [userPool.userPoolArn],
+              effect: Effect.ALLOW,
+            }),
+          ],
+        })
+      );
+    }
```
Equartey, line 84:
```suggestion
    createUserFunction.addToRolePolicy(
      new PolicyStatement({
        actions: ['cognito-idp:AdminCreateUser'],
        resources: [userPool.userPoolArn],
        effect: Effect.ALLOW,
      }),
    );
```
suggestion: This needs to be tested, but I wonder if you can add the policy like this?  Could reuse for the one below too.

packages/passwordless-auth-construct/src/sign-up/create_user_service.ts
Equartey, line 1: question: Looks like this file is a duplicate. Can it be removed?

packages/passwordless-auth-construct/src/services/cognito_user_service.test.ts
```diff
@@ -0,0 +1,221 @@
...
+            Name: 'custom:passwordless_sign_up',
+            Value: '{"allowSignInAttempt":true,"deliveryMedium":"EMAIL"}',
+          },
+        ],
+        UserPoolId: 'userPoolId',
+        Username: 'username',
+      };
+
+      const actualCreateUserCommand = sendMock.mock.calls[0].arguments[0]
+        .input as unknown as AdminCreateUserCommandInput;
```
Equartey, line 65:
```suggestion
        .input as AdminCreateUserCommandInput;
```
nit: I think unknown can be omitted

Equartey: Same for the other occurrences in this file

packages/passwordless-auth-construct/src/services/cognito_user_service.test.ts
```diff
@@ -0,0 +1,221 @@
...
+
+      strictEqual(sendMock.mock.callCount(), 1);
+      deepStrictEqual(actualCreateUserCommand, expectedAttributes);
+    });
+
+    void it('should return successful for user created with email', async () => {
```
Equartey, line 71:
```suggestion
    void it('should return successful for user created with phone', async () => {
```

Equartey (commented):
Looks great, we're so close!

packages/passwordless-auth-construct/src/custom-auth/custom_auth_service.ts
Equartey: after re-reading this snippet, looks like a no-op is sufficient.

packages/passwordless-auth-construct/src/services/cognito_user_service.ts
elorzafe: The problem is setting the region on the client, i couldnt see other examples that were using the region

packages/passwordless-auth-construct/src/sign-up/construct.ts
elorzafe: I will see if that works

packages/passwordless-auth-construct/src/sign-up/create_user_service.ts
elorzafe: This is the lambda handler for the API Gateway

packages/passwordless-auth-construct/src/sign-up/create_user_service.ts
elorzafe: this `try/catch` is for JSON.parse error

packages/passwordless-auth-construct/src/sign-up/construct.ts
elorzafe: now I remember that doesnt work because of cyclic dependencies

elorzafe: A default role is created according to cdk docs 
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda-readme.html#execution-role

I will change it to use the `!`

packages/passwordless-auth-construct/src/custom-auth/custom_auth_service.ts
```diff
@@ -210,9 +218,44 @@ export class CustomAuthService {
...
+            username: event.userName,
+            attributeName,
+            userPoolId: event.userPoolId,
+            region: event.region,
+          });
+        }
+      }
+    } catch (_err) {
+      // best effort to parse passwordless_sign_up attribute
+    }
```
Equartey:
```suggestion
    let passwordlessConfiguration: Record<string, string> | undefined;
    try {
      passwordlessConfiguration =
        event.request.userAttributes[PASSWORDLESS_SIGN_UP_ATTR_NAME] &&
        JSON.parse(
          event.request.userAttributes[PASSWORDLESS_SIGN_UP_ATTR_NAME]
        );
    } catch (_err) {
      // best effort to parse passwordless_sign_up attribute
    }

    if (!passwordlessConfiguration?.allowSignInAttempt) {
      return verifyResult;
    }
    const attributeName =
      passwordlessConfiguration.deliveryMedium === 'SMS'
        ? 'phone_number_verified'
        : 'email_verified';

    const attributeVerified =
      event.request.userAttributes[attributeName] === 'true';
    // Only update verified attribute the first time
    if (answerCorrect && !attributeVerified) {
      logger.debug(`Updating user attribute to verified: ${attributeName}`);
      await this.userService.markAsVerifiedAndDeletePasswordlessAttribute({
        username: event.userName,
        attributeName,
        userPoolId: event.userPoolId,
        region: event.region,
      });
    }
```

Equartey (approved):
🚀

packages/passwordless-auth-construct/API.md
```diff
@@ -38,6 +40,7 @@ export type OtpAuthOptions = {
 export type PasswordlessAuthProps = {
     magicLink?: MagicLinkAuthOptions;
     otp?: OtpAuthOptions;
+    signUpNoPassword?: boolean;
```
edwardfoyle, line 43:
```suggestion
    passwordlessSignUp?: boolean;

```

.eslint_dictionary.json
```diff
@@ -19,6 +20,7 @@
   "codegen",
   "cognito",
   "commonjs",
+  "Cors",
```
edwardfoyle, line 23: nit:
```suggestion
  "cors",
```

.eslint_dictionary.json
```diff
@@ -87,6 +89,7 @@
   "pathname",
   "pipelined",
   "posix",
+  "Preflight",
```
edwardfoyle, line 92: also nit:
```suggestion
  "preflight",
```

packages/passwordless-auth-construct/src/construct.ts
```diff
@@ -151,11 +159,21 @@ export class AmplifyPasswordlessAuth extends Construct {
...
-          verifyAuthChallengeResponse,
+          verifyAuthChallengeResponse: this.verifyAuthChallengeResponse,
         },
         props.magicLink
       );
     }
+
+    // Configure Sign Up without password
+    if (props.signUpNoPassword) {
+      new AmplifySignUpPasswordless(
```
edwardfoyle, line 170: along with the prop rename, rename these related classes and names to `AmplifyPasswordlessSignUp`

packages/passwordless-auth-construct/src/services/cognito_user_service.ts
```diff
@@ -0,0 +1,128 @@
+// ../amplify-backend/packages/passwordless-auth-construct/lib/sign-up/createUserService.js
```
edwardfoyle, line 1: nit: remove

packages/passwordless-auth-construct/src/custom-auth/custom_auth_service.ts
```diff
@@ -210,9 +219,48 @@ export class CustomAuthService {
 
     const method = this.validateSignInMethod(signInMethod);
 
-    return this.challengeServiceFactory
+    const verifyResult = this.challengeServiceFactory
       .getService(method)
       .verifyChallenge(event);
+
+    const answerCorrect = (await verifyResult).response.answerCorrect;
+    logger.debug(`Answer is correct: ${answerCorrect ? 'true' : 'false'}`);
```
edwardfoyle, line 227:
```suggestion
    logger.debug(`Challenge response is ${answerCorrect ? 'correct' : 'incorrect'}`);
```

packages/passwordless-auth-construct/src/custom-auth/custom_auth_service.ts
```diff
@@ -210,9 +219,48 @@ export class CustomAuthService {
 
     const method = this.validateSignInMethod(signInMethod);
 
-    return this.challengeServiceFactory
+    const verifyResult = this.challengeServiceFactory
```
edwardfoyle, line 222: await this and remove the await on line 226

packages/passwordless-auth-construct/src/custom-auth/custom_auth_service.ts
```diff
@@ -210,9 +219,48 @@ export class CustomAuthService {
...
+
+    let passwordlessConfiguration: Record<string, string> | undefined;
+
+    try {
+      passwordlessConfiguration =
+        event.request.userAttributes[PASSWORDLESS_SIGN_UP_ATTR_NAME] &&
+        JSON.parse(
+          event.request.userAttributes[PASSWORDLESS_SIGN_UP_ATTR_NAME]
+        );
+    } catch (_err) {
```
edwardfoyle, line 237: There are a few places where this pattern is used. Would be better to make a utility function like `tryParseJson` that returns undefined instead of throwing if a string is not parseable.

packages/passwordless-auth-construct/src/services/cognito_user_service.ts
```diff
@@ -0,0 +1,128 @@
...
+    });
+
+    await this.cognitoClient.send(deleteAttrCommand);
+  }
+
+  /**
+   * Create User
+   * @param params - Create User parameter
+   * @param params.username - The username to be verify the attribute
+   * @param params.email - The attribute that is going to used as email (optional)
```
edwardfoyle, line 65:
```suggestion
   * @param params.email - The attribute that is going to be used as an email (optional)
```

packages/passwordless-auth-construct/src/services/cognito_user_service.ts
```diff
@@ -0,0 +1,128 @@
...
+      UserAttributeNames: [PASSWORDLESS_SIGN_UP_ATTR_NAME],
+    });
+
+    await this.cognitoClient.send(deleteAttrCommand);
+  }
+
+  /**
+   * Create User
+   * @param params - Create User parameter
+   * @param params.username - The username to be verify the attribute
```
edwardfoyle, line 64: This sentence doesn't make sense

packages/passwordless-auth-construct/src/services/cognito_user_service.ts
```diff
@@ -0,0 +1,128 @@
...
+   * @param params.email - The attribute that is going to used as email (optional)
+   * @param params.phone_number - The attribute that is going to used as phone number (optional)
```
edwardfoyle, line 66:
```suggestion
   * @param params.phone_number - The attribute that is going to be used as a phone number (optional)
```

packages/passwordless-auth-construct/src/services/cognito_user_service.ts
```diff
@@ -0,0 +1,128 @@
...
+      UserAttributes: userAttributes,
+      MessageAction: 'SUPPRESS',
+    });
+
+    try {
+      await this.cognitoClient.send(command);
+      return;
+    } catch (err) {
+      logger.debug(err);
+      throw new Error('User already exists');
```
edwardfoyle, line 125: This could be any error, not necessarily user already exists. Maybe we should just log and rethrow the underlying error?

packages/passwordless-auth-construct/src/services/cognito_user_service.ts
```diff
@@ -0,0 +1,128 @@
...
+    const command = new AdminCreateUserCommand({
+      Username: username,
+      UserPoolId: userPoolId,
+      UserAttributes: userAttributes,
+      MessageAction: 'SUPPRESS',
+    });
+
+    try {
+      await this.cognitoClient.send(command);
+      return;
```
edwardfoyle, line 122: nit: unnecessary return

packages/passwordless-auth-construct/src/services/cognito_user_service.ts
```diff
@@ -0,0 +1,128 @@
...
+      userAttributes.push({
+        Name: 'phone_number',
+        Value: phone_number,
+      });
+
+      passwordlessConfiguration.allowSignInAttempt = true;
+      passwordlessConfiguration.deliveryMedium = DeliveryMediumType.SMS;
+    }
+
+    // TODO: what happens if both are included, which one should be valid?
```
edwardfoyle, line 104: does this need to be addressed before merging?

packages/passwordless-auth-construct/src/services/cognito_user_service.ts
```diff
@@ -0,0 +1,128 @@
...
+      userAttributes.push({
+        Name: PASSWORDLESS_SIGN_UP_ATTR_NAME,
+        Value: JSON.stringify(passwordlessConfiguration),
+      });
+    }
+
+    const command = new AdminCreateUserCommand({
+      Username: username,
+      UserPoolId: userPoolId,
+      UserAttributes: userAttributes,
```
edwardfoyle, line 116: do we need to account for additional attributes beyond just phone and email here?

packages/passwordless-auth-construct/src/sign-up/construct.ts
```diff
@@ -0,0 +1,103 @@
...
+  /**
+   * Create a new SignUp Passwordless construct
+   * verifyAuthChallengeResponse function is needed in order to attach policy
+   * userPool is needed to identity the resource and create the policy for verifyAuthChallengeResponse
+   */
+  constructor(
+    scope: Construct,
+    id: string,
+    verifyAuthChallengeResponse: NodejsFunction,
+    userPool: IUserPool
```
edwardfoyle, line 31: To be CDK idiomatic, these two props should be rolled into one props object. Also, the `verifyAuthChallengeResponse` type should be `IFunction` not `NodejsFunction`

packages/passwordless-auth-construct/src/sign-up/construct.ts
```diff
@@ -0,0 +1,103 @@
...
+            resources: [userPool.userPoolArn],
+            effect: Effect.ALLOW,
+          }),
+        ],
+      })
+    );
+
+    // According to cdk docs there always is a default role https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda-readme.html#execution-role
+    // eslint-disable-next-line  @typescript-eslint/no-non-null-assertion
+    verifyAuthChallengeResponse.role!.attachInlinePolicy(
```
edwardfoyle, line 88: If we only need the function execution role, that should be what's passed into the ctor rather than the whole function

packages/passwordless-auth-construct/src/sign-up/construct.ts
```diff
@@ -0,0 +1,103 @@
...
+    id: string,
+    verifyAuthChallengeResponse: NodejsFunction,
+    userPool: IUserPool
+  ) {
+    super(scope, id);
+
+    // default memory allocation for lambda functions
+    const defaultMemorySize = 128;
+
+    const commonOptions: NodejsFunctionProps = {
```
edwardfoyle, line 38: looks like this is only used once? Can it just be added inline to the function props below?

packages/passwordless-auth-construct/src/sign-up/construct.ts
```diff
@@ -0,0 +1,103 @@
...
+      entry: path.join(dirname, '.', 'create_user_handler.js'),
```
edwardfoyle, line 39:
```suggestion
      entry: path.join(dirname, 'create_user_handler.js'),
```

packages/passwordless-auth-construct/src/sign-up/construct.ts
```diff
@@ -0,0 +1,103 @@
...
+  constructor(
+    scope: Construct,
+    id: string,
+    verifyAuthChallengeResponse: NodejsFunction,
+    userPool: IUserPool
+  ) {
+    super(scope, id);
+
+    // default memory allocation for lambda functions
+    const defaultMemorySize = 128;
```
edwardfoyle, line 36: This doesn't need to be set explicitly

packages/passwordless-auth-construct/src/sign-up/construct.ts
```diff
@@ -0,0 +1,103 @@
...
+    const putCreateUserIntegration = new apigateway.LambdaIntegration(
+      createUserFunction,
+      {
+        requestTemplates: { 'application/json': '{ "statusCode": "200" }' },
+      }
+    );
+
+    api.root.addMethod('PUT', putCreateUserIntegration);
+
+    new CfnOutput(this, 'apiUrl', { value: api.url });
```
edwardfoyle, line 70: if this URL need to make it's way back to the client config file, we'll need to use [`addBackendOutputEntry`](https://github.com/aws-amplify/amplify-backend/blob/main/packages/backend-output-storage/src/stack_metadata_output_storage_strategy.ts#L25). You can see how this is used in the data construct [here](https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts#L262)

We could also mark this as a TODO for later

packages/passwordless-auth-construct/src/sign-up/create_user_handler.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    return {
+      statusCode: 200,
+      body: JSON.stringify({}),
+      headers: {
+        'Access-Control-Allow-Origin': '*',
+      },
+    };
+  } catch (err) {
```
edwardfoyle, line 54: rather than wrap the whole function in a try catch, there should be upfront request validation that early returns

packages/passwordless-auth-construct/src/sign-up/create_user_handler.ts
```diff
@@ -0,0 +1,60 @@
...
+    return {
+      statusCode: 400,
+      body: 'Invalid parameters',
```
edwardfoyle, line 57: We should provide a more useful error message (or at the very least log something more useful)

packages/passwordless-auth-construct/src/sign-up/create_user_handler.ts
```diff
@@ -0,0 +1,60 @@
...
+      await cognitoCreateUserService.createUser({
+        username: params.username,
+        userPoolId: params.userPoolId,
+        email: params.email,
+        phone_number: params.phone_number,
+      });
+    } catch (err) {
+      return {
+        statusCode: 400,
+        body: (err as Error).message,
```
edwardfoyle, line 40: Should this be JSON stringified?

packages/passwordless-auth-construct/src/sign-up/create_user_handler.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    if (
+      !params ||
+      (!params.email && !params.phone_number) ||
+      !params.userPoolId ||
+      !params.username
+    ) {
+      return {
+        statusCode: 400,
+        body: 'Missing parameters',
```
edwardfoyle, line 23: We should say what parameters were missing

packages/passwordless-auth-construct/src/sign-up/create_user_handler.ts
```diff
@@ -0,0 +1,60 @@
...
+      };
+    }
+
+    try {
+      const cognitoCreateUserService = new CognitoUserService(
+        new CognitoIdentityProviderClient()
+      );
```
edwardfoyle, line 30: This should be initialized outside of the handler

packages/passwordless-auth-construct/src/sign-up/create_user_handler.ts
```diff
@@ -0,0 +1,60 @@
...
+      await cognitoCreateUserService.createUser({
+        username: params.username,
+        userPoolId: params.userPoolId,
+        email: params.email,
+        phone_number: params.phone_number,
+      });
+    } catch (err) {
+      return {
+        statusCode: 400,
```
edwardfoyle, line 39: This isn't necessarily a 400. It could be a 5xx or a 4xx that indicates a bug in our code rather than a customer 4xx

packages/passwordless-auth-construct/src/types.ts
```diff
@@ -342,3 +345,38 @@ export type KmsConfig = {
 export enum PasswordlessErrorCodes {
   CODE_MISMATCH_EXCEPTION = 'CodeMismatchException',
 }
+
+export type CreateUserParams = {
+  userPoolId: string;
+  username: string;
+  // this is the format for Cognito UserPool API, eslint not happy otherwise
+  // eslint-disable-next-line
```
edwardfoyle, line 353: Our construct doesn't have to use the same prop names as the Cognito UserPool API. We can translate between `phoneNumber` and `phone_number` when making service calls

packages/passwordless-auth-construct/src/types.ts
```diff
@@ -342,3 +345,38 @@ export type KmsConfig = {
...
+  userPoolId: string;
+};
+
+export type UserService = {
+  /**
+   * Create User
+   * @param username - The username to be verify the attribute
+   * @param email - The attribute that is going to used as email (optional)
+   * @param phone_number - The attribute that is going to used as phone number (optional)
+   * @param userPoolId - The UserPool ID
```
edwardfoyle, line 370: This JSDoc should be moved to CreateUserParams

packages/passwordless-auth-construct/src/types.ts
```diff
@@ -342,3 +345,38 @@ export type KmsConfig = {
...
+   */
+  createUser: (params: CreateUserParams) => Promise<void>;
+  /**
+   * Update user and mark attribute as verified
+   * @param username - The username to be verify the attribute
+   * @param attributeName - The attribute that is going to be mark as verified
+   * @param userPoolId - The UserPool ID
```
edwardfoyle, line 377: Same here, move to MarkVerifiedAndDeletePasswordlessParams

edwardfoyle (commented):
Left mostly small comments but a few that are more functional.

Also just leaving a note for posterity that we should think more about the implications of exposing a public API for this. Some concerns are the ability to switch domains later and throttling / abuse controls on the API.

packages/passwordless-auth-construct/src/services/cognito_user_service.ts
elorzafe: the idea is to avoid giving client to much info about the users

packages/passwordless-auth-construct/src/services/cognito_user_service.ts
elorzafe: that wasnt part of the original requirements, when the user sign in the first time will have to provide that info by responding challenges

packages/passwordless-auth-construct/src/services/cognito_user_service.ts
elorzafe: We can discuss that with the team after merge

packages/passwordless-auth-construct/src/custom-auth/custom_auth_service.ts
elorzafe: Can we discuss that in a follow up PR, just to verify if other packages needs this as well

packages/passwordless-auth-construct/src/sign-up/construct.ts
```diff
@@ -0,0 +1,88 @@
...
+        requestTemplates: { 'application/json': '{ "statusCode": "200" }' },
+      }
+    );
+
+    api.root.addMethod('PUT', putCreateUserIntegration);
+
+    // TODO: expose this on the external stack
+    new CfnOutput(this, 'apiUrl', { value: api.url });
+
+    // eslint-disable-next-line  @typescript-eslint/no-non-null-assertion
```
sobolk, line 60: please add comment explaining why linter rule is disabled OR find different syntax.

packages/passwordless-auth-construct/src/sign-up/construct.ts
sobolk: Please also create GH issue for this and link in the comment. This should be release blocker.

packages/passwordless-auth-construct/src/sign-up/create_user_handler.ts
```diff
@@ -0,0 +1,90 @@
...
+  } = {};
+
+  try {
+    params = event && event.body && JSON.parse(event.body);
+  } catch (err) {
+    logger.debug(err);
+
+    return {
+      statusCode: 400,
+      body: 'Invalid parameters',
```
sobolk, line 32: should we include stack trace or some other indicator for which parameter is invalid? This is going to be hard to debug otherwise.

should we also assert content type of incoming request?

sobolk (changes requested):

packages/auth-construct/src/construct.ts
```diff
@@ -350,6 +350,9 @@ export class AmplifyAuth
         phoneNumber: DEFAULTS.IS_REQUIRED_ATTRIBUTE.phoneNumber(phoneEnabled),
         ...(props.userAttributes ? props.userAttributes : {}),
       },
+      customAttributes: {
+        passwordless_sign_up: new cognito.StringAttribute({ mutable: true }),
+      },
```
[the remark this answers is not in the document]
elorzafe: The problem with user attributes is if the customer disables `signUp` it will be change that will require to create a new user pool.

packages/passwordless-auth-construct/src/sign-up/create_user_handler.ts
elorzafe: I will change to `Invalid JSON`, the following lines has the individual missed parameters

packages/auth-construct/src/construct.ts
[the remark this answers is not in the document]
sobolk: This is going to be an interesting problem to solve then. Since this is "standalone auth construct" code, i.e. direct cdk users who only use "auth construct" and not opt into passwordless ("passwordless-auth-construct") shouldn't see that.

This might be a bit tricky since auth construct doesn't seem to have a away to define custom attributes as far as I can tell. Therefore I suggest to add TODO and create release blocker github issue to figure this out later before we merge this feature to main.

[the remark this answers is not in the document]
sobolk: As for immutability , we have similar problem with "domain" in this PR (reverted for now due to test failures, but it'll come back) https://github.com/aws-amplify/amplify-backend/pull/756/files#diff-0ae889ed8a9821b6bcb870a09d9f9570ae785235d7b7269524307f4f7bf4d319 .
(see comments around `domainPrefix?: string`).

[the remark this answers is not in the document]
sobolk: Also have we considered storing this in some other place like DDB table?
(passwordless auth already defines some ddb tables for this kind of data)

elorzafe:
> Left mostly small comments but a few that are more functional.
> 
> Also just leaving a note for posterity that we should think more about the implications of exposing a public API for this. Some concerns are the ability to switch domains later and throttling / abuse controls on the API.

I created this issue to double check your concern across the team
https://github.com/aws-amplify/amplify-backend/issues/913

sobolk (approved):

PR #836, opened 2023-12-19
Initial implementation for function secret access

rtpascual:
## Problem

`defineFunction` does not have access to secrets created with `npx amplify sandbox secret`.

## Changes

- Add support for passing in secrets through the `environment` parameter of `defineFunction`.
- Move parameter path methods from `backend-secret` to `platform-core` to be used across multiple packages.
- Add bundling options for function to add a code snippet to resolve secrets for users.

## Validation

Used a test project with local changes to deploy a function with secrets using `defineFunction` and verified the following:
- Environment variables for the placeholder text and secret path are populating in the Lambda console as expected.
- Test executing the function to verify secrets are resolved to their expected values as seen in the Parameter Store.

## Checklist

### DIFF
--- a/packages/backend-function/src/factory.ts
+++ b/packages/backend-function/src/factory.ts
@@ -8,11 +8,12 @@ import {
   ResourceProvider,
 } from '@aws-amplify/plugin-types';
 import { Construct } from 'constructs';
-import { NodejsFunction } from 'aws-cdk-lib/aws-lambda-nodejs';
+import { NodejsFunction, OutputFormat } from 'aws-cdk-lib/aws-lambda-nodejs';
 import * as path from 'path';
 import { getCallerDirectory } from './get_caller_directory.js';
-import { Duration } from 'aws-cdk-lib';
+import { Duration, Stack } from 'aws-cdk-lib';
 import { Runtime } from 'aws-cdk-lib/aws-lambda';
+import * as iam from 'aws-cdk-lib/aws-iam';
 
 /**
  * Entry point for defining a function in the Amplify ecosystem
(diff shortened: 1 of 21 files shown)


packages/backend-function/src/factory.ts
```diff
@@ -209,14 +222,43 @@ class AmplifyFunction
...
     super(scope, id);
+    const functionLambda = new NodejsFunction(scope, `${id}-lambda`, {
+      entry: props.entry,
+      environment: props.environment as { [key: string]: string }, // for some reason TS can't figure out that this is the same as Record<string, string>
+      timeout: Duration.seconds(props.timeoutSeconds),
+      memorySize: props.memoryMB,
+      runtime: nodeVersionMap[props.runtime],
+      bundling: {
+        // Added '\' to the end of each line for readability here
+        // eslint-disable-next-line spellcheck/spell-checker
```
rtpascual, line 233: eslint was complaining about `ssm` here, let me know if this should be added to the dictionary.

packages/backend-function/src/factory.ts
```diff
@@ -209,14 +222,43 @@ class AmplifyFunction
...
+        banner: `import { SSM } from '@aws-sdk/client-ssm'; const client = new SSM();\
+        const envArray = ${JSON.stringify(Object.keys(props.environment))};\
```
edwardfoyle, line 235: Once this snippet is moved to it's own file, we may need to store this list as a separate env var like `SECRET_NAME_LIST=["VAR_1", "VAR_2", 'etc']` that we can read to determine which secrets to look up.

Also, not every environment key is a secret.

edwardfoyle (commented):
Looking good, just left some small comments.

Only somewhat big thing is I think this change warrants some e2e coverage. I'd recommend adding it to the existing `data_storage_auth_with_triggers.ts` test which already sets up some secrets to use for auth. You'd need to update the "[specialTestFunction](https://github.com/aws-amplify/amplify-backend/blob/main/packages/integration-tests/src/test-projects/data-storage-auth-with-triggers-ts/amplify/function.ts#L3)" definition to reference one of those secrets and then verify that it was able to fetch it at runtime. You could verify by adding the resolved secret value to the response payload [here](https://github.com/aws-amplify/amplify-backend/blob/main/packages/integration-tests/src/test-projects/data-storage-auth-with-triggers-ts/amplify/func-src/response_generator.ts#L13) and then verify you got the right thing [here](https://github.com/aws-amplify/amplify-backend/blob/main/packages/integration-tests/src/test-projects/data-storage-auth-with-triggers-ts/amplify/func-src/response_generator.ts#L13)

packages/backend-function/src/factory.ts
```diff
@@ -209,14 +222,43 @@ class AmplifyFunction
...
+    });
+
+    functionLambda.grantPrincipal.addToPrincipalPolicy(
+      new iam.PolicyStatement({
+        effect: iam.Effect.ALLOW,
+        actions: ['ssm:GetParameters'],
+        resources: [
+          `arn:aws:ssm:${Stack.of(scope).region}:${
+            Stack.of(scope).account
+          }:parameter/*`,
```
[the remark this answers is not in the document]
rtpascual: I was trying to be specific by copying the resources [here](https://github.com/aws-amplify/amplify-backend/blob/main/packages/backend/src/engine/backend-secret/backend_secret_fetcher_provider_factory.ts#L48), this is the `arn` provided by the error message when the lambda didn't have the permissions to access a parameter.

edwardfoyle:
One other thing that we handle with auth secrets and need to handle here as well is falling back to "shared" secrets if a branch-specific secret doesn't exist. You can see where we do this in the secret resolver custom resource [here](https://github.com/aws-amplify/amplify-backend/blob/main/packages/backend/src/engine/backend-secret/lambda/backend_secret_fetcher.ts#L77-L78)

I'm thinking we should figure out some way to refactor that code such that it can be used in the custom resource lambda and also injected into the banner of customer lambdas.

packages/backend-function/src/factory.ts
rtpascual: Gotcha. And yeah not every environment key is a secret, `getParameters` will list them as `InvalidParameters` in the response while the env vars that are secret paths will have their results in `Parameters`.

packages/backend-function/src/factory.ts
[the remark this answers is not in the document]
edwardfoyle: The resource scope that you linked to is better (scoping to the specific namespace and name). If we can scope to the specific parameters here we should. The other location is a little different but you could argue we should scope that better as well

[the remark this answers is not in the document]
rtpascual: Alright I'll try to narrow down the scope a bit more

packages/backend-function/src/factory.ts
edwardfoyle: We shouldn't rely on that as an implicit signal. While unlikely, someone could have some other parameter path stored in an env var that shouldn't be included here. We should only call SSM for the secret paths that we care about.

rtpascual:
Closing this PR for #845 in order to change the source branch for e2e test checks.

rtpascual:
Closing this PR for #845 in order to change the source branch for e2e test checks.

PR #854, opened 2023-12-29
add --config-out-dir arg to pipeline-deploy

johnpc:
## Problem

`amplify pipeline-deploy` deploys cdk resources in CI/CD environments and runs `amplify generate config` automatically. However, `pipeline-deploy` does not currently allow an `--out-dir` to be specified, so the config is always written to the root of the project which is not always desired.

N/A

## Changes

This pr adds the `--out-dir` argument to the `pipeline-deploy` command.

## Validation

* unit tests
* performed deployment with `--out-dir` set
* performed deployment without `--out-dir` set

## Checklist

### DIFF
--- a/.changeset/moody-grapes-move.md
+++ b/.changeset/moody-grapes-move.md
@@ -2,4 +2,4 @@
 '@aws-amplify/backend-cli': patch
 ---
 
-add --out-dir to pipeline-deploy command
+add --config-out-dir to pipeline-deploy command
--- a/.changeset/moody-grapes-move.md
+++ b/.changeset/moody-grapes-move.md
@@ -1,0 +1,5 @@
+---
+'@aws-amplify/backend-cli': patch
+---
+
+add --out-dir to pipeline-deploy command
--- a/packages/cli/src/commands/pipeline-deploy/pipeline_deploy_command.test.ts
+++ b/packages/cli/src/commands/pipeline-deploy/pipeline_deploy_command.test.ts
@@ -92,7 +92,7 @@ void describe('deploy command', () => {
       () => Promise.resolve()
     );
     await getCommandRunner(true).runCommand(
-      'pipeline-deploy --app-id abc --branch test-branch --out-dir src'
+      'pipeline-deploy --app-id abc --branch test-branch --config-out-dir src'
     );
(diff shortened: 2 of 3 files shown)


packages/cli/src/commands/pipeline-deploy/pipeline_deploy_command.ts
```diff
@@ -79,6 +83,12 @@ export class PipelineDeployCommand
         type: 'string',
         array: false,
       })
+      .option('out-dir', {
+        describe:
+          'A path to directory where config is written. If not provided defaults to current process working directory.',
+        type: 'string',
+        array: false,
+      })
```
sobolk, line 91: I think we should be consistent with `sandbox` command here and use this more explicit naming https://github.com/aws-amplify/amplify-backend/blob/6a2f9aa8fe81bf48ac725470a1bd64622966da46/packages/cli/src/commands/sandbox/sandbox_command.ts#L143-L156 
The shorter naming is fine in `generate config` as it's easier to reason what that arg is there. But apparently there was some in `sandbox` hence longer name there.
And also consider adding second parameter - format as well.

Notice that we didn't find a good way to reuse parameters across commands (mixins?), so I recommend to just copy from sandbox.

.changeset/moody-grapes-move.md
```diff
@@ -0,0 +1,5 @@
+---
+'@aws-amplify/backend-cli': patch
+---
+
+add --out-dir to pipeline-deploy command
```
sobolk, line 5:
```suggestion
add --config-out-dir to pipeline-deploy command
```

packages/cli/src/commands/pipeline-deploy/pipeline_deploy_command.test.ts
```diff
@@ -84,4 +84,35 @@ void describe('deploy command', () => {
     ]);
     assert.equal(generateClientConfigMock.mock.callCount(), 1);
   });
+
+  void it('allows --out-dir argument', async () => {
```
sobolk, line 88:
```suggestion
  void it('allows --config-out-dir argument', async () => {
```

sobolk (approved):

sobolk (approved):

edwardfoyle (approved):

PR #901, opened 2024-01-10
test: add deploy test

0618:
## Problem

Move https://github.com/aws-amplify/amplify-backend/pull/891 to here because IAM role can't configure credentials on a test from fork.

- add `e2e-flow` test, which is a copy of `test-e2e/create-amplify.test`
- add `deploy` test to `create-amplify` script
- [x]  make the deploy test work

## Changes

## Checklist


packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
```diff
@@ -0,0 +1,258 @@
...
+      'package.json',
+      'tsconfig.json',
+    ];
+
+    assert.deepStrictEqual(
+      files.sort(),
+      expectedAmplifyFiles.map((suffix) => path.join(pathPrefix, suffix))
+    );
+
+    if (packageManager === 'yarn-modern') {
```
0618, line 166: This part can't be moved to `setupPackageManager`, because has to be done after `create-amplify`

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
```diff
@@ -0,0 +1,258 @@
...
+          }
+        );
+
+        await execa('node', ['--version'], {
+          cwd: tempDir,
+        });
+      }
+    }
+
+    if (packageManager === 'pnpm') {
```
0618, line 233: This can be removed since it doesn't fix `pnpm` failure on windows

0618: The `pnpm` failure is because it doesn't suppor nested node_modules. [ref](https://pnpm.io/faq)

packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -126,7 +127,7 @@ export class CDKDeployer implements BackendDeployer {
     }
     try {
       await this.executeChildProcess(
-        'npx',
+        this.packageManager === 'npm' ? 'npx' : this.packageManager,
```
sobolk, line 130: Can we do this in ctor and save result in private prop, so that we eliminate clones of this logic?

packages/backend-deployer/src/cdk_deployer_singleton_factory.ts
```diff
@@ -43,9 +43,14 @@ export class BackendDeployerFactory {
    */
   static getInstance = (): BackendDeployer => {
     if (!BackendDeployerFactory.instance) {
+      const { PACKAGE_MANAGER_EXECUTABLE = 'npm' } = process.env;
+      const packageManager = PACKAGE_MANAGER_EXECUTABLE.startsWith('yarn')
+        ? 'yarn'
+        : PACKAGE_MANAGER_EXECUTABLE;
```
sobolk, line 49: I'm assuming this is temporary solution but I think we should:
1. Move logic that reads `process.env` to "cli like" components (backend-cli, sandbox)
2. Once we get test passing we'll need to work in backend-cli to infer package manager kind and make this information flow into deployer.

This could be as part of this PR or follow up.

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
```diff
@@ -0,0 +1,258 @@
...
+      'config',
+      'set',
+      'registry',
+      'http://localhost:4873',
+    ]);
+    await execa(packageManager, ['config', 'get', 'registry']);
+  }
+
+  return { packageManagerExecutable, packageManager };
+};
```
sobolk, line 93: Couple of things:
1. given the volume I suggest to extract this to separate file.
2. it seems that `PACKAGE_MANAGER_EXECUTABLE` env var should really be `PACKAGE_MANAGER` 
3. this code seems to look like some sort of mapping, i.e. it maps `PACKAGE_MANAGER_EXECUTABLE/PACKAGE_MANAGER` env var into certain actions that are relevant for given package manager. and the way this code will likely evolve is that we'll be adding more varieties. So instead of scattering if/else conditions around I'd just group logice per package manager kind. Something like this:

```
export const setupPackageManager = async (...) => {
   const packageManager = (process.env.PACKAGE_MANAGER_EXECUTABLE ??
    'npm') as PackageManager;
   switch(packageManager) {
      case 'npm':
          packageManagerExecutable = 'npx';
          await initializeNpm();
          break;
      case 'yarn-modern:
          packageManagerExecutable = 'yarn';
          await initializeYarnModern();
          break;
      case 'yarn-classic:
          packageManagerExecutable = 'yarn';
          await initializeYarnClassic();
          break;
   }
  
}

```

This will make future editors of this function understand better what they have to do to add support for another package manager.

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
```diff
@@ -0,0 +1,258 @@
...
+    await execa('npm', ['run', 'clean:npm-proxy'], { stdio: 'inherit' });
+    await execa('npm', ['run', 'vend'], { stdio: 'inherit' });
+
+    cfnClient = new CloudFormationClient(e2eToolingClientConfig);
+    testBranch = await amplifyAppPool.createTestBranch();
+    branchBackendIdentifier = {
+      namespace: testBranch.appId,
+      name: testBranch.branchName,
+      type: 'branch',
+    };
```
sobolk, line 116: this works because we have one test here, but we should move this to `beforeEach` so that future author adding another test has easier job.

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
```diff
@@ -0,0 +1,258 @@
...
+  });
+
+  after(async () => {
+    await cfnClient.send(
+      new DeleteStackCommand({
+        StackName: BackendIdentifierConversions.toStackName(
+          branchBackendIdentifier
+        ),
+      })
+    );
```
sobolk, line 126: if we do `beforeAll` from upper comment we should move this piece to `afterEach`

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
```diff
@@ -0,0 +1,258 @@
...
+
+void describe('installs expected packages and scaffolds expected files', async () => {
+  let branchBackendIdentifier: BackendIdentifier;
+  let testBranch: TestBranch;
+  let cfnClient: CloudFormationClient;
+  const tempDir = await fsp.mkdtemp(
+    path.join(os.tmpdir(), 'test-create-amplify')
+  );
+  const { packageManagerExecutable, packageManager } =
+    await setupPackageManager(tempDir);
```
sobolk, line 103: I was facing problems when I tried to use `async/await` in describe blocks. Perhaps this is behind pnpm failures (it's a guess though).

I suggest to define 
```
let packageManagerExecutable: <type>;
let packageManager: <type>
```

and call `setupPackageManager()` in `before` block.

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
```diff
@@ -0,0 +1,258 @@
...
+      })
+    );
+    await fsp.rm(tempDir, { recursive: true });
+    // stop the npm proxy
+    await execa('npm', ['run', 'stop:npm-proxy'], { stdio: 'inherit' });
+  });
+
+  void it(`starting project`, async () => {
+    await execa(
+      packageManager.startsWith('yarn') ? 'yarn' : packageManager,
```
sobolk, line 134: why don't we use `packageManagerExecutable` from above ?

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
```diff
@@ -0,0 +1,258 @@
...
+        stdio: 'inherit',
+      }
+    );
+
+    const amplifyPathPrefix = path.join(tempDir, 'amplify');
+
+    const pathPrefix = path.join(tempDir, 'amplify');
+
+    const files = await glob(path.join(amplifyPathPrefix, '**', '*'), {
+      // eslint-disable-next-line spellcheck/spell-checker
```
sobolk, line 147: please add `nodir` to dictionary. if it's part of glob interface then it's going to be re-occuring sooner or later.

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
sobolk: In that case I suggest to create `continuePackageManagerSetupAfterCreateAmplify()` to host this logic.

This however makes me wonder why do we need this. I.e. is create amplify flow missing some steps or do we need to do this because of our testing strategy ?

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
```diff
@@ -0,0 +1,258 @@
...
+  return { packageManagerExecutable, packageManager };
+};
+
+void describe('installs expected packages and scaffolds expected files', async () => {
+  let branchBackendIdentifier: BackendIdentifier;
+  let testBranch: TestBranch;
+  let cfnClient: CloudFormationClient;
+  const tempDir = await fsp.mkdtemp(
+    path.join(os.tmpdir(), 'test-create-amplify')
+  );
```
sobolk, line 101: This should also be in `beforeEach` for future proofing.

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
```diff
@@ -0,0 +1,258 @@
...
+            'pluralize',
+            'zod',
+            '@aws-amplify/platform-core',
+            'esbuild',
+          ],
+          {
+            cwd: tempDir,
+            stdio: 'inherit',
+          }
+        );
```
sobolk, line 225: what happens if we don't do this? will brand new project fail deployment?
if yes then we should troubleshoot this further to figure out what's proper fix for this.

I.e. `<packageManagerExecutable> create amplify && <packageManagerExecutable> amplify sandbox (or pipeline deploy)`  should just work.

This looks like we might need to either do something with our dependencies  OR enchance create-amplify.

Btw. for the CDK part I suggest that you pull latest main - we made `create amplify` install CDK libs in customer's project recently.

sobolk (commented):
This looks like going into right direction.

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
0618: Done 433a884eca

packages/backend-deployer/src/cdk_deployer_singleton_factory.ts
sobolk: If you choose follow up. Please `// TODO` this.

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
0618: Done 0eb8919bb7

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
0618: Done df93412961.

I just put them into `beforeEach` and `afterEach`

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
0618: Done df93412961

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
0618: done df93412961

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
0618: because for `npm`, we don't want to use `npx`

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
0618: Done

packages/backend-deployer/src/cdk_deployer_singleton_factory.ts
0618: Done e485bf5de8

packages/backend-deployer/src/cdk_deployer.ts
0618: This PR focus on adding the test. I'll make a follow-up PR to change the code

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
0618: Simplified it [65df176](https://github.com/aws-amplify/amplify-backend/pull/901/commits/65df1765a01a3e3a0e7e004022ca9765b0fe8f17)

I'll make a follow-up PR to change create-amplify and pipeline-deploy.

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
0618: Good point. It turned out after we use `node_modules` for `yarn-modern`, we don't need to add most of the dependencies. [65df176](https://github.com/aws-amplify/amplify-backend/pull/901/commits/65df1765a01a3e3a0e7e004022ca9765b0fe8f17)

packages/create-amplify/src/package-manager-controller/package_manager_controller.ts
```diff
@@ -15,10 +15,10 @@ export abstract class PackageManagerController {
    */
   constructor(
     readonly projectRoot: string,
-    protected executable: string,
-    protected binaryRunner: string,
-    protected initDefault: string[],
-    protected installCommand: string,
+    protected readonly executable: string,
```
0618, line 18: this change is for a left-over comment https://github.com/aws-amplify/amplify-backend/pull/793#discussion_r1437173240

packages/create-amplify/src/package-manager-controller/package_manager_controller_factory.ts
```diff
@@ -21,20 +21,6 @@ export class PackageManagerControllerFactory {
     private readonly userAgent: string
   ) {}
 
-  /**
```
0618, line 24: re-order the functions to make linter happy

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        '--project',
+        amplifyPathPrefix,
+      ],
+      {
+        cwd: tempDir,
+        stdio: 'inherit',
+      }
+    );
+
+    if (packageManager === 'yarn-modern') {
```
0618, line 109: we may need to install the dependencies at `create-amplify` instead

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
edwardfoyle, line 1: `e2e_flow` seems too generic of a name for this file. Maybe `create_amplify_and_deploy`?

I'm also not crazy about the `e2e-flow` directory. It's not clear how it's different from the existing `test-e2e` directory. Maybe this should just be a file within `test-e2e`? Or if we want to keep it separate for the sake of the test matrix, maybe we need a new directory like `package-manager-sanity-checks`?

packages/integration-tests/src/e2e-flow/setup_package_manager.ts
```diff
@@ -0,0 +1,120 @@
...
+  packageManager: PackageManager;
+  packageManagerExecutable: PackageManagerExecutable;
+}> => {
+  const packageManager = (process.env.PACKAGE_MANAGER ??
+    'npm') as PackageManager;
+  const packageManagerExecutable = packageManager.startsWith('yarn')
+    ? 'yarn'
+    : packageManager === 'npm'
+    ? 'npx'
+    : (packageManager as PackageManagerExecutable);
```
sobolk, line 91: can we merge this `if` with switch below? will be easier to read than chain of `? :`s

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      ['create', 'amplify', '--yes'],
+      {
+        cwd: tempDir,
+        stdio: 'inherit',
+      }
+    );
+
+    const amplifyPathPrefix = path.join(tempDir, 'amplify');
+
+    const pathPrefix = path.join(tempDir, 'amplify');
```
edwardfoyle, line 71: do we need both of these?

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      'tsconfig.json',
+    ];
+
+    assert.deepStrictEqual(
+      files.sort(),
+      expectedAmplifyFiles.map((suffix) => path.join(pathPrefix, suffix))
+    );
+
+    // assert that project compiles successfully
+    await execa(
```
edwardfoyle, line 93: pipeline-deploy will also run tsc. Could we just do pipeline-deploy and take this out?

packages/integration-tests/src/e2e-flow/setup_package_manager.ts
```diff
@@ -0,0 +1,120 @@
...
+  {
+    const { stdout } = await execa('npm', ['config', 'get', 'cache']);
+    const npxCacheLocation = path.join(stdout.toString().trim(), '_npx');
+    if (existsSync(npxCacheLocation)) {
+      await fsp.rm(npxCacheLocation, { recursive: true });
+    }
+  }
+};
+
+const initializePnpm = async (packageManager: string) => {
```
edwardfoyle, line 20: seems like the function name is saying that packageManager is always going to be 'pnpm' here. Can we remove the function arg?

packages/integration-tests/src/e2e-flow/setup_package_manager.ts
```diff
@@ -0,0 +1,120 @@
+import { execa } from 'execa';
+import * as fsp from 'fs/promises';
+import { existsSync } from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+
+type PackageManager = 'npm' | 'yarn-classic' | 'yarn-modern' | 'pnpm';
+type PackageManagerExecutable = 'npx' | 'yarn' | 'pnpm';
+
+const initializeNpm = async () => {
```
edwardfoyle, line 10: looks like all of the other functions are pointing the package registry to localhost. does this function need to do the same?

packages/integration-tests/src/e2e-flow/setup_package_manager.ts
```diff
@@ -0,0 +1,120 @@
...
+      await initializeYarnClassic(execaOptions);
+      break;
+
+    case 'yarn-modern':
+      execaOptions.cwd = dir;
+      await initializeYarnModern(execaOptions);
+      break;
+
+    default:
+      break;
```
edwardfoyle, line 116: we should probably throw an error here

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
edwardfoyle: can we mark this as a TODO?

packages/integration-tests/src/e2e-flow/setup_package_manager.ts
0618: Good catch. Our `vent` does it already for `npm`

.github/workflows/poc-e2e-flow-test.yml
```diff
@@ -61,7 +64,18 @@ jobs:
...
+      - name: Configure test execution credentials
+        uses: aws-actions/configure-aws-credentials@04b98b3f9e85f563fb061be8751a0352327246b0 # version 3.0.1
+        with:
+          role-to-assume: ${{ secrets.E2E_RUNNER_ROLE_ARN }}
+          aws-region: us-west-2
+      - name: Run E2E flow tests with ${{ matrix.pkg-manager }}
         shell: bash
         run: |
-          PACKAGE_MANAGER_EXECUTABLE=${{matrix.pkg-manager}} npm run test:dir packages/integration-tests/src/test-e2e/create_amplify.test.ts
+          PACKAGE_MANAGER=${{matrix.pkg-manager}} npm run test:dir packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
```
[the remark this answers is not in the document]
0618: I do think we can have a better name than `e2e-flow`. 

How about `e2e-shallow/package-manager.test.ts`? So that potentially we can add more shallow e2e test for something else.

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
0618: Let's move the conversation to this thread https://github.com/aws-amplify/amplify-backend/pull/901/files#r1456681682

packages/integration-tests/src/e2e-flow/setup_package_manager.ts
0618: Good catch. Done 17f406b70c69579ad2aa87c69f57018c8c2dd2f9

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
0618: Removed 17f406b70c69579ad2aa87c69f57018c8c2dd2f9

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
```diff
@@ -0,0 +1,135 @@
...
+  CloudFormationClient,
+  DeleteStackCommand,
+} from '@aws-sdk/client-cloudformation';
+import { BackendIdentifierConversions } from '@aws-amplify/platform-core';
+import { getClientConfigPath } from '@aws-amplify/client-config';
+import { TestBranch, amplifyAppPool } from '../amplify_app_pool.js';
+import { e2eToolingClientConfig } from '../e2e_tooling_client_config.js';
+import { setupPackageManager } from './setup_package_manager.js';
+
+void describe('installs expected packages and scaffolds expected files', async () => {
```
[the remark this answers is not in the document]
0618: Good catch. updated 17f406b70c69579ad2aa87c69f57018c8c2dd2f9

packages/integration-tests/src/e2e-flow/setup_package_manager.ts
0618: Done 60d33bc179

packages/integration-tests/src/e2e-flow/setup_package_manager.ts
0618: Done 60d33bc179

packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -126,7 +127,7 @@ export class CDKDeployer implements BackendDeployer {
     }
     try {
       await this.executeChildProcess(
-        'npx',
+        this.packageManager === 'npm' ? 'npx' : this.packageManager, // TODO: once the e2e flow test is complete, we need to refactor this to use a ctor
```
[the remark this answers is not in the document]
0618: Done e13058e61b

.github/workflows/poc-e2e-flow-test.yml
```diff
@@ -61,7 +64,18 @@ jobs:
...
+        uses: ./.github/actions/setup_profile
+        with:
+          role-to-assume: ${{ secrets.E2E_TOOLING_ROLE_ARN }}
+          aws-region: us-west-2
+          profile-name: e2e-tooling
+      - name: Configure test execution credentials
+        uses: aws-actions/configure-aws-credentials@04b98b3f9e85f563fb061be8751a0352327246b0 # version 3.0.1
+        with:
+          role-to-assume: ${{ secrets.E2E_RUNNER_ROLE_ARN }}
+          aws-region: us-west-2
```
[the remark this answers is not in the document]
0618: We could. But is it necessary?  `aws-actions/configure-aws-credentials` is already a re-usable action

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
0618: Good catch! Removed one

packages/integration-tests/src/e2e-flow/setup_package_manager.ts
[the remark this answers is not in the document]
0618: Added a todo on top

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
0618: Removed all of them 038ac85e15

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
0618: Done 038ac85e15

It should just work now

.github/workflows/poc-e2e-flow-test.yml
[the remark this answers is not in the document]
edwardfoyle: I don't think it's necessary yet, just a thought. If we end up repeating this again then it's probably time to make an action

.github/workflows/poc-e2e-flow-test.yml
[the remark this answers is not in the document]
edwardfoyle: What's the overlap between this test and the canary test? Could that be a single test suite that we run in both places?

packages/integration-tests/src/e2e-flow/setup_package_manager.ts
edwardfoyle: Since we aren't using vend for local testing anymore, maybe it makes sense to rework that command to only start the local registry and publish, and then these methods are responsible for updating package manager config to point to localhost? I'm not sure, I can see arguments both ways

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
0618: The logic is moved to `create-amplify` now

packages/integration-tests/src/e2e-flow/e2e_flow.test.ts
0618: Done 222249a973 <- See if that's what you meant?

.github/workflows/poc-e2e-flow-test.yml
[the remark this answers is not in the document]
0618: this test only run `create amplify` and `pipeline-deploy`. We keep the minimum (if any) assertions. 

It might not be necessary to run at both places IMO.

packages/integration-tests/src/e2e-flow/setup_package_manager.ts
0618: I vote for the laizier one 😁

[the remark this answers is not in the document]
0618: Do we still need the comment on line 9 after the refactor?

.github/workflows/poc-e2e-flow-test.yml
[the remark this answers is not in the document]
sobolk: I was also thinking about re-using canary test for this purpose. And... whether we should be testing package managers in canary tests too?

If we do this then package manager tests would just get extra coverage for sandbox by looking at https://github.com/aws-amplify/amplify-backend/blob/main/packages/integration-tests/src/test-live-dependency-health-checks/health_checks.test.ts . Which might not be a bad idea.

On the other hand I can't say how both suites will evolve, so maybe keeping them separately for now is not a bad idea.

They're both simple so revisiting this decision later is not a big undertaking. So I'm generally fine either way here.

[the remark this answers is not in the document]
sobolk: As for naming:
1. if we end up re-using canary test then naming is already set there and seems to be good.
2. if we keep this test separately I'd just use naming specific to purpose and revisit it if we ever get to build up this suite. I like `package-manager-sanity-checks` from other comment for directory naming.

packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -33,8 +33,12 @@ export class CDKDeployer implements BackendDeployer {
...
   constructor(
     private readonly cdkErrorMapper: CdkErrorMapper,
-    private readonly backendLocator: BackendLocator
-  ) {}
+    private readonly backendLocator: BackendLocator,
+    private readonly packageManager: string
+  ) {
+    this.packageManager =
+      this.packageManager === 'npm' ? 'npx' : this.packageManager;
+  }
```
sobolk, line 41: A note for future follow up when we plumb this from CLI.

The deployer only cares about "package manager executable". So perhaps this resolution should happen in CLI.

Or, perhaps we should start thinking about moving package manager abstraction (controller?) to `platform-core` and have `PackageManagerController` injected here from CLI?

packages/create-amplify/src/package-manager-controller/yarn_modern_package_manager_controller.ts
```diff
@@ -35,6 +36,14 @@ export class YarnModernPackageManagerController extends PackageManagerController
     );
   }
 
+  installDependencies: (
+    packageNames: string[],
+    type: DependencyType
+  ) => Promise<void> = async (packageNames, type) => {
+    await this.addDependencies(this.projectRoot);
+    await super.installDependencies(packageNames, type);
```
sobolk, line 44: nice :-)
This is why we invented that abstraction.

packages/integration-tests/src/process-controller/process_controller.ts
```diff
@@ -159,3 +164,75 @@ export const cdkCli = (
     env: options?.env,
   });
 };
+
+/**
+ *  packageManagerExecute - Factory function that returns a ProcessController for the specified package manager
+ */
+export const packageManagerExecute = (
```
sobolk, line 171:
```suggestion
export const executeWithPackageManager = (
```

packages/integration-tests/src/process-controller/process_controller.ts
```diff
@@ -159,3 +164,75 @@ export const cdkCli = (
...
+): ProcessController => {
+  const packageManagerExecutable = packageManager.startsWith('yarn')
+    ? 'yarn'
+    : packageManager;
+  const command = execaSync('npx', ['which', packageManagerExecutable], {
+    cwd: dir,
+  }).stdout.trim();
+  if (!command) {
+    throw new Error('Unable to locate packageManager binary');
+  }
```
sobolk, line 232: that doesn't seem to be used.

packages/integration-tests/src/process-controller/process_controller.ts
```diff
@@ -159,3 +164,75 @@ export const cdkCli = (
...
```
sobolk, line 187: we can remove this too.

packages/integration-tests/src/process-controller/process_controller.ts
```diff
@@ -159,3 +164,75 @@ export const cdkCli = (
...
+  return new ProcessController(packageManagerBinary, args, {
+    cwd: dir,
+    env: options?.env,
+  });
+};
+
+/**
+ * packageManagerCli - Factory function that returns a ProcessController for the specified package manager
+ */
+export const packageManagerCli = (
```
sobolk, line 216:
```suggestion
export const runPackageManager = (
```

packages/integration-tests/src/e2e-flow/setup_package_manager.ts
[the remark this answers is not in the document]
sobolk: Let's keep it for now. If we take no action we can remove this when going to main.

packages/backend-deployer/src/cdk_deployer.ts
0618: Yea, we should move the abstruction to `platform-core`. I'll do it in a follow-up PR.

packages/integration-tests/src/process-controller/process_controller.ts
0618: `packageManagerCli` is consistent with `amplifyCLI` and `cdkCli` though

0618: done d33f115b9e

packages/integration-tests/src/process-controller/process_controller.ts
0618: done d33f115b9e

packages/integration-tests/src/process-controller/process_controller.ts
0618: done d33f115b9e

packages/integration-tests/src/process-controller/process_controller.ts
0618: done d33f115b9e

packages/backend-deployer/src/cdk_deployer.ts
0618: added a comment d33f115b9e

.github/workflows/poc-e2e-flow-test.yml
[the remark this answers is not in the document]
0618: Thanks for the discussion. I renamed the file and added a TODO 8e6cc14b26

sobolk (approved):

packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -33,8 +33,14 @@ export class CDKDeployer implements BackendDeployer {
    */
   constructor(
     private readonly cdkErrorMapper: CdkErrorMapper,
-    private readonly backendLocator: BackendLocator
-  ) {}
+    private readonly backendLocator: BackendLocator,
+    private readonly packageManager: string
+  ) {
+    // TODO: use PackageManagerController to get the executable.
```
sobolk, line 39: I wonder if it's going to be `get the executable` or perhaps `packageManagerController.executeWithPackageManager(command, args)`.
e.g. `packageManagerController.executeWithPackageManager('tsc', tscArgs)`;

0618: That sounds a good idea 👍

.changeset/honest-pumas-fold.md
edwardfoyle, line 1: Just a suggestion for future reviews on feature branches: I think it's fine to merge main directly into the feature branch without a PR, then make PRs on top of those merges. This cuts down on PR diffs that are from changes that have already been merged to main. This will also make your life easier when merging the feature branch back to main as it will greatly reduce the number of merge conflicts.

packages/integration-tests/src/package_manager_sanity_checks.test.ts
```diff
@@ -0,0 +1,119 @@
...
+import {
+  runPackageManager,
+  runWithPackageManager,
+} from './process-controller/process_controller.js';
+
+/**
+ * TODO: This test will be refactored to use the test from health-checks.test.ts and run with different package managers.
+ */
+
+void describe('create-amplify and pipeline deploy', async () => {
```
edwardfoyle, line 31: nit: perhaps `describe('getting started happy path')`

packages/integration-tests/src/package_manager_sanity_checks.test.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  beforeEach(async () => {
+    tempDir = await fsp.mkdtemp(path.join(os.tmpdir(), 'test-create-amplify'));
+    const packageManagerInfo = await setupPackageManager(tempDir);
+    packageManagerExecutable = packageManagerInfo.packageManagerExecutable;
+    packageManager = packageManagerInfo.packageManager;
+
+    // start a local npm proxy and publish the current codebase to the proxy
+    await execa('npm', ['run', 'clean:npm-proxy'], { stdio: 'inherit' });
+    await execa('npm', ['run', 'vend'], { stdio: 'inherit' });
```
edwardfoyle, line 47: Doesn't matter yet since there's only one test here, but starting/stopping the npm proxy should probably go in `before()` and `after()` rather than `beforeEach()` and `afterEach()`

packages/integration-tests/src/process-controller/process_controller.ts
```diff
@@ -159,3 +164,59 @@ export const cdkCli = (
     env: options?.env,
   });
 };
+
+/**
+ *  runWithPackageManager - Factory function that returns a ProcessController for the specified package manager
+ */
```
edwardfoyle, line 170: The difference between this function and the one below is very subtle. I'd fill out these JSDocs a little more to explain the difference

packages/integration-tests/src/setup_package_manager.ts
```diff
@@ -0,0 +1,115 @@
...
+    await fsp.rm(npxCacheLocation, { recursive: true });
+  }
+};
+
+const initializePnpm = async () => {
+  const packageManager = 'pnpm';
+  await execa('npm', ['install', '-g', packageManager], {
+    stdio: 'inherit',
+  });
+  await execa(packageManager, ['--version']);
```
edwardfoyle, line 26: shouldn't we be using `runPackageManager` for these commands?

edwardfoyle (approved):
Left some nits and a couple comments that should be addressed, but I'm fine if they go in a follow up PR

packages/integration-tests/src/package_manager_sanity_checks.test.ts
0618: Done add3437799

packages/integration-tests/src/package_manager_sanity_checks.test.ts
0618: Done add3437799

packages/integration-tests/src/setup_package_manager.ts
0618: Not needed since `packageManager = 'pnpm'` in this case

packages/integration-tests/src/process-controller/process_controller.ts
0618: Done 4c4069bcb4

edwardfoyle (approved):

PR #903, opened 2024-01-11
chore: remove Logger class & refactor Printer

bombguy:
## Problem

## Changes

## Checklist


packages/cli-core/src/printer/printer.ts
```diff
@@ -37,16 +61,171 @@ export class Printer {
...
 }
+
+const argv = await yargs(process.argv.slice(2)).options({
+  debug: {
+    type: 'boolean',
+    default: false,
+  },
+}).argv;
+
+const minimumLogLevel = argv.debug ? LogLevel.DEBUG : LogLevel.INFO;
```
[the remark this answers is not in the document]
bombguy: I will work on this on a followup PR. 

Currently, to enable this debug flag, you would need to do forward the flag like this `npm create amplify -y -- --debug`. I agree that its a good idea to expose the debug flag at top level.

packages/cli-core/src/printer/printer.ts
```diff
@@ -37,16 +61,171 @@ export class Printer {
...
+
+const instance = new Printer(minimumLogLevel);
+
+export default instance;
```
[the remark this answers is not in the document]
bombguy: I removed the export instead. Printer class is now static singleton, developers should be just calling static methods from Printer.

packages/cli-core/src/printer/printer.ts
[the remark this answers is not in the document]
bombguy: @sobolk I don't think that will solve the flag forwarding issue in the create-amplify flow

packages/cli-core/src/printer/printer.ts
```diff
@@ -7,6 +8,40 @@ export type RecordValue = string | number | string[] | Date;
...
+  ) {
+    Printer.minimumLogLevel = minimumLogLevel;
+    Printer.writeStream = writeStream;
+    this.refreshRate = 500;
+  }
+
+  /**
+   * returns static singleton.
+   */
+  static make(
```
edwardfoyle, line 35: What I meant by my other comment is we should make all the methods on this class instance methods, not static methods. eg
```ts
class Printer {
  private constructor() {}

  info: () => {}

  debug: () => {}

  etc
}

export const printer = new Printer();
```

Then the usage in other files is
```ts
import { printer } from '@aws-amplify/cli-core`

printer.info()
```

We should also only export a Printer type and have a `DefaultPrinter` that implements that type

packages/cli-core/src/printer/printer.ts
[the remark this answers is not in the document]
sobolk: In that case we should perhaps just scan `process.argv` without depending on `yargs` here to find the flag.

packages/cli-core/src/printer/printer.ts
```diff
@@ -37,16 +72,152 @@ export class Printer {
...
+    return Printer.writeStream.isTTY;
+  }
 }
+
+export enum LogLevel {
+  INFO,
+  DEBUG,
+}
+
+export enum EscapeSequence {
```
edwardfoyle, line 207: this should not be exported

packages/cli-core/src/printer/printer.ts
```diff
@@ -37,16 +72,152 @@ export class Printer {
...
    */
   static printNewLine = () => {
-    console.log(EOL);
+    this.writeStream.write(EOL);
   };
+
+  /**
+   * Logs a message to the console.
+   */
+  static log(message: string, level: LogLevel = LogLevel.INFO, eol = true) {
```
edwardfoyle, line 106: we should have _either_ `log(message, level)` _or_ `info(message)`, `debug(message)` but not both. I think I have a slight preference for the former because it's fewer methods on the class, but I don't have a strong opinion one way or the other. @sobolk any opinion here?

packages/cli-core/src/printer/printer.ts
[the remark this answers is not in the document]
edwardfoyle: Perhaps the yargs parsing for the debug flag should be moved to cli-core and reused by create-amplify

packages/cli-core/src/printer/printer.ts
edwardfoyle: With the recent offline discussion, the last part "We should also only export a Printer type and have a DefaultPrinter that implements that type" is now irrelevant. We should export the Printer class.

packages/cli-core/src/printer/printer.ts
sobolk: All the loggers I know have both APIs. See
https://github.com/winstonjs/winston?tab=readme-ov-file#using-the-default-logger
https://logging.apache.org/log4j/2.x/javadoc/log4j-api/org/apache/logging/log4j/Logger.html
https://learn.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-8.0

Having said that, doing one patter and introducing other later is always an option. 
But my sophisticated algorithm chose `log` so I'd be fine with that first.
image

packages/cli-core/src/printer/printer.ts
sobolk: I'd still define `type Printer` and use it to declare type in caller packages.
Due to https://github.com/microsoft/TypeScript/issues/53558 (especially if we want to pass printer from cli to sandbox and they'd end up reading type defs from different places in node modules)

sobolk: Perhaps we need https://www.slf4j.org/api/org/slf4j/LoggerFactory.html :D

packages/cli/src/printer.ts
```diff
@@ -0,0 +1,7 @@
+import { LogLevel, Printer } from '@aws-amplify/cli-core';
+
+const minimumLogLevel = process.argv.includes('--debug')
+  ? LogLevel.DEBUG
+  : LogLevel.INFO;
+
+export const printer = new Printer(minimumLogLevel);
```
edwardfoyle, line 7: I'm on the fence about whether we should construct this here and import it as needed, or figure out a way to construct it in the main parser and inject it via ctors into the commands: https://github.com/aws-amplify/amplify-backend/blob/main/packages/cli/src/main_parser_factory.ts

At the very least, we should add the `--debug` flag to the main parser so it shows up in `--help`

@sobolk any thoughts here?

packages/cli-core/API.md
```diff
@@ -23,12 +25,25 @@ export enum COLOR {
...
-    static printNewLine: () => void;
-    static printRecord: <T extends Record<string | number, RecordValue>>(object: T) => void;
-    static printRecords: <T extends Record<string | number, RecordValue>>(objects: T[]) => void;
+    constructor(minimumLogLevel: LogLevel, stdout?: NodeJS.WriteStream, stderr?: NodeJS.WriteStream);
+    indicateProgress(message: string, callback: () => Promise<void>): Promise<void>;
+    log(message: string, level?: LogLevel, eol?: boolean): void;
+    print: (message: string, colorName?: COLOR) => void;
+    printNewLine: () => void;
+    printRecord: <T extends Record<string | number, RecordValue>>(object: T) => void;
+    printRecords: <T extends Record<string | number, RecordValue>>(objects: T[]) => void;
```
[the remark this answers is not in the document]
bombguy: we can just keep one

packages/cli/src/main_parser_factory.test.ts
```diff
@@ -4,7 +4,7 @@ import { TestCommandRunner } from './test-utils/command_runner.js';
 import { createMainParser } from './main_parser_factory.js';
 import { version } from '#package.json';
 
-void describe('main parser', { concurrency: false }, () => {
+void describe('main parser', { concurrency: false }, async () => {
```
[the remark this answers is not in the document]
bombguy: It was an artfact from different approach earlier where I was trying to parse yargs while creating parser. Will remove this difff

packages/cli/src/printer.ts
bombguy: I was trying to do this and it become a bit of nightmare trying to refactor all the changes. This will force us to attempt parse twice.

packages/cli-core/src/printer/printer.ts
```diff
@@ -4,27 +4,27 @@ import { EOL } from 'os';
...
+    private readonly refreshRate: number = 500
+  ) {}
 
   /**
-   * Prints an array of objects/records to console.
+   * Prints an array of objects/records to output stream.
    */
-  static printRecords = <T extends Record<string | number, RecordValue>>(
+  printRecords = <T extends Record<string | number, RecordValue>>(
     objects: T[]
```
edwardfoyle, line 28: We could make this a spread input so you don't have to wrap a single entry in an array

packages/cli/src/printer.ts
edwardfoyle: Okay, I think I'm good with this approach for now. We should still add the `--debug` flag to the main parser though (even if it's not used for anything within the yargs process) and leave a comment pointing to where it is used here.

packages/create-amplify/src/amplify_project_creator.ts
```diff
@@ -32,19 +32,23 @@ export class AmplifyProjectCreator {
     private readonly initialProjectFileGenerator: InitialProjectFileGenerator,
     private readonly npmInitializedEnsurer: NpmProjectInitializer,
     private readonly gitIgnoreInitializer: GitIgnoreInitializer,
-    private readonly projectRoot: string
+    private readonly projectRoot: string,
+    private readonly printer: Printer
```
edwardfoyle, line 36: Since we're going with the `import { printer } from './printer'` strategy, we should be consistent here as well (ie don't inject in the ctor, just import it)

edwardfoyle (approved):
LGTM, good work!

PR #905, opened 2024-01-11
updates the create flow with color and structure

ykethan:
fixes https://github.com/aws-amplify/amplify-backend/issues/722

## Changes

Updates the create flow with color to highlight messages, commands and dependencies being installed 
add a spinner to show the installation process
uses [Kleur](https://github.com/lukeed/kleur) for colors

update:

https://github.com/aws-amplify/amplify-backend/assets/87995712/c997399a-e50e-4f40-aba5-a11383643b73

white background:

https://github.com/aws-amplify/amplify-backend/assets/87995712/085272a6-daf4-4043-92c8-031e455fd934

## Checklist


packages/create-amplify/src/amplify_project_creator.ts
```diff
@@ -40,51 +43,83 @@ export class AmplifyProjectCreator {
...
+            .yellow()
+            .bold(
+              'cd .' + this.projectRoot.replace(process.cwd(), '') + ''
+            )} and then:`;
+
+    logger.log(kleur.cyan('Welcome to AWS Amplify!\n'));
+
+    const instructionSteps = [
+      cdCommand,
+      `- Get started with your project by running ${kleur
```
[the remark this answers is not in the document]
edwardfoyle: Coming back to this after reading the rest of the PR. I think we should have a single format object in `cli-core` that has all of the different common formats that we want. This keeps all of the formatting logic in a single place and also gives us a single location to enforce cross-cutting functionality like turning off colors if the shell doesn't support it.

eg:
```ts
// cli-core

const supportsColor = // maybe use https://www.npmjs.com/package/supports-color?

const identityFunction = (anything: unknown) => anything;

export const format = {
  amplifyCommand: supportsColor ? (command: string) => kleur.yellow().bold(`'npx amplify ${command}'`) : identityFunction,
  sectionHeader: supportsColor ? kleur.blue().bold : identityFunction,
  list: (lines: string[]) => lines.map(line => ` - ${line}`).join(os.EOL)
}
```
Then the usage would be
```ts
import { format, logger } from '@aws-amplify/cli-core';

logger.log(`Get started with ${format.amplifyCommand('sandbox')}`)
```

edwardfoyle (commented):
In addition to the individual comments below, a couple high level things:
1. Lets not go crazy with the colors. Use colors as a tool to make information more digestable (eg delimiting sections of output or highlighting commands to run). If everything is colored, then nothing stands out.
2. Reserve red, orange and yellow for things that are errors or warnings
3. You might want to wait for [this refactor](https://github.com/aws-amplify/amplify-backend/pull/903) to land before going too much further to prevent a bunch of merge conflicts

packages/create-amplify/src/amplify_project_creator.ts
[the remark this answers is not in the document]
ykethan: the kleur: https://github.com/lukeed/kleur#conditional-support, should already check for the color support or would you think we should still do [supports-color](https://www.npmjs.com/package/supports-color)

packages/cli-core/src/format/format.test.ts
```diff
@@ -2,8 +2,44 @@ import * as os from 'node:os';
...
+  void it('should format amplify command without binary runner', () => {
+    const command = 'help';
+    const expectedOutput = cyan(`npx amplify help`);
+    const actualOutput = format.amplifyCommand(command);
+    assert.strictEqual(actualOutput, expectedOutput);
+  });
+
+  void it('should format amplify command with binary runner', () => {
+    const command = 'help';
+    const binaryRunner = 'npx';
```
edwardfoyle, line 17: we should test something besides the default

packages/cli-core/src/package-manager-controller/package_manager_controller_base.ts
```diff
@@ -52,9 +53,17 @@ export abstract class PackageManagerControllerBase
   /**
    * getWelcomeMessage - returns a welcome message for the customer
    */
-  getWelcomeMessage = () =>
-    `Run \`${this.binaryRunner} amplify help\` for a list of available commands. 
-Get started by running \`${this.binaryRunner} amplify sandbox\`.`;
+  getWelcomeMessage = () => {
+    const welcomeInfo = [
+      `Get started by running ${
+        format.amplifyCommand('sandbox', this.binaryRunner) || ''
```
edwardfoyle, line 59: I wonder if an API like `format.runner(this.binaryRunner).amplifyCommand('sandbox')` would be better. Then you can get an amplifyCommand formatter for a specific runner and reuse it like so:
```ts
const { amplifyCommand } = format.runner(this.binaryRunner);
const welcomeInfo = [
  `Get started with ${amplifyCommand('sandbox')}`,
  `Run ${amplifyCommand('help') for a list of available commands`
]
}

```

packages/cli-core/src/printer/printer.test.ts
```diff
@@ -59,38 +59,70 @@ void describe('Printer', () => {
     assert.strictEqual(mockedWrite.mock.callCount(), 0);
   });
 
-  void it('indicateProgress logs message & animates ellipsis if on TTY', async () => {
+  //   void it('indicateProgress does not animates ellipsis if not TTY & prints log message once', async () => {
```
edwardfoyle, line 62: remove unused code

packages/create-amplify/src/amplify_project_creator.ts
```diff
@@ -46,49 +49,74 @@ export class AmplifyProjectCreator {
...
-${this.packageManagerController.getWelcomeMessage()}`
-    );
-
-    printer.log(
-      `Amplify (Gen 2) collects anonymous telemetry data about general usage of the CLI.
-
-Participation is optional, and you may opt-out by using \`amplify configure telemetry disable\`.
-
-To learn more about telemetry, visit ${LEARN_MORE_USAGE_DATA_TRACKING_LINK}`
+      grey(
```
ykethan, line 112: should this be moved to package_manager_controller

packages/create-amplify/src/amplify_project_creator.ts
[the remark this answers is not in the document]
edwardfoyle: If kleur already checks that's fine. We could have a test for it though just to make sure nothing breaks when updating dependencies

packages/cli-core/src/printer/printer.ts
```diff
@@ -88,6 +81,24 @@ export class Printer {
...
+  async indicateProgress(
+    action: () => Promise<void>,
+    message: string,
+    successMessage: string
+  ) {
+    try {
+      this.startAnimatingSpinner(message);
+      await action();
+      this.stopAnimatingSpinner(successMessage);
+    } catch (error) {
```
edwardfoyle, line 96: no need for catch, just stop the animation in a finally block

packages/cli-core/src/printer/printer.ts
```diff
@@ -88,6 +81,24 @@ export class Printer {
...
+      this.stopAnimatingSpinner('An error occurred');
```
edwardfoyle, line 97: This is going to print a checkmark in front of "An error occurred"

packages/create-amplify/src/amplify_project_creator.ts
```diff
@@ -46,49 +49,74 @@ export class AmplifyProjectCreator {
...
+    await printer.indicateProgress(
+      () =>
+        this.packageManagerController.installDependencies(
           this.defaultDevPackages,
           'dev'
-        );
-      }
+        ),
+      'Installing devDependencies...',
+      'DevDependencies installed'
```
edwardfoyle, line 68: Do we need this as a separate message? If we remove the ... from the installing message we can have `<spinner> Installing devDependencies` => `<check marck> Installing devDependencies` which seems fine to me

packages/create-amplify/src/amplify_project_creator.ts
```diff
@@ -46,49 +49,74 @@ export class AmplifyProjectCreator {
...
       process.cwd() === this.projectRoot
-        ? ''
-        : `Navigate to your project directory using
-'cd .${this.projectRoot.replace(process.cwd(), '')}'.
-Then get started with the following commands:
-`;
-
+        ? null
+        : `Navigate to your project directory using ${cyan(
+            bold('cd .' + this.projectRoot.replace(process.cwd(), '') + '')
```
edwardfoyle, line 97: we should have another format method for non-amplify commands

packages/create-amplify/src/amplify_project_creator.ts
```diff
@@ -46,49 +49,74 @@ export class AmplifyProjectCreator {
...
-
-    printer.log(
-      `Amplify (Gen 2) collects anonymous telemetry data about general usage of the CLI.
-
-Participation is optional, and you may opt-out by using \`amplify configure telemetry disable\`.
-
-To learn more about telemetry, visit ${LEARN_MORE_USAGE_DATA_TRACKING_LINK}`
+      grey(
+        `Amplify (Gen 2) collects anonymous telemetry data about general usage of the CLI. Participation is optional, and you may opt-out by using ${cyan(
+          'amplify configure telemetry disable'
```
edwardfoyle, line 114: this should have the npx prefix

packages/create-amplify/src/amplify_project_creator.ts
edwardfoyle, line 112: This should be another format utility like `format.note()` or something like that

packages/cli-core/src/printer/printer.test.ts
```diff
@@ -59,38 +59,48 @@ void describe('Printer', () => {
     assert.strictEqual(mockedWrite.mock.callCount(), 0);
   });
 
-  void it('indicateProgress logs message & animates ellipsis if on TTY', async () => {
+  void it('start animating spinner with message and stops animation in TTY terminal', async () => {
```
edwardfoyle, line 62: we need another test for when the action fails

packages/create-amplify/src/amplify_project_creator.ts
ykethan:
```
⠹ Installing devDependencies gets replaced with ✔️  DevDependencies installed
```
felt would be minor dx improvement

packages/create-amplify/src/amplify_project_creator.ts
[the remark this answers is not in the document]
ykethan: https://github.com/ykethan/amplify-backend/blob/6ab4c209343d3864f5a408b21c5f7376416b0b9d/packages/cli-core/src/colors.ts#L2

noticed this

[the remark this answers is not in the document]
edwardfoyle: I think it's fine to pull in this dependency, but perhaps we should clean up the printer API to remove references to that color enum. eg [here](https://github.com/ykethan/amplify-backend/blob/6ab4c209343d3864f5a408b21c5f7376416b0b9d/packages/cli-core/src/printer/printer.ts#L43)

That can either be rolled into this PR or go as a separate PR after this one

packages/cli-core/API.md
```diff
@@ -29,6 +29,14 @@ export enum COLOR {
 
 // @public
 export const format: {
+    runner: (binaryRunner: string) => {
+        amplifyCommand: (command: string) => string;
+    };
+    infoMessage: (message: string) => string;
```
edwardfoyle, line 35: `infoMessage` could be confused with an info log line. I think I prefer my original suggestion of `note` or maybe `aside` if we're trying to convey that it's not a super important message

packages/cli-core/API.md
```diff
@@ -52,7 +60,7 @@ export class PackageManagerControllerFactory {
 // @public
 export class Printer {
     constructor(minimumLogLevel: LogLevel, stdout?: NodeJS.WriteStream, stderr?: NodeJS.WriteStream, refreshRate?: number);
-    indicateProgress(message: string, callback: () => Promise<void>): Promise<void>;
+    indicateProgress(action: () => Promise<void>, message: string): Promise<void>;
```
edwardfoyle, line 63: we can switch these args to avoid an unnecessary diff

packages/cli-core/src/format/format.test.ts
```diff
@@ -31,4 +68,44 @@ void describe('format', () => {
...
+
+void describe('format utility with manual color toggling', async () => {
+  // disable color, https://github.com/lukeed/kleur#individual-colors
+  beforeEach(() => {
+    $.enabled = false;
+  });
+
+  afterEach(() => {
+    $.enabled = true;
+  });
```
edwardfoyle, line 88: these should be `beforeAll` and `afterAll`

packages/cli-core/src/format/format.test.ts
```diff
@@ -31,4 +68,44 @@ void describe('format', () => {
...
+
+  void it('should format success message in green color', () => {
+    const message = 'Success message';
+    const expectedOutput = green(message);
+    const actualOutput = format.success(message);
+    assert.strictEqual(actualOutput, expectedOutput);
+  });
+});
+
+void describe('format utility with manual color toggling', async () => {
```
edwardfoyle, line 80:
```suggestion
void describe('format when terminal colors disabled', async () => {
```

packages/cli-core/src/format/format.test.ts
```diff
@@ -31,4 +68,44 @@ void describe('format', () => {
...
+  // disable color, https://github.com/lukeed/kleur#individual-colors
+  beforeEach(() => {
+    $.enabled = false;
+  });
+
+  afterEach(() => {
+    $.enabled = true;
+  });
+
+  void it('section header should not include ANSI color codes in strings when color is disabled', () => {
```
edwardfoyle, line 90:
```suggestion
  void it('prints plain section header', () => {
```

packages/cli-core/src/format/format.test.ts
```diff
@@ -31,4 +68,44 @@ void describe('format', () => {
...
+    const coloredMessage = format.sectionHeader(message);
+
+    assert.strictEqual(
+      coloredMessage.includes('\x1b['),
+      false,
+      'Color codes should not be present'
+    );
+  });
+
+  void it('amplify command should not include ANSI color codes in strings when color is disabled', () => {
```
edwardfoyle, line 101:
```suggestion
  void it('prints plain command', () => {
```

packages/cli-core/src/format/format.test.ts
```diff
@@ -31,4 +68,44 @@ void describe('format', () => {
...
+
+  void it('section header should not include ANSI color codes in strings when color is disabled', () => {
+    const message = 'Hello';
+    const coloredMessage = format.sectionHeader(message);
+
+    assert.strictEqual(
+      coloredMessage.includes('\x1b['),
+      false,
+      'Color codes should not be present'
+    );
```
edwardfoyle, line 98:
```suggestion
    assert.strictEqual(coloredMessage, 'Hello');
```

packages/cli-core/src/format/format.test.ts
```diff
@@ -31,4 +68,44 @@ void describe('format', () => {
...
+
+  void it('amplify command should not include ANSI color codes in strings when color is disabled', () => {
+    const message = 'Hello';
+    const coloredMessage = format.runner('yarn').amplifyCommand(message);
+
+    assert.strictEqual(
+      coloredMessage.includes('\x1b['),
+      false,
+      'Color codes should not be present'
+    );
```
edwardfoyle, line 109:
```suggestion
    assert.strictEqual(coloredMessage, 'Hello');
```

packages/cli-core/src/printer/printer.test.ts
```diff
@@ -59,38 +59,57 @@ void describe('Printer', () => {
...
-    await new Printer(LogLevel.INFO).indicateProgress(
-      'loading a long list',
-      () => new Promise((resolve) => setTimeout(resolve, 3000))
-    );
-    // filter out the escape characters.
+
+    const message = 'Message 1';
+
+    await new Printer(LogLevel.INFO).indicateProgress(async () => {
+      await new Promise((resolve) => setTimeout(resolve, 2000));
```
edwardfoyle, line 68: do we need such a long timeout in the tests?

packages/cli-core/src/printer/printer.ts
```diff
@@ -10,6 +10,11 @@ export class Printer {
   // Properties for ellipsis animation
   private timer: ReturnType<typeof setTimeout>;
   private timerSet: boolean;
+  /**
+   * Spinner frames
+   */
+  private spinnerFrames = ['⠋', '⠙', '⠹', '⠸', '⠼', '⠴', '⠦', '⠧', '⠇', '⠏'];
+  private currentSpinnerMessage = '';
```
edwardfoyle, line 17: no longer needed

packages/cli-core/src/printer/printer.ts
```diff
@@ -88,6 +81,18 @@ export class Printer {
...
 
+  /**
+   * Logs a message with animated spinner
+   */
+  async indicateProgress(action: () => Promise<void>, message: string) {
+    try {
+      this.startAnimatingSpinner(message);
+      await action();
+    } finally {
+      this.stopAnimatingSpinner(message);
```
edwardfoyle, line 92: This is still going to print a check mark even if the action fails. We should probably decouple printing the final message from stopping the spinner. ie `stopAnimatingSpinner` should clear the line and reset the cursor but not print anything.

Then it's on the caller to print a success/fail message. Likely a failure message doesn't need to be printed as the thrown error is already going to be printed. This also solves the problem you had of wanting to print a nice success message like `devDependencies installed!`

packages/cli-core/src/printer/printer.ts
```diff
@@ -103,36 +107,54 @@ export class Printer {
...
-      this.stdout.write(message + '.'.repeat(++frameCount % frameLength));
+      const frame = this.spinnerFrames[frameIndex];
+      this.stdout.write(`${frame} ${message}`);
+      frameIndex = (frameIndex + 1) % this.spinnerFrames.length;
     }, this.refreshRate);
   }
 
   /**
-   * Stops animating ellipsis and replace with a log message.
+   * Stops animating spinner and replace with a log message.
```
edwardfoyle, line 155:
```suggestion
   * Stops animating spinner.
```

packages/cli-core/src/printer/printer.ts
```diff
@@ -88,6 +80,18 @@ export class Printer {
     }
   }
 
+  /**
+   * Logs a message with animated spinner
```
edwardfoyle, line 84:
```suggestion
   * Logs a message with animated spinner
   * If stdout is not a TTY, the message is logged at the info level without a spinner
```

packages/cli-core/src/printer/printer.ts
```diff
@@ -103,36 +107,54 @@ export class Printer {
...
+  /**
+   * Start animating spinner at the end of a log message.
+   */
+  private startAnimatingSpinner(message: string) {
     if (!this.isTTY()) {
       this.log(message, LogLevel.INFO);
       return;
     }
 
     if (this.timerSet) {
```
edwardfoyle, line 136: This isn't technically part of this PR, but there's a subtle inconsistency here where if stdout is a TTY, it will allow multiple timers running at once (because of the return above). Instead we should check for timerSet first, then isTTY so that the behavior is consistent in both environments.

packages/create-amplify/src/amplify_project_creator.ts
```diff
@@ -50,49 +52,69 @@ export class AmplifyProjectCreator {
...
     const cdPreamble =
       process.cwd() === this.projectRoot
-        ? ''
-        : `Navigate to your project directory using
-'cd .${this.projectRoot.replace(process.cwd(), '')}'.
-Then get started with the following commands:
-`;
+        ? null
+        : `Navigate to your project directory using ${format.command(
+            'cd .' + this.projectRoot.replace(process.cwd(), '') + ''
```
edwardfoyle, line 94:
```suggestion
            `cd .${this.projectRoot.replace(process.cwd(), '')}`
```

edwardfoyle (approved):
LGTM, just a few fit and finish comments

packages/cli-core/src/printer/printer.ts
edwardfoyle: Just realized this isn't actually a problem because if it's a TTY, it will never get to the point of setting the timer in the first place. But it's fine, you don't need to change it back

edwardfoyle (approved):
LGTM, good work!

packages/cli-core/package.json
```diff
@@ -19,6 +19,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@inquirer/prompts": "^3.0.0",
-    "execa": "^8.0.1"
+    "execa": "^8.0.1",
+    "kleur": "^4.1.5"
```
0618, line 23: Why do we choose to use kleur instead of other packages? 

It'd prevent us from a lot of headaches if we don't introduce any packages at all https://stackoverflow.com/questions/9781218/how-to-change-node-jss-console-font-color

.eslint_dictionary.json
```diff
@@ -52,6 +52,7 @@
   "globals",
   "grantable",
   "graphql",
+  "grey",
```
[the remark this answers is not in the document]
ykethan: using gray should be fine to use as well. difference being british english vs american english. 
can change it if you believe gray might be better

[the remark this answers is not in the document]
0618: I'm just curious if we could add less words. It's not a big deal

packages/cli-core/package.json
ykethan: evaluated kleur, picocolors and chalk, kluer has chalk like syntax and is lighter. 
also kluer has zero dependancies
Benchmarks: https://github.com/lukeed/kleur#benchmarks 
we can switch to picocolors: https://github.com/alexeyraspopov/picocolors#benchmarks as well.

sobolk (approved):

0618 (approved):
kleur looks not actively maintained. Once it has security issue, it'd be hard for us to patch. 

Since it's light and doesn't have any dependencies, I guess it might not have too much security concern because if there is, it might have already been found.

PR #906, opened 2024-01-11
Add @aws-amplify/backend-deployer to @aws-amplify/cli as a dependency

0618:
## Problem

We need to add @aws-amplify/backend-deployer to @aws-amplify/cli because it's used in https://github.com/aws-amplify/amplify-backend/blob/cefddd1b4f2113a6fa99dd00a811c3a7d9393e39/packages/cli/src/commands/pipeline-deploy/pipeline_deploy_command_factory.ts#L2 

without it, yarn complains:

> Error: @aws-amplify/backend-cli tried to access @aws-amplify/backend-deployer, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

## Changes

## Checklist

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -29,6 +29,7 @@
   },
   "homepage": "https://github.com/aws-amplify/amplify-backend#readme",
   "dependencies": {
+    "@aws-amplify/backend-deployer": "^0.4.5",
     "@aws-amplify/backend-output-schemas": "^0.5.1",
     "@aws-amplify/backend-secret": "^0.4.1",
     "@aws-amplify/cli-core": "^0.2.0",
--- a/.changeset/fair-shoes-yell.md
+++ b/.changeset/fair-shoes-yell.md
@@ -1,0 +1,5 @@
+---
+'@aws-amplify/backend-cli': patch
+---
+
+add @aws-amplify/backend-deployer as a dependency to @aws-amplify/cli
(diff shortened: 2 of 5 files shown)


packages/cli/package.json
```diff
@@ -29,6 +29,7 @@
   },
   "homepage": "https://github.com/aws-amplify/amplify-backend#readme",
   "dependencies": {
+    "@aws-amplify/backend-deployer": "^0.4.5",
```
sobolk, line 32: This is a good catch.

I wonder why this rule didn't catch this earlier https://github.com/aws-amplify/amplify-backend/blob/3eb59ff4c709a9fe0fb02638312bb6c0fad01be8/.eslintrc.cjs#L122 .

Otherwise I wonder if we should declare `backend-deployer` as `peerDependency` given that in customer's app it's supposed to be delivered by `backend` package. @edwardfoyle what do you think?

0618: I put it under "dependencies" because I saw other `aws-amplify/*` packages are all in this category...

edwardfoyle: It should be under dependencies. We should figure out why the lint rule didn't catch it.

sobolk: I've run the rule in debugger and it seems that rule silently skips it because it can't load the package.
See screenshots.
image
image

We should log GH issue to figure this out.

sobolk (approved):

packages/cli/package.json
0618: Issue is created https://github.com/aws-amplify/amplify-backend/issues/908

sobolk:
Please make checks happy.

sobolk (approved):

edwardfoyle (approved):

PR #946, opened 2024-01-24
[Prototype] Custom Config

sobolk:
## Problem

This is a prototype of custom config.

## Changes

## Checklist


packages/backend-output-schemas/src/index.ts
```diff
@@ -67,6 +67,10 @@ export const unifiedBackendOutputSchema = z.object({
   [authOutputKey]: versionedAuthOutputSchema.optional(),
   [graphqlOutputKey]: versionedGraphqlOutputSchema.optional(),
   [storageOutputKey]: versionedStorageOutputSchema.optional(),
+  ['AWS::Amplify::Custom']: z.object({
+    version: z.literal('1'),
+    payload: z.record(z.string(), z.string()),
+  }).optional(),
```
sobolk, line 73: Sample payload:

```
{
  version: '1',
  payload: {
    amplifycustommyApiUrl: '{"value":"https://REDACTED.execute-api.us-west-2.amazonaws.com/prod/"}',
    amplifycustommyApiKey: '{"value":"REDACTED"}',
    amplifycustommyTestMapRegion: '{"value":"us-west-2","options":{"clientConfigDestinations":[{"clientConfigFormat":"json","path":["geo","amazon_location_service","region"]}]}}',
    amplifycustommyTestMapName: '{"value":"myTestMap","options":{"clientConfigDestinations":[{"clientConfigFormat":"json","path":["geo","amazon_location_service","maps","default"]}]}}',
    amplifycustommyTestMapStyle: '{"value":"VectorEsriNavigation","options":{"clientConfigDestinations":[{"clientConfigFormat":"json","path":["geo","amazon_location_service","maps","items","myTestMap","style"]}]}}',
    amplifycustommyTestSearchIndex: '{"value":"amplifycustomconfigpocappREDACTED","options":{"clientConfigDestinations":[{"clientConfigFormat":"json","path":["geo","amazon_location_service","search_indices","default"]},{"clientConfigFormat":"json","path":["geo","amazon_location_service","search_indices","items[]"]}]}}'
  }
}
```

packages/backend-output-storage/src/stack_metadata_output_storage_strategy.ts
```diff
@@ -36,4 +37,29 @@ export class StackMetadataBackendOutputStorageStrategy
...
+    return {
+      appendOutput: (key: string, value: string) => {
+        if (key in outputs) {
+          throw new Error(`Output ${key} is already defined`);
+        }
+        outputs[key] = value;
+        new CfnOutput(this.stack, key, { value });
+      }
+    }
+  };
```
sobolk, line 64: The existing `addBackendOutputEntry` doesn't work well for `payload` that's produced in multiple places. Therefore I invented a new type of BackendOutputEntry that can be appended to.

In general I was considering two approaches:
1. To put all custom outputs under `AWS::Amplify::Custom` metadata key.
    1. This seems to simplify handling on client config side.
2. To have `AWS::Amplify::Custom::{key}` per each custom output.
    1. This avoids creating `addAppendableBackendOutputEntry`
    2. But it complicates client config generation part, i.e. having dynamic key prefix complicates zod parsing of outputs and "client config contributor" resolution to handle given entry.

packages/backend/src/backend.ts
```diff
@@ -3,8 +3,32 @@ import { Stack } from 'aws-cdk-lib';
...
+  // But then... when we unify client config then clientConfigFormat doesn't make any sense and we'd have to deprecate it.
+  //
+  // For the purpose of prototype I'm going to ignore clientConfigFormat and assume it just covers JS schema and we map.
+  clientConfigFormat: string;
+  path: Array<string>;
+}
+
+export type CustomOutputOptions = {
+  clientConfigDestinations?: Array<ClientConfigDestination>
+}
```
sobolk, line 30: Please read this TODO.

packages/backend/src/backend_factory.ts
```diff
@@ -108,6 +115,32 @@ export class BackendFactory<
...
+        this.outputStorageStrategy.addAppendableBackendOutputEntry(
+          'AWS::Amplify::Custom'
+        );
+    }
+
+    const outputData = JSON.stringify({ value: value, options });
+    console.log(outputData);
+    // TODO how do we do custom namespacing ??
+    this.customOutputsEntry.appendOutput(`amplifycustom${key}`, outputData);
+  };
```
sobolk, line 143: 1. It seems that we can't just use output key to smuggle information about destination client config path due to CFN validation.
2. To overcome that I just put serialized json with output value AND metadata that describes where this should go in client config.
3. Is versioning of `AWS::Amplify::Custom` good enough to deal with these serialized jsons?

packages/client-config/src/client-config-contributor/custom_client_config_contributor.ts
```diff
@@ -0,0 +1,68 @@
...
+            currentRef[lastPathSegment] = customOutputData.value;
+          }
+        });
+      }
+    });
+
+    console.log(result);
+
+    return result;
+  };
```
sobolk, line 67: This file is interesting.

The idea is that:
1. Custom outputs are put into `custom.X` path in client config by default
2. If directive was sent from `setCustomOutput` api these values can be placed into specific locations (i.e. to handle Geo use case where we need to match schema expected by frontend libs that exist today).

The actual mechanism to target specific config keys is TBD and a bit hacky as of now.

packages/custom-config-poc-app/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+# in separate shell
+# start in repo root
+cd packages/custom-config-poc-app
+npm run build
+npm run dev
+
+# in browser or separate shell
+open http://localhost:3000/
+```
```
sobolk, line 21: if you want to play locally.

packages/custom-config-poc-app/amplify/backend.ts
```diff
@@ -0,0 +1,145 @@
...
+      api,
+      stage: api.deploymentStage,
+    },
+  ],
+});
+
+usagePlan.addApiKey(apiKey);
+
+backend.setCustomOutput('myApiUrl', api.url);
+backend.setCustomOutput('myApiKey', apiKeyValue);
```
sobolk, line 63: Sample usage of API:
1. `api.url` is CDK token when this executes
2. `apiKeyValue` represents value available when this code runs (feed by env vars to avoid posting secret, but can be just hardcoded string as well).

packages/custom-config-poc-app/amplify/backend.ts
```diff
@@ -0,0 +1,145 @@
...
+  ]
+});
+backend.setCustomOutput('myTestMapStyle', testMapStyle, {
+  clientConfigDestinations: [
+    {
+      clientConfigFormat: 'json',
+      path: ['geo', 'amazon_location_service', 'maps', 'items', testMap.mapName, 'style']
+    }
+  ]
+});
```
sobolk, line 116: Advanced usage with desired keys targeting.

packages/custom-config-poc-app/amplify/backend.ts
```diff
@@ -0,0 +1,145 @@
...
+      clientConfigFormat: 'json',
+      path: ['geo', 'amazon_location_service', 'search_indices', 'default']
+    },
+    {
+      clientConfigFormat: 'json',
+      // TODO how do we handle arrays ? Should we use json path syntax ?
+      path: ['geo', 'amazon_location_service', 'search_indices', 'items[]']
+    }
+  ]
+});
```
sobolk, line 143: Another advanced usage with key targeting.

sobolk (commented):
It works.
image

packages/backend/src/backend_factory.ts
```diff
@@ -121,5 +155,19 @@ export const defineBackend = <T extends DefineBackendProps>(
...
+      // into some output
+      // and just merge it with client config at read.
+
+      // This likely could use client config unification.
+      // I.e. backend def -> unified client config partial to outputs -> deployment ->
+      // -> outputs -> read unified client config -> map to js/ts/dart/whatever.
+
+      // Should we start crafting unified schema nad return it from generateClientConfig API ?
+      // and map to old/v6 formats in generateClientConfigToFile ?
+    }
```
sobolk, line 171: The disadvantage that `setCustomOutput` with that `path` pointers to config has is that it's hard to figure out what paths should be used.

This presents an alternative approach where we unify client config schema and use it to provide typings.

packages/custom-config-poc-app/amplify/backend.ts
```diff
@@ -0,0 +1,194 @@
...
+  //       "default": placeIndex.placeIndexName,
+  //       "items": [
+  //         placeIndex.placeIndexName
+  //       ]
+  //     }
+  //   }
+  // }
+});
+
+// ####### Alternative proposal End #####
```
sobolk, line 194: This is not plumbed through. But could potentially provide better DX.

packages/custom-config-poc-app/amplify/backend.ts
```diff
@@ -0,0 +1,194 @@
...
+      resources: [testMap.attrArn],
+    }),
+  ],
+});
+
+backend.setCustomOutput('myTestMapRegion', locationStack.region, {
+  clientConfigDestinations: [
+    {
+      // TODO figure out if we really need per client config type here
+      // See comments in packages/backend/src/backend.ts
```
edwardfoyle, line 100: Based on the comments above and seeing this, we should avoid this if at all possible. It seems like a lot of noise to expose in a public API

packages/custom-config-poc-app/amplify/backend.ts
```diff
@@ -0,0 +1,194 @@
...
+backend.setCustomOutput('myTestSearchIndex', placeIndex.placeIndexName, {
+  clientConfigDestinations: [
+    {
+      clientConfigFormat: 'json',
+      path: ['geo', 'amazon_location_service', 'search_indices', 'default'],
+    },
+    {
+      clientConfigFormat: 'json',
+      // TODO how do we handle arrays ? Should we use json path syntax ?
+      path: ['geo', 'amazon_location_service', 'search_indices', 'items[]'],
```
edwardfoyle, line 157: I'd be a fan of json path syntax (or at least a subset of that syntax). Also, if we determine we can get rid of the `clientConfigFormat` thing, I'd suggest an API like
```ts
backend.setCustomOutput('geo.amazon_location_service.search_indices[0].myTestSearchIndex', placeIndex.placeIndexName)
```

packages/custom-config-poc-app/amplify/backend.ts
edwardfoyle: I think this is better DX if you are trying to specify or override lots of config values, but is actually more cumbersome if you're just trying to add or override one or two things.

That being said, I think I like this DX better than specifying the key path. It reads more like a config object than just setting a bunch of values.

It definitely presents some challenges in terms of serializing all of this to some sort of CFN output that can be reconstructed later, but I think that's solvable. Although I'm not sure how we'd deal with using CDK tokens as object keys...

packages/client-config/src/client-config-contributor/custom_client_config_contributor.ts
```diff
@@ -0,0 +1,68 @@
...
+      return {};
+    }
+
+    console.log(customOutputs);
+
+    const result: CustomClientConfig & Record<string, unknown> = { custom: {} };
+
+    Object.entries(customOutputs.payload).forEach(([key, value]) => {
+      const customOutputData = JSON.parse(value);
+      // TODO how do we do custom namespacing ??
```
edwardfoyle, line 35: Is the concern here that a custom key name may collide with one of our internal names? CDK will fail if you try to add multiple CFN outputs with the same name. But if we want to allow this behavior, then yes we'd need some sort of namespacing. That could probably go as a prerequisite PR to the rest of this change.

packages/client-config/src/client-config-contributor/custom_client_config_contributor.ts
```diff
@@ -0,0 +1,68 @@
...
+      result.custom[key.replace('amplifycustom', '')] = customOutputData.value;
+
+      const customOutputOptions: CustomOutputOptions | undefined = customOutputData.options;
+
+      if (customOutputOptions){
+        // TODO should this be in addition to placing value in custom.X section or should these be XOR ?
```
edwardfoyle, line 41: I would suggest XOR

packages/client-config/src/client-config-contributor/custom_client_config_contributor.ts
```diff
@@ -0,0 +1,68 @@
...
+          let currentArrayRef: Array<unknown> | undefined = undefined;
+          for (let i = 0 ; i < clientConfigDestination.path.length - 1; i++) {
+            const pathSegment = clientConfigDestination.path[i];
+            if (!currentRef[pathSegment]) {
+              currentRef[pathSegment] = {}
+            }
+            currentRef = currentRef[pathSegment] as Record<string, unknown>;
+          }
+          const lastPathSegment = clientConfigDestination.path[clientConfigDestination.path.length - 1];
+          // TODO this is a hack
```
edwardfoyle, line 53: recent CVE not withstanding, [lodash set](https://lodash.com/docs/4.17.15#set) could be the right tool for the job here

packages/backend/src/backend_factory.ts
edwardfoyle: 3. It should be. If we change anything about the format within the "AWS::Amplify::Custom" block we'd need to increment that version. That would cover any changes to those serialized JSON objects.

packages/client-config/src/client-config-contributor/custom_client_config_contributor.ts
edwardfoyle: Only thing I don't see here is how existing config values can be overwritten. Seems like we might need to have this custom contributor run last in the "contributor pipeline" so that it can apply customer specified overrides to existing values. Or maybe there's some other mechanism we're going to use for that?

sobolk: Yep. It's a position of the contributor in the pipeline (being last) that makes it work in "override scenario".

And that can remain as a solution if we decide that "path" should be JS path (or whatever we think the unified format is at given point of time).

This can get complicate if we allow "path per config type", in that case we'd need to build some mechanism to override js->other type result. (but maybe we'll never get there, so I not implemented that yet).

packages/custom-config-poc-app/amplify/backend.ts
sobolk: Even with "path" implementation at some point I started to put result of `JSON.stringify()` that contained CDK tokens into Outputs. 

I got positively surprised that it worked (see sample on receiving end https://github.com/aws-amplify/amplify-backend/pull/946#discussion_r1466982659 ).

This made me thing that we can just put serialized JSON with CDK tokens and they'll get replaced at some point in synth/deploy process.


As for dealing with breaking this down I think we could make this work in such a way that:
```
backend.addToClientConfig({
  custom: {
    myApiKey: apiKeyValue,
  },
  });
  
 backend.addToClientConfig({
  custom: {
    myApiUrl: api.url,
  },
  }); 
```

keeps accumulating stuff until synthesis happens (this might need a bit of implementation trickery but should be doable).

packages/custom-config-poc-app/amplify/backend.ts
sobolk: The immediate problem that I bumped into with this is CFN validation for output names(keys), that I mention in code comments attached to https://github.com/aws-amplify/amplify-backend/pull/946#discussion_r1466991617 .

So either this api has to take both cfn output key/name and target config path. Or we'll need to generate cfn output key/name under the hood. OR just put everything into one big json and put under reserved output key.

packages/client-config/src/client-config-contributor/custom_client_config_contributor.ts
sobolk: Yes, that is the concern.
Which also raises a question if we should be namespacing known verticals outputs. Today we have only auth and data but at some point there will be more.

packages/client-config/src/client_config_generator_factory.ts
```diff
@@ -32,6 +33,7 @@ export class ClientConfigGeneratorFactory {
       new AuthClientConfigContributor(),
       new GraphqlClientConfigContributor(modelSchemaAdapter),
       new StorageClientConfigContributor(),
+      new CustomClientConfigContributor(),
     ]);
```
sobolk, line 37: This contributor being last makes "override known keys" scenario work.

packages/client-config/src/client-config-contributor/custom_client_config_contributor.ts
sobolk: https://github.com/aws-amplify/amplify-backend/pull/946#discussion_r1467159112

packages/custom-config-poc-app/amplify/backend.ts
sobolk: Thinking about this more.
Perhaps we just keep stashing json serialized result of calls to `backend.setCustomOutput / addToClientConfig` into stack outputs that we name internally (i.e. `customOutput1, customOutput2 ...`) and apply them in "custom contributor" sequentially then merge result into resulting config object.

packages/client-config/src/client-config-contributor/custom_client_config_contributor.ts
sobolk: However, then if customer passes an output key to `backend.addCustomOutput` they'd expect it to see in CFN console I guess.

So perhaps we shouldn't namespace and let them figure out potential collisions and rely on CFN validation.
Regardless of that I think we should do namespacing for our verticals (at least those we own) to minimize collision probability.

packages/backend/src/backend_factory.ts
```diff
@@ -108,6 +120,51 @@ export class BackendFactory<
...
+
+    const outputData = JSON.stringify(clientConfigPart);
+    this.customOutputsEntryAlternative.appendOutput(
+      `alternativeCustomApproach${this.customOutputsAlternativeEntryCount}`,
+      outputData
+    );
+    this.customOutputsAlternativeEntryCount++;
+  };
 }
```
sobolk, line 168: Added a sample of how this could work.

A sample output looks like this:
```
{"myApiUrl":"https://REDACTED.execute-api.us-west-2.amazonaws.com/prod/","myApiKey":"REDACTED"}}
```

packages/custom-config-poc-app/amplify/backend.ts
edwardfoyle: I suppose CDK is putting CFN Sub parameters in the output values which are resolved at deploy time.

Another thought I had was rather than stuffing metadata in the output (which is supposed to just be the "value" of the output), perhaps we could extend the metadata section of the CFN template to include metadata on which path a particular config value is supposed to be written to.

One downside of this is I'm not sure the automagic CDK token resolution will work in stack metadata. But it would have the benefit of not having to stuff serialized objects in CFN output

packages/custom-config-poc-app/amplify/backend.ts
edwardfoyle: Yeah that could work although I think I would rather stuff this metadata into the stack metadata rather than outputs. See https://github.com/aws-amplify/amplify-backend/pull/946#discussion_r1468120158

Currently we store a mapping of which stack outputs belong to which categories. This would be an extension of that mapping where we specify a config path for a particular CFN output

packages/client-config/src/client-config-contributor/custom_client_config_contributor.ts
edwardfoyle: FWIW, the only output I have seen that has any chance of collision is "region". All of our other output names are highly domain-specific.

Personally I think it's fine to not namespace and if there's a collision synth will fail. Then you just change your output to a slightly different name and carry on.

sobolk: that solves custom outputs.

However, what I'm worried about is that some day 5th category is implemented by some team and never tested with `data`. Then attempt to use that new category with `data` will result in permanent conflict and p0 bug ...

packages/custom-config-poc-app/amplify/backend.ts
sobolk: This is how output section in template looks like. I'll give it (metadata) a try but my hopes are low.
image

sobolk: another limitation :
this will go south if output key doesn't change but config destination does (just happened in my stack :( )
image

sobolk: as expected.
image

sobolk: (note to myself - I rotated the key from one of the screenshots above, it's no longer valid)

packages/client-config/src/client-config-contributor/custom_client_config_contributor.ts
edwardfoyle: Yeah that's a fair callout

packages/custom-config-poc-app/amplify/backend.ts
```diff
@@ -0,0 +1,219 @@
...
+// So the limit seems to be around 450000 characters
+
+// client config from create-amplify template has 3342 bytes.
+
+// const veryLongOutput2 = 'a'.repeat(450000);
+// new CfnOutput(locationStack, 'howbigcanoutputbe2', {
+//   value: veryLongOutput2,
+// });
+
+// ####### How big can this be End ######
```
sobolk, line 219: Looks like all outputs combined per stack can have 450000 characters (bytes, if dumped to file size is same).
Sample client config used in this app is 1607 bytes
Client config from `create-amplify` app is 3342 bytes.

So it seem outputs can hold ~134 copies of config from `create-amplify`

packages/custom-config-poc-app/amplify/backend.ts
```diff
@@ -0,0 +1,254 @@
...
+      path: 'geo.amazon_location_service.search_indices[0]',
+    },
+    {
+      clientConfigFormat: ClientConfigFormat.JSON_MOBILE,
+      path: 'geo.amazon_location_service.some_other_path',
+    },
+  ],
+});
+
+// ###### More DX samples End #######
```
sobolk, line 254: More samples for different solution options

packages/backend/src/backend.ts
```diff
@@ -1,10 +1,56 @@
...
 export type BackendBase = {
   createStack: (name: string) => Stack;
+  setCustomOutput: (key: string, value: string, options?: CustomOutputOptions) => void;
+  setOutput: (clientConfigPart: Partial<ClientConfig>) => void;
+
+  addOutput1: (clientConfigPart: Partial<ClientConfig>) => void;
+  addOutput2: (clientConfigPart: AuthClientConfig | GraphqlClientConfig | CustomClientConfig) => void;
+  addOutput3: (path: string, value: string) => void;
+  addOutput4: (outputId: string, value: string, options?: OutputOptions) => void;
 };
```
sobolk, line 20: couple of backend api proposals.

packages/custom-config-poc-app/amplify/backend.ts
```diff
@@ -0,0 +1,254 @@
...
+import { ClientConfigFormat } from '@aws-amplify/client-config';
+
+const backend = defineBackend({
+  auth,
+  myFunc,
+});
+
+// ####### API Gateway + Lambda Start #######
+
+const apiGatewayStack = backend.createStack('APIGateway');
```
johnpc, line 19: Why create a new stack? Why not reuse the stack with something like `backend.myFunc.resources.lambda.stack`?

I worry that using a new named stack might conflict between sandboxes / amplify-deployed cdk?

sobolk: This code isn't going to be part of solution.
It's just piece of POC Gen2 app that I used for testing.
Both approaches are possible.

sobolk:
Final implementation here https://github.com/aws-amplify/amplify-backend/pull/978

PR #952, opened 2024-01-26
adds info command

ykethan:
## Problem

## Changes

Information includes: System info, Binaries, select NPM Packages, non sensitive AWS and CDK environment variables.
This adds a package called backend-diagnose that runs envinfo and cdk doctor to collect relevant info.

image

## Checklist

### DIFF
--- a/packages/cli/src/commands/info/info_command.ts
+++ b/packages/cli/src/commands/info/info_command.ts
@@ -1,0 +1,51 @@
+import * as os from 'node:os';
+import { Argv, CommandModule } from 'yargs';
+import {
+  formatCdkInfo,
+  formatEnvInfo,
+  getCdkInfo,
+  getEnvInfo,
+} from '../../info/index.js';
+import { LogLevel, Printer } from '@aws-amplify/cli-core';
+
+/**
+ * Represents the InfoCommand class.
+ */
+export class InfoCommand implements CommandModule<object> {
+  /**
+   * @inheritDoc
+   */
+  readonly command: string;
+
+  /**
+   * @inheritDoc
+   */
+  readonly describe: string;
+
+  /**
+   * Creates top level entry point for generate command.
+   */
+  constructor() {
+    this.command = 'info';
+    this.describe = 'Generates information for Amplify troubleshooting';
+  }
+
+  /**
+   * @inheritDoc
+   */
+  handler = async (): Promise<void> => {
+    const printer = new Printer(LogLevel.INFO);
+    const envInfo = await getEnvInfo();
+    const formattedEnvInfo: string = formatEnvInfo(envInfo);
+    const cdkInfo = await getCdkInfo();
+    const formattedCdkInfo: string = formatCdkInfo(cdkInfo);
+    printer.print(`${formattedEnvInfo}${os.EOL}${formattedCdkInfo}`);
+  };
+
+  /**
+   * @inheritDoc
+   */
+  builder = (yargs: Argv): Argv => {
+    return yargs.version(false).help();
+  };
+}
--- a/packages/cli/src/info/cdk_info.ts
+++ b/packages/cli/src/info/cdk_info.ts
@@ -1,0 +1,63 @@
+import * as os from 'node:os';
+import { execa as _execa } from 'execa';
+import { indent } from './indent.js';
+
+/**
+ * Format CDK information.
+ * @param execa - The execa to run the cdk command.
+ * @returns The cdk doctor output.
+ */
+export const getCdkInfo = async (execa = _execa): Promise<string> => {
+  const cdkDoctorArgs: string[] = ['cdk', 'doctor', '--', ' --no-color'];
+  let output;
+  try {
+    // type-cast the value to string, we know it will be a string given the options
+    output = (
+      await execa('npx', cdkDoctorArgs, {
+        stderr: 'pipe',
+      })
+    ).stderr;
+  } catch (error) {
+    // should this be replaced with an AmplifyError
+    throw new Error(`Failed to run command: ${error as string}`);
+  }
+  return output;
+};
+
+/**
+ * Format CDK information.
+ * @param info - The CDK information to format.
+ * @returns The formatted CDK information.
+ */
+export const formatCdkInfo = (info: string) => {
+  const sensitiveKeys = [
+    'AWS_ACCESS_KEY_ID',
+    'AWS_SECRET_ACCESS_KEY',
+    'AWS_SESSION_TOKEN',
+  ];
+
+  const lines = info.split(/\r\n|\r|\n/);
+  const formattedLines = lines.map((line) => {
+    //removes emoji from output
+    let formattedLine = line.replace(/[^\x20-\x7E]/g, '').trim();
+
+    const containsSensitiveInfo = sensitiveKeys.some((key) =>
+      formattedLine.includes(key)
+    );
+    if (containsSensitiveInfo) {
+      return null;
+    }
+
+    if (formattedLine.startsWith('AWS_') || formattedLine.startsWith('CDK_')) {
+      formattedLine = indent(formattedLine);
+    } else if (formattedLine.startsWith('- ')) {
+      formattedLine = indent(formattedLine.substring(2));
+    }
+
+    return formattedLine;
+  });
+
+  return formattedLines
+    .filter((line) => line !== null && !line.startsWith('CDK Version'))
+    .join('\n');
+};
--- a/packages/cli/src/info/env_info.ts
+++ b/packages/cli/src/info/env_info.ts
@@ -1,0 +1,91 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import * as os from 'node:os';
+import envinfo from 'envinfo';
+import { indent } from './indent.js';
+
+export type EnvInfoBinary = {
+  version: string;
+  path: string;
+};
+
+export type EnvInfoNpmPackage =
+  | {
+      wanted: string;
+      installed: string;
+    }
+  | string;
+
+// write this manually given types are not available
+export type EnvInfo = {
+  System: {
+    OS: string;
+    CPU: string;
+    Memory: string;
+    Shell: EnvInfoBinary;
+  };
+  Binaries: {
+    Node: EnvInfoBinary;
+    Yarn: EnvInfoBinary;
+    npm: EnvInfoBinary;
+    pnpm: EnvInfoBinary;
+  };
+  npmPackages: Record<string, EnvInfoNpmPackage>;
+};
+
+/**
+ * Get environment information
+ */
+export const getEnvInfo = async (): Promise<EnvInfo> => {
+  const info = await envinfo.run(
+    {
+      System: ['OS', 'CPU', 'Memory', 'Shell'],
+      Binaries: ['Node', 'Yarn', 'npm', 'pnpm'],
+      npmPackages: [
+        '@aws-amplify/backend',
+        '@aws-amplify/backend-cli',
+        'typescript',
+        'aws-cdk',
+        'aws-cdk-lib',
+        'aws-amplify',
+      ],
+      // npmGlobalPackages: ['@aws-amplify/cli'],
+    },
+    { json: true, showNotFound: true }
+  );
+  return JSON.parse(info);
+};
+
+/**
+ * Format environment information.
+ * @param info - The environment information.
+ * @returns The formatted environment information.
+ */
+export const formatEnvInfo = (info: EnvInfo) => {
+  const system = [
+    'System:',
+    ...Object.entries(info.System).map(([part, details]) => {
+      if (typeof details !== 'string') {
+        return indent(`${part}: ${details.path}`);
+      }
+      return indent(`${part}: ${details}`);
+    }),
+  ];
+  const binaries = [
+    'Binaries:',
+    ...Object.entries(info.Binaries).map(([name, binary]) => {
+      return indent(`${name}: ${binary.version} - ${binary.path}`);
+    }),
+  ];
+  const npmPackages = [
+    'NPM Packages:',
+    ...Object.entries(info.npmPackages).map(([name, details]) => {
+      if (typeof details !== 'string') {
+        return indent(`${name}: ${details.installed}`);
+      }
+      return indent(`${name}: ${details}`);
+    }),
+  ];
+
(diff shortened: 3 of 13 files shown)


packages/cli/src/commands/info/info_command.ts
```diff
@@ -0,0 +1,51 @@
...
+   * @inheritDoc
+   */
+  handler = async (): Promise<void> => {
+    const printer = new Printer(LogLevel.INFO);
+    const envInfo = await getEnvInfo();
+    const formattedEnvInfo: string = formatEnvInfo(envInfo);
+    const cdkInfo = await getCdkInfo();
+    const formattedCdkInfo: string = formatCdkInfo(cdkInfo);
+    printer.print(`${formattedEnvInfo}${os.EOL}${formattedCdkInfo}`);
+  };
```
sobolk, line 43: Please refactor this in such a way that:
1. Classes/types are created and injected into command for testability. See how `ConfigureProfileCommand` is implemented for example. Some ideas `class LocalEnvironmentDetailsReader`, `type LocalEnvironmentDetails { environment: {}, cdk: {}`, `class LocalEnvironmentDetailsFormatter`.
2. Please avoid acronyms when possible. I.e. CDK is fine, but `environment` could be better.
3. Printer should also be injected, there should be an instance somewhere. (Look around in cli package to see the pattern).

packages/cli/src/info/cdk_info.ts
```diff
@@ -0,0 +1,62 @@
...
+export const getCdkInfo = async (execa = _execa): Promise<string> => {
+  const cdkDoctorArgs: string[] = ['cdk', 'doctor', '--', ' --no-color'];
+  let output;
+  try {
+    // type-cast the value to string, we know it will be a string given the options
+    output = (
+      await execa('npx', cdkDoctorArgs, {
+        stderr: 'pipe',
+      })
+    ).stderr;
```
sobolk, line 18: is output going to stderr ?
noting to stdout?

perhaps we could use `all` to combine streams to single output, see https://github.com/sindresorhus/execa?tab=readme-ov-file#all-2 ?

packages/cli/src/info/cdk_info.ts
```diff
@@ -0,0 +1,62 @@
...
+  } catch (error) {
+    // should this be replaced with an AmplifyError
+    throw new Error(`Failed to run command: ${error as string}`);
+  }
```
sobolk, line 22: Please remove try catch block. This error handling doesn't bring any value.
If cdk doctor doesn't work this would be a fault not AmplifyError anyway.

packages/cli/src/info/env_info.ts
```diff
@@ -0,0 +1,91 @@
+/* eslint-disable @typescript-eslint/naming-convention */
```
sobolk, line 1: What is the reason to not follow naming convention ?

if this is because we add types for output we don't controll then please move types to separate file.

packages/cli/src/info/indent.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Adds two spaces of indentation to the given line.
+ * @param line The line to be indented.
+ * @returns The indented line.
+ */
+export const indent = (line: string) => `  ${line}`;
```
sobolk, line 6: this should be printer functionality.

packages/cli/src/info/index.ts
```diff
@@ -0,0 +1,8 @@
+export { getCdkInfo, formatCdkInfo } from './cdk_info.js';
+export {
+  getEnvInfo,
+  formatEnvInfo,
+  EnvInfo,
+  EnvInfoBinary,
+  EnvInfoNpmPackage,
+} from './env_info.js';
```
sobolk, line 8: I'd remove this file. We don't do `index.ts` for internal components.

packages/cli/src/commands/info/info_command.ts
edwardfoyle: Specifically, [this printer instance](https://github.com/aws-amplify/amplify-backend/blob/main/packages/cli/src/printer.ts) should be used

packages/cli/src/info/indent.ts
edwardfoyle: This could also be another argument for the "format" object I suggested [here](https://github.com/aws-amplify/amplify-backend/pull/905#discussion_r1449321497). Then usage would be something like `printer.print(format.indent('stuff'))`

packages/cli-core/API.md
```diff
@@ -38,6 +38,7 @@ export enum LogLevel {
 // @public
 export class Printer {
     constructor(minimumLogLevel: LogLevel, stdout?: NodeJS.WriteStream, stderr?: NodeJS.WriteStream, refreshRate?: number);
+    format: (message: string, indent: number) => string;
```
sobolk, line 41: Could you please follow @edwardfoyle 's suggestion from https://github.com/aws-amplify/amplify-backend/pull/952#discussion_r1468129526 ? I think his proposal is good and will scale better in the future when we add more "formatting options".

packages/cli/src/info/cdk_info.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Provides information about CDK.
+ */
+export class CdkInfoProvider {
+  /**
+   * Format CDK information.
+   * @param execa - The execa to run the cdk command.
+   * @returns The cdk doctor output.
+   */
+  async getCdkInfo(execa = _execa): Promise<string> {
```
sobolk, line 13: if this is returning `string` then could we please make `formatCdkInfo` private and return formatted output from here?
Is there any case we'd want unformatted output somewhere ?

packages/cli/src/info/cdk_info.ts
```diff
@@ -0,0 +1,63 @@
+import { execa as _execa } from 'execa';
+import { printer } from '../printer.js';
+
+/**
+ * Provides information about CDK.
+ */
+export class CdkInfoProvider {
```
sobolk, line 7: please adjust file name after refactoring.

packages/cli/src/info/env_info.ts
```diff
@@ -0,0 +1,67 @@
...
+    );
+    return JSON.parse(info);
+  }
+
+  /**
+   * Format environment information.
+   * @param info - The environment information.
+   * @returns The formatted environment information.
+   */
+  formatEnvInfo(info: EnvInfo): string {
```
sobolk, line 38: if this class is called `EnvironmentInfoProvider` then this method should be extracted out from here.
Can you add `class EnvironmentInfoFormatter` so we can have `environmentInforFormatter.format(envInvo)` ?

packages/cli/src/info/env_info_types.ts
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+export type EnvInfoBinary = {
```
sobolk, line 2:
```suggestion
// Types in this file are for return type from 'envinfo' package. We don't own schema.
/* eslint-disable @typescript-eslint/naming-convention */
export type EnvInfoBinary = {
```

Please add comment explaining why.

sobolk (commented):
This is looking much better.

packages/cli/src/info/cdk_info.ts
edwardfoyle: Also, execa should be injected into the ctor rather than the method (it's fine to still have the default value since it's only used for testing)

packages/cli/src/info/env_info.ts
```diff
@@ -0,0 +1,67 @@
...
+        Binaries: ['Node', 'Yarn', 'npm', 'pnpm'],
+        npmPackages: [
+          '@aws-amplify/backend',
+          '@aws-amplify/backend-cli',
+          'typescript',
+          'aws-cdk',
+          'aws-cdk-lib',
+          'aws-amplify',
+        ],
+        // npmGlobalPackages: ['@aws-amplify/cli'],
```
edwardfoyle, line 26: remove if not needed (this shouldn't be needed)

packages/cli/src/info/env_info.ts
edwardfoyle: Or similar to Kamil's comment above, if we don't have a need for unformatted output, just have `EnvironmentInfoProvider.getFormattedEnvInfo` or such.

packages/cli/src/commands/info/info_command.ts
```diff
@@ -0,0 +1,51 @@
...
+  readonly describe: string;
+
+  /**
+   * Creates top level entry point for generate command.
+   */
+  constructor(
+    private readonly environmentInfoProvider: EnvironmentInfoProvider,
+    private readonly cdkInfoProvider: CdkInfoProvider
+  ) {
+    this.command = 'info';
```
edwardfoyle, line 28: just to play devil's advocate, is there anything else we may want to put behind an "info" command in the future? Maybe info about the currently deployed backend, or info about a local diff. I'm not opposed to `amplify info` just want to make sure we aren't backing ourselves in a corner. We should also make sure we have PM alignment on the name if we haven't already done that.

josefaidt: `info` seems fine but to Ed's note I could see this growing or us wanting something else/more later. We've had thoughts of `diagnose`/`doctor`/something similar, and the motivation for `info` is to ease env info collection for filing issues. Maybe this is something that's rolled up in to a similar command?
```
amplify diagnose --env-info
```

packages/cli-core/API.md
ykethan: didn't add the ones with color might be better to add them as part of https://github.com/aws-amplify/amplify-backend/pull/905

packages/cli/src/commands/info/info_command.ts
edwardfoyle: Sounds like `info` is probably good for now. We can add args / deprecate / shuffle things around in future major versions if necessary

packages/cli-core/src/format/format.ts
```diff
@@ -0,0 +1,19 @@
+import * as os from 'node:os';
+
+/**
+ * Provides formatting utilities.
+ */
+export const format = {
+  list: (lines: string[]) =>
+    lines.map((line: string) => ` - ${line}`).join(os.EOL),
+  indent: (message: string, indentLevel: number) => {
```
edwardfoyle, line 9: We can omit `indentLevel` and just always indent by 2 spaces. If you want to indent more, you chain calls to indent together.

packages/cli-core/src/format/format.ts
```diff
@@ -0,0 +1,19 @@
...
+    if (indentLevel < 0) {
+      return message;
+    }
+    const spaces = ' '.repeat(indentLevel);
+    return message
+      .split('\n')
```
edwardfoyle, line 15: to make this platform agnostic, we should replace `os.EOL` with `\n`, then split at `\n` and rejoin with `os.EOL`

packages/cli-core/src/format/format.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    const lines = ['Item 1', 'Item 2', 'Item 3'];
+    const expectedOutput = ` - Item 1${os.EOL} - Item 2${os.EOL} - Item 3`;
+    const actualOutput = format.list(lines);
+    assert.strictEqual(actualOutput, expectedOutput);
+  });
+
+  void test.it(
+    'should indent the message with the specified number of spaces',
+    () => {
+      const message = 'Hello\nWorld';
```
edwardfoyle, line 17: use `os.EOL` so that our test matrix will ensure cross-platform support

packages/cli/src/info/cdk_info_provider.test.ts
```diff
@@ -0,0 +1,43 @@
+import * as assert from 'node:assert';
+import * as test from 'node:test';
+import { execa } from 'execa';
+import { CdkInfoProvider } from './cdk_info_provider.js';
+
+void test.describe('CDK Info', () => {
+  const mockValue: string = `
+ℹ️ CDK Version: 2.110.1 (build 0d37f0d)
+ℹ️ AWS environment variables:
+  - AWS_SESSION_TOKEN = <redacted>
```
edwardfoyle, line 10: we should include all sensitive keys in the test payload and ensure that they are all removed in the output

packages/cli/src/info/cdk_info_provider.ts
```diff
@@ -0,0 +1,67 @@
...
+   */
+  private formatCdkInfo(info: string): string {
+    const sensitiveKeys = [
+      'AWS_ACCESS_KEY_ID',
+      'AWS_SECRET_ACCESS_KEY',
+      'AWS_SESSION_TOKEN',
+    ];
+
+    const lines = info.split(/\r\n|\r|\n/);
+    const formattedLines = lines.map((line) => {
```
edwardfoyle, line 40: we should start with filtering out sensitive keys before anything else
```suggestion
    const formattedLines = lines
      .filter(line => !sensitiveKeys.some(key => line.includes(key)))
      .map(line => {
        // rest of stuff
      })
```

ykethan: image

by default the cdk command already redact the info on sensitive variables. filtering the variables is to ensure we have a cleaner output.

edwardfoyle (approved):
LGTM!

edwardfoyle (approved):

packages/cli-core/src/format/format.test.ts
```diff
@@ -0,0 +1,34 @@
+import * as os from 'node:os';
+import * as assert from 'node:assert';
+import { describe, it } from 'node:test';
+import { format } from './format.js';
+
+void describe('format utilities', () => {
```
sobolk, line 6:
```suggestion
void describe('format', () => {
```

packages/cli-core/src/format/format.ts
```diff
@@ -0,0 +1,19 @@
+import * as os from 'node:os';
+
+/**
+ * Provides formatting utilities.
```
sobolk, line 4:
```suggestion
 * Formats various inputs into single string.
```

sobolk (changes requested):
https://www.google.com/search?q=utils+code+smell

sobolk (dismissed):

sobolk (approved):

edwardfoyle (approved):

PR #974, opened 2024-02-01
fix: Flatten OIDC props to eliminate importing OidcAttributeRequestMethod

awsluja:
## Problem
Users would need to import OidcAttributeRequestMethod to specify values for attributeRequestMethod.

## Changes
The attributeRequestMethod type has been converted to a string union in order to eliminate the need to import OidcAttributeRequestMethod.

## Validation
Tests updated to ensure that mapping accounts for all properties and translates string union back to OidcAttributeRequestMethod under the hood.

## Checklist

### DIFF
--- a/packages/auth-construct/src/construct.test.ts
+++ b/packages/auth-construct/src/construct.test.ts
@@ -1230,6 +1230,13 @@ void describe('Auth construct', () => {
     void it('supports oidc and email', () => {
       const app = new App();
       const stack = new Stack(app);
+      const authorizationURL = 'http://localhost:3000/authorization';
+      const jwksURL = 'https://localhost:3000/jwksuri';
+      const tokensURL = 'http://localhost:3000/token';
+      const userInfoURL = 'http://localhost:3000/userinfo';
+      const mockIdentifiers = ['one', 'two'];
+      const mockScopes = ['scope1', 'scope2'];
+      const attributeRequestMethod = 'POST';
       new AmplifyAuth(stack, 'test', {
         loginWith: {
           email: true,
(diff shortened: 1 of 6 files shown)


packages/auth-construct/src/construct.ts
```diff
@@ -696,14 +697,29 @@ export class AmplifyAuth
...
+        oidc.attributeRequestMethod === undefined
+          ? 'GET' // default if not defined
+          : oidc.attributeRequestMethod;
       result.oidc = new cognito.UserPoolIdentityProviderOidc(
         this,
         `${this.name}OidcIDP`,
         {
           userPool,
-          ...external.oidc,
+          attributeRequestMethod:
```
0618, line 710: if we do the `const oidc = {...external.oidc}`, we can do `...oidc` here to make it easier to read.

awsluja: we can't do ...oidc; we are remapping property types here

edwardfoyle (approved):

packages/auth-construct/src/construct.test.ts
```diff
@@ -1250,10 +1271,106 @@ void describe('Auth construct', () => {
...
+          authorize_url: authorizationURL,
+          client_id: oidcClientId,
+          client_secret: oidcClientSecret,
+          jwks_uri: jwksURI,
+          oidc_issuer: oidcIssuerUrl,
+          token_url: tokensURL,
+        },
+        ProviderName: oidcProviderName,
+        ProviderType: 'OIDC',
+      });
```
sobolk, line 1373: the inconsistency of casing of these props makes me cry.
but I guess we have no power here.
should we just craft our types?

packages/auth-construct/src/types.ts
```diff
@@ -139,8 +139,19 @@ export type FacebookProviderProps = Omit<
  */
 export type OidcProviderProps = Omit<
   cognito.UserPoolIdentityProviderOidcProps,
-  'userPool'
->;
+  'userPool' | 'attributeRequestMethod'
+> & {
```
sobolk, line 143: maybe instead of fixing cdk types we should start investing in our types + mapping?
this call is easier to make pre-GA.

sobolk (approved):
LGTM.
But... we should re-consider our strategy for cdk types... see comments.

packages/auth-construct/src/construct.test.ts
awsluja: these come from the cfn templates; I don't think we have control over that

packages/auth-construct/src/types.ts
awsluja: Originally did this to reduce maintenance overhead, we save a lot on types by doing this specially for oauth

PR #984, opened 2024-02-02
generate type definition file for static environment variables for functions

rtpascual:
## Problem

Customers may not be aware of environment variables available for Lambda runtimes when writing function code.

## Changes

Use `node.addValidation` as a mechanism to generate a type definition file for each function after synthesis
- The file is generated as `/.amplify/function-env/<function-name>.ts`
- Add paths to `tsconfig.json` generated by `create amplify` so customers can import types with `import { env } from '@env/<function-name>';`

Type definition file contents:
```ts
/** Lambda runtime environment variables, see https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime */
export const env = process.env as {
/** The handler location configured on the function. */
_HANDLER: string;

/** The X-Ray tracing header. This environment variable changes with each invocation. */
_X_AMZN_TRACE_ID: string;

/** The default AWS Region where the Lambda function is executed. */
AWS_DEFAULT_REGION: string;

/** The AWS Region where the Lambda function is executed. If defined, this value overrides the AWS_DEFAULT_REGION. */
AWS_REGION: string;

/** The runtime identifier, prefixed by AWS_Lambda_ (for example, AWS_Lambda_java8). */
AWS_EXECUTION_ENV: string;

/** The name of the function. */
AWS_LAMBDA_FUNCTION_NAME: string;

/** The amount of memory available to the function in MB. */
AWS_LAMBDA_FUNCTION_MEMORY_SIZE: string;

/** The version of the function being executed. */
AWS_LAMBDA_FUNCTION_VERSION: string;

/** The initialization type of the function, which is on-demand, provisioned-concurrency, or snap-start. */
AWS_LAMBDA_INITIALIZATION_TYPE: string;

/** The name of the Amazon CloudWatch Logs group for the function. */
AWS_LAMBDA_LOG_GROUP_NAME: string;

/** The name of the Amazon CloudWatch Logs stream for the function. */
AWS_LAMBDA_LOG_STREAM_NAME: string;

/** AWS access key. */
AWS_ACCESS_KEY: string;

/** AWS access key ID. */
AWS_ACCESS_KEY_ID: string;

/** AWS secret access key. */
AWS_SECRET_ACCESS_KEY: string;

/** AWS Session token. */
AWS_SESSION_TOKEN: string;

/** The host and port of the runtime API. */
AWS_LAMBDA_RUNTIME_API: string;

/** The path to your Lambda function code. */
LAMBDA_TASK_ROOT: string;

/** The path to runtime libraries. */
LAMBDA_RUNTIME_DIR: string;

/** The locale of the runtime. */
LANG: string;

/** The execution path. */
PATH: string;

/** The system library path. */
LD_LIBRARY_PATH: string;

/** The Node.js library path. */
NODE_PATH: string;

/** For X-Ray tracing, Lambda sets this to LOG_ERROR to avoid throwing runtime errors from the X-Ray SDK. */
AWS_XRAY_CONTEXT_MISSING: string;

/** For X-Ray tracing, the IP address and port of the X-Ray daemon. */
AWS_XRAY_DAEMON_ADDRESS: string;

/** The environment's time zone. */
TZ: string;
};
```
DX:
Screenshot 2024-02-20 at 8 35 42 PM
Screenshot 2024-02-20 at 8 36 35 PM

## Validation

Local project with type definition file generated:
- Confirmed DX from types
- Lambda executes successfully

## Checklist


packages/backend-function/src/function_env_type_generator.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  const fsWriteFileSyncMock = mock.method(fs, 'writeFileSync', () => null);
+
+  beforeEach(() => {
+    fdCloseMock.mock.resetCalls();
+    fsOpenSyncMock.mock.resetCalls();
+    fsMkdirSyncMock.mock.resetCalls();
+    fsWriteFileSyncMock.mock.resetCalls();
+  });
+
+  void it('generates a type definition file', () => {
```
edwardfoyle, line 19: we should also have a test that actually writes the file and imports it to verify that it is valid syntax. I think it's okay for that to be a unit test even though it's doing some file i/o because it should be very fast

packages/backend-function/src/function_env_type_generator.ts
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Generates a type definition file for environment variables
+ */
+export class FunctionEnvironmentTypeGenerator {
+  /**
+   * Lambda runtime environment variables and their descriptions.
+   * See https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime
+   */
+  private staticEnvironmentVariables: Record<string, string> = {
```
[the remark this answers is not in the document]
rtpascual: I'll separate this out to its own file

packages/backend-function/src/function_env_type_generator.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    for (const key in this.staticEnvironmentVariables) {
+      const comment = `/** ${this.staticEnvironmentVariables[key]} */`;
+      const declaration = `${key}: string;`;
+
+      declarations.push(`${comment}\n${declaration}`);
+    }
+
+    const content =
+      'export const env = process.env as {\n' +
```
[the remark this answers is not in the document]
rtpascual: We can but this will eventually also have env vars they define for the function and resource access related env vars. I guess we can update the JS doc when those changes are added.

packages/sandbox/src/file_watching_sandbox.ts
```diff
@@ -47,7 +47,7 @@ export const getBootstrapUrl = (region: string) =>
  */
 export class FileWatchingSandbox extends EventEmitter implements Sandbox {
   private watcherSubscription: Awaited<ReturnType<typeof subscribe>>;
-  private outputFilesExcludedFromWatch = ['.amplify'];
+  private outputFilesExcludedFromWatch = ['.amplify', '**/amplify/*_env.ts'];
```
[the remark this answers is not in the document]
rtpascual: I can try putting something together for this

packages/backend-function/src/function_env_type_generator.test.ts
rtpascual: Ok I wasn't sure about this one which was why I just tested if one of the static env vars is passed to `writeFileSync`

packages/sandbox/src/file_watching_sandbox.ts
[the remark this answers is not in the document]
rtpascual: Another thing is the entry to `.gitignore` has the coupling and would have to be updated if we ever make changes. This one is more ok because developers will notice if the type def files stop being ignored.

.gitignore
```diff
@@ -21,3 +21,4 @@ e2e-tests
 concurrent_workspace_script_cache.json
 
 testDir
+**/amplify/*_env.ts
```
edwardfoyle, line 24: Do we need this in our gitignore? Shouldn't the gitignore of `test-projects/*` cover this?

packages/platform-core/src/function_type_def_convention_provider.ts
```diff
@@ -0,0 +1,24 @@
...
+    functionName: string
+  ): string {
+    return `${path.dirname(functionEntryPath)}/amplify/${functionName}_env.ts`;
+  }
+
+  /**
+   * Get the ignore pattern for type definition files
+   */
+  static getFunctionTypeDefIgnorePattern(): string {
+    // returns '**/<typeDefParentFolderName>/*<typeDefFileNameSuffix>.ts'
```
edwardfoyle, line 21:
```suggestion
    // returns '**/amplify/*<typeDefFileNameSuffix>.ts'
```

packages/platform-core/src/function_type_def_convention_provider.ts
```diff
@@ -0,0 +1,24 @@
+import path from 'path';
+
+/**
+ * Provides static methods for type definition conventions for path and ignore pattern
+ */
+export class FunctionTypeDefConventionProvider {
```
edwardfoyle, line 6: We should add a couple tests for this class

packages/backend-function/src/function_env_type_generator.test.ts
```diff
@@ -0,0 +1,59 @@
...
+      fsWriteFileSyncMock.mock.calls[0].arguments[1]
+        ?.toString()
+        .includes(sampleStaticEnv)
+    );
+
+    mock.restoreAll();
+  });
+
+  void it('generated type definition file has valid syntax', async () => {
+    const targetDirectory = await fs.promises.mkdtemp('func_env_type_gen_test');
```
edwardfoyle, line 44: import from `fs/promises` instead

.gitignore
rtpascual: Unit tests also cause the typedef generation to happen, for example in `backend-function/src/test-assets/default-lambda`. Not sure if we wanted those to be committed or not.

packages/backend-function/src/function_env_type_generator.test.ts
rtpascual: I was using `fs` since the test above was mocking `sync` methods, unless we are ok with additional import of `fs/promises`

edwardfoyle: Yeah the additional import should be fine

.gitignore
edwardfoyle: We should make sure the unit tests clean up after themselves. But yeah it's okay to have the rule just so that if the unit tests fail, we don't end up committing test output

edwardfoyle (approved):

edwardfoyle (approved):

.gitignore
sobolk: If unit tests are localized in `backend-function` should we localize this expression too ?

packages/backend-function/src/factory.ts
```diff
@@ -260,6 +261,15 @@ class AmplifyFunction
...
     };
+
+    this.node.addValidation({
+      validate: (): string[] => {
+        const functionEnvironmentTypeGenerator =
+          new FunctionEnvironmentTypeGenerator(id, props.entry);
+        functionEnvironmentTypeGenerator.generateTypeDefFile();
+        return [];
+      },
+    });
```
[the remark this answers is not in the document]
rtpascual: If I throw in `generateTypeDefFile` we get this
```
Error: Error generating typedef file
    at FunctionEnvironmentTypeGenerator.generateTypeDefFile (/Users/rotp/workplace/amplify-backend/packages/backend-function/src/function_env_type_generator.ts:50:11)
    at Object.validate (/Users/rotp/workplace/amplify-backend/packages/backend-function/src/factory.ts:269:42)
    at Node.validate (/Users/rotp/workplace/amplify-backend/node_modules/constructs/src/construct.ts:412:24)
    at /Users/rotp/workplace/amplify-backend/node_modules/aws-cdk-lib/core/lib/private/synthesis.js:2:3722
    at visit (/Users/rotp/workplace/amplify-backend/node_modules/aws-cdk-lib/core/lib/private/synthesis.js:4:255)
    at visit (/Users/rotp/workplace/amplify-backend/node_modules/aws-cdk-lib/core/lib/private/synthesis.js:4:334)
    at visit (/Users/rotp/workplace/amplify-backend/node_modules/aws-cdk-lib/core/lib/private/synthesis.js:4:334)
    at visit (/Users/rotp/workplace/amplify-backend/node_modules/aws-cdk-lib/core/lib/private/synthesis.js:4:334)
    at validateTree (/Users/rotp/workplace/amplify-backend/node_modules/aws-cdk-lib/core/lib/private/synthesis.js:2:3657)
    at synthesize (/Users/rotp/workplace/amplify-backend/node_modules/aws-cdk-lib/core/lib/private/synthesis.js:1:1561)
```

packages/sandbox/src/file_watching_sandbox.ts
```diff
@@ -27,7 +27,10 @@ import {
   FilesChangesTracker,
   createFilesChangesTracker,
 } from './files_changes_tracker.js';
-import { AmplifyError } from '@aws-amplify/platform-core';
+import {
+  AmplifyError,
+  FunctionTypeDefConventionProvider,
+} from '@aws-amplify/platform-core';
```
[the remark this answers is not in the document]
rtpascual: This is the only thing from `backend-function` so far that `sandbox` would be dependent on, maybe we can revisit this conversation if more things show up? Either `sandbox` will start depending on `backend-function` or we will have a new `function-core` package.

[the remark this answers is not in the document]
sobolk: I'd rather make `sandbox` depend on `backend-function` or duplicate this logic in both places.
We shouldn't set precedence that it's ok to dump vertical logic into platform.

@edwardfoyle @Amplifiyer thoughts?

.gitignore
rtpascual: There is also a unit test generating typedefs in `integration-tests` as well, if I remove this in gitignore here are the untracked files:
```
packages/backend-function/src/test-assets/default-lambda/amplify/
packages/backend-function/src/test-assets/lambda-with-dependencies/amplify/
packages/integration-tests/src/test-projects/data-storage-auth-with-triggers-ts/amplify/data/echo/amplify/
packages/integration-tests/src/test-projects/data-storage-auth-with-triggers-ts/amplify/func-src/amplify/
packages/integration-tests/src/test-projects/standalone-data-auth-modes/amplify/data/amplify/
```

The problem with the unit test in `integration-tests` is that we can't mock `fs` so we would need to mock `FunctionEnvironmentTypeGenerator` which would add a dependency

sobolk: how about `**/test-assets/**/amplify` and `**/test-projects/**/amplify`(or so) ?
this at least scopes it down to be sure it's test resources.

josefaidt:
```ts
import type { LambdaEnv, StorageHandler } from "@aws-amplify/backend"
import type { myFunction } from "./resource"

declare global {
  namespace NodeJs {
    interface ProcessEnv extends LambdaEnv<typeof myFunction["environment"]> {}
  }
}

export const handler: StorageHandler = async () => {}
```

ashish-aws:
Josef has some thoughts on this, lets talk about this tomorrow at standup @rtpascual (do not merge before that)

packages/sandbox/src/file_watching_sandbox.ts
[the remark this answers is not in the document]
Amplifiyer: We shouldn't be adding verticals stuff in platform-core. It's fine to take a dependency on `backend-function`.

packages/backend-function/src/factory.ts
```diff
@@ -260,6 +261,17 @@ class AmplifyFunction
...
+    // This is a hack to generate a type definition file after synthesis
+    this.node.addValidation({
+      validate: (): string[] => {
+        new FunctionEnvironmentTypeGenerator(
+          id,
+          props.entry
+        ).generateTypeDefFile();
+        return [];
+      },
+    });
```
Amplifiyer, line 274: Shouldn't this be tied to an `amplify generate ...` command rather than to synthesis? What'd be the DX like in current case?
1. Use `defineFunction` in your project
2. Run `amplify sandbox`
3. Come back to your project and use the environment variables?

rtpascual: Yup that would be the DX for this.

If we tied this to just the `amplify generate ...` command then customers would have to run this after every change that adds/removes function env vars (once we start adding dynamic env vars) to benefit from the types. We can certainly add a `generate` command after but we discussed with Josef and decided having it tied to the end of synthesis for now.

packages/backend-function/src/factory.ts
```diff
@@ -300,6 +301,14 @@ class AmplifyFunction
     };
 
     this.storeOutput(outputStorageStrategy);
+
+    // This is a hack to generate a type definition file after synthesis
```
edwardfoyle, line 305: Hack is maybe a little strong here. More like "using CDK validation mechanism as a way to generate a type definition file at the end of synthesis"

packages/cli-core/package.json
```diff
@@ -18,6 +18,7 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
+    "@aws-amplify/backend-function": "^0.8.0-beta.1",
```
edwardfoyle, line 21: cli-core must not depend on backend-function. We could either
1. hardcode the path in the core package with a comment explaining the coupling to backend-functions or
2. Put the paths in platform-core and import in both backend-function and cli-core

packages/backend-function/API.md
```diff
@@ -23,6 +23,13 @@ export type FunctionProps = {
     runtime?: NodeVersion;
 };
 
+// @public
+export class FunctionTypeDefConventionProvider {
```
edwardfoyle, line 27: if this needs to be exposed, it should go to platform-core

edwardfoyle (changes requested):

packages/cli-core/package.json
rtpascual: I am leaning towards 1. @sobolk and @Amplifiyer also had strong opinions on adding verticals stuff in `platform-core` [here](https://github.com/aws-amplify/amplify-backend/pull/984/#discussion_r1480341390)

sobolk (approved):

packages/cli-core/package.json
sobolk: yes.

edwardfoyle (commented):
Change LGTM, but we should figure out how to incorporate this into the e2e tests [here](https://github.com/aws-amplify/amplify-backend/blob/912034ec9015e6ef9426919249dee06ba54cbb89/packages/integration-tests/src/test-projects/data-storage-auth-with-triggers-ts/amplify/func-src/response_generator.ts#L9). ie instead of getting that value directly from process.env, we should import our generated shim and ensure that getting the value from env works as expected.

I'm also okay if we want to track this in a follow up task, but we definitely need some e2e coverage that bundling the functions is working with this file

rtpascual:
@edwardfoyle Yes this will be in a follow up task with the work for ensuring resource access environment variables are included in the typedef files.

edwardfoyle:
oh duh, that's not implemented yet :)

edwardfoyle (approved):

PR #991, opened 2024-02-06
feat(data): enable destructive schema updates for all deployments and enable table replacement upon GSI updates in sandbox

### LINKED ISSUE #894: Allow destructive updates is always disabled when in sandbox.

AaronZyLee:
## Problem

- Destructive schema updates are always allowed in backend data definition
- Enable the feature of replacing the table when GSI updates are detected 
  - This is enabled in sandbox ONLY
  - It is always enabled in sandbox
  - CDK Aspects is used for modifying this attribute in the construct

Fix #894 
## Changes

Developers will be able to change the key schema of the table directly and there is no need to wipe out the resources first. Since the destructive update is always allowed, under the hood the DynamoDB table will be deleted first then created again with the same physical Id.

In addition, sandbox now treat every GSI update as a table replacement event, which is a much faster way. Since the table data will be lost, this is only available in sandbox mode.

### 1. Example of changing the model identifier
The following is validated by using the local dev version of `amplify sandbox` for the following schema change.
Original schema
```TypeScript
const schema = a.schema({
  Todo: a
    .model({
      content: a.string(),
      title: a.string()
    }).authorization([a.allow.public()]),
});
```
Updated schema
```TypeScript
const schema = a.schema({
  Todo: a
    .model({
      todoId: a.id().required(),
      name: a.string().required(),
      content: a.string(),
      title: a.string()
    }).identifier(['todoId', 'name'])
    .authorization([a.allow.public()]),
});
```
The sandbox deployment output
```
⚠️ The following non-hotswappable changes were found:
    logicalID: TodoTable, type: Custom::AmplifyDynamoDBTable, rejected changes: attributeDefinitions,keySchema, reason: This resource type is not supported for hotswap deployments

Could not perform a hotswap deployment, as the stack amplify-gen2cli-lizeyu-sandbox-70ff726f12 contains non-Asset changes
Falling back to doing a full deployment
```
One thing to note the warning is expected as the attributes of custom resource are not supported for hotswap. It finally falls into a full CFN deployment.

### 2. Example of enabling table replacement for GSI updates in sandbox mode
The following is validated by using the local dev version of `amplify sandbox` for the following schema change.
Original Schema
```TypeScript
const schema = a.schema({
  Todo: a
    .model({
      content: a.string(),
      description: a.string(),
      done: a.boolean(),
      priority: a.enum(['low', 'medium', 'high'])
    })
    .authorization([a.allow.owner()])
});
```
Updated Schema
```TypeScript
const schema = a.schema({
  Todo: a
    .model({
      content: a.string(),
      description: a.string(),
      done: a.boolean(),
      priority: a.enum(['low', 'medium', 'high'])
    })
    .authorization([a.allow.owner()])
    .secondaryIndexes([
      a.index("content").name("byContent"),
      a.index("description").name("byDes"),
    ])
    ,
});
```
Use the following commands for sandbox deploy
```
$ amplify sandbox
```
Then examine the DynamoDB console. Observed that the table is deleted first then created
Screenshot 2024-03-01 at 12 50 37 PM

## Checklist


edwardfoyle:
What's the difference between this PR and [this one](https://github.com/aws-amplify/amplify-backend/pull/896/files)?

edwardfoyle:
Also see: https://github.com/aws-amplify/amplify-backend/pull/896#discussion_r1446711362

AaronZyLee:
> Also see: [#896 (comment)](https://github.com/aws-amplify/amplify-backend/pull/896#discussion_r1446711362)

It is different as the values are only set to true when the `sandbox` deployment mode is detected. The `sandboxEnabled` in the previous PR is not correct IMO as it uses the following code snippet to determine the value
```
    const sandboxModeEnabled = isUsingDefaultApiKeyAuth(
      this.providedAuthConfig,
      this.props.authorizationModes
    );
```
This approach mostly follows the discussions of the PR https://github.com/aws-amplify/amplify-backend/pull/667#discussion_r1395920428

edwardfoyle:
The important bit from that other comment is that most likely we should _always_ allow destructive updates. How do customers actually make destructive updates if that's what they want? We should have checks for this, but it should not be the job of the IaC definition to check, it should be the job of our CI/CD offering.

AaronZyLee:
> The important bit from that other comment is that most likely we should _always_ allow destructive updates. How do customers actually make destructive updates if that's what they want? We should have checks for this, but it should not be the job of the IaC definition to check, it should be the job of our CI/CD offering.

Thanks for the head up. I am on the board of always allowing the destructive updates and let the CI/CD handle the breaking change check. I will address this in the PR.

In addition, there is one additional flag `replaceTableUponGsiUpdate` in the managed table, which will always delete the table first and recreate it when the GSI updates are detected. This saves a lot of time of the deployment as it does not need to go through multiple GSI updates. As it still wipes out the table completely, it is definitely a behavior we want for developers in the sandbox mode only.

edwardfoyle:
> In addition, there is one additional flag replaceTableUponGsiUpdate in the managed table, which will always delete the table first and recreate it when the GSI updates are detected. 

Agreed that should only be enabled in sandbox. But I'm wondering if that should actually be a separate flag / env var to sandbox? Reason being that I may have test data that I actually want to keep in sandbox. Of course, you should be able to repopulate test data, but it could be annoying if tables are always deleted even if you just want to update a single GSI

.changeset/lucky-tigers-carry.md
```diff
@@ -0,0 +1,6 @@
+---
+'@aws-amplify/backend-data': minor
+'@aws-amplify/sandbox': minor
```
sobolk, line 3: why is sandbox included in changeset?

packages/backend-data/src/factory.ts
```diff
@@ -171,15 +177,57 @@ class DataGenerator implements ConstructContainerEntryGenerator {
...
     return amplifyApi;
   };
 }
 
+const REPLACE_TABLE_UPON_GSI_UPDATE_ATTRIBUTE_NAME: keyof TranslationBehavior =
+  'replaceTableUponGsiUpdate';
+
+enum BackendDataEnvironmentVariables {
+  // Environment variables allowed in sandbox mode only
+  REPLACE_TABLE_UPON_GSI_UPDATE = 'REPLACE_TABLE_UPON_GSI_UPDATE',
```
sobolk, line 202:
```suggestion
  AMPLIFY_BACKEND_DATA_REPLACE_TABLE_UPON_GSI_UPDATE = 'AMPLIFY_BACKEND_DATA_REPLACE_TABLE_UPON_GSI_UPDATE',
```

packages/backend-data/src/factory.ts
```diff
@@ -171,15 +177,57 @@ class DataGenerator implements ConstructContainerEntryGenerator {
...
+}
+/**
+ * Aspect class to modify the amplify managed DynamoDB table
+ * to allow table replacement upon GSI update when the environment variable is set to true
+ * This is only allowed to be modified in sandbox mode
+ */
+class ReplaceTableUponGsiUpdateChecker implements IAspect {
```
sobolk, line 209:
```suggestion
class ReplaceTableUponGsiUpdateOverrideAspect implements IAspect {
```

sobolk (commented):
Please add tests.

packages/backend-data/src/factory.ts
```diff
@@ -171,15 +177,57 @@ class DataGenerator implements ConstructContainerEntryGenerator {
...
+  public visit(scope: IConstruct): void {
+    const replaceTableUponGsiUpdate: boolean =
+      process.env[
+        BackendDataEnvironmentVariables.REPLACE_TABLE_UPON_GSI_UPDATE
```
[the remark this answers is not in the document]
edwardfoyle: Another option: the data construct could read from a CDK context value to determine this behavior. Then here we only need to set the CDK context value rather than have this aspect override.

Even better would be to expose this as a normal prop on the data construct that we can set rather than have this "backdoor"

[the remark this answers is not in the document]
AaronZyLee: I think it is better not to leave this param in the data construct as it is a param we only want to take effect in the sandbox mode and it should never happen in other type of deployment. The main purpose for this flag is to iterate fast deployment with the sacrifice of the table data, which is not suitable for branch deployment.

For the idea of using CDK context value, it was adopted in a [previous PR](https://github.com/aws-amplify/amplify-category-api/pull/2078) but it would pass the reading context value logic into the data construct. It was not a desired option at that time.

packages/backend-data/src/factory.ts
AaronZyLee: The env var is removed as it is always allowed in sandbox

packages/backend-data/src/factory.ts
AaronZyLee: Same as above

.changeset/lucky-tigers-carry.md
AaronZyLee: The sandbox mode behaves differently when the GSI updates are detected. Once this PR takes effect, the table will be replaced instead of change the GSIs to achieve a faster deployment experience.

packages/backend-data/src/factory.ts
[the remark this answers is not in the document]
AaronZyLee: Confirmed from product: always allow GSI update replacing table in sandbox

.changeset/lucky-tigers-carry.md
sobolk: That is fine, but the behavior change comes from `backend-data`, not sandbox.
In theory, it's possible to use new `sandobox` with old `backend-data` - in that case the old GSI mode will take effect.

We should remove sandbox from changeset.

sobolk (approved):

edwardfoyle (approved):

PR #1002, opened 2024-02-07
feat: add support for generating user pool groups

awsluja:
## Problem
Developers currently need to manually configure user groups, which requires several additional steps.

## Changes
Developers can now configure user groups easily, by providing a list of user group names.
User groups will be automatically created with those names, and user group roles will be configured for each group.

Developers can then configure those roles with custom policies.
image

## Validation

Example:
Screenshot 2024-02-07 at 1 52 35 PM

Resources deployed:
Screenshot 2024-02-07 at 1 49 25 PM

Trust policy:
Screenshot 2024-02-07 at 1 50 19 PM

## Checklist


packages/auth-construct/src/construct.ts
```diff
@@ -43,8 +45,9 @@ import { build as arnBuilder } from '@aws-sdk/util-arn-parser';
 
 type DefaultRoles = { auth: Role; unAuth: Role };
 type IdentityProviderSetupResult = {
-  oauthMappings: Record<string, string>;
+  oAuthMappings: Record<string, string>;
   providersList: string[];
+  oAuthSettings: cognito.OAuthSettings | undefined;
```
awsluja, line 50: One note about this change - I move oAuthSettings into the setupExternalProviders to allow for a reduction in the constructor logic.

packages/plugin-types/src/auth_resources.ts
```diff
@@ -54,4 +55,21 @@ export type AuthResources = {
...
+    groupName: string;
+    /**
+     * The generated CfnUserPoolGroup for this group
+     */
+    cfnUserGroup: CfnUserPoolGroup;
+    /**
+     * The generated Role for this group
+     */
+    role: IRole;
+  }[];
```
edwardfoyle, line 74: @josefaidt are you sure you want the override access pattern to be `backend.auth.resources.groups[0].role` rather than `backend.auth.resoruces.groups.admin.role`? I feel pretty strongly that the latter is more readable than wondering what the 0th index of the array is

josefaidt: I agree that it is easier to work with an object, and something that was brought up earlier, but landed on this approach since it aligns with the input type.

Ideally you would use `Array.find` and a truthy check/non-null assertion to override groups

```ts
const admins = backend.auth.resources.groups.find(g => g.name === "admins")!
```

edwardfoyle: Chatted offline and we agreed to go with the object access pattern on the resources object

awsluja: Thanks for the feedback; I've updated the PR with the suggested access pattern

packages/auth-construct/src/types.ts
```diff
@@ -313,6 +313,12 @@ export type AuthProps = {
    * 'NONE' - None – users will have to contact an administrator to reset their passwords
    */
   accountRecovery?: keyof typeof cognito.AccountRecovery;
+  /**
+   * Provide a list of user group names which will be created for you.
```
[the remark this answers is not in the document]
Amplifiyer: Also I'm guessing that the order of the input matters? If yes, we should mention that in the comments in detail.

Plus add that the new groups are created with roles having 0 permissions. Use this ... to add permissions to your groups.

packages/plugin-types/API.md
```diff
@@ -39,6 +40,13 @@ export type AuthResources = {
     authenticatedUserIamRole: IRole;
     unauthenticatedUserIamRole: IRole;
     cfnResources: AuthCfnResources;
+    groups: {
+        [key: string]: {
+            groupName: string;
```
Amplifiyer, line 45: If the key is also `groupName` do we need this in object as well?

packages/auth-construct/src/construct.ts
```diff
@@ -101,11 +104,17 @@ export class AmplifyAuth
 
   private readonly computedUserPoolProps: UserPoolProps;
 
-  private readonly domainPrefix: string | undefined;
+  private readonly name: string;
 
-  private readonly oAuthSettings: cognito.OAuthSettings | undefined;
+  private readonly domainPrefix: string | undefined;
```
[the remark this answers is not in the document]
awsluja: In this case, yes, we re-read from during store output; I have another PR where I am switching to read from CFN resources for store-output and will be eliminating the need for this instance property then

packages/plugin-types/API.md
awsluja: This is just for clarity, [key: string] doesn't support JS doc so it isn't known unless you are looking at the docs: 
image

awsluja: ie:
image

packages/auth-construct/src/construct.ts
```diff
@@ -605,14 +619,26 @@ export class AmplifyAuth
...
+    if (external.logoutUrls.length === 0) {
+      throw Error(
+        'You must define logoutUrls when configuring external login providers.'
+      );
+    }
+    if (external.callbackUrls.length === 0) {
+      throw Error(
+        'You must define callbackUrls when configuring external login providers.'
+      );
+    }
```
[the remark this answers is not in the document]
awsluja: can I do this as a follow up? there are 8 instances in auth construct that I'll need to change to conform to the AmplifyError

packages/plugin-types/API.md
edwardfoyle: Could we do `[groupName: string]` as the key definition?

packages/auth-construct/src/construct.ts
```diff
@@ -282,6 +253,48 @@ export class AmplifyAuth
...
   };
 
+  /**
+   * Auto generate the user pool groups and group roles
+   */
+  private setupUserPoolGroups = (
+    groups: string[] | undefined,
+    identityPool: CfnIdentityPool
+  ) => {
+    if (groups) {
```
edwardfoyle, line 263: nit: prefer early return on `if (!groups)`

Alternatively, we could replace the line below with `(groups || []).forEach(...)` and remove this if entirely

packages/plugin-types/API.md
Amplifiyer: Just to refer/destructure this type, customers "have to" know that the key is `groupName`, right?

edwardfoyle (dismissed):
LGTM, just a couple small things. Also make sure to get Praveen's stamp before merging

packages/plugin-types/API.md
edwardfoyle: We could also add JSDocs on the group field itself explaining that the value is a record that maps group name to corresponding roles and userpool group

awsluja: Updating to this, and removing the 'groupName' property. Thanks for the suggestions, 
image

packages/auth-construct/src/construct.ts
awsluja: updated

packages/auth-construct/src/construct.ts
```diff
@@ -260,39 +260,37 @@ export class AmplifyAuth
...
+        `${this.name}${groupName}Group`,
+        {
+          userPoolId: this.userPool.userPoolId,
+          groupName: groupName,
+          roleArn: groupRole.roleArn,
+          precedence: index,
+        }
+      );
+      this.groups[groupName] = {
+        groupName: groupName,
```
edwardfoyle, line 289:
```suggestion
```

awsluja: updated

edwardfoyle (approved):

packages/auth-construct/src/construct.ts
```diff
@@ -893,40 +939,41 @@ export class AmplifyAuth
...
     }
-    if (oauthMappings[authProvidersList.google]) {
-      output.googleClientId = oauthMappings[authProvidersList.google];
+    if (oAuthMappings[authProvidersList.google]) {
+      output.googleClientId = oAuthMappings[authProvidersList.google];
     }
-    if (oauthMappings[authProvidersList.apple]) {
-      output.appleClientId = oauthMappings[authProvidersList.apple];
+    if (oAuthMappings[authProvidersList.apple]) {
+      output.appleClientId = oAuthMappings[authProvidersList.apple];
```
0618, line 953: How about...?

```suggestion
    const oAuthMappings = this.providerSetupResult.oAuthMappings;
    Object.entries(authProvidersList).forEach(([authProvider, endpoint]) => {
        if (oauthMappings[endpoint]) {
        output[`${authProvider}ClientId` as keyof AuthOutput['payload']] = oauthMappings[endpoint];
        }
    })
```

packages/auth-construct/src/construct.ts
```diff
@@ -893,40 +939,41 @@ export class AmplifyAuth
...
           }.amazoncognito.com`;
         }
 
         output.oauthScope = JSON.stringify(
-          this.oAuthSettings.scopes?.map((s) => s.scopeName) ?? []
+          oAuthSettings.scopes?.map((s) => s.scopeName) ?? []
         );
-        output.oauthRedirectSignIn = this.oAuthSettings.callbackUrls
-          ? this.oAuthSettings.callbackUrls.join(',')
+        output.oauthRedirectSignIn = oAuthSettings.callbackUrls
```
0618, line 972: Do we need to change all the cases from `oauth` to `oAuth`?

If we do, could you make a global search?

packages/auth-construct/src/construct.ts
awsluja: Very clean suggestion , I'll include this in the refactor PR with Praveens other suggestions if thats OK

packages/auth-construct/src/construct.ts
```diff
@@ -605,14 +615,26 @@ export class AmplifyAuth
...
+      throw Error(
+        'You must define logoutUrls when configuring external login providers.'
+      );
+    }
+    if (external.callbackUrls.length === 0) {
+      throw Error(
+        'You must define callbackUrls when configuring external login providers.'
+      );
+    }
     if (external.google) {
```
0618, line 638: `if(external....){...` has a lot of repeat code as well. We can follow up to clean them up

awsluja: yeah, though a lot of these have a lot of custom mapping logic, some of are direct mappings and others are customized

0618 (approved):
Some comments need to be addressed in a follow-up PR. Otherwise, LGTM

packages/auth-construct/src/construct.ts
0618: still can be cleaned up, at least, make a function for each of them

Issue #994, opened 2024-02-06
using pnpm, tsc init step of `create-amplify@0.6.0` fails


josefaidt:
### Environment information

```plain text
System:
    OS: macOS 14.3
    CPU: (10) arm64 Apple M1 Pro
    Memory: 1.42 GB / 32.00 GB
    Shell: 3.7.0 - /opt/homebrew/bin/fish
  Binaries:
    Node: 18.10.0 - ~/Library/Caches/fnm_multishells/77574_1680319338649/bin/node
    Yarn: 1.22.19 - ~/Library/Caches/fnm_multishells/77574_1680319338649/bin/yarn
    npm: 8.19.2 - ~/Library/Caches/fnm_multishells/77574_1680319338649/bin/npm
    pnpm: 8.14.3 - ~/Library/Caches/fnm_multishells/77574_1680319338649/bin/pnpm
    bun: 1.0.21 - ~/.bun/bin/bun
    Watchman: Not Found
  npmPackages:
    @aws-amplify/backend: ^0.11.0 => 0.11.0 
    @aws-amplify/backend-cli: ^0.11.0 => 0.11.0 
    aws-amplify: ^6.0.13 => 6.0.13 
    aws-cdk: ^2 => 2.126.0 
    aws-cdk-lib: ^2 => 2.126.0 
    typescript: ^5.0.0 => 5.3.3
```

### Description

```
➜  p create amplify@latest
.../Library/pnpm/store/v3/tmp/dlx-55937  | Progress: resolved 1, reused 0, downloaded .../Library/pnpm/store/v3/tmp/dlx-55937  | Progress: resolved 3, reused 2, downloaded .../Library/pnpm/store/v3/tmp/dlx-55937  | Progress: resolved 10, reused 4, downloaded.../Library/pnpm/store/v3/tmp/dlx-55937  | Progress: resolved 62, reused 56, downloade.../Library/pnpm/store/v3/tmp/dlx-55937  | Progress: resolved 94, reused 87, downloade.../Library/pnpm/store/v3/tmp/dlx-55937  | Progress: resolved 122, reused 118, downloa.../Library/pnpm/store/v3/tmp/dlx-55937  | Progress: resolved 145, reused 138, downloa.../Library/pnpm/store/v3/tmp/dlx-55937  | Progress: resolved 151, reused 145, downloa.../Library/pnpm/store/v3/tmp/dlx-55937  | +154 +++++++++++++++
[...]

PR #1008, opened 2024-02-08
fix(package-manager): pnpm tsc not found

0618:
## Problem

https://github.com/aws-amplify/amplify-backend/issues/994

## Changes

The issue is because `execa` uses the binaries from the execution environment instead of the target local environment. 

Just need to pass `preferLocal` and `localDir` to use the local binaries.

Ref: https://github.com/sindresorhus/execa?tab=readme-ov-file#preferlocal

**Why didn't e2e test catch it?**

It's because the e2e test use `/Users/xxxxxx/amplify-backend/node_modules/.bin/tsc` when it doesn't uses localDir.

**Changes**
- delete the `.bin` folder of the repo to prevent the test project from using them, so that we'll be able to detect this kind of error in the test
- fix `pnpm` by adding `preferLocal` and `localDir` options

## Validation

Tested on my local machine.

## Checklist


packages/cli-core/src/package-manager-controller/package_manager_controller_base.ts
```diff
@@ -112,7 +112,8 @@ Get started by running \`${this.binaryRunner} amplify sandbox\`.`;
       targetDir,
       this.binaryRunner,
       tscArgs,
-      this.execa
+      this.execa,
+      { preferLocal: true, localDir: targetDir }
```
sobolk, line 116: is this the only place impacted?
would this also matter for deployer which runs `tsc` ?

is there something we could do to e2e tests to make them detect this?

0618: Good question. I'm wondering that too. I have a commit on my local to improve the e2e, but I need more test before pushing.

rtpascual: Could we maybe use `npx which tsc` and verify the path is localDir instead of `/Users/xxxxxx/amplify-backend/node_modules/.bin/tsc`?

0618: @rtpascual yea, that's what I did when debugging. Not sure if we need that in the code

rtpascual: Just for e2e tests to confirm they are using the correct `tsc`, not sure if there is a better way

0618: Now the test should be able to detect if any bin's missing. 

I made commit 99677e753b to delete the .bin of the repo

0618:
Move to https://github.com/aws-amplify/amplify-backend/pull/1016 for e2e test

PR #1016, opened 2024-02-11
fix(package-manager): pnpm tsc not found

0618:
continue https://github.com/aws-amplify/amplify-backend/pull/1008 because e2e test doesn't run against forks.

## Problem

https://github.com/aws-amplify/amplify-backend/issues/994

## Changes

The issue is because `execa` uses the binaries from the execution environment instead of the target local environment. 

Just need to pass `preferLocal` and `localDir` to use the local binaries.

Ref: https://github.com/sindresorhus/execa?tab=readme-ov-file#preferlocal

**Why didn't e2e test catch it?**

It's because the e2e test used `/Users/xxxxxx/amplify-backend/node_modules/.bin/tsc` when it doesn't use localDir.

**Changes**
- delete the `.bin` folder of the repo to prevent the test project from using them, so that we'll be able to detect this kind of error in the test
- fix `pnpm` by adding `preferLocal` and `localDir` options

## Validation

Tested on my local machine.

## Checklist


packages/integration-tests/src/package_manager_sanity_checks.test.ts
```diff
@@ -69,7 +76,11 @@ void describe('getting started happy path', async () => {
   void it('creates new project and deploy them without an error', async () => {
     await runPackageManager(
       packageManager,
-      ['create', 'amplify', '--yes'],
+      [
+        'create',
+        `amplify${packageManager.startsWith('yarn') ? '' : '@beta'}`,
```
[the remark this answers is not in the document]
0618: `yarn create amplify@beta` doesn't work 😭  which is why we have to treat it differently. 


I think it's a bug of yarn because it does install correctly, but it calls a wrong binary. 

```
❯ yarn create amplify@0.6.1-beta.0 --yes
yarn create v1.22.21
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
warning "create-amplify > @aws-amplify/plugin-types@0.8.0" has unmet peer dependency "aws-cdk-lib@^2.110.1".
warning "create-amplify > @aws-amplify/plugin-types@0.8.0" has unmet peer dependency "constructs@^10.0.0".
[4/4] 🔨  Building fresh packages...

success Installed "create-amplify@0.6.1-beta.0" with binaries:
      - create-amplify
/bin/sh: /Users/xxxxxxxx/.yarn/bin/create-amplify@0.6.1-beta.0: No such file or directory
error Command failed.
Exit code: 127
Command: /Users/xxxxxxxx/.yarn/bin/create-amplify@0.6.1-beta.0
Arguments: --yes
Directory: /Volumes/workplace/playground/amplify-2024-02-12
Output:

info Visit https://yarnpkg.com/en/docs/cli/create for documentation about this command.
```

So I think, a workaround would be -- if it's `yarn`, we install beta first, then `yarn create amplify`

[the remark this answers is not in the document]
edwardfoyle: We should probably do the workaround. Otherwise we are effectively not running this test on yarn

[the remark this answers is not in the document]
0618: Done 52dc1f706b

packages/cli-core/src/package-manager-controller/yarn_modern_package_manager_controller.ts
```diff
@@ -84,7 +84,7 @@ export class YarnModernPackageManagerController extends PackageManagerController
     await this.executeWithDebugLogger(
       targetDir,
       'yarn',
-      ['add', '-D', 'tsx', 'esbuild'],
+      ['add', '-D', 'corepack'],
```
edwardfoyle, line 87: what is this being used for?

packages/integration-tests/src/package_manager_sanity_checks.test.ts
```diff
@@ -34,10 +34,17 @@ void describe('getting started happy path', async () => {
...
+     * delete .bin folder in the repo because
+     * 1) e2e tests don't use them
+     * 2) execa would use them if it can not find the binary in the test project
+     */
+    await execaCommand(`rm -rf ${process.cwd()}/node_modules/.bin`);
   });
 
   after(async () => {
     // stop the npm proxy
+    await execa('npm', ['run', 'setup:local'], { stdio: 'inherit' }); // add tsx back since we removed all the binaries
```
edwardfoyle, line 47: What's the problem we are trying to solve here?

I'm also a little concerned about using `setup:local` in e2e tests. The intent of that script is to setup a local dev environment and coupling that to running e2e tests could lead to undesirable or unexpected behavior in the tests as that script evolves

packages/integration-tests/src/package_manager_sanity_checks.test.ts
```diff
@@ -67,11 +74,22 @@ void describe('getting started happy path', async () => {
...
 
   void it('creates new project and deploy them without an error', async () => {
-    await runPackageManager(
-      packageManager,
-      ['create', 'amplify', '--yes'],
-      tempDir
-    ).run();
+    // TODO: remove the condition once GA https://github.com/aws-amplify/amplify-backend/issues/1013
+    if (packageManager === 'yarn-classic') {
+      await execa('yarn', ['global', 'add', 'create-amplify@beta']);
```
edwardfoyle, line 79: does this have to be global?

packages/cli-core/src/package-manager-controller/yarn_modern_package_manager_controller.ts
0618: It's for yarn https://yarnpkg.com/getting-started/install

But good question..it's supposed to be installed globally with node and it didn't complain before.

Let me check...

packages/integration-tests/src/package_manager_sanity_checks.test.ts
0618: Nope. I'll change it.

packages/integration-tests/src/package_manager_sanity_checks.test.ts
0618: `./node_modules/.bin` was removed, but we need `tsx` to run `npm run stop:npm-proxy`.

I did `npm install tsx`, but it'd update package.json and package-lock.json with a newer version, which 1) engineer may accidentally include in a commit (I did), 2) have to run `npm run setup:local` again to install the binaries after the test.

packages/integration-tests/src/package_manager_sanity_checks.test.ts
0618: Yes, we have to use `global` here because `yarn create` uses `global` install 😭

packages/integration-tests/src/package_manager_sanity_checks.test.ts
edwardfoyle: Could we just run `npm install` instead?

packages/integration-tests/src/package_manager_sanity_checks.test.ts
```diff
@@ -34,10 +34,17 @@ void describe('getting started happy path', async () => {
     // start a local npm proxy and publish the current codebase to the proxy
     await execa('npm', ['run', 'clean:npm-proxy'], { stdio: 'inherit' });
     await execa('npm', ['run', 'vend'], { stdio: 'inherit' });
+    /**
+     * delete .bin folder in the repo because
+     * 1) e2e tests don't use them
+     * 2) execa would use them if it can not find the binary in the test project
```
edwardfoyle, line 40:
```suggestion
     * 1) e2e tests should not depend on them
     * 2) package managers will use them if it can not find a binary in the test project
```

packages/integration-tests/src/package_manager_sanity_checks.test.ts
edwardfoyle: ah gotcha. Maybe leave a small comment explaining that

packages/cli-core/src/package-manager-controller/yarn_modern_package_manager_controller.ts
0618: removed f14f692c41

packages/integration-tests/src/package_manager_sanity_checks.test.ts
0618: It's not 100% that the package manger uses it. I think it's more about `execa`

0618: Done eaf8440c1f

packages/integration-tests/src/package_manager_sanity_checks.test.ts
```diff
@@ -34,10 +34,22 @@ void describe('getting started happy path', async () => {
...
   });
 
   after(async () => {
+    // TODO: remove the condition once GA https://github.com/aws-amplify/amplify-backend/issues/1013
+    // create-amplify has been installed globally, we need to remove it before running `setup:local`
+    if (packageManager === 'yarn-classic') {
+      await execa('yarn', ['global', 'remove', 'create-amplify']);
+    }
     // stop the npm proxy
+    await execa('npm', ['run', 'setup:local'], { stdio: 'inherit' }); // add tsx back since we removed all the binaries
```
edwardfoyle, line 52: Just making sure my previous comment didn't slip through the cracks: can we use `npm install` here instead of `setup:local`?

edwardfoyle (approved):
One more comment, but if that's not possible, LGTM

packages/integration-tests/src/package_manager_sanity_checks.test.ts
0618: Done 60ef5803e2

awsluja (approved):

PR #1057, opened 2024-02-22
poc(client-config): unified schema based gen2 client config prototype

Amplifiyer:
## Description

Support both Gen1 (aws-amplify/amplify-cli) frontend client configuration as well as a new unified client configuration (Gen2)

## Changes

* This is a POC and doesn't have many tests included.
* This POC has two unified client config versions, just to demonstrate how multiple of them would coexist as well as the effort to add more versions of client config
* The schema and the types are included in the library. In the future they may live in a different package/repository
* The types are auto-generated from JSON schema using `quicktype` command line utility
* One of the goal of this POC is to ensure that Gen1 config and Gen2 config are as much separated as possible. For that
  * A new set of Gen2 client config contributors are created and the older legacy contributors are deleted
  * A new converter converts Gen2 client config to legacy client config
* All auto generated types are exported as they are needed in the `backend` package for overriding the config values. See the following screenshots where customer is trying to override and `new_property` only exists in the version 2
  * Screenshot 2024-02-22 at 5 47 00 PM
  * Screenshot 2024-02-22 at 5 46 49 PM
* The version is numerical instead of semver. Semver doesn't really work very well when we have to "go back" in versions, i.e. customers needing to access/generate client config for older versions.

## Validation

Tested manually with config generation


packages/client-config/API.md
```diff
@@ -62,16 +172,106 @@ export enum ClientConfigFormat {
...
+
+// @public (undocumented)
+export type Gen2ClientConfig = clientConfigTypesV1.ConfigTypesV1 | clientConfigTypesV2.ConfigTypesV2;
+
+// @public
+export const generateClientConfig: (credentialProvider: AwsCredentialIdentityProvider, backendIdentifier: DeployedBackendIdentifier, version?: number) => Promise<ClientConfig>;
 
 // @public
-export const generateClientConfigToFile: (credentialProvider: AwsCredentialIdentityProvider, backendIdentifier: DeployedBackendIdentifier, outDir?: string, format?: ClientConfigFormat, log?: ((message: string) => void) | undefined) => Promise<void>;
+export const generateClientConfigToFile: (credentialProvider: AwsCredentialIdentityProvider, backendIdentifier: DeployedBackendIdentifier, outDir?: string, format?: ClientConfigFormat, log?: ((message: string) => void) | undefined, version?: number) => Promise<void>;
```
sobolk, line 274: after reading a bit maybe `version` should be typed with enum ?

packages/client-config/src/client-config-writer/gen2_client_config_formatter.ts
```diff
@@ -0,0 +1,39 @@
...
+      case ClientConfigFormat.DART: {
+        return `const amplifyConfig = '''${JSON.stringify(
+          clientConfig,
+          null,
+          2
+        )}''';`;
+      }
+      case ClientConfigFormat.JSON_MOBILE:
+      case ClientConfigFormat.JSON:
+        return JSON.stringify(clientConfig, null, 2);
```
sobolk, line 34: nice.

sobolk:
> The types are auto-generated from JSON schema using quicktype command line utility

does generation happen as part of build? or is this manual step?
if the later we'll need some kind of check.

packages/cli/src/client-config/client_config_generator_adapter.ts
```diff
@@ -34,14 +34,16 @@ export class ClientConfigGeneratorAdapter {
   generateClientConfigToFile = async (
     backendIdentifier: DeployedBackendIdentifier,
     outDir?: string,
-    format?: ClientConfigFormat
+    format?: ClientConfigFormat,
+    version?: number
```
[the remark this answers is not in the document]
Amplifiyer: it's up to us to decide. My gut feeling says, `semver` doesn't work here. So let's keep it simple (it's a CLI command input as well)

packages/cli/src/commands/generate/config/generate_config_command.ts
```diff
@@ -99,6 +101,12 @@ export class GenerateConfigCommand
           'A path to directory where config is written. If not provided defaults to current process working directory.',
         type: 'string',
         array: false,
+      })
+      .option('version', {
+        describe:
+          'Version of the client config. Version 0 represents classic amplify-cli client config (Default)',
```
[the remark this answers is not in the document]
Amplifiyer: It's not supposed to match the JS library. Also android/iOS and flutter are on different MV right now. This is a new start
0 => Classic CLI (we can still make changes if needed here as this is not versioned anyways)
1+ => Gen2

packages/client-config/API.md
```diff
@@ -17,6 +17,48 @@ export type AnalyticsClientConfig = {
...
+    social_providers?: string[];
+    user_pool_client_id?: string;
+    user_pool_id?: string;
+    user_sign_up_attributes?: string[];
+    user_username_attributes?: UserUsernameAttribute[];
+    user_verification_mechanisms?: UserVerificationMechanism[];
+};
+
+// @public
+type Auth_2 = {
```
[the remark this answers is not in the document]
Amplifiyer: This is api-extractor doing it's thing and only for internally satisfying dependencies. they are exported back under namespaces. `_2` is not seen by customer.

packages/client-config/API.md
```diff
@@ -45,8 +87,76 @@ export type AuthClientConfig = {
...
+    // (undocumented)
+    UsWest2 = "us-west-2"
+}
+
+// @public (undocumented)
+type Bucket = {
+    aws_region?: AwsRegion;
+    name?: string;
+    prefixes?: string[];
+    [property: string]: any;
```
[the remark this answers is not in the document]
Amplifiyer: This is for any "extra property". This is being removed.

packages/client-config/src/client-config-schema/config_types_v1.ts
```diff
@@ -0,0 +1,139 @@
...
+   */
+  _version: '1';
+  /**
+   * Outputs generated from defineAuth
+   */
+  auth?: Auth;
+  /**
+   * Outputs generated from defineAuth
+   */
+  custom?: { [key: string]: any };
```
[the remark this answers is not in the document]
Amplifiyer: These are auto generated :(

Amplifiyer:
> > The types are auto-generated from JSON schema using quicktype command line utility
> 
> does generation happen as part of build? or is this manual step? if the later we'll need some kind of check.

They are currently generated manually. In future we will have a process where types for all platform will be generated automatically (not as part of CLI build but perhaps an external build).

packages/client-config/API.md
```diff
@@ -45,8 +87,76 @@ export type AuthClientConfig = {
...
+// @public (undocumented)
+type Bucket_2 = {
+    aws_region?: AwsRegion_2;
+    name?: string;
+    prefixes?: string[];
+    [property: string]: any;
+};
+
+// @public (undocumented)
+export type ClientConfig = Gen1ClientConfig | Gen2ClientConfig;
```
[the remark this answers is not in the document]
Amplifiyer: In my current proposal, we treat every single change in Gen2 client config as major version. 
i.e. Gen1 -> Classic Config format
Gen2 -> New Gen2 based schema format.

We will never have Gen3 in this system.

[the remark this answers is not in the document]
sobolk: I see. Then maybe we should call it `ClientConfig` and `ClientConfigLegacy` ?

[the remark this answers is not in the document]
sobolk: and deprecate legacy shape. (we drop it at some point when bumping mv ?)

packages/cli/src/client-config/client_config_generator_adapter.ts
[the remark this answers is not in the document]
edwardfoyle: I would still suggest typing this as a string, or perhaps as a string union of `'1' | '2'`. In addition to being slightly more flexible, it prevents us from having to do string to number conversions

packages/client-config/API.md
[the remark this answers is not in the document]
sobolk: I see, so this is just result of name collision. 
I'd append suffix to all gen1 types (and maybe mark them as deprecated - now or later).

btw. is there any difference between gen1 and gen2 schema ? these types look very similar.

packages/client-config/API.md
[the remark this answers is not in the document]
Amplifiyer: Yup, that's a good idea. Definitely clearer than Gen1/Gen2

packages/cli/src/commands/generate/config/generate_config_command.ts
```diff
@@ -99,6 +101,12 @@ export class GenerateConfigCommand
           'A path to directory where config is written. If not provided defaults to current process working directory.',
         type: 'string',
         array: false,
+      })
+      .option('version', {
```
edwardfoyle, line 105: since --version is typically reserved for printing the program version, maybe we should call this `output-version`?

packages/client-config/src/gen2-client-config-contributor/gen2_client_config_contributor_v2.ts
Amplifiyer, line 1: This is a copy of [gen2_client_config_contributor_v1.ts](https://github.com/aws-amplify/amplify-backend/pull/1057/files#diff-bbe688df61df3cb72ed38f00ad0236d121b6cb9f69814a6ea304c07ac603fc5d) basically to show how two versions of Gen2 config will be maintained.

packages/client-config/API.md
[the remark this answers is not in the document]
Amplifiyer: Also see here https://github.com/aws-amplify/amplify-backend/pull/1057/files#r1499770179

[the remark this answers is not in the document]
sobolk: I see, so this is just to show versioning framework but new schema is still tbd. correct ?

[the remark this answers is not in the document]
Amplifiyer: yes correct, the new schema should be available soon (a little bit of changes than what is here). In the final PR, for now, we will have only one Gen2 client config.

packages/client-config/API.md
```diff
@@ -45,8 +87,76 @@ export type AuthClientConfig = {
...
+    // (undocumented)
+    AwsIam = "AWS_IAM",
+    // (undocumented)
+    AwsLambda = "AWS_LAMBDA",
+    // (undocumented)
+    OpenidConnect = "OPENID_CONNECT"
+}
+
+// @public (undocumented)
+enum AwsRegion {
```
edwardfoyle, line 119: I would suggest that we type aws region properties as string. Providing an enum type becomes a never-ending game of catchup

packages/client-config/src/client-config-schema/config_types_v2.ts
```diff
@@ -0,0 +1,139 @@
...
+   */
+  _version: '2';
+  /**
+   * Outputs generated from defineAuth
+   */
+  auth?: Auth;
+  /**
+   * Outputs generated from defineAuth
+   */
+  custom?: { [key: string]: any };
```
edwardfoyle, line 16: nit
```suggestion
  custom?: { [key: string]: unknown };
```

packages/client-config/src/client-config-schema/config_types_v1.ts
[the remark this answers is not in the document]
edwardfoyle: If these files are autogenerated, should that generation be part of the package build step and these files aren't actually checked in?

packages/client-config/src/gen2-client-config-contributor/gen2_client_config_contributor_v2.ts
edwardfoyle, line 1: I'm a little confused about the difference between `gen2_client_config_contributor_v1` and `gen2_client_config_contributor_v2`? Seems like there are too many competing versions in the name. Since all of this exists in gen2, should we just have "v1_client_config_contributor" and "v2_client_config_contributor"?

packages/client-config/src/generate_client_config_to_file.ts
```diff
@@ -18,20 +19,24 @@ export const generateClientConfigToFile = async (
...
+  version?: number
 ): Promise<void> => {
   const packageJson = await readPackageJson();
 
   const clientConfigWriter = new ClientConfigWriter(
     getClientConfigPath,
-    new ClientConfigFormatter(
-      new ClientConfigConverter(packageJson.name, packageJson.version)
-    )
+    version && version >= 1
```
edwardfoyle, line 29: we should explicitly check for known versions and throw on anything unknown / unsupported

edwardfoyle (commented):
Mostly looks good. Need to do another pass with a little more focus, but left some initial comments

packages/client-config/src/client-config-schema/config_types_v1.ts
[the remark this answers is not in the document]
Amplifiyer: no, these are generated once, and in future perhaps in a separate process. The schema files won't be checked in but the types files will be.

packages/cli/src/client-config/client_config_generator_adapter.ts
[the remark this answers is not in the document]
Amplifiyer: We weren't doing string to number conversions anywhere. The reasons I chose number is it's more flexible to specify that this CLI version supports a min and max version and validate that as well. What do you think? For now I changed it to string.

packages/cli/src/commands/generate/config/generate_config_command.ts
Amplifiyer: good point, changed.

packages/client-config/API.md
Amplifiyer: added a comment in the schema doc for PMs to look at.

packages/client-config/src/client-config-schema/config_types_v2.ts
Amplifiyer: auto-generated. We can perform changes but that becomes cumbersome to remember to always update the code generation (unless we whip out our own codegen)

packages/client-config/API.md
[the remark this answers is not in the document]
Amplifiyer: Changed it to `ClientConfigLegacy` and `ClientConfigGen2`. `ClientConfig` is the union type of these two

packages/client-config/src/gen2-client-config-contributor/gen2_client_config_contributor_v2.ts
Amplifiyer: `gen2_client_config_contributor_v1` and `gen2_client_config_contributor_v2` represented two (hypothetical) versions of gen2 unified config. In the final PR there will be only one. The two are added here to show how new versions will be added in the future.


Removed the `gen2_` from the file names and the folder is renamed to `client_config_contributor_gen2` and `client_config_contributor_legacy`

packages/client-config/API.md
```diff
@@ -45,8 +87,76 @@ export type AuthClientConfig = {
...
+    // (undocumented)
+    UsEast1 = "us-east-1",
+    // (undocumented)
+    UsEast2 = "us-east-2",
+    // (undocumented)
+    UsWest1 = "us-west-1",
+    // (undocumented)
+    UsWest2 = "us-west-2"
```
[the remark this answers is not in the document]
Amplifiyer: on the other end, we should perhaps remove it completely https://github.com/aws-amplify/amplify-backend/pull/1057#discussion_r1499801648

packages/client-config/API.md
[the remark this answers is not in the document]
sobolk: Thinking a bit more.
Do we even need `ClientConfigLegacy` and `ClientConfigGen2` ?
Perhaps we should just change return type of `generateClientConfig` to desired new shape and just have option to render different version in `generateClientConfigToFile` . This is assuming that new format will have right amount of data and can be transformed into one under the hood.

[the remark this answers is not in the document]
sobolk: Especially if we say that after introducing `ClientConfigGen2` there's never new type, but we version it via client config package version. <- this makes me think we should find solution where there's only one set of types for programatic api but many options for writing to file.

packages/client-config/API.md
josefaidt: that's a good callout to pivot this to a string. It appears CDK went a similar path and opted not to surface an enum

packages/client-config/API.md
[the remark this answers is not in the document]
Amplifiyer: Discussed offline. Will remove `ClientConfigLegacy` in favor of just keeping `ClientConfigGen2` in both places `backend.addOutput` and return type of `generateClientConfig`.

packages/client-config/src/client-config-schema/client_config_v2.ts
Amplifiyer, line 1: Types in this file are TBD. Here just for POC demonstration

packages/client-config/src/generate_client_config.ts
```diff
@@ -18,14 +21,15 @@ import {
  */
 export const generateClientConfig = async (
   credentialProvider: AwsCredentialIdentityProvider,
-  backendIdentifier: DeployedBackendIdentifier
+  backendIdentifier: DeployedBackendIdentifier,
+  version: ClientConfigVersion
 ): Promise<ClientConfig> => {
```
sobolk, line 26: Would it be possible to infer narrowed down return type from `version` parameter ?

packages/backend/src/engine/custom_outputs_accumulator.test.ts
```diff
@@ -104,7 +104,9 @@ void describe('Custom outputs accumulator', () => {
     );
 
     accumulator.addOutput({
-      aws_user_pools_id: 'some_user_pool_id',
+      auth: {
+        user_pool_id: 'some_user_pool_id',
+      },
```
sobolk, line 109: not sure where to leave this comment.
we should add validation: if somebody `addOutput({ version: 1, ...})` and then generates config with `--version 2` then this should throw runtime error somehow.

sobolk: also attempt to use `backend.addOutput({ version: 1, ...}); backend.addOutput({ version: 2, ...})` should also throw.

Amplifiyer: Added both validations.

packages/client-config/src/generate_client_config.ts
Amplifiyer: The method return type has to be a union since the value of version is only available at runtime :(. Consumers can however use typeguards with a bunch of if-else to type it to the narrow type.

sobolk: would this work if `version` is enumerated https://stackoverflow.com/questions/54165536/typescript-function-return-type-based-on-input-parameter ?

sobolk (commented):
The poc looks good. Looking forward to see final version.

packages/client-config/src/generate_client_config.ts
Amplifiyer: nice! conditional types :)
However I still don't see how it helps narrowing the types as the problem now just shifted to the consumer, i.e. unless they have the constant version passed in, they can't see the narrow type and it will be a union.

Tried it here https://github.com/aws-amplify/amplify-backend/pull/1057/commits/75c04e3a3c1083fa581a7ccdccbe7987d96106da

sobolk: While it doesn't solve the problem 100% it does help in cases you mention - when version is known upfront and is const.

Btw. this makes me think - should we default version to "latest" in this API ? In that case calls without specifying it would default to narrowed type for latest right ?

Amplifiyer: The default is TBD, most likely it will be legacy version. 

Moreover I preferred setting default at the command level (in the cli package) rather than defaulting in the config generation, meaning the default value is still a dynamic variable for this API.

sobolk: I was thinking about programatic usage of this API (outside of CLI context).

sobolk: I.e. default for CLI doesn't have to be the same as programmatic default.

PR #1090, opened 2024-02-29
feat: attach policy & ssm params to access userpool from auth resource

bombguy:
### `defineAuth` should accept configuration to add access to resources.
Screenshot 2024-02-29 at 9 40 26 AM

Should attach correct policy with access to the userpool to the specified resource. Matches the policy set from gen 1 CLI
https://github.com/aws-amplify/amplify-cli/blob/8ad57bf99a404f3c92547c8a175458016f682fac/packages/amplify-category-auth/src/provider-utils/awscloudformation/service-walkthroughs/auth-questions.ts#L623

Should add SSM param that points to the userpool resource to the specified resource.

## Validation

Screenshot 2024-02-29 at 9 47 20 AM
Screenshot 2024-02-29 at 9 48 02 AM

## Checklist


packages/backend-auth/src/auth_access_policy_arbiter.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    // no-op if access definition is not found for role.
+    if (!accessDefinition) {
+      return;
+    }
+
+    const accessAcceptor = accessDefinition.getResourceAccessAcceptor(
+      this.getInstanceProps
+    );
+    if (!accessAcceptor) {
```
edwardfoyle, line 56: do we need this check? If the accessAcceptor is missing that should be a fault rather than silently doing nothing

packages/backend-auth/src/userpool_access_policy_factory.ts
```diff
@@ -0,0 +1,148 @@
...
+    'cognito-idp:DeleteUser',
+  ],
+  list: [
+    'cognito-identity:List*',
+    'cognito-idp:AdminList*',
+    'cognito-idp:List*',
+    'cognito-sync:List*',
+    'iam:ListOpenIdConnectProviders',
+    'iam:ListRoles',
+    'sns:ListPlatformApplications',
```
edwardfoyle, line 146: These actions should almost certainly not be part of this policy

packages/backend-auth/src/userpool_access_policy_factory.ts
```diff
@@ -0,0 +1,148 @@
...
+    'cognito-idp:DeleteUserPool',
+    'cognito-idp:AdminDeleteUser',
+    'cognito-idp:DeleteIdentityProvider',
+    'cognito-idp:DeleteUser',
+  ],
+  list: [
+    'cognito-identity:List*',
+    'cognito-idp:AdminList*',
+    'cognito-idp:List*',
+    'cognito-sync:List*',
```
edwardfoyle, line 143: we should avoid * actions. Better to list out exactly what we want

packages/backend-auth/src/auth_access_policy_arbiter.ts
```diff
@@ -0,0 +1,86 @@
...
+      return;
+    }
+    const policy = this.userpoolAccessPolicyFactory.createPolicy(
```
edwardfoyle, line 59: left a similar comment below, but the "user" vs "group" scope needs to apply to the policy

edwardfoyle (changes requested):

packages/backend-auth/API.md
```diff
@@ -11,15 +11,31 @@ import { AuthResources } from '@aws-amplify/plugin-types';
...
 import { FunctionResources } from '@aws-amplify/plugin-types';
 import { GoogleProviderProps } from '@aws-amplify/auth-construct-alpha';
 import { OidcProviderProps } from '@aws-amplify/auth-construct-alpha';
+import { ResourceAccessAcceptor } from '@aws-amplify/plugin-types';
 import { ResourceAccessAcceptorFactory } from '@aws-amplify/plugin-types';
 import { ResourceProvider } from '@aws-amplify/plugin-types';
 import { TriggerEvent } from '@aws-amplify/auth-construct-alpha';
 
+// @public (undocumented)
+export type AccessDefinition = {
```
sobolk, line 26: Can we align naming of all these new types with storage https://github.com/aws-amplify/amplify-backend/blob/main/packages/backend-storage/API.md ?

I.e. it seems that in storage we use prefix in type names.

packages/backend-auth/API.md
```diff
@@ -86,6 +103,17 @@ export type OidcProviderFactoryProps = Omit<OidcProviderProps, 'clientId' | 'cli
     clientSecret: BackendSecret;
 };
 
+// @public (undocumented)
+export type ResourceAccessBuilder = {
+    to: (actions: UserpoolAction[]) => AccessDefinition;
+};
+
+// @public (undocumented)
+export type Role = 'users' | 'groups';
```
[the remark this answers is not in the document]
sobolk: Kind of hard to name it.

If we don't know perhaps we should inline this into `Partial<Record<'users' | 'groups', AccessDefinition>>` above. and punt the problem.

`users` and `groups` are `UserPoolEnitity` perhaps ?

packages/backend-auth/src/access_builder.test.ts
```diff
@@ -0,0 +1,57 @@
...
+} from '@aws-amplify/plugin-types';
+import assert from 'node:assert';
+
+void describe('allowAccessBuilder', () => {
+  const resourceAccessAcceptorMock = mock.fn();
+
+  const getResourceAccessAcceptorMock = mock.fn(
+    // allows us to get proper typing on the mock args
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    (_: string) => resourceAccessAcceptorMock
```
sobolk, line 17: this should be solvable if you provide typing for function, 
see example here https://github.com/aws-amplify/amplify-backend/blob/9ad867e17d216b2292fb4f942a84f728e2471a5f/packages/client-config/src/client-config-writer/client_config_writer.test.ts#L18 .

packages/backend-auth/src/access_builder.test.ts
```diff
@@ -0,0 +1,57 @@
...
+  );
+
+  const getConstructFactoryMock = mock.fn(
+    // this lets us get proper typing on the mock args
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    <T extends ResourceProvider>(_: string) => ({
```
sobolk, line 23: same here.

packages/backend-auth/API.md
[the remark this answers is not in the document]
Amplifiyer: As per IAM terminology, these are either Principal or Identity https://docs.aws.amazon.com/IAM/latest/UserGuide/intro-structure.html#intro-structure-terms

packages/backend-auth/src/userpool_access_policy_factory.ts
```diff
@@ -0,0 +1,148 @@
+import { IUserPool } from 'aws-cdk-lib/aws-cognito';
+import { Policy, PolicyStatement } from 'aws-cdk-lib/aws-iam';
+import { Stack } from 'aws-cdk-lib';
+import { AmplifyFault } from '@aws-amplify/platform-core';
+import { UserpoolAction } from './types.js';
+
+export type Permission = {
+  actions: UserpoolAction[];
+  /**
+   * An s3 prefix that defines the scope of the actions
```
Amplifiyer, line 10: There are a bunch of these storage artifacts here that should be cleaned up

packages/backend-auth/src/userpool_access_policy_factory.ts
```diff
@@ -0,0 +1,148 @@
...
+  };
+
+  private getStatement = (action: UserpoolAction) =>
+    new PolicyStatement({
+      actions: actionPolicyMap[action],
+      resources: [this.userpool.userPoolArn],
+    });
+}
+
+const actionPolicyMap: Record<UserpoolAction, string[]> = {
```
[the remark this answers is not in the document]
Amplifiyer: Aren't we essentially breaking the "Principle of Least Privilege" here by combining these multiple permissions against generic umbrella terms `create`, `list` etc?

The actions here should be very specifically named that maps to the strict set of permissions required for those actions.

packages/backend-auth/API.md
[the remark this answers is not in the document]
edwardfoyle: this is different than that. Here we are granted access to either the user actions (CreateUser, GetUser, etc) on the userpool or the group actions (CreateGroup, GetGroup, etc) on the userpool. This is not referring to an IAM User or IAM Group

packages/backend-auth/src/userpool_access_policy_factory.ts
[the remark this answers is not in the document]
edwardfoyle: I think it's fine for a single "action" to map to a handful of different IAM actions. for example maybe "creating users" requires AdminCreateUser, ConfirmSignUp and SetUserMFAPreference (just making that up). It would be okay to combine that all together under one "create user" use case (along with documentation on exactly what IAM actions will be enabled). But this huge list of actions that doesn't seem to have high cohesion is definitely not what we want.

packages/backend-auth/API.md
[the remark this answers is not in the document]
Amplifiyer: I thought it was CognitoUser or CognitoGroups (i.e. still Cognito Identity). In case this is referring to actions, perhaps

`export type UserPoolAction = 'userActions' | 'groupActions';`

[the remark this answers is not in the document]
edwardfoyle: That's kinda where my head was at with the `ActionScope` suggestion. For example, a definition of
```ts
{
  users: allow.resource(myFunc).to(['read'])
}
```
Would mean "scope the read access of myFunc to UserPool users".

And
```ts
{
  groups: allow.resource(otherFunc).to(['read'])
}
```
Would mean "scope the read access of otherFunc to UserPool roles".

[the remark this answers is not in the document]
sobolk: `export type UserPoolAction = 'userActions' | 'groupActions';` this sounds good.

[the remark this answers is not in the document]
sobolk: `UserPoolActionScope` is fine too.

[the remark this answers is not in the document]
Amplifiyer: I'm less worried about the type but the value here. I think this conveys a little more accurately
```ts
{
  userActions: [
    allow.resource(myFunc).to(['read'])
  ],
  groupActions: [
    allow.resource(otherFunc).to(['read'])
  ]
}
```

packages/backend-auth/src/userpool_access_policy_factory.ts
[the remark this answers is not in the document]
Amplifiyer: Yes combining the IAM actions is fine, but not under a generic action such as `create` or `read`. This is too ambiguous for resource that has multiple things to create/read etc.

packages/backend-auth/API.md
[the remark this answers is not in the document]
edwardfoyle: I'm good with the above. Not sure we want to leak "UserPool" into the type name at this level of abstraction though. Maybe `AuthManagementActions = 'userActions' | 'groupActions'`?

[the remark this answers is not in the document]
bombguy: not relevant now, action scopes are removed

packages/backend-auth/src/auth_access_policy_arbiter.ts
```diff
@@ -1,65 +1,41 @@
...
-    private readonly accessDefinition: Partial<Record<Role, AccessDefinition>>,
-    private readonly ssmEnvironmentEntriesGenerator: SsmEnvironmentEntriesGenerator,
+    private readonly accessDefinition: AuthAccessDefinition[],
     private readonly getInstanceProps: ConstructFactoryGetInstanceProps,
-    private readonly userpool: IUserPool,
-    private readonly userpoolAccessPolicyFactory = new UserPoolAccessPolicyFactory(
-      userpool
-    )
+    private readonly ssmEnvironmentEntries: SsmEnvironmentEntry[],
+    private readonly userpoolAccessPolicyFactory: UserPoolAccessPolicyFactory
```
edwardfoyle, line 19: nit:
```suggestion
    private readonly userPoolAccessPolicyFactory: UserPoolAccessPolicyFactory
```

packages/backend-auth/src/auth_access_policy_arbiter.ts
```diff
@@ -1,65 +1,41 @@
...
+  acceptResourceAccess = (accessDefinition: AuthAccessDefinition) => {
     const accessAcceptor = accessDefinition.getResourceAccessAcceptor(
       this.getInstanceProps
     );
-    if (!accessAcceptor) {
-      return;
-    }
     const policy = this.userpoolAccessPolicyFactory.createPolicy(
-      new Set(accessDefinition.actions)
+      accessDefinition.actions
```
edwardfoyle, line 35: We should round-trip this into a Set to ensure there are no duplicates

packages/backend-auth/src/factory.test.ts
```diff
@@ -130,19 +130,57 @@ void describe('AmplifyAuthFactory', () => {
...
+              'cognito-idp:AdminRespondToAuthChallenge',
+              'cognito-idp:AdminSetUserMFAPreference',
+              'cognito-idp:AdminSetUserPassword',
+              'cognito-idp:AdminSetUserSettings',
+              'cognito-idp:AdminUpdateDeviceStatus',
+              'cognito-idp:AdminUpdateUserAttributes',
+              'cognito-idp:AdminUserGlobalSignOut',
+            ],
+            Effect: 'Allow',
+            Resource: `${backendAuth.resources.userPool.userPoolArn}`,
```
edwardfoyle, line 175: nit:
```suggestion
            Resource: backendAuth.resources.userPool.userPoolArn,
```

packages/backend-auth/src/factory.ts
```diff
@@ -156,12 +161,18 @@ class AmplifyAuthGenerator implements ConstructContainerEntryGenerator {
     // this produces the access definition that will be used to create the storage policies
     const accessDefinition = this.props.access(this.allowAccessBuilder);
 
+    const ssmEnvironmentEntries =
+      ssmEnvironmentEntriesGenerator.generateSsmEnvironmentEntries({
+        [`${this.defaultName}_USERPOOL_ARN`]:
```
edwardfoyle, line 166: Looks like the SDK uses the UserPool Id, not ARN. That's what we should set here

For example https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/cognito-identity-provider/command/AdminCreateUserCommand/

packages/backend-auth/src/factory.ts
```diff
@@ -156,12 +161,18 @@ class AmplifyAuthGenerator implements ConstructContainerEntryGenerator {
...
+          authConstructMixin.resources.userPool.userPoolArn,
+      });
+
+    authConstruct.resources.cfnResources;
```
edwardfoyle, line 170: unused?

packages/backend-auth/src/userpool_access_policy_factory.ts
```diff
@@ -22,127 +14,99 @@ export class UserPoolAccessPolicyFactory {
...
+
+      if (typeof mappedAction === 'string') {
+        policyActions.add(mappedAction);
+      } else {
+        mappedAction.forEach((action) => policyActions.add(action));
+      }
+    });
+
+    if (policyActions.size === 0) {
       throw new AmplifyFault('EmptyPolicyFault', {
```
edwardfoyle, line 37: Would this hit if a customer specified `allow.resource(myFunc).to([])`? If so, this should be an Error not a Fault

packages/backend-auth/src/userpool_access_policy_factory.ts
```diff
@@ -22,127 +14,99 @@ export class UserPoolAccessPolicyFactory {
...
-    'cognito-idp:AdminCreateUser',
-    'cognito-idp:CreateUserImportJob',
-    'cognito-idp:AdminSetUserSettings',
-    'cognito-idp:AdminLinkProviderForUser',
-    'cognito-idp:CreateIdentityProvider',
+type IamActionMap = {
+  [action in ActionIam]: string;
+} & {
+  [amplifyActionSet in ActionMeta]: string[];
+};
```
edwardfoyle, line 63: for simplicity, I'd recommend making this a `Record<AuthAction, string[]>`. For the 1:1 mappings, just put the single value in an array

packages/backend-auth/src/userpool_access_policy_factory.ts
```diff
@@ -22,127 +14,99 @@ export class UserPoolAccessPolicyFactory {
...
+
+    actions.forEach((authAction) => {
+      const mappedAction = iamActionMap[authAction];
+
+      if (typeof mappedAction === 'string') {
+        policyActions.add(mappedAction);
+      } else {
+        mappedAction.forEach((action) => policyActions.add(action));
+      }
+    });
```
edwardfoyle, line 34: With my suggestion below on the actionMap, this can simplify to something like
```suggestion
const policyActions = new Set(actions.flatMap(action => iamActionMap[action]))
```

packages/backend-auth/src/userpool_access_policy_factory.ts
```diff
@@ -22,127 +14,99 @@ export class UserPoolAccessPolicyFactory {
...
-    actions.forEach((action) => {
-      statements.push(this.getStatement(action));
-    });
+    const policy = new Policy(
+      this.stack,
+      `${this.namePrefix}${this.policyCount++}`,
+      {
+        statements: [
+          new PolicyStatement({
+            actions: [...policyActions],
```
edwardfoyle, line 48:
```suggestion
            actions: policyActions,
```

packages/backend-auth/src/userpool_access_policy_factory.ts
```diff
@@ -22,127 +14,99 @@ export class UserPoolAccessPolicyFactory {
...
-    'cognito-idp:AdminLinkProviderForUser',
-    'cognito-idp:CreateIdentityProvider',
+type IamActionMap = {
+  [action in ActionIam]: string;
+} & {
+  [amplifyActionSet in ActionMeta]: string[];
+};
+
+const iamActionMap: IamActionMap = {
+  manageUser: [
```
edwardfoyle, line 66: This list still seems like too much for "manageUser". For example [AdminGetDevice](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminGetDevice.html) doesn't seem like it has anything to do with user management. Same goes for the other "device related" APIs. While those devices are associted with users, I would suggest a separate meta action for "userDeviceManagement" if / when there is demand

edwardfoyle, line 66:
```suggestion
  manageUsers: [
```

packages/backend-auth/src/userpool_access_policy_factory.ts
```diff
@@ -22,127 +14,99 @@ export class UserPoolAccessPolicyFactory {
...
-    'cognito-idp:CreateGroup',
-    'cognito-idp:CreateUserPool',
+    'cognito-idp:AdminEnableUser',
     'cognito-idp:AdminForgetDevice',
-    'cognito-idp:AdminAddUserToGroup',
+    'cognito-idp:AdminGetDevice',
+    'cognito-idp:AdminGetUser',
+    'cognito-idp:AdminListDevices',
+    'cognito-idp:AdminListGroupsForUser',
+    'cognito-idp:AdminListUserAuthEvents',
```
edwardfoyle, line 79: Reading the docs for this action I don't think it belongs in user management: https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminListUserAuthEvents.html

it looks more like a "userAudit" action

packages/backend-auth/src/userpool_access_policy_factory.ts
```diff
@@ -22,127 +14,99 @@ export class UserPoolAccessPolicyFactory {
...
+    'cognito-idp:AdminRemoveUserFromGroup',
```
edwardfoyle, line 80: I'd suggest a separate "manageUserGroupMembership" meta action for adding / removing users from groups

packages/backend-auth/src/userpool_access_policy_factory.ts
```diff
@@ -22,127 +14,99 @@ export class UserPoolAccessPolicyFactory {
...
-    'cognito-idp:ConfirmDevice',
-    'cognito-idp:AdminResetUserPassword',
-    'cognito-idp:ResendConfirmationCode',
-  ],
-  update: [
-    'cognito-idp:ForgotPassword',
-    'cognito-idp:UpdateAuthEventFeedback',
-    'cognito-idp:UpdateResourceServer',
-    'cognito-idp:UpdateUserPoolClient',
+    'cognito-idp:AdminSetUserPassword',
```
edwardfoyle, line 84: For setting and resetting passwords I'd recommend "manageUserPasswordRecovery"

packages/backend-auth/src/userpool_access_policy_factory.ts
bombguy: policyActions is a Set, actions is expecting string[]

packages/backend-auth/src/factory.test.ts
```diff
@@ -113,6 +114,55 @@ void describe('AmplifyAuthFactory', () => {
...
+            ],
+            Effect: 'Allow',
+            Resource: backendAuth.resources.userPool.userPoolArn,
+          },
+        ],
+        Version: '2012-10-17',
+      }
+    );
+    assert.ok(
+      mockAcceptResourceAccess.mock.calls[1].arguments[0] instanceof Policy
```
edwardfoyle, line 162: nit: add another assertion on the second policy document

packages/backend-auth/src/factory.ts
```diff
@@ -148,6 +165,29 @@ class AmplifyAuthGenerator implements ConstructContainerEntryGenerator {
         },
       }),
     };
+    if (!this.props.access) {
+      return authConstructMixin;
+    }
+    // props.access is the access callback defined by the customer
+    // here we inject the roleAccessBuilder into the callback and run it
```
edwardfoyle, line 172: nit: 
```suggestion
    // here we inject the allowAccessBuilder into the callback and run it
```

packages/backend-auth/src/factory.ts
```diff
@@ -148,6 +165,29 @@ class AmplifyAuthGenerator implements ConstructContainerEntryGenerator {
...
+    // this produces the access definition that will be used to create the storage policies
```
edwardfoyle, line 173:
```suggestion
    // this produces the access definition that will be used to create the auth access policies
```

packages/backend-auth/src/userpool_access_policy_factory.ts
```diff
@@ -0,0 +1,122 @@
...
+  manageUsers: [
+    'cognito-idp:AdminConfirmSignUp',
+    'cognito-idp:AdminCreateUser',
+    'cognito-idp:AdminDeleteUser',
+    'cognito-idp:AdminDeleteUserAttributes',
+    'cognito-idp:AdminDisableUser',
+    'cognito-idp:AdminEnableUser',
+    'cognito-idp:AdminGetUser',
+    'cognito-idp:AdminListGroupsForUser',
+    'cognito-idp:AdminRespondToAuthChallenge',
```
edwardfoyle, line 81: I'd suggest taking this out and leaving it as a standalone action

packages/backend-auth/src/userpool_access_policy_factory.ts
```diff
@@ -0,0 +1,122 @@
...
+    'cognito-idp:AdminSetUserMFAPreference',
+    'cognito-idp:AdminSetUserSettings',
+    'cognito-idp:AdminUpdateUserAttributes',
+    'cognito-idp:AdminUserGlobalSignOut',
```
edwardfoyle, line 85: Same here, suggest this as a standalone action only

packages/backend-auth/src/userpool_access_policy_factory.ts
```diff
@@ -0,0 +1,122 @@
...
+
+type IamActionMap = {
+  [action in ActionIam]: string;
+} & {
+  [amplifyActionSet in ActionMeta]: string[];
+};
+
+const iamActionMap: IamActionMap = {
+  manageUsers: [
+    'cognito-idp:AdminConfirmSignUp',
```
edwardfoyle, line 73: I'd suggest taking this out and leaving it as a standalone action

packages/backend-auth/src/userpool_access_policy_factory.ts
josefaidt: agree to taking it out, but entirely (updated the spreadsheet @bombguy )

packages/backend-auth/src/userpool_access_policy_factory.ts
josefaidt: @edwardfoyle are you thinking a standalone meta action? We've bucketed every individual action so far

edwardfoyle: No, just take it out of "manageUsers" but leave it [here](https://github.com/aws-amplify/amplify-backend/pull/1090/files#diff-0141bb8674518ff023c7a33a2bf8bef5a8189f0c6c1ba39558a10195dbc5b886R121)

packages/backend-auth/src/userpool_access_policy_factory.ts
josefaidt: @edwardfoyle similar question but I feel this one should stay a part of "manageUsers" for manual confirmations

packages/backend-auth/src/userpool_access_policy_factory.ts
josefaidt: perhaps we should remove this one entirely?

edwardfoyle: Sure

packages/backend-auth/src/userpool_access_policy_factory.ts
edwardfoyle: My vote would still be to have this separate. Reading the docs, this doesn't seem like it would fall within normal "add/remove users" type scenarios, which is what I'm imagining for "manageUsers". It seems easy enough to add this as an additional permission if you want to do signup confirmations

edwardfoyle: Chatted offline -- if we can confirm that this permission is needed when creating users, then I'm good with leaving this as part of the "manageUsers" grouping

rtpascual (approved):
LGTM, see if we need to have a docs PR similar to what is done for storage: https://github.com/aws-amplify/docs/pull/7009

PR #1096, opened 2024-03-01
Added a function to list stacks which failed to delete

bzsurbhi:
## Problem
Currently, there could be a possibility that an associated stack fails to delete when deleting a branch in Gen2. We do not show those stacks in the Gen2 console as of now.

## Changes
Added a new function to list stacks which failed to delete

## Validation
- Added unit tests
- Tested locally

## Checklist


packages/deployed-backend-client/src/deployed_backend_client.ts
```diff
@@ -134,12 +136,67 @@ export class DefaultDeployedBackendClient implements DeployedBackendClient {
     };
   };
 
+  /**
+   * Returns Amplify stacks which failed to delete for the account and region
+   */
+  listDeleteFailedStacks =
```
Amplifiyer, line 142: This file can use some refactoring. We are duplicating the entire logic as the `listSandboxes`.

We should create one parameterized method that generates the stackMetaData based on inputs and the client (console) can pass the relevant values (e.g. a status filter). We don't need to create new APIs for each filter or type of deployment.

edwardfoyle: hard +1. My suggestion would be to refactor this method and `listSandboxes` into a single `listBackends` method that has the filter args that we want (structured in a way that we can add additional types of filters as the need arises).

We have also taken the stance that the API of this package should not leak the "stack" concept. Here we are leaking both stack and stack status (listDeleteFailedStack).
FWIW, we've already broken that rule a little bit with the `DeployedBackendResource` type, but that doesn't mean we should continue leaking more.

bzsurbhi: Refactored into a single method

packages/deployed-backend-client/API.md
```diff
@@ -173,22 +173,23 @@ export type FunctionConfiguration = {
...
+    status: BackendDeploymentStatus;
+    backendId: BackendIdentifier | undefined;
 };
 
 // @public (undocumented)
-export type ListSandboxesResponse = {
-    sandboxes: SandboxMetadata[];
-    nextToken: string | undefined;
+export type ListBackendsRequest = {
+    deploymentType: string;
```
edwardfoyle, line 185: This should be `DeploymentType` from `plugin-types`

packages/deployed-backend-client/src/deployed_backend_client.ts
```diff
@@ -82,23 +82,38 @@ export class DefaultDeployedBackendClient implements DeployedBackendClient {
...
+    listBackendsRequest?: ListBackendsRequest
+  ): Promise<ListBackendsResponse> => {
+    const stackMetadata: ListBackendsMetadata[] = [];
 
+    let nextToken = listBackendsRequest?.nextToken;
+    const deploymentType = listBackendsRequest?.deploymentType;
+    let stackStatusFilter: StackStatus[] = [];
+    if (deploymentType == 'branch') {
+      stackStatusFilter = ['DELETE_FAILED'];
+    }
```
edwardfoyle, line 95: Why are we coupling the deployment type to a delete_failed stack status? We should probably have two filter criteria on ListBackendRequest, one for backend type and another for backend deployment status. We should use our BackendDeploymentStatus type rather than expose CFN status directly

packages/deployed-backend-client/src/deployed_backend_client.ts
```diff
@@ -121,23 +136,25 @@ export class DefaultDeployedBackendClient implements DeployedBackendClient {
...
 
     return {
-      sandboxes: stackMetadata,
+      backends: stackMetadata,
       nextToken,
     };
   };
 
-  private isSandboxStack = (stackName: string | undefined): boolean => {
+  private isSandboxOrBranchStack = (
```
edwardfoyle, line 152:
```suggestion
  private getBackendStackType = (
```

packages/deployed-backend-client/src/deployed_backend_client_factory.ts
```diff
@@ -26,14 +26,15 @@ export enum ApiAuthType {
...
+export type ListBackendsMetadata = {
   name: string;
   lastUpdated: Date | undefined;
   status: BackendDeploymentStatus;
   backendId: BackendIdentifier | undefined;
 };
 
-export type ListSandboxesRequest = {
+export type ListBackendsRequest = {
+  deploymentType: string;
```
edwardfoyle, line 37: I wonder if this should be an optional field and if not specified we return data from all Amplify stacks?

packages/deployed-backend-client/src/deployed_backend_client_factory.ts
```diff
@@ -26,14 +26,15 @@ export enum ApiAuthType {
...
   nextToken?: string;
```
edwardfoyle, line 38: We probably need to remove this and return some sort of Iterable object with a `.next(token?: string): BackendMetadata[]` method to get the next batch of results. Reason being, what happens when someone passes in a nextToken but with different filter criteria?

packages/deployed-backend-client/src/deployed_backend_client_factory.ts
```diff
@@ -26,14 +26,15 @@ export enum ApiAuthType {
   AMAZON_COGNITO_USER_POOLS = 'AMAZON_COGNITO_USER_POOLS',
 }
 
-export type SandboxMetadata = {
+export type ListBackendsMetadata = {
```
edwardfoyle, line 29:
```suggestion
export type BackendMetadata = {
```

packages/deployed-backend-client/src/deployed_backend_client_factory.ts
```diff
@@ -26,14 +26,15 @@ export enum ApiAuthType {
...
   name: string;
   lastUpdated: Date | undefined;
   status: BackendDeploymentStatus;
   backendId: BackendIdentifier | undefined;
```
edwardfoyle, line 33: Should backendType be included in this object now?

packages/deployed-backend-client/src/deployed_backend_client_factory.ts
edwardfoyle: We could also consider using [yield](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/yield) to return an Iterable that can keep track of the nextToken internally

packages/deployed-backend-client/src/deployed_backend_client_factory.ts
```diff
@@ -26,16 +27,16 @@ export enum ApiAuthType {
   AMAZON_COGNITO_USER_POOLS = 'AMAZON_COGNITO_USER_POOLS',
 }
 
-export type ListBackendsMetadata = {
+export type BackendSummariesMetadata = {
```
edwardfoyle, line 30:
```suggestion
export type BackendSummaryMetadata = {
```

packages/deployed-backend-client/API.md
```diff
@@ -126,7 +140,9 @@ export enum ConflictResolutionMode {
 
 // @public (undocumented)
 export type DeployedBackendClient = {
-    listBackends: (listBackendsRequest?: ListBackendsRequest) => Promise<ListBackendsResponse>;
+    listBackends: (listBackendsRequest?: ListBackendsRequest) => AsyncGenerator<{
+        backends: BackendSummariesMetadata[];
+    }, void, unknown>;
```
edwardfoyle, line 145: while this will work, I think it would be a nicer API to consume if it returned `AsyncGenerator<BackendSummariesMetadata>`. This would allow iterating over the return value like
```ts
for await (const entry of listBackends(props)) {
  // each entry is a backend summary
}
```
instead of having to do a nested loop
```ts
for await (const responseChunk of listBackends(props)) {
  for (const entry of responseChunk.backends) {
    // get the individual entries here
  }
}
```

packages/deployed-backend-client/src/deployed_backend_client.ts
```diff
@@ -141,9 +141,7 @@ export class DefaultDeployedBackendClient implements DeployedBackendClient {
       nextToken = listStacksResponse.nextToken;
 
       if (stackMetadata.length !== 0) {
-        yield {
-          backends: stackMetadata,
-        };
+        yield stackMetadata;
       }
```
edwardfoyle, line 145: What I meant was yield individual metadata objects. eg
```ts
for (const entry of stackMetadata) {
  yield entry
}
```

This prevents the caller from having to do a nested loop to get the entries

sobolk (changes requested):

packages/deployed-backend-client/API.md
```diff
@@ -126,7 +140,7 @@ export enum ConflictResolutionMode {
 
 // @public (undocumented)
 export type DeployedBackendClient = {
-    listSandboxes: (listSandboxesRequest?: ListSandboxesRequest) => Promise<ListSandboxesResponse>;
+    listBackends: (listBackendsRequest?: ListBackendsRequest) => AsyncGenerator<BackendSummaryMetadata[], void, unknown>;
```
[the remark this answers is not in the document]
sobolk: The thing is that API should somehow allow both async iteration and give possibility to caller to do paging themselves. 
Or at least the api should always be open to implement both strategies if needed.

packages/deployed-backend-client/API.md
sobolk: I added my comments here https://github.com/aws-amplify/amplify-backend/pull/1096#discussion_r1517952728 .

TLDR. API should still allow caller handled pagination while offering easier way to craft for loop.
I think our goal should be to enable writing for loop without nesting, and that is DX we should optimize for. But there should be option to iterate page by page (which is secondary scenario and can be a bit more obscure).

packages/deployed-backend-client/API.md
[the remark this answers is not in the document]
edwardfoyle: That's a good call. With that in mind, my suggestion would be
```ts
type ListBackendsResult = {
  getBackendSummaryPage: AsyncGenerator<BackendSummary[]>
}
```
Which can then be extended to include a generator for individual summaries if / when the need arises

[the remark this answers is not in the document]
sobolk: `getBackendSummaryPages` or `getBackendSummaryByPage`

packages/deployed-backend-client/src/deployed_backend_client_factory.ts
```diff
@@ -95,10 +96,14 @@ export enum BackendDeploymentStatus {
   UNKNOWN = 'UNKNOWN',
 }
 
+export enum BackendStatusFilter {
```
Amplifiyer, line 99:
```suggestion
export enum BackendStatus {
```
It's not a filter but a status, somebody filters with these statuses

packages/deployed-backend-client/src/deployed_backend_client.ts
```diff
@@ -83,22 +84,34 @@ export class DefaultDeployedBackendClient implements DeployedBackendClient {
...
+      const listStacksResponse = await this.listStacks(nextToken, statusFilter);
+
+      const listStacksSummaries =
+        deploymentType === 'sandbox'
+          ? listStacksResponse.stackSummaries.filter(
+              (stackSummary: StackSummary) => {
+                return stackSummary.StackStatus !== StackStatus.DELETE_COMPLETE;
+              }
+            )
+          : listStacksResponse.stackSummaries;
```
Amplifiyer, line 108: We shouldn't filter statuses for one particular deployment type if the caller didn't ask it. Especially now that we have a generic method with a stack status as input.

bzsurbhi: This a not(!) filter where we are excluding the stacks with 'DELETE_COMPLETE' status. If we don't add this filter here, we'll have to pass an array of all other(more than 20) statuses in the listStacks() call

packages/deployed-backend-client/src/deployed_backend_client.ts
```diff
@@ -82,23 +83,43 @@ export class DefaultDeployedBackendClient implements DeployedBackendClient {
...
+   * Returns a list of stacks for specific deployment type and status
+   * @yields
    */
-  listSandboxes = async (
-    listSandboxesRequest?: ListSandboxesRequest
-  ): Promise<ListSandboxesResponse> => {
-    const stackMetadata: SandboxMetadata[] = [];
-    let nextToken = listSandboxesRequest?.nextToken;
-
+  async *listBackendsSummaries(listBackendsRequest?: ListBackendsRequest) {
```
edwardfoyle, line 99:
```suggestion
  private async *listBackendsInternal(listBackendsRequest?: ListBackendsRequest) {
```

packages/deployed-backend-client/src/deployed_backend_client.ts
edwardfoyle: I think it makes sense to exclude DELETE_COMPLETE unless the filter explicitly requests it. But we should do it in the listStacks command, not as a client side filter. You can get all of the statuses except this one with the following code
```ts
StackStatusFilter: Object.values(StackStatus).filter(status => status !== StackStatus.DELETE_COMPLETE)
```

edwardfoyle (dismissed):

edwardfoyle:
Merge from main should fix the PR size check

packages/deployed-backend-client/API.md
```diff
@@ -173,22 +187,14 @@ export type FunctionConfiguration = {
...
 };
 
 // @public (undocumented)
-export type SandboxMetadata = {
-    name: string;
-    lastUpdated: Date | undefined;
-    status: BackendDeploymentStatus;
-    backendId: BackendIdentifier | undefined;
+export type ListBackendsResponse = {
+    getBackendSummaryByPage: AsyncGenerator<BackendSummaryMetadata[]>;
```
sobolk, line 197: If this is just property then it should be named `backendSummaryByPage`.
If we want to keep `getBackendSummaryByPage` then this should become 
`getBackendSummaryByPage: () => AsyncGenerator<BackendSummaryMetadata[]>;`

Fwiw. I think this should be a function.

sobolk (approved):

edwardfoyle (approved):

Issue #1150, opened 2024-03-14
allow customization of verification email links


josefaidt:
### Environment information

```plain text
System:
  OS: macOS 14.3.1
  CPU: (10) arm64 Apple M1 Pro
  Memory: 2.16 GB / 32.00 GB
  Shell: /opt/homebrew/bin/fish
Binaries:
  Node: 20.11.1 - ~/Library/Caches/fnm_multishells/76639_1710429675053/bin/node
  Yarn: 1.22.19 - ~/Library/Caches/fnm_multishells/76639_1710429675053/bin/yarn
  npm: 10.2.4 - ~/Library/Caches/fnm_multishells/76639_1710429675053/bin/npm
  pnpm: 8.15.4 - ~/Library/Caches/fnm_multishells/76639_1710429675053/bin/pnpm
NPM Packages:
  @aws-amplify/backend: 0.13.0-beta.5
  @aws-amplify/backend-cli: 0.12.0-beta.5
  aws-amplify: 6.0.20
  aws-cdk: 2.132.1
  aws-cdk-lib: 2.132.1
  typescript: 5.4.2
AWS environment variables:
  AWS_PROFILE = josef-gen2
  AWS_STS_REGIONAL_ENDPOINTS = regional
  AWS_NODEJS_CONNECTION_REUSE_ENABLED = 1
  AWS_SDK_LOAD_CONFIG = 1
No CDK environment variables
```

### Description

Cognito does not provide a raw URL, but rather allows customizing the message by replacing the inner contents of the shortcode
> When you choose Link as the verification type, your custom message must include a placeholder in the format `{##Verify Your Email##}`. You can change the text string between the placeholder characters, for example `{##Click here##}`. A verification link titled Verify Your Email replaces this placeholder.
https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pool-settings-message-customizations.html#cognito-user-pool-settings-email-verification-message-customization

However the auth construct uses the default, static value `{##Verify Your Email##}` https://github.com/aws-amplify/amplify-backend/blob/main/packages/auth-construct/src/construct.ts#L63

This is especially important for i18n

josefaidt:
[...]

PR #1181, opened 2024-03-21
feat: add support for email verification link text customization

awsluja:
## Problem
Currently, developers can not customize the text displayed for the email verification link. This is because the value was hard-coded to {##Verify Email##} prior to these changes. Hard-coding was done due a [bug with CDK](https://github.com/aws/aws-cdk/issues/23828) which prevented customizations at the time. That [bug was fixed recently](https://github.com/aws/aws-cdk/pull/28832).

https://github.com/aws-amplify/amplify-backend/issues/1150
## Changes
This change adds support for customizing the text used in the email verification link.

## Validation
Tests have been updated to ensure that the link text is correctly formed and that default values for the link text are still correct if there are no customizations.

Example for 'LINK':
image

Example for 'CODE':
image

## Checklist


packages/auth-construct/API.md
```diff
@@ -42,14 +42,10 @@ export type AuthProps = {
 };
 
 // @public
-export type EmailLogin = true | EmailLoginSettings;
+export type EmailLogin = true | EmailLoginSettings | Record<string, never>;
```
sobolk, line 45: What does `Record<string, never>` mean in this context ? Why would `EmailLogin` loosely typed?

packages/auth-construct/src/construct.test.ts
```diff
@@ -243,7 +243,8 @@ void describe('Auth construct', () => {
   void it('creates email login mechanism with specific settings', () => {
     const app = new App();
     const stack = new Stack(app);
-    const emailBodyFunction = (code: string) => `custom email body ${code}`;
+    const emailBodyFunction = (code: () => string) =>
```
0618, line 246: nit: if we convert `code` from a `string` to a `function`. should we also rename `code` to `getCode`?

packages/auth-construct/src/construct.test.ts
```diff
@@ -363,8 +365,8 @@ void describe('Auth construct', () => {
   void it('does not throw if valid email verification message for LINK', () => {
     const app = new App();
     const stack = new Stack(app);
-    const emailBodyFunction = (link: string) =>
-      `valid message ${link} with link`;
+    const emailBodyFunction = (link: (customLinkText?: string) => string) =>
```
0618, line 368: same for `link`

packages/auth-construct/src/types.ts
```diff
@@ -6,31 +6,52 @@ import {
...
+   */
+  verificationEmailStyle?: 'CODE';
+  /**
+   * Customize your verification emails.
+   * Use the code parameter to inject verification codes into the user verification email.
+   * @example
+   * verificationEmailStyle: "CODE",
+   * verificationEmailBody: (code) => `Your verification code is ${code()}.`
+   */
+  verificationEmailBody?: (code: () => string) => string;
```
0618, line 43: why do we have to change the parameter from `string` to a `function`? 

If we have to, would it be compatible with the previous version? (I guess it's ok to be uncompatible since it's in `beta`)

packages/auth-construct/API.md
awsluja: It represents the empty object type, which allowed us to retain empty object support; but I don't need it anymore. I needed it at one point because I had made one of the properties of EmailLoginSettings required.

packages/auth-construct/src/construct.test.ts
awsluja: let me check with Josef

packages/auth-construct/src/types.ts
```diff
@@ -6,31 +6,52 @@ import {
...
    * @example
    * verificationEmailStyle: "LINK",
-   * verificationEmailBody: (link: string) => `Your verification link is ${link}.`
+   * verificationEmailBody: (link) => `Your verification link is ${link()}.`
+   * @example
+   * To customize the link text, you can provide custom link text to the .
+   * verificationEmailBody: (link) => `Your custom verification link is ${link('custom link text')}.`
    */
-  verificationEmailBody?: (codeOrLink: string) => string;
+  verificationEmailBody?: (link: (customLinkText?: string) => string) => string;
```
josefaidt, line 25: can we rename `customLinkText` to `text`?

packages/auth-construct/src/construct.test.ts
josefaidt: folks can name the arg something different if they'd like, but this is mostly to align with changing `link` to a function as well despite not having an argument on `code`

edwardfoyle (dismissed):

packages/auth-construct/src/types.ts
awsluja: updated

packages/auth-construct/src/types.ts
awsluja: we're ok with this since were still in beta

sobolk (approved):

edwardfoyle (approved):

PR #1227, opened 2024-03-29
add amplify secret name to the generated typescript data schema file

sundersc:
## Problem

Adds the amplify secret name to the generated typescript data schema file.
- The core functionality exists in the API repo `@aws-amplify/graphql-schema-generator` package. Gen2 CLI is responsible for passing the amplify secret name option from the below command to the library.

```sh
npx amplify generate schema-from-database --connection-uri-secret CONN_STRING_SECRET --out schema.rds.ts
```

The underlying library uses the secret name in the generated file as shown in the below link. This file is later used to deploy the backend.
https://github.com/aws-amplify/amplify-category-api/blob/2c8ee6b16c1bb2b5dececa70354d7f4fff616497/packages/amplify-graphql-schema-generator/src/__tests__/__snapshots__/generate-ts-data-schema.test.ts.snap#L115

The PR also upgrades the `@aws-amplify/graphql-schema-generator` version to `0.8.0` which auto detects the database vpc configuration from the given connection uri.

NA

## Changes
- Unit tests
- Manual test

NA. No changes in the command structure.

## Checklist

### DIFF
--- a/packages/cli/src/commands/generate/schema-from-database/generate_schema_command.ts
+++ b/packages/cli/src/commands/generate/schema-from-database/generate_schema_command.ts
@@ -74,8 +74,10 @@ export class GenerateSchemaCommand
     );
 
     await this.schemaGenerator.generate({
-      secretName,
-      connectionUri: connectionUriSecret.value,
+      connectionUri: {
+        secretName,
+        value: connectionUriSecret.value,
+      },
       out: outputFile,
     });
   };
--- a/packages/cli/src/commands/generate/schema-from-database/generate_schema_command.ts
+++ b/packages/cli/src/commands/generate/schema-from-database/generate_schema_command.ts
@@ -74,6 +74,7 @@ export class GenerateSchemaCommand
     );
 
     await this.schemaGenerator.generate({
+      secretName,
       connectionUri: connectionUriSecret.value,
       out: outputFile,
     });
--- a/packages/schema-generator/API.md
+++ b/packages/schema-generator/API.md
@@ -12,8 +12,10 @@ export class SchemaGenerator {
 
 // @public (undocumented)
 export type SchemaGeneratorConfig = {
-    secretName: string;
-    connectionUri: string;
+    connectionUri: {
+        secretName: string;
+        value: string;
+    };
     out: string;
 };
 
--- a/packages/schema-generator/API.md
+++ b/packages/schema-generator/API.md
@@ -12,6 +12,7 @@ export class SchemaGenerator {
 
 // @public (undocumented)
 export type SchemaGeneratorConfig = {
+    secretName: string;
     connectionUri: string;
     out: string;
 };
--- a/packages/schema-generator/src/generate_schema.ts
+++ b/packages/schema-generator/src/generate_schema.ts
@@ -3,8 +3,10 @@ import fs from 'fs/promises';
 import { AmplifyUserError } from '@aws-amplify/platform-core';
 
 export type SchemaGeneratorConfig = {
-  secretName: string;
-  connectionUri: string;
+  connectionUri: {
+    secretName: string;
+    value: string;
+  };
   out: string;
 };
 
@@ -17,12 +19,12 @@ type AmplifyGenerateSchemaError =
  */
 export class SchemaGenerator {
   generate = async (props: SchemaGeneratorConfig) => {
-    const dbConfig = parseDatabaseUrl(props.connectionUri);
+    const dbConfig = parseDatabaseUrl(props.connectionUri.value);
 
     try {
       const schema = await TypescriptDataSchemaGenerator.generate({
         ...dbConfig,
-        connectionUriSecretName: props.secretName,
+        connectionUriSecretName: props.connectionUri.secretName,
       });
       await fs.writeFile(props.out, schema);
     } catch (err) {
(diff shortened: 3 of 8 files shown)


packages/schema-generator/API.md
```diff
@@ -12,6 +12,7 @@ export class SchemaGenerator {
 
 // @public (undocumented)
 export type SchemaGeneratorConfig = {
+    secretName: string;
```
palpatim, line 15: Why do we require both secretName and connectionUri? Shouldn't those both be optional, so that a caller can specify one or the other?

sundersc: The underlying libraries `schema-generator` package in Gen2 repo and `@ws-amplify/graphql-schema-generator` do not know what secret is or how & where to read it. The reason we input the secret name is the generated file need to populate it in the output file.

Gen2 `@aws-amplify/backend-cli` package makes use of the secret client (`import { SecretClient } from '@aws-amplify/backend-secret';`) to read the values and pass it to the library. We do not want the underlying libraries to know anything about how these secrets are managed.

packages/cli/src/commands/generate/schema-from-database/generate_schema_command.ts
```diff
@@ -74,6 +74,7 @@ export class GenerateSchemaCommand
     );
 
     await this.schemaGenerator.generate({
+      secretName,
       connectionUri: connectionUriSecret.value,
```
phani-srikar, line 78: since we're already fetching the secret value at runtime which is `connectionUri` here, why do we need to pass the `secretName` also in addition to that? Can't the underlying libraries directly use the `connectionUri`?

packages/schema-generator/src/generate_schema.ts
```diff
@@ -19,7 +20,10 @@ export class SchemaGenerator {
     const dbConfig = parseDatabaseUrl(props.connectionUri);
 
     try {
-      const schema = await TypescriptDataSchemaGenerator.generate(dbConfig);
+      const schema = await TypescriptDataSchemaGenerator.generate({
+        ...dbConfig,
+        connectionUriSecretName: props.secretName,
```
phani-srikar, line 25: Is this the Amplify secret name or the SSM path to the connection URI? Since it's calling the low level API used by both Gen2 and GraphQL CDK construct users, I assume it is the later?

packages/schema-generator/API.md
palpatim: Got it. As long as the customer isn't having to supply both, I am OK with this.

packages/cli/src/commands/generate/schema-from-database/generate_schema_command.ts
sundersc: We need the secret name to populate in the output file. Refer to the snapshot in the below link.
https://github.com/aws-amplify/amplify-category-api/blob/2c8ee6b16c1bb2b5dececa70354d7f4fff616497/packages/amplify-graphql-schema-generator/src/__tests__/__snapshots__/generate-ts-data-schema.test.ts.snap#L115

For why we don't want to read the connection uri from the secret name in the library, refer to my comment [here](https://github.com/aws-amplify/amplify-backend/pull/1227#discussion_r1544973748).

packages/schema-generator/src/generate_schema.ts
sundersc: It is the amplify secret name. Here, the argument `connectionUriSecretName` is just a string, we use this value in the generated file as [shown in this snapshot](https://github.com/aws-amplify/amplify-category-api/blob/2c8ee6b16c1bb2b5dececa70354d7f4fff616497/packages/amplify-graphql-schema-generator/src/__tests__/__snapshots__/generate-ts-data-schema.test.ts.snap#L115) (which returns a value of type `BackendSecret`). Later when we wire `defineData` with the CDK construct, we need to invoke `BackendSecret.resolvePath` to pass the actual SSM path to the CDK construct.

phani-srikar: Thanks for clarifying.

phani-srikar (dismissed):

packages/schema-generator/API.md
sobolk, line 15: The API as is says that both are mandatory parameters.
If I understand this correctly the customer interacting with CLI won't get to this level of details and they'll just operate at "backend secret" abstraction level as explained by @sundersc .

At the generator abstraction level however it seems that both parameters are related and mandatory (i.e. resolved by cli and passed down). 

Therefore:
Can we call it `connectionUriSecretName` ?
Or if both are required and related can we group them together in a form of:
```
export type SchemaGeneratorConfig = {
  connectionUri: {
     secretName: string;
     value: string;
  }
}
```

It would be best if relationship between these two were visible while reading this API. Additionally if there's area that can't be easily expressed by API (the upstream data flow for instance) it should be explained in jsdocs for this type.

sobolk (commented):
LGTM. I.e. the functionality and data flow makes sense.
Please see comment at API.md file.

packages/schema-generator/API.md
sundersc: Combined the secret name and the value into a single property `connectionUri` as suggested.

sobolk (approved):

edwardfoyle (approved):

PR #1233, opened 2024-04-02
Expand canaries to new regions

sobolk:
## Problem

We're running canaries only in `us-west-2` and therefore we might be late to detect service roll outs that may impact us.

## Changes

This PR has the following changes:
1. It adds three more regions to both canary checks and resource cleanups
2. Adds/fixes existing actions/scripts where `us-west-2` was hardcoded
3. Adds retry mechanism at process level that retries commands 2 times by default. This is to keep probability of false positive low while we expand horizontally.

## Validation

canaries: https://github.com/aws-amplify/amplify-backend/actions/runs/8526147426
cleanup: https://github.com/aws-amplify/amplify-backend/actions/runs/8526285215

## Checklist


scripts/retry.js
```diff
@@ -0,0 +1,71 @@
...
+  'Usage:\n./scripts/retry.js <command>\n./scripts.retry.js --maxAttempts <number> <command>';
+
+class RetryOptions {
+  constructor(maxAttempts, command) {
+    this.maxAttempts = maxAttempts;
+    this.command = command;
+  }
+}
+
+const parseInput = () => {
```
edwardfoyle, line 26: We can use the built-in [parseArgs](https://nodejs.org/api/util.html#utilparseargsconfig) utility to make args parsing a little cleaner / more robust

edwardfoyle (commented):
LGTM, just one suggestion for arg parsing

scripts/retry.js
sobolk: I'll try. If this can surface the vararg-ish tail (the command part) nicely then I'll use it.

sobolk: https://github.com/aws-amplify/amplify-backend/pull/1233/commits/02a8f092dd1500d0ad01e28d2e35fa6e8a75048e

scripts/retry.js
```diff
@@ -24,22 +25,24 @@ class RetryOptions {
 }
 
 const parseInput = () => {
-  if (process.argv.length < 3) {
+  try {
+    const parsedResults = parseArgs({
+      allowPositionals: true,
+      options: {
+        maxAttempts: {
+          type: 'string',
```
sobolk, line 33: image

sobolk: no number, we have to parse :(

edwardfoyle (approved):

scripts/retry.js
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env node
+
```
0618, line 2: Why don't we use typescript for this file?

0618 (approved):

scripts/retry.js
sobolk: This is explained in the comment below.

The problem is that to use typescript we have to run `npm install` first. However, there are cases where we'd like to use retry without installing dependencies. Including wrapping `npm install` command itself as it tend to flake from time to time due to network glitches.

sobolk: The alternative that I was considering was a bash script. But I figured that dependency-less pure JS is easier for us to write/read/understand.

PR #1292, opened 2024-04-12
chore: auto-generate cognito domain when external providers are configured

Amplifiyer:
## Problem

>Customers encounter unexpected client behavior after setting a social provider in defineAuth but not specifying a domainPrefix, which is required when using social providers. Customers are more than likely to configure a domainPrefix that somewhat aligns with their app name and branch, however also encounter friction when attempting to do so as that requires the use of environment variables and setting based on env context (without access to the cdk context pre-construct).

>Customers are also more likely to have a need for configuring a custom domain for their auth by binding it to their app domain. This will also help ensure there is less friction when publishing to app stores, etc.

## Changes

1. Always auto-generate a cognito domain based on the stack name which should be unique for a branch-appId or a sandbox. Use a 20 char hash as the domain prefix.
2. Custom domains can be configured via overrides right now and will be supported first class in future. Since both the custom domain and cognito domain will be configured/present at the same time, surface both in the client config to avoid confusion.

## Validation

Unit tests and manually tested

## Checklist

### DIFF
--- a/packages/auth-construct/src/construct.ts
+++ b/packages/auth-construct/src/construct.ts
@@ -40,6 +40,7 @@ import {
 } from '@aws-amplify/backend-output-storage';
 import * as path from 'path';
 import { coreAttributeNameMap } from './string_maps.js';
+import { createHash } from 'crypto';
 
 type DefaultRoles = { auth: Role; unAuth: Role };
 type IdentityProviderSetupResult = {
@@ -102,14 +103,14 @@ export class AmplifyAuth
 
   private readonly name: string;
 
-  private readonly domainPrefix: string | undefined;
-
   private readonly groups: {
     [key: string]: {
       cfnUserGroup: CfnUserPoolGroup;
       role: Role;
     };
   } = {};
+
+  private domainPrefix: string | undefined;
 
   /**
    * Create a new Auth construct with AuthProps.
@@ -123,7 +124,6 @@ export class AmplifyAuth
     super(scope, id);
 
     this.name = props.name ?? '';
-    this.domainPrefix = props.loginWith.externalProviders?.domainPrefix;
 
     // UserPool
     this.computedUserPoolProps = this.getUserPoolProps(props);
@@ -581,12 +581,12 @@ export class AmplifyAuth
       return result;
     }
     // make sure logout/callback urls are not empty
-    if (external.logoutUrls.length === 0) {
+    if (external.logoutUrls && external.logoutUrls.length === 0) {
       throw Error(
         'You must define logoutUrls when configuring external login providers.'
       );
     }
-    if (external.callbackUrls.length === 0) {
+    if (external.callbackUrls && external.callbackUrls.length === 0) {
       throw Error(
         'You must define callbackUrls when configuring external login providers.'
       );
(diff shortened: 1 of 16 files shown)


packages/auth-construct/src/construct.ts
```diff
@@ -751,15 +751,15 @@ export class AmplifyAuth
       result.providersList.push('SAML');
     }
 
-    // UserPool Domain
-    if (this.domainPrefix && result.providersList.length > 0) {
+    // Auto generate Cognito Domain if external providers are configured
+    if (result.providersList.length > 0) {
```
edwardfoyle, line 755: We should probably use [this](https://github.com/aws-amplify/amplify-backend/blob/ab739ea1ce7f0e297a28020976e51f8b1a2138e9/packages/backend/src/backend_id_scoped_stable_backend_identifiers.ts#L24-L26) recently added hash method. It's injected into the `generateContainerEntry` method in the `backend-auth` factory. I think the construct should continue to have an optional prop for the domainPrefix that can be computed by the factory and passed in.

packages/backend-output-schemas/src/auth/v1.ts
```diff
@@ -25,7 +25,7 @@ export const authOutputSchema = z.object({
 
     socialProviders: z.string().optional(), // JSON array as string
 
-    oauthDomain: z.string().optional(),
+    oauthCognitoDomain: z.string().optional(),
```
edwardfoyle, line 28: should this be required now?

packages/client-config/src/client-config-schema/schema_v1.json
```diff
@@ -105,10 +105,14 @@
               "minItems": 0,
               "uniqueItems": true
             },
-            "domain": {
+            "cognito_domain": {
               "description": "Cognito Domain used for identity providers",
```
edwardfoyle, line 109: nit: just glancing at this, "Custom" and "Cognito" look very similar. Maybe change this description to `Autogenerated domain used for identity providers`?

packages/backend-output-schemas/src/auth/v1.ts
Amplifiyer: It's only required if external providers are present. Since this is a flat config, we can't make it required.

packages/client-config/src/client-config-schema/schema_v1.json
Amplifiyer: Updated.

packages/auth-construct/src/construct.ts
Amplifiyer: Good point, updated.

packages/backend-output-schemas/src/auth/v1.ts
edwardfoyle: Ah right 👍

edwardfoyle (approved):

packages/backend-auth/src/factory.ts
```diff
@@ -130,6 +131,10 @@ class AmplifyAuthGenerator implements ConstructContainerEntryGenerator {
       ),
       outputStorageStrategy: this.getInstanceProps.outputStorageStrategy,
     };
+    if (authProps.loginWith.externalProviders) {
+      authProps.loginWith.externalProviders.domainPrefix =
+        stableBackendIdentifiers.getStableBackendHash();
```
awsluja, line 136: Is the consensus here that a hash of type, namespace, and name is sufficient to avoid collisions?
image

0618 (approved):

packages/backend-auth/src/factory.ts
Amplifiyer: yes, a combination of `type`, `namespace`, and `name` results in a unique stackname

johnf:
> 2\. Custom domains can be configured via overrides

For anyone wondering how to do this, I used

```ts
const { cfnUserPool } = backend.auth.resources.cfnResources;
const { AWS_BRANCH, USER } = process.env;
const env = !AWS_BRANCH ? USER : AWS_BRANCH.replace('backend-admin-', '');

const userPoolDomain = cfnUserPool.node.scope?.node.findChild('UserPoolDomain').node.defaultChild as CfnUserPoolDomain;
userPoolDomain.addPropertyOverride('Domain', `admin-${env}`);
```

There is probably a more correct way but this worked for me.

PR #1344, opened 2024-04-19
fix: disable implicit grant oauth flows by default

awsluja:
## Problem
A bug in the setupExternalProviders would return undefined for the 'oAuthSettings' value if external providers were not configured.
[The undefined value led to enabling both implicit grant and token oauth flows.](https://github.com/aws/aws-cdk/blob/91246acde1ab0512ea6b375f66c283516cb6f2b0/packages/aws-cdk-lib/aws-cognito/lib/user-pool-client.ts#L388)

## Changes

## Validation
Tests updated/added to ensure that implicit grant flow is disabled, even if external providers are not configured.

## Checklist

### DIFF
--- a/packages/auth-construct/src/construct.test.ts
+++ b/packages/auth-construct/src/construct.test.ts
@@ -967,6 +967,24 @@ void describe('Auth construct', () => {
       });
     });
 
+    void it('disables implicit grant OAUTH flow by default', () => {
+      const app = new App();
+      const stack = new Stack(app);
+      new AmplifyAuth(stack, 'test');
+      const template = Template.fromStack(stack);
+      template.hasResourceProperties('AWS::Cognito::UserPool', {
+        Policies: {
+          PasswordPolicy: {
+            MinimumLength: 8,
+            RequireLowercase: true,
+            RequireNumbers: true,
+            RequireSymbols: true,
+            RequireUppercase: true,
+          },
+        },
+      });
+    });
+
     void it('creates a default client with cognito provider', () => {
       const app = new App();
       const stack = new Stack(app);
--- a/packages/auth-construct/src/construct.test.ts
+++ b/packages/auth-construct/src/construct.test.ts
@@ -972,16 +972,31 @@ void describe('Auth construct', () => {
       const stack = new Stack(app);
       new AmplifyAuth(stack, 'test');
       const template = Template.fromStack(stack);
-      template.hasResourceProperties('AWS::Cognito::UserPool', {
-        Policies: {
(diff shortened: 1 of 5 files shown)


packages/auth-construct/src/construct.ts
```diff
@@ -770,9 +772,7 @@ export class AmplifyAuth
       scopes: external.scopes
         ? this.getOAuthScopes(external.scopes)
         : DEFAULT_OAUTH_SCOPES,
-      flows: {
-        authorizationCodeGrant: true,
-      },
+      flows: DEFAULTS.OAUTH_FLOWS,
```
Amplifiyer, line 775: Isn't this redundant since you initialized `result.oAuthSettings.flows` with this value anyways?

packages/auth-construct/src/construct.test.ts
```diff
@@ -967,6 +967,39 @@ void describe('Auth construct', () => {
...
     });
 
+    void it('disables implicit grant OAUTH flow by default', () => {
+      const app = new App();
+      const stack = new Stack(app);
+      new AmplifyAuth(stack, 'test');
+      const template = Template.fromStack(stack);
+      template.hasResourceProperties('AWS::Cognito::UserPoolClient', {
+        AllowedOAuthFlows: ['code'],
+      });
```
Amplifiyer, line 977: wouldn't we want to assert the absence of implicit grant or presence of `implicitCodeGrant: false`? This test would perhaps pass without this change as well.

awsluja: In this case, we really are asserting that 'code' is the only allowed OAuth flow in the UserPoolClient, since that is the only other available option. 
I can rename the test description for clarity

packages/auth-construct/src/construct.ts
awsluja: Cognito's oAuthSettings type uses read-only properties, so a new oAuthSettings object is being assigned here. Even if it was not read-only, copy assignment would be required for the other remaining propertyes (which don't show up in the PR):
```
result.oAuthSettings = {
  callbackUrls: external.callbackUrls,
  logoutUrls: external.logoutUrls,
  scopes: external.scopes,
```
I don't think the trade is worth it here

packages/auth-construct/src/construct.test.ts
awsluja: updated

Amplifiyer (approved):

edwardfoyle (approved):

PR #1376, opened 2024-04-25
feat(backend-data): alias identityPool to iam

iartemiev:
## Changes

* extended public-facing `defaultAuthorizationMode` type to include `identityPool`
* the new value maps to `AWS_IAM` in the Data CDK construct
* added unit test for mapping

## Validation

* unit test

## Checklist

### DIFF
--- a/packages/backend-data/API.md
+++ b/packages/backend-data/API.md
@@ -36,7 +36,7 @@ export type DataProps = {
 export type DataSchemaInput = string | DerivedModelSchema | DerivedCombinedSchema;
 
 // @public
-export type DefaultAuthorizationMode = 'iam' | 'userPool' | 'oidc' | 'apiKey' | 'lambda';
+export type DefaultAuthorizationMode = 'identityPool' | 'userPool' | 'oidc' | 'apiKey' | 'lambda';
 
 // @public
 export const defineData: (props: DataProps) => ConstructFactory<AmplifyData>;
--- a/packages/backend-data/API.md
+++ b/packages/backend-data/API.md
@@ -36,7 +36,7 @@ export type DataProps = {
 export type DataSchemaInput = string | DerivedModelSchema | DerivedCombinedSchema;
 
 // @public
-export type DefaultAuthorizationMode = 'identityPool' | 'userPool' | 'oidc' | 'apiKey' | 'lambda';
+export type DefaultAuthorizationMode = 'iam' | 'identityPool' | 'userPool' | 'oidc' | 'apiKey' | 'lambda';
 
 // @public
 export const defineData: (props: DataProps) => ConstructFactory<AmplifyData>;
--- a/.changeset/forty-files-mix.md
+++ b/.changeset/forty-files-mix.md
@@ -1,0 +1,5 @@
+---
+'@aws-amplify/backend-data': minor
+---
+
+alias identityPool to iam
--- a/.changeset/forty-files-mix.md
+++ b/.changeset/forty-files-mix.md
@@ -2,4 +2,4 @@
 '@aws-amplify/backend-data': minor
 ---
 
-alias identityPool to iam
+support identityPool defaultAuthMode aliased to iam
--- a/package-lock.json
+++ b/package-lock.json
@@ -18229,6 +18229,7 @@
       "version": "11.2.0",
       "resolved": "https://registry.npmjs.org/fs-extra/-/fs-extra-11.2.0.tgz",
       "integrity": "sha512-PmDi3uwK5nFuXh7XDTlVnS17xJS7vW36is2+w3xcv8SVxiB4NyATf4ctkVY5bkSjX0Y4nbvZCq1/EjtEyr9ktw==",
+      "dev": true,
       "dependencies": {
         "graceful-fs": "^4.2.0",
         "jsonfile": "^6.0.1",
@@ -18844,6 +18845,7 @@
       "version": "9.0.6",
       "resolved": "https://registry.npmjs.org/immer/-/immer-9.0.6.tgz",
       "integrity": "sha512-G95ivKpy+EvVAnAab4fVa4YGYn24J1SpEktnJX7JJ45Bd7xqME/SCplFzYFmTbrkwZbQ4xJK1xMTUYBkN6pWsQ==",
+      "dev": true,
       "funding": {
         "type": "opencollective",
         "url": "https://opencollective.com/immer"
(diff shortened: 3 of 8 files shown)


packages/backend-data/API.md
```diff
@@ -36,7 +36,7 @@ export type DataProps = {
 export type DataSchemaInput = string | DerivedModelSchema | DerivedCombinedSchema;
 
 // @public
-export type DefaultAuthorizationMode = 'iam' | 'userPool' | 'oidc' | 'apiKey' | 'lambda';
+export type DefaultAuthorizationMode = 'identityPool' | 'userPool' | 'oidc' | 'apiKey' | 'lambda';
```
sobolk, line 39: I find mapping `identityPool` to `AWS_IAM` in this particular context a bit confusing in this context.

I understand the meaning of `identityPool` in `@auth` directive context. It just says that `identityPool` entities are authorized or not there.

However in this context, where this knob drives behavior of underlying Appsync api this might be confusing because `identityPool` is subset of larger IAM auth.

Consider the following scenario:
1. Customer creates gen2 project
2. They remove/not use Amplify's auth (identity pool concept is removed from their project)
3. They want to have both api key and generic iam auth (which is implicit in gen2) and let's say oidc
4. They want to set default auth to IAM
5. How would they know they should use `identityPool` to achieve that behavior?

iartemiev: Discussed with @renebrandel and we're going to keep both, `identityPool` and `iam` as valid values. `identityPool` will map to `AWS_IAM` still. 

Does that alleviate the concern, @sobolk?

sobolk: It does but it creates another one.

If we add both `identityPool` and `iam` which pretty much have same effect on the default then...
1. How do we tell difference between them for the purpose of setting default mode?
2. How do we tell customer that defaulting to `identityPool` means defaulting to any IAM auth really?

iartemiev: 1. As far as AppSync is concerned, they're the same thing. AppSync doesn't differentiate between an IAM role vended from IDP vs a manually-assumed role. So the default auth mode for the API will be IAM in either case.
2. I believe that's what we're trying to abstract for customers here, hence the mapping. Customers using `identityPool` in their `.authorization(...)` or `@auth` rule can use the same label for the default auth mode without having to think about it just being IAM under the hood. @renebrandel correct me if I'm off here

sobolk: But the knob is called `DefaultAuthorizationMode` which implies that intent is to set Appsync's default authorization mode. Which has some connection with `identityPool` depending on schema. 
The sentence "I want to set api's default authorization mode to `identityPool`" doesn't seem to make much sense and have somewhat surprising result if somebody doesn't read the docs.
I think this should remain as `iam` (only) and get documented, i.e. it's relation to identityPool auth.

sobolk: Btw. adding `identityPool` later in addition to `iam` is non breaking change. Should we hold and see if this is really confusing and try to rely on docs first?

renebrandel: The problem is that customers struggle associating the 'identityPool' with 'iam'. It's not clear that these things are linked up.

The 'defaultAuthorizationMode' will in the future also impact the client library (and the outputs), where we'd want to ensure the auth mode always matches the provider. As in the client library the "identityPool" is used as the authorization mode:

```ts
const { data } = client.models.Todo.list({
  authMode: "identityPool"
})
```

We are already adding the client library auth modes so for customers they can always be sure that whatever "auth mode provider" is defined in their authorization rules, is the auth mode they should use to make the request.

sobolk: @edwardfoyle @Amplifiyer @srquinn21 what do you think?

Amplifiyer: I can see the argument that customers not knowing how `identityPool` is associated with `iam`. We also try to abstract it with `defineAuth` that there are IAM roles under the hood.

sobolk:
```
const { data } = client.models.Todo.list({
  authMode: "identityPool"
})
```

Wouldn't this work even if IAM is secondary auth mode at appsync?

sobolk: Consider the following scenario:
1. I have `defineAuth` in my project
2. I have `defineData` with:
    1. I have model `Todo @auth(identityPool)`
    2. I have model `TodoWithoutAuthDirective`
3. I have multiple auth modes enabled, for example OIDC and/or  API Key
4. Setting `defaultAuthMode = identityPool` enables IAM access for `TodoWithoutAuthDirective`

Is this intuitive and safe ?

josefaidt: Does `iam`/`identityPool` mean "either one of the auth roles plugged into your Identity pool"? What about `authenticatedIamRole` and `unauthenticatedIamRole`?

sobolk: No. They're mapped to Appsync knobs.
image

sobolk: Ok. I think I was able to convince myself.

I have run the following experiment:

**App 1 with:**

```
const schema = `
type Todo @model @auth(rules: [{ allow: public, provider: identityPool }]) {
        id: ID!
        title: String!
}

type TodoWithoutAuthDirective @model {
        id: ID!
        title: String!
}
`

  schema,
  authorizationModes: {
    apiKeyAuthorizationMode: {
      expiresInDays: 7
    },
    oidcAuthorizationMode: {
      clientId: 'abcd',
      oidcIssuerUrl: 'https://localhost/',
      oidcProviderName: 'foo',
      tokenExpireFromIssueInSeconds: 60,
      tokenExpiryFromAuthInSeconds: 60,
    }
  }
});
```

results in:
image


**App 2 with:**

```
const schema = `
type Todo @model @auth(rules: [{ allow: public, provider: identityPool }]) {
        id: ID!
        title: String!
}

type TodoWithoutAuthDirective @model {
        id: ID!
        title: String!
}
`

  schema,
  authorizationModes: {
    defaultAuthorizationMode: 'iam',
    apiKeyAuthorizationMode: {
      expiresInDays: 7
    },
    oidcAuthorizationMode: {
      clientId: 'abcd',
      oidcIssuerUrl: 'https://localhost/',
      oidcProviderName: 'foo',
      tokenExpireFromIssueInSeconds: 60,
      tokenExpiryFromAuthInSeconds: 60,
    }
  }
});
```

results in:
image


**The conclusion**
It appears that no matter what `defaultAuthorizationMode` all models are granted IAM access regardless of default mode setting. This is because we always set `enableIamAuthorizationMode: true` when calling data construct.

Please consider this discussion resolved then, as scenario where IAM access is granted by accident is moot point.

sobolk (dismissed):

svidgen (dismissed):

sobolk (approved):

svidgen (approved):

PR #1483, opened 2024-05-09
fix: auth construct uses cfn resources to generate output instead of input props

awsluja:
## Problem
See: https://github.com/aws-amplify/amplify-backend/issues/762

Previously, developers that made changes to their Auth resource's passwordPolicy using CDK would not see those changes reflected in the stored outputs file.
This would happen if they added social providers as well, as the socialProviders array would not contain the social provider if it was added outside of the construct.

Additionally, the tests have been updated to use the latest strategy for applying overrides, which ensures that those changes will be applied to the output.
image

https://github.com/aws-amplify/amplify-backend/issues/762
## Changes

## Validation

Tests have been added for overriding password policies and having those changes reflected in the stored output.
image

Tests have been added to ensure social providers are included if they are added via cdk.
image

## Checklist


packages/auth-construct/src/construct.test.ts
```diff
@@ -630,75 +624,159 @@ void describe('Auth construct', () => {
...
-    void it('stores outputs in platform - minimum config', () => {
-      const authConstruct = new AmplifyAuth(stack, 'test', {
+    void it('outputs defaults for minimal email config', () => {
+      new AmplifyAuth(stack, 'test', {
         loginWith: {
           email: true,
         },
-        outputStorageStrategy: stubBackendOutputStorageStrategy,
       });
+      const template = Template.fromStack(stack);
```
awsluja, line 641: note: instead of stubbing the outputStorageStrategy, we are using Template.fromStack to generate the final output, and validating that instead

sobolk (dismissed):

packages/auth-construct/src/construct.ts
```diff
@@ -900,132 +902,209 @@ export class AmplifyAuth
...
+          const policy = (cfnUserPool.policies as CfnUserPool.PoliciesProperty)
+            .passwordPolicy as CfnUserPool.PasswordPolicyProperty;
+          const requirements = [];
+          policy.requireNumbers && requirements.push('REQUIRES_NUMBERS');
+          policy.requireLowercase && requirements.push('REQUIRES_LOWERCASE');
+          policy.requireUppercase && requirements.push('REQUIRES_UPPERCASE');
+          policy.requireSymbols && requirements.push('REQUIRES_SYMBOLS');
+          return JSON.stringify(requirements);
+        }
+        return undefined;
```
0618, line 989:
```suggestion
        if (!cfnUserPool.policies) {
          return;
        }
        const {requireNumbers, requireLowercase, requireUppercase, requireSymbols} = (cfnUserPool.policies as CfnUserPool.PoliciesProperty)
            .passwordPolicy as CfnUserPool.PasswordPolicyProperty;
          return JSON.stringify([
            ...(requireNumbers ? ['REQUIRES_NUMBERS']: []),
            ...(requireLowercase ? ['REQUIRES_LOWERCASE'] : []),
            ...(requireUppercase ? ['REQUIRES_UPPERCASE'] : []),
            ...(requireSymbols ? ['REQUIRES_SYMBOLS'] : []),
        ]);
```

packages/auth-construct/src/construct.ts
```diff
@@ -900,132 +902,209 @@ export class AmplifyAuth
...
-      output.mfaConfiguration = this.computedUserPoolProps.mfa;
+    // extract social providers from UserPool resource
+    output.socialProviders = Lazy.string({
+      produce: () => {
+        const outputProviders = [];
+        const userPoolProviders = this.resources.userPool.identityProviders;
+        if (userPoolProviders) {
+          for (const provider of userPoolProviders) {
+            const providerType =
+              provider.node['_children']['Resource']['providerType'];
```
0618, line 1023: Pls use dot notations in this file as well

packages/auth-construct/src/construct.ts
```diff
@@ -900,132 +902,209 @@ export class AmplifyAuth
...
+              outputProviders.push(
+                provider.node['_children']['Resource']['providerName']
+              );
+            }
+          }
+          if (outputProviders.length > 0) {
+            return JSON.stringify(outputProviders);
+          }
+        }
+        return undefined;
```
0618, line 1052: Pls move the `return undefined;` to top and change it to `return`

packages/auth-construct/src/construct.test.ts
```diff
@@ -630,75 +624,159 @@ void describe('Auth construct', () => {
...
+      const outputs = template['template']['Outputs'];
```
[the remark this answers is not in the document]
awsluja: using brackets here because the properties are not exposed via dot

packages/client-config/src/client-config-contributor/client_config_contributor_v1.ts
```diff
@@ -74,12 +74,6 @@ export class AuthClientConfigContributor implements ClientConfigContributor {
       authOutput.payload.mfaTypes
     );
 
-    parseAndAssignObject(
```
edwardfoyle, line 77: why is this removed?

packages/auth-construct/src/construct.ts
```diff
@@ -900,132 +902,209 @@ export class AmplifyAuth
       Stack.of(this)
     )
   ): void => {
```
edwardfoyle, line 904: Since this file is already quite large, this method could be a good candidate to break out into another file. Not a blocker though

edwardfoyle (approved):

packages/auth-construct/src/construct.ts
```diff
@@ -900,132 +902,209 @@ export class AmplifyAuth
...
+    // these properties cannot be overwritten
     const output: AuthOutput['payload'] = {
       userPoolId: this.resources.userPool.userPoolId,
       webClientId: this.resources.userPoolClient.userPoolClientId,
       identityPoolId: this.resources.cfnResources.cfnIdentityPool.ref,
       authRegion: Stack.of(this).region,
-      allowUnauthenticatedIdentities:
+    };
+
+    // the properties below this line can be overwritten, so they are exposed via cdk LAZY
```
Amplifiyer, line 913: Is it not possible to extract the entire `cfnUserPool` once instead of for each property?

packages/auth-construct/src/construct.test.ts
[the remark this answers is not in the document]
0618: Weird. Why not? They are just strings. Did you mean the type error? Can we do something like `template?.template`?

packages/auth-construct/src/construct.ts
```diff
@@ -900,132 +902,209 @@ export class AmplifyAuth
...
-      this.providerSetupResult.oAuthSettings.callbackUrls
-    ) {
-      const oAuthSettings = this.providerSetupResult.oAuthSettings;
-      if (this.domainPrefix) {
-        output.oauthCognitoDomain = `${this.domainPrefix}.auth.${
-          Stack.of(this).region
-        }.amazoncognito.com`;
-      }
+    const getUserPoolClientResource = () => {
+      return this.resources.userPoolClient.node['_children']['Resource'];
```
Amplifiyer, line 1057: We shouldn't do that. An `_` typically represents an internal variable/API that is not meant for consumption by clients, i.e. cdk might change this or behavior without requiring a Major version.

packages/auth-construct/src/construct.ts
awsluja: updated

packages/auth-construct/src/construct.ts
```diff
@@ -900,132 +902,209 @@ export class AmplifyAuth
...
-        this.computedUserPoolProps.passwordPolicy.minLength?.toString();
+    // extract the passwordPolicy from the UserPool policies
+    output.passwordPolicyMinLength = Lazy.string({
+      produce: () => {
+        if (cfnUserPool.policies) {
+          const policy = (cfnUserPool.policies as CfnUserPool.PoliciesProperty)
+            .passwordPolicy as CfnUserPool.PasswordPolicyProperty;
+          return policy.minimumLength?.toString();
+        }
+        return undefined;
```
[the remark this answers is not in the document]
awsluja: updated

packages/auth-construct/src/construct.ts
```diff
@@ -900,132 +902,209 @@ export class AmplifyAuth
...
+          const schema =
+            cfnUserPool.schema as CfnUserPool.SchemaAttributeProperty[];
+          schema.forEach((attribute) => {
+            if (attribute.required && attribute.name) {
+              requiredAttributes.push(attribute.name);
+            }
+          });
+          return JSON.stringify(
+            requiredAttributes.map((attr) => attr.toLowerCase())
+          );
```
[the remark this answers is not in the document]
awsluja: great suggestions, updated

packages/client-config/src/client-config-contributor/client_config_contributor_v1.ts
awsluja: its a duplicate of the above lines

packages/auth-construct/src/construct.ts
awsluja: good point; I will try to find another way to extract providerType

awsluja: I found a way to do this with findResource and casting to the resource type

packages/auth-construct/src/construct.ts
awsluja: yes, just updated thanks

packages/auth-construct/src/construct.ts
```diff
@@ -1026,30 +1032,23 @@ export class AmplifyAuth
...
-      return this.resources.userPoolClient.node['_children']['Resource'];
-    };
-
     output.oauthCognitoDomain = Lazy.string({
       produce: () => {
-        const userPoolDomain =
-          this.resources.userPool.node['_children']['UserPoolDomain'];
+        const userPoolDomain = this.resources.userPool.node.tryFindChild(
+          'UserPoolDomain'
+        ) as UserPoolDomain | undefined;
```
Amplifiyer, line 1049: We should use typeguards instead and throw errors if the object does not pertain to the type casted.

packages/auth-construct/src/construct.test.ts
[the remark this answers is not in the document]
awsluja: I was able to eliminate this instance by using template.findOutputs('*') which returns an object containing all of the outputs

packages/auth-construct/src/construct.ts
awsluja: I updated the return statement, but the 2nd portion of this suggestion seems a bit hacky so I just left the original for that

packages/auth-construct/src/construct.ts
awsluja: updated to use findResource() combined with casting; that helped eliminate all usage of '_children' etc

packages/auth-construct/src/construct.ts
awsluja: doing this would produce an IResolvable for the cfnUserPool , which is not something we can access properties of (if we were trying to build the output from the IResolvable)

packages/auth-construct/src/construct.ts
awsluja: thanks, I updated this to check for types

packages/auth-construct/src/construct.ts
```diff
@@ -900,132 +904,194 @@ export class AmplifyAuth
...
+      produce: () => {
+        return cfnUserPoolClient.allowedOAuthFlows
+          ? cfnUserPoolClient.allowedOAuthFlows.join(',')
+          : '';
+      },
+    });
+
+    output.oauthClientId = Lazy.string({
+      produce: () => {
+        return this.resources.userPoolClient.userPoolClientId;
```
Amplifiyer, line 1092:
```suggestion
        return cfnUserPoolClient.userPoolClientId;
```

packages/auth-construct/src/construct.ts
```diff
@@ -900,132 +904,194 @@ export class AmplifyAuth
...
+          ? cfnUserPoolClient.logoutUrLs.join(',')
+          : '';
+      },
+    });
+
+    output.oauthResponseType = Lazy.string({
+      produce: () => {
+        return cfnUserPoolClient.allowedOAuthFlows
+          ? cfnUserPoolClient.allowedOAuthFlows.join(',')
+          : '';
```
Amplifiyer, line 1086: Should the default be `code` as it was before?

packages/auth-construct/src/construct.ts
```diff
@@ -1017,86 +1009,81 @@ export class AmplifyAuth
...
-                provider.node['_children']['Resource']['providerName']
-              );
-            }
+        if (!userPoolProviders || userPoolProviders.length === 0) {
+          return;
+        }
+        for (const provider of userPoolProviders) {
+          const providerResource = provider.node.findChild(
+            'Resource'
+          ) as CfnUserPoolIdentityProvider;
```
edwardfoyle, line 1018: Do we need any validations here that the object has the expected shape?

edwardfoyle (dismissed):

packages/auth-construct/src/construct.ts
```diff
@@ -900,132 +904,194 @@ export class AmplifyAuth
...
-        output.oauthCognitoDomain = `${this.domainPrefix}.auth.${
+    output.oauthCognitoDomain = Lazy.string({
+      produce: () => {
+        const userPoolDomain =
+          this.resources.userPool.node.tryFindChild('UserPoolDomain');
+        if (!userPoolDomain) {
+          return;
+        }
+        if (!(userPoolDomain instanceof UserPoolDomain)) {
+          throw Error('UserPoolDomain is not of type UserPoolDomain.');
```
Amplifiyer, line 1052:
```suggestion
          throw Error('Could not find UserPoolDomain resource in the stack.');
```

0618 (approved):

packages/auth-construct/src/construct.ts
```diff
@@ -900,132 +904,194 @@ export class AmplifyAuth
...
+      produce: () => {
+        const outputProviders = [];
+        const userPoolProviders = this.resources.userPool.identityProviders;
+        if (!userPoolProviders || userPoolProviders.length === 0) {
+          return;
+        }
+        for (const provider of userPoolProviders) {
+          const providerResource = provider.node.findChild(
+            'Resource'
+          ) as CfnUserPoolIdentityProvider;
```
Amplifiyer, line 1018: We should add type checks for all places where we do `.node.findChild`

awsluja: updated

packages/auth-construct/src/construct.ts
awsluja: the default is set above and should be returned in the left hand side of the ternary operation
image

packages/auth-construct/src/construct.ts
awsluja: updated to return cfnUserPoolClient.ref;

packages/auth-construct/src/construct.ts
awsluja: updated

edwardfoyle (commented):
Looks like the e2e tests are not happy. eg https://github.com/aws-amplify/amplify-backend/actions/runs/9292498952/job/25573634149#step:7:237

0618 (dismissed):

packages/deployed-backend-client/src/stack_metadata_output_retrieval_strategy.ts
```diff
@@ -95,12 +95,14 @@ export class StackMetadataBackendOutputRetrievalStrategy
     Object.entries(backendOutputMetadata).forEach(([outputKeyName, entry]) => {
       const outputData = entry.stackOutputs.reduce(
         (accumulator, outputName) => {
+          if (stackOutputRecord[outputName] === undefined) {
+            return {
+              ...accumulator,
+            };
```
edwardfoyle, line 101: nit

```suggestion
            return accumulator;
```

awsluja: updated

packages/auth-construct/src/construct.ts
```diff
@@ -1059,7 +1059,7 @@ export class AmplifyAuth
         const userPoolDomain =
           this.resources.userPool.node.tryFindChild('UserPoolDomain');
         if (!userPoolDomain) {
-          return;
+          return '';
```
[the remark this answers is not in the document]
awsluja: We must specify some entry, for Lazy values to be used. Empty string is supported, but using undefined will cause:
"Deployment failed: Error [ValidationError]: Template format error: Every Outputs member must contain a Value object"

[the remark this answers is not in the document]
edwardfoyle: Does that mean that [this line](https://github.com/aws-amplify/amplify-backend/pull/1483/files/d054b1a19818b39c5d17c915cbac365120888b7d..8b5a13b0c2e5610fb25e897dd192da3619c3fdf9#diff-8cbab95ba8f311985c905dbe6e63fbaa448a8cec22c852b662760cddaba4293dR98) should also check for empty string and omit it if it's empty?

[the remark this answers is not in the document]
awsluja: I'll check for both undefined and empty string; that would catch both

sobolk (approved):

0618 (approved):

sobolk (approved):

edwardfoyle (approved):

PR #1492, opened 2024-05-10
feat: add support for function logs streaming to sandbox

Amplifiyer:
## Changes

Add support for streaming function logs in sandbox
* CLI
  * Default to disabled with an explicit opt-in `--stream-function-logs`
  * Allow customers to specify a list of function names for which logs should be streamed
  * Sandbox with `--once` should not stream function logs
  * Allow specifying a file path to stream all the function's execution log events to. 

* Display
  * Display log events with the "friendly name" of the function in a specific color. Total of 5 colors used in round robin
  * Display a timestamp for each log event
  * Colors are turned off for file writing.

* Functionality
  * While sandbox is idle, function logs are streamed in parallel for all the functions as they are executed.
  * Log streaming is paused during a deployment and resumes from the time the deployment was initiated to avoid intermingling the logs
  * New functions, deleted functions are automatically updated after a deployment, i.e. if customer adds a new function, it's logs will be automatically streamed (if no filter was used)

* New permissions need to be added to the managed policy AmplifyBackendDeployFullAccess
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "FunctionLogsStreaming",
            "Effect": "Allow",
            "Action": [
                "lambda:ListTags",
                "logs:FilterLogEvents"
            ],
            "Resource": [
                "arn:aws:logs:*:*:log-group:/aws/lambda/amplify-*:*",
                "arn:aws:lambda:*:*:function:amplify-*"
            ]
        }
    ]
}
```

## Validation

Unit tests added

## Checklist


packages/sandbox/src/cloudwatch_logs_monitor.ts
```diff
@@ -0,0 +1,297 @@
...
+    return Promise.all(promises);
+  };
+
+  /**
+   * Print out one CloudWatch event using the local printer.
+   */
+  private print = (event: CloudWatchLogEvent): void => {
+    const cloudWatchEventDisplay = this.allLogGroups.find(
+      (logGroup) => logGroup.logGroupName === event.logGroupName
+    )?.cloudWatchEventDisplay;
```
edwardfoyle, line 220: We should store this lookup as a map rather than iterating over the array on every print call

packages/sandbox/src/cloudwatch_logs_monitor.ts
```diff
@@ -0,0 +1,297 @@
...
+    if (!cloudWatchEventDisplay) {
+      return;
+    }
+    const friendlyResourceName =
+      cloudWatchEventDisplay[event.logGroupName].friendlyName;
+    const color = cloudWatchEventDisplay[event.logGroupName].color;
```
edwardfoyle, line 226: nit
```suggestion
    const { color, friendlyName } = cloudWatchEventDisplay[event.logGroupName];
```

packages/sandbox/src/file_watching_sandbox.ts
```diff
@@ -234,6 +255,7 @@ export class FileWatchingSandbox extends EventEmitter implements Sandbox {
         // not reset tracker prematurely
         this.shouldValidateAppSources
       );
+      // await new Promise((resolve) => setTimeout(resolve, 1000));
```
edwardfoyle, line 258: no longer needed?

packages/sandbox/src/lambda_function_log_streamer.ts
```diff
@@ -0,0 +1,169 @@
...
+          LogLevel.DEBUG
+        );
+      }
+    }
+
+    // finally start listening
+    this.logsMonitor?.activate();
+  };
+
+  public deactivate = () => {
```
edwardfoyle, line 111:
```suggestion
  public stopWatchingLogs = () => {
```
To match the other method and to avoid confusion with `CloudWatchLogEventMonitor.deactivate`

packages/sandbox/src/lambda_function_log_streamer.ts
```diff
@@ -0,0 +1,169 @@
...
+      return;
+    }
+
+    for (const entry of deployedFunctionNameToArnMap) {
+      const listTagsResponse = await this.lambda.send(
+        new ListTagsCommand({
+          Resource: entry.arn,
+        })
+      );
+      const friendlyFunctionName =
```
edwardfoyle, line 76: using `getBackendMetadata()` already does this filtering

Amplifiyer: Do we get the tagged names there? I thought it was only the physical Arn

https://github.com/aws-amplify/amplify-backend/blob/57a5138dbbc92b385e2e46258fa165dfb422afe7/packages/deployed-backend-client/src/deployed_backend_client.ts#L346-L363

packages/sandbox/package.json
```diff
@@ -39,6 +42,7 @@
     "@types/parse-gitignore": "^1.0.0"
   },
   "peerDependencies": {
-    "aws-cdk": "^2.127.0"
+    "aws-cdk": "^2.127.0",
+    "aws-cdk-lib": "^2.127.0"
```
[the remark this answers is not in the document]
Amplifiyer: As I mentioned in the other comment, `DeployedBackendClient.getBackendMetadata()` is too heavy to be used here. Moreover I looked at the `ArnParser` implemented there and it doesn't cover few cases such as AWS Partition.

My goal is to move this `aws-cdk-lib` ARN parsing in the `deployed-backend-client` and deprecating what's already there. But that will come in a follow up refactor as this PR is already too big

packages/sandbox/src/lambda_function_log_streamer.ts
```diff
@@ -0,0 +1,169 @@
...
+    functionNamesToFilter?: string[]
+  ) => {
+    const backendOutput: BackendOutput =
+      await this.backendOutputClient.getOutput(sandboxBackendId);
+
+    const definedFunctionsPayload =
+      backendOutput[functionOutputKey]?.payload.definedFunctions;
+    const deployedFunctionNames = definedFunctionsPayload
+      ? (JSON.parse(definedFunctionsPayload) as string[])
+      : [];
```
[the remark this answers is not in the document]
edwardfoyle: Maybe we need some sort of "filter" prop for `getBackendMetadata` where it can be instructed to only load metadata from certain verticals? So in this case, we could use it to only load function metadata

packages/cli-core/src/format/format.ts
```diff
@@ -71,6 +80,18 @@ export class Format {
     Object.entries(record)
       .map(([key, value]) => `${key}: ${String(value)}`)
       .join(os.EOL);
+  color = (message: string, color: Color) => color(message);
 }
 
+export type Color = Colorize;
+export type ColorName = keyof typeof colors;
+
+export const colors = {
```
edwardfoyle, line 89: this is still exporting kleur functions. Why not have
```ts
export enum Color {
  GREEN = 'green';
  YELLOW = 'yellow';
  etc
}
```

Then internally map those enum values to the kleur function.

packages/sandbox/package.json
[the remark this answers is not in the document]
edwardfoyle: I think we should consider ways to make `getBackendMetadata` work for this use case (or add additional methods on `DeployedBackendClient`). We could make such refactors a prerequisite to this change.

packages/sandbox/src/lambda_function_log_streamer.ts
edwardfoyle: ah, yeah if we need the friendly name, we'd need to do some extra processing or expand the interface the deployed backend client

packages/sandbox/src/lambda_function_log_streamer.ts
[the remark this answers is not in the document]
Amplifiyer: We don't even need the function stack as technically what we are loading is not a `metadata` unless we start adding the "friendly name" in the metadata or outputs section. In that case we can continue to use `getOutput` here.

I also don't see any issues with depending on `backend-output-schemas` here. `backendOutputClient.getOutput` is the right abstraction here.

packages/sandbox/src/lambda_function_log_streamer.ts
Amplifiyer: I don't think we are at a point that warrants to expand the interface of deployed backend client. Also one can argue that `tags` (or other resource specific properties) are not considered `metadata` and shouldn't be part of this package. We can look into adding the `friendly-name` to stack outputs to avoid `lambda:list-tags`, but I don't think we are there yet.

packages/cli-core/src/format/format.ts
Amplifiyer: Updated to not export the `colors` object but just a type and enumerated color names.

packages/sandbox/src/cloudwatch_logs_monitor.ts
```diff
@@ -0,0 +1,297 @@
...
+
+  private active = false;
+
+  private printer = printer; // default stdout
+
+  /**
+   * Initializes the start time to be `now`
+   */
+  constructor(readonly cloudWatchLogsClient: CloudWatchLogsClient) {
+    this.startTime = Date.now();
```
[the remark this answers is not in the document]
edwardfoyle: This is why it mostly doesn't matter for now. Because instantiation and activation are happening within a couple ms of each other. But just from an API perspective, it's a bit odd for instantiation to start a clock that will impact the behavior of the methods. Typically ctors are for injecting dependencies

packages/cli-core/src/printer/printer.ts
```diff
@@ -19,10 +20,19 @@ export class Printer {
...
+    private readonly stdout:
+      | WriteStream
+      | NodeJS.WritableStream = process.stdout,
+    private readonly stderr:
+      | WriteStream
+      | NodeJS.WritableStream = process.stderr,
     private readonly refreshRate: number = 500
-  ) {}
+  ) {
+    // if not running in tty, turn off colors
```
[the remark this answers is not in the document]
edwardfoyle: Aren't the methods in `format` already a noop if colors are not enabled? Why do we need additional logic for it here?

packages/sandbox/src/lambda_function_log_streamer.ts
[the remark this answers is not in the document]
edwardfoyle: Taking a closer look at this, it looks like we're only using the functionOutputKey from that package. However, `getOutput` should already have typed keys so we can just use the string directly here and we'll get typechecking without having to import the schema package.

packages/cli-core/src/printer/printer.ts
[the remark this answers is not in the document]
Amplifiyer: `format` uses `supportColors` which only looks at the environment/terminal you are running in and decides whether to enable or disable colors. E.g. in CI/CD the colors might be disabled. See https://github.com/isaacs/color-support

In the printer, we might be piping the data in some other place as well (e.g. writing to a file) `supportColors` doesn't know that and will keep the colors enabled.

packages/sandbox/src/lambda_function_log_streamer.ts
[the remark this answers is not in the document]
Amplifiyer: I will update it to using the key `"AWS::Amplify::Function"` directly, but wouldn't it just add duplicity. Not that we are ever going to change it but I don't understand the drawback of this dependency?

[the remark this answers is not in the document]
edwardfoyle: It's not a huge deal. In my head, the `deployed-backend-client` is the "entry point" by which all readers should get info about the backend. the `schema` package is meant to be a contract between the `backend-client` package and the places where the output is written (the constructs). Ideally the schema shouldn't be needed directly by consumers of `deployed-backend-client`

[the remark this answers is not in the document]
Amplifiyer: It makes sense to have one writer, but I believe it's fine to have multiple readers. We already have `client-config` and `model-generator`

packages/cli/src/commands/sandbox/sandbox_command.ts
```diff
@@ -196,6 +219,31 @@ export class SandboxCommand
...
+        .option('function-name', {
+          describe:
+            'Name of functions to stream execution logs from. Usage --function-name func1 --function-name func2. Default: Stream all functions logs',
+          array: true,
+          type: 'string',
+          group: 'Options to configure streaming of lambda function logs',
+          implies: ['stream-function-logs'],
+          requiresArg: true,
+        })
+        .option('stream-output', {
```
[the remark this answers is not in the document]
Amplifiyer: This is now changed to `log-file-out`

packages/cli/src/commands/sandbox/sandbox_command.ts
```diff
@@ -196,6 +219,31 @@ export class SandboxCommand
...
           global: false,
         })
+        .option('stream-function-logs', {
+          describe:
+            'Whether to stream function execution logs. Default: false. Use --function-names to filter for specific functions to stream logs from',
+          boolean: true,
+          global: false,
+          group: 'Options to configure streaming of lambda function logs',
+        })
+        .option('function-name', {
```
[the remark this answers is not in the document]
Amplifiyer: We are keeping the extra argument for filtering, but it's now accepts a regex and is called `logs-filter`. This will be applicable for any cloudwatch logs we extend to in future (such as data resolver logs)

packages/sandbox/package.json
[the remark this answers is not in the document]
sobolk: I agree that adding dependency on `aws-cdk-lib` for `ArnParser` seems too heavy.
Perhaps we should invest in `ArnParser` and move it to `platform-core`?

packages/sandbox/src/cloudwatch_logs_monitor.ts
```diff
@@ -0,0 +1,284 @@
...
+} from '@aws-sdk/client-cloudwatch-logs';
+import fs from 'fs';
+import path from 'path';
+
+/**
+ * After reading events from all CloudWatch log groups how long should we wait to read more events.
+ *
+ * If there is some error with reading events (i.e. Throttle) then this is also how long we wait until we try again
+ */
+const SLEEP_MS = 2_000;
```
sobolk, line 23: should this be another parameter to sandbox command?

packages/sandbox/src/cloudwatch_logs_monitor.ts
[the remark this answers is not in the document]
sobolk: This should go to `activate` . (and if we do this it could be removed from `deactivate`).

packages/sandbox/src/cloudwatch_logs_monitor.ts
```diff
@@ -0,0 +1,284 @@
...
+      ? outputLocation
+      : path.resolve(process.cwd(), outputLocation);
+    this.printer = new Printer(
+      LogLevel.INFO,
+      fs.createWriteStream(targetPath, { flags: 'a', autoClose: true })
+    );
+  };
+
+  /**
+   * Pick the next color in the object `colors` in round robin fashion
```
sobolk, line 158: Fwiw, as somebody who's partially colorblind I would encourage to add secondary disambiguation mechanism other than colors.
For example ability to prefix log lines with function name, something like:
```
[func1] log1
[func2] log2
```
and so on.

packages/sandbox/src/cloudwatch_logs_monitor.ts
```diff
@@ -0,0 +1,284 @@
...
+    return colorNames[
+      this.allLogGroups.length % colorNames.length
+    ] as ColorName;
+  };
+
+  private scheduleNextTick = (sleep: number): void => {
+    setTimeout(() => void this.tick(), sleep);
+  };
+
+  private tick = async (): Promise<void> => {
```
sobolk, line 170: nit: `onTick`

packages/sandbox/src/cloudwatch_logs_monitor.ts
```diff
@@ -0,0 +1,284 @@
...
+      });
+    } catch (e) {
+      throw new AmplifyFault(
+        'CloudWatchStreamingFault',
+        { message: `Error occurred while monitoring logs: %s` },
+        e instanceof Error ? e : undefined
+      );
+    }
+
+    this.scheduleNextTick(SLEEP_MS);
```
sobolk, line 187: Is there any reason why we didn't try `setTimeout` or `setInterval` ? Or is this just equivalent?

packages/sandbox/src/cloudwatch_logs_monitor.ts
```diff
@@ -0,0 +1,284 @@
...
+
+  /**
+   * Reads all new log events from a set of CloudWatch Log Groups in parallel
+   */
+  private readNewEvents = async (): Promise<Array<CloudWatchLogEvent>> => {
+    const promises: Array<Promise<Array<CloudWatchLogEvent>>> = [];
+    for (const logGroup of this.allLogGroups) {
+      promises.push(this.readEventsFromLogGroup(logGroup));
+    }
+    return (await Promise.all(promises)).flat();
```
sobolk, line 198: What happens if these calls start to fail, get throttled?
Would sandbox stop streaming or would it keep trying until problem goes away?

Should sandbox try for predefned period of time or number of times and give up at some point?

packages/sandbox/src/cloudwatch_logs_monitor.ts
```diff
@@ -0,0 +1,284 @@
...
+          if (event.timestamp && endTime < event.timestamp) {
+            endTime = event.timestamp;
+          }
+        }
+      }
+      // As long as there are _any_ events in the log group `filterLogEvents` will return a nextToken.
+      // This is true even if these events are before `startTime`. So if we have 100 events and a nextToken
+      // then assume that we have hit the limit and let the user know some messages have been suppressed.
+      // We are essentially showing them a sampling (10000 events printed out is not very useful)
+      if (filteredEvents.length === 100 && response.nextToken) {
```
sobolk, line 264: I think that this shouldn't apply when we're streaming to file.

File is a different case and somebody may want to grep it/search it and missing entries will be surprising there.

packages/sandbox/src/file_watching_sandbox.ts
```diff
@@ -209,6 +227,7 @@ export class FileWatchingSandbox extends EventEmitter implements Sandbox {
    */
   stop = async () => {
     this.printer.log(`[Sandbox] Shutting down`, LogLevel.DEBUG);
+    this.functionsLogStreamer?.stopWatchingLogs();
```
sobolk, line 230: should this be `stopStreamingLogs` for naming consistency sake?

sobolk (commented):
Looks good overall.

packages/sandbox/src/cloudwatch_logs_monitor.ts
sobolk: also, for console. should sampling be configurable through command parameters?

packages/cli/src/commands/sandbox/sandbox_command.test.ts
```diff
@@ -118,6 +122,9 @@ void describe('sandbox command', () => {
     assert.match(output, /--outputs-format/);
     assert.match(output, /--outputs-out-dir/);
     assert.match(output, /--once/);
+    assert.match(output, /--stream-function-logs/);
+    assert.match(output, /--logs-filter/);
+    assert.match(output, /--logs-out-file/);
```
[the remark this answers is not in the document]
Amplifiyer: That's how it was earlier. But the product requirements are to combine the `--logs-filter` and `--logs-out-file` for all types of logs and have a on/off switch for each type of log.

packages/cli/src/commands/sandbox/sandbox_command.ts
```diff
@@ -210,7 +257,13 @@ export class SandboxCommand
...
           return true;
         })
-        .conflicts('once', ['exclude', 'dir-to-watch'])
+        .conflicts('once', [
+          'exclude',
+          'dir-to-watch',
+          'stream-function-logs',
+          'logs-filter',
+          'logs-out-file',
+        ])
```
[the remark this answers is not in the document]
Amplifiyer: you can then just start sandbox, if there are no changes, cdk skips the deployment

packages/sandbox/src/cloudwatch_logs_monitor.ts
Amplifiyer: yes that's how it is

```console
illed Duration: 69 ms   Memory Size: 512 MB     Max Memory Used: 90 MB
[defaultNodeFunction] 2:35:31 PM START RequestId: a8d1913a-598f-43e0-8a96-f7e566a63108 Version: $LATEST
[defaultNodeFunction] 2:35:32 PM END RequestId: a8d1913a-598f-43e0-8a96-f7e566a63108
[defaultNodeFunction] 2:35:32 PM REPORT RequestId: a8d1913a-598f-43e0-8a96-f7e566a63108 Duration: 1056.85 ms   Billed Duration: 1057 ms Memory Size: 512 MB     Max Memory Used: 90 MB
```

packages/sandbox/src/cloudwatch_logs_monitor.ts
Amplifiyer: it's just a wrapper over `setTimeout`

packages/sandbox/src/cloudwatch_logs_monitor.ts
sobolk: Awesome, thanks.

packages/sandbox/src/cloudwatch_logs_monitor.ts
Amplifiyer: > also, for console. should sampling be configurable through command parameters?

This feature is exclusively for sandbox, so won't apply for console at all.

> I think that this shouldn't apply when we're streaming to file.

There is a security aspect here as well to prevent overwhelming the machine with too many logs (DoS)

packages/sandbox/src/file_watching_sandbox.ts
Amplifiyer: yeah I can change both start and stop to say `stream` rather than watch.

packages/sandbox/src/cloudwatch_logs_monitor.ts
sobolk: > This feature is exclusively for sandbox, so won't apply for console at all.

I meant `console.log` vs `file.write` scenario on local computer.

packages/sandbox/src/cloudwatch_logs_monitor.ts
Amplifiyer: Sandbox will keep trying with it's polling frequency, sandbox won't get interrupted through since log watching is it's running asynchronously.

Let me look into adding more error handling to perhaps stop streaming if too many unrecoverable errors happen.

packages/sandbox/src/cloudwatch_logs_monitor.ts
Amplifiyer: Oh I see, that is not part of the requirements, but can be easily added in the future if needed.

packages/sandbox/src/cloudwatch_logs_monitor.ts
Amplifiyer: Isn't `onTick` a event listener and the callback is just a `tick` which is what is this method.

packages/sandbox/src/cloudwatch_logs_monitor.ts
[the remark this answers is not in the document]
Amplifiyer: The use case for this is to keep track of when we "deactivated" the log streamer such that when we resume, we resume from that point onward and not when the caller asked to resume.

This helps in pausing the streaming during the deployment. As soon as the deployment is detected, we pause the log streaming, resets the start_time to now so that when we resume, we don't miss the logs during the deployment time.

Doing this way prevents the caller to maintain timings and let this module own it.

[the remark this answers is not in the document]
sobolk: Should we have `start`,`stop` and `pause` then ? or perhaps `start` and `pause` for now.
Perhaps this naming would be less ambiguous. I'm not sure if `deactivate` means `stop` or `pause`.

.github/workflows/health_checks.yml
```diff
@@ -187,7 +187,10 @@ jobs:
         with:
           ref: ${{ steps.get_baseline_commit_sha.outputs.baseline_commit_sha }}
       - uses: ./.github/actions/setup_node
-      - uses: ./.github/actions/restore_build_cache
+      - name: Install and build baseline version
+        run: |
+          npm ci
+          npm run build
```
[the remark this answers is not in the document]
Amplifiyer: https://github.com/aws-amplify/amplify-backend/pull/1624

packages/cli-core/src/printer/printer.ts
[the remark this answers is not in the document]
Amplifiyer: I removed it from printer such that printer will print whatever is given to it. Instead of fiddling with kleur's `$` the caller can then choose to not use `format` or colors.

packages/sandbox/src/cloudwatch_logs_monitor.ts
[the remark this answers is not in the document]
Amplifiyer: Yup, changing to `pause`

.github/workflows/health_checks.yml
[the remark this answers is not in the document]
Amplifiyer: I'll remove this here once #1624 is merged, otherwise the e2e fails.

packages/sandbox/package.json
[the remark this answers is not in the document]
Amplifiyer: That would be worse though adding `aws-cdk-lib` dependency on `platform-core`?

For now, I found the same parser in aws-sdk, so instead switched it to that.

packages/sandbox/src/cloudwatch_logs_monitor.ts
Amplifiyer: Fixed the error handling. We will now pause the streaming when an exception happens. Subsequent deployments if any will again resume the streaming if it has been fixed.

packages/cli/src/commands/sandbox/sandbox_command.test.ts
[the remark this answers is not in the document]
sobolk: > for all types of logs and have a on/off switch for each type of log.

I think this is premature and should be done after we have 2nd category of logs.
But if this backfires we can hide/deprecate these options if necessary so not a blocker.

packages/sandbox/src/cloudwatch_logs_monitor.ts
sobolk: I think this is going to be most wanted feature for people who stream logs to local file.
Mind creating GH issue?

packages/sandbox/src/cloudwatch_logs_monitor.ts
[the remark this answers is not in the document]
sobolk: I was thinking about this more and I'm wondering if over time this could evolve to something like this:
```
monitor.start() // always sets startTime as now
// observe logs then pause
monitor.pause() // pauses streaming
// observe another batch of logs
monitor.resume() // un-pauses streaming from last checkpoint
// pause again and go for lunch
monitor.pause() // pause again
// after lunch resume but from current time as going for lunch would accumulate too much logs.
monitor.restart() OR another monitor.start() // to reset checkpoint to now
```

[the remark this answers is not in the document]
Amplifiyer: Perhaps, but I think that's more complicated than what we have today, `start` and `pause`

packages/sandbox/src/cloudwatch_logs_monitor.ts
Amplifiyer: It is a trigger though, it invokes the `this.readNextBatchOfEvents`. This is keeping scheduler `scheduleNextTick` and `tick` separate than the reading cloudwatch logic, i.e. `this.readNextBatchOfEvents`

packages/sandbox/src/cloudwatch_logs_monitor.ts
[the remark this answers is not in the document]
sobolk: Yes. The goal of my comment was to see if we should set ourselves towards that direction. I.e. to MVP would be to have `start`, `pause` and `resume` . Or perhaps just `pause` and `resume` if we treat ctor call as implicit start.

packages/sandbox/src/cloudwatch_logs_monitor.ts
Amplifiyer: https://github.com/aws-amplify/amplify-backend/issues/1625

packages/sandbox/package.json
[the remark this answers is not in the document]
sobolk: > That would be worse though adding aws-cdk-lib dependency on platform-core?

I'm not sure about that. Problem is that CDK lib is kind of special dependency for us and by current design it's customer's project that set it's version. I.e. platform-core would need to take peer dep on it and we'd need to make sure we don't use symbols from version newer than baseline. Which is all solvable but feels heavy to just parse arn.

> For now, I found the same parser in aws-sdk, so instead switched it to that.

I think this is good. `aws-sdk` is a not as special as our cdk dep.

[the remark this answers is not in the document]
sobolk: And that CDK lib comes in a form of small dedicated package. Not "whole sdk".

packages/sandbox/src/cloudwatch_logs_monitor.ts
sobolk: I see. This is private function, so please make a call. This is non-blocking nit.

packages/sandbox/src/cloudwatch_logs_monitor.ts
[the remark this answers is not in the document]
Amplifiyer: The `ctor` can't do a start since we don't have the `logGroups` information. Also I think `activate` captures both start/resume.

Moreover, none of this is part of public API interface, so will be easy to change whenever we need.

sobolk (approved):

packages/sandbox/package.json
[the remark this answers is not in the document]
Amplifiyer: > Problem is that CDK lib is kind of special dependency for us and by current design it's customer's project that set it's version

Correct, that's why it was a peer dependency here as well. But agree, aws-sdk is definitely better.

packages/cli/src/commands/sandbox/sandbox_command.ts
```diff
@@ -196,6 +219,30 @@ export class SandboxCommand
           boolean: true,
           global: false,
         })
+        .option('stream-function-logs', {
+          describe:
+            'Whether to stream function execution logs. Default: false. Use --function-names to filter for specific functions to stream logs from',
```
edwardfoyle, line 224:
```suggestion
            'Whether to stream function execution logs. Default: false. Use --logs-filter in addition to this flag to stream specific function logs',
```

packages/cli/src/commands/sandbox/sandbox_command.ts
```diff
@@ -196,6 +219,30 @@ export class SandboxCommand
...
+          boolean: true,
+          global: false,
+          group: 'Logs streaming',
+        })
+        .option('logs-filter', {
+          describe: `Regex pattern to filter logs from only matched functions. E.g. to stream logs for a function, specify it's name, and to stream logs from all functions starting with auth specify 'auth' Default: Stream all logs`,
```
edwardfoyle, line 230: Shouldn't a regex for "functions starting with auth" be `^auth.*`? Also, why are we supporting regex and array input? You can make a regex to support multiple disjoint strings with `(foo|bar|baz)`

IMO if we are going to support regex, it should just be a single arg

packages/sandbox/src/file_watching_sandbox.ts
```diff
@@ -133,6 +134,12 @@ export class FileWatchingSandbox extends EventEmitter implements Sandbox {
 
     await this.printSandboxNameInfo(options.identifier);
 
+    if (options.functionStreamingOptions?.logsOutFile) {
+      this.functionsLogStreamer.setOutputLocation(
+        options.functionStreamingOptions.logsOutFile
+      );
+    }
```
edwardfoyle, line 141: This seems like something that should be set in the ctor of the LambdaFunctionLogStreamer

packages/sandbox/src/file_watching_sandbox.ts
```diff
@@ -164,7 +174,15 @@ export class FileWatchingSandbox extends EventEmitter implements Sandbox {
         await this.deploy(options);
       }
       latch = 'open';
+
+      // Idle state, let customers know and start streaming function logs if requested
       this.emitWatching();
+      if (options.functionStreamingOptions?.enabled) {
```
edwardfoyle, line 180: Seems like the FunctionLogStreamer should be given all of the streaming options in its ctor and then it can decide internally if `startStreamingLogs` and `stopStreamingLogs` should be a noop or not based on the given config. Then all the FileWatchingSandbox has to know about is calling start/stop streaming logs at the right times

packages/sandbox/src/lambda_function_log_streamer.ts
```diff
@@ -0,0 +1,178 @@
...
+          LogLevel.DEBUG
+        );
+        shouldStreamLogs = true;
+      }
+
+      if (shouldStreamLogs) {
+        this.logsMonitor?.addLogGroups(
+          friendlyFunctionName,
+          // a CW log group is implicitly created for each lambda function with the lambda function's name
+          `/aws/lambda/${entry.name}`
```
edwardfoyle, line 111: _technically_ customers could override this. Not sure if it's worth another API call to lookup the log stream that's configured for the lambda?

packages/sandbox/src/sandbox.ts
```diff
@@ -37,6 +37,13 @@ export type SandboxOptions = {
   format?: ClientConfigFormat;
   profile?: string;
   watchForChanges?: boolean;
+  functionStreamingOptions?: SandboxFunctionStreamingOptions;
```
edwardfoyle, line 40: related to a comment above, I would remove this from the SandboxOptions and instead inject it directly into the LambdaFunctionLogStreamer

packages/cli/src/commands/sandbox/sandbox_command.ts
Amplifiyer: 1. For a full regex match yes, but I'm going for a partial string match as in my opinion it's easier to specify and also a default for pretty much all regex matchers. I don't see why we would prevent partial matching.
2. Similarly providing array just makes it easier to provide input instead of creating complex regex pattern. @josefaidt to provide more input here.

packages/sandbox/src/file_watching_sandbox.ts
Amplifiyer: None of the sandbox options are provided to the `SandboxSingletonFactory` like `outDir`. I have changed it to pass this option in the start/activate like the `BackendDeployer` instead of just a method for `setOutputLocation`

packages/sandbox/src/lambda_function_log_streamer.ts
Amplifiyer: Correct and I think it's fine to start with this and expand later on how customers are changing their log configuration (e.g. they can also decide to base 64 encode their logs). FWIW, cdk also doesn't retrieve actual CW log group but assumes this default.

packages/sandbox/src/file_watching_sandbox.ts
Amplifiyer: Sure, updated.

packages/cli/src/commands/sandbox/sandbox_command.ts
josefaidt: :wave: yes the important piece of regex is the wildcard match (e.g. `auth*` picks up all functions named with `auth` like `auth-post-confirmation` and `auth-pre-signup`) but it is much easier to specify multiple patterns than it is to craft complex regex when writing the command

```
--logs-filter auth* --logs-filter resolver*
```

I think regex can be a bit of a hurdle though compared to a glob pattern

sobolk (approved):

packages/sandbox/src/lambda_function_log_streamer.ts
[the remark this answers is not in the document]
edwardfoyle: That's true but they both use `deployed-backend-client` to get backend output. Neither of them goes directly to CFN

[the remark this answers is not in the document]
Amplifiyer: This is also coming from `deployed-backed-client`
```ts
const backendOutput: BackendOutput = await this.backendOutputClient.getOutput(sandboxBackendId);
```

Same as https://github.com/aws-amplify/amplify-backend/blob/20bf679430a15277a8c50525dfadfe30a8f4cc1d/packages/cli/src/commands/generate/forms/generate_forms_command.ts#L69 and https://github.com/aws-amplify/amplify-backend/blob/20bf679430a15277a8c50525dfadfe30a8f4cc1d/packages/client-config/src/generate_client_config.ts#L52

packages/cli/src/commands/sandbox/sandbox_command.ts
edwardfoyle: Do we want to support glob patterns or regex then? The current impl is using regex which would require `--logs-filter auth.* logs-filter resolver.*`

Amplifiyer: glob patterns don't make sense for non pathlike strings. Globs are typically (or only?) used for pathname expansions.

packages/sandbox/src/lambda_function_log_streamer.ts
```diff
@@ -0,0 +1,185 @@
...
+  ) {}
+
+  /**
+   * Starts streaming logs in the given sandbox.
+   * @param sandboxBackendId The sandbox backend identifier.
+   * @param streamingOptions Options to configure the log streaming.
+   */
+  startStreamingLogs = async (
+    sandboxBackendId: BackendIdentifier,
+    streamingOptions?: SandboxFunctionStreamingOptions
```
edwardfoyle, line 40: Since these parameters don't change over the lifecycle of this class, it seems like they should be part of the ctor

packages/sandbox/src/cloudwatch_logs_monitor.ts
```diff
@@ -0,0 +1,293 @@
...
+    this.startTime = Date.now();
+  }
+
+  /**
+   * resume writing/printing events
+   * If output location file is specified, the logs will be appended to that file.
+   * If the file doesn't exist it will be created.
+   * @param outputLocation file location
+   */
+  activate = (outputLocation?: string): void => {
```
edwardfoyle, line 104: Should outputLocation be part of the ctor?

edwardfoyle (commented):
Looks good, just left a few comments regarding what props get passed in to methods vs ctors

edwardfoyle (approved):

packages/cli/src/commands/sandbox/sandbox_command.ts
josefaidt: There are use cases for glob patterns outside filenames
https://www.sqlitetutorial.net/sqlite-glob/
https://duckdb.org/docs/sql/functions/pattern_matching.html#glob

like [minimatch](https://github.com/isaacs/minimatch) without the globstar

Amplifiyer: Minimatch is basically a glob implementation, still only relevant for pathlike string matching, not for any arbitrary matches.

In the SQL world, the `glob` is basically implemented as a regex, e.g. there is no such thing as `**`.

sobolk (approved):

PR #1576, opened 2024-05-28
use ssm parameter to detect CDK boostrap

sobolk:
## Problem

CDK can be bootstrapped in a region using custom stack name. I.e. other than `CDKToolkit`

https://github.com/aws-amplify/amplify-backend/issues/1510

## Changes

This PR changes detection logic to use SSM parameter that contains version.

SSM access is already required for other functionalities to work.

## Checklist

### DIFF
--- a/packages/sandbox/src/file_watching_sandbox.ts
+++ b/packages/sandbox/src/file_watching_sandbox.ts
@@ -299,6 +299,7 @@ export class FileWatchingSandbox extends EventEmitter implements Sandbox {
       const { Parameters: parameters } = await this.ssmClient.send(
         new GetParametersByPathCommand({
           Path: CDK_BOOTSTRAP_VERSION_PARAMETER_PREFIX,
+          Recursive: true,
         })
       );
 
--- a/packages/sandbox/src/file_watching_sandbox.ts
+++ b/packages/sandbox/src/file_watching_sandbox.ts
@@ -15,7 +15,7 @@ import _open from 'open';
 // EventEmitter is a class name and expected to have PascalCase
 // eslint-disable-next-line @typescript-eslint/naming-convention
 import EventEmitter from 'events';
-import { GetParametersByPathCommand, SSMClient } from '@aws-sdk/client-ssm';
+import { GetParameterCommand, SSMClient } from '@aws-sdk/client-ssm';
 import {
   AmplifyPrompter,
   LogLevel,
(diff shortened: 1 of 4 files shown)


edwardfoyle (commented):
Functionally the change LGTM, but it looks like the `AmplifyBackendDeployFullAccess` policy only grants permission to read SSM params that start with `/amplify`:
```
{
            "Sid": "AmplifySSM",
            "Effect": "Allow",
            "Action": [
                "ssm:GetParametersByPath",
                "ssm:GetParameters",
                "ssm:GetParameter"
            ],
            "Resource": [
                "arn:aws:ssm:*:*:parameter/amplify/*"
            ],
            "Condition": {
                "StringEquals": {
                    "aws:ResourceAccount": "${aws:PrincipalAccount}"
                }
            }
        },
```

sobolk:
> Functionally the change LGTM, but it looks like the `AmplifyBackendDeployFullAccess` policy only grants permission to read SSM params that start with `/amplify`:
> 
> ```
> {
>             "Sid": "AmplifySSM",
>             "Effect": "Allow",
>             "Action": [
>                 "ssm:GetParametersByPath",
>                 "ssm:GetParameters",
>                 "ssm:GetParameter"
>             ],
>             "Resource": [
>                 "arn:aws:ssm:*:*:parameter/amplify/*"
>             ],
>             "Condition": {
>                 "StringEquals": {
>                     "aws:ResourceAccount": "${aws:PrincipalAccount}"
>                 }
>             }
>         },
> ```

Thanks. I was assuming we don't have this problem without actually checking it.

packages/sandbox/src/file_watching_sandbox.ts
```diff
@@ -35,8 +32,9 @@ import {
   BackendIdentifierConversions,
 } from '@aws-amplify/platform-core';
 
-export const CDK_BOOTSTRAP_STACK_NAME = 'CDKToolkit';
-export const CDK_BOOTSTRAP_VERSION_KEY = 'BootstrapVersion';
+// CDK stores bootstrap version in parameter store. Example parameter name looks like /cdk-bootstrap/<hash>/version
```
awsluja, line 35: is this documented somewhere by CDK

sobolk: It's reverse engineered from https://github.com/aws/aws-cdk/blob/526d4ad8a8528d7f7c5665673ad39dd62fda3a17/packages/aws-cdk/lib/api/bootstrap/bootstrap-template.yaml#L616-L622 .
We haven't found a better way.

sobolk: Here's some additional context.

From https://docs.aws.amazon.com/cdk/v2/guide/bootstrapping.html
> --qualifier is a string that is added to the names of all resources in the bootstrap stack. A qualifier lets you avoid resource name clashes when you provision multiple bootstrap stacks in the same environment. The default is hnb659fds (this value has no significance).
> Changing the qualifier also requires that your CDK app pass the changed value to the stack synthesizer. For more information, see [Stack synthesizers](https://docs.aws.amazon.com/cdk/v2/guide/bootstrapping.html#bootstrapping-synthesizers).

From https://docs.aws.amazon.com/cdk/v2/guide/bootstrapping.html#bootstrapping-synthesizers
> The qualifier is added to the name of bootstrap resources to distinguish the resources in separate bootstrap stacks. To deploy two different versions of the bootstrap stack in the same environment (AWS account and Region), the stacks must have different qualifiers.
> This feature is intended for name isolation between automated tests of the CDK itself. Unless you can very precisely scope down the IAM permissions given to the AWS CloudFormation execution role, there are no permission isolation benefits to having two different bootstrap stacks in a single account. Therefore, there's usually no need to change this value.
> To change the qualifier, configure the DefaultStackSynthesizer either by instantiating the synthesizer with the property:

The important pieces seem to be **This feature is intended for name isolation between automated tests of the CDK itself** and **The default is hnb659fds (this value has no significance)** .

So to use other value than default would require passing qualifier to all stack synthesizers and it seems we should only look up for default setup to check if CDK boostrap is Gen2 compatible

edwardfoyle (approved):

abhi7cr (approved):
LGTM

awsluja (approved):

Issue #1543, opened 2024-05-21
RFC: Enable Python, Go, and others with `defineFunction`


josefaidt:
Hey folks :wave: we're looking to add support for "custom" runtimes with `defineFunction`! This is a follow-up to the following issues, and serves as a request for feedback on how we're thinking of addressing the feature requests

* https://github.com/aws-amplify/amplify-backend/issues/1486
* https://github.com/aws-amplify/amplify-backend/issues/949

We are looking to bridge the gap between `defineFunction` and defining functions using CDK, which then require sideloading to Amplify resources with CDK (e.g. adding a Python function as a trigger for Storage's underlying S3 bucket). To achieve this, we're thinking about an escape hatch to wrap initialized CDK constructs with `defineFunction`:

```ts
import { defineFunctionWithAnyLanguage as defineFunction } from "@aws-amplify/backend"
import { Runtime } from "aws-cdk-lib/aws-lambda"
import { GoFunction } from "@aws-cdk/aws-lambda-go-alpha"
import { PythonFunction } from "@aws-cdk/aws-lambda-python-alpha"

// a regular function
defineFunction({
  name: "my-regular-function",
})

// a go function
defineFunction((scope) => {
  return new GoFunction(scope, "GoFunction", {
    entry: "app/cmd/api",
  })
})

// a python function
defineFunction((scope) => {
  return new PythonFunction(scope, "PythonFunction", {
    entry: '/path/to/my/function',
    runtime: Runtime.PYTHON_3_8,
  })
})
```

Many of the features of `defineFunction` today are specific to the TypeScript developer experience such as secrets, automatic secret resolution, typed environment variables, and sensible default configuration for bundling. Using this escape hatch would opt out of consuming these features.

However, this gives you a way to use Python, Go, or other runtimes with your Functions, and attach to other Amplify resources without CDK. This enables use cases such as:

* Python resolvers for Data models, custom queries
* Go resolvers for Data models, custom queries
* Python-based triggers for Storage

Let us know what you think!

LukaASoban:
This looks great! Exactly what I was hoping for (similar way of defining the function itself) 

Would we also be able to use these functions as handlers for graphql resolvers in the same way as before? For example below I could pass in the function to the handler?

```
    /**
     * Custom query to return all UserProfiles by email or phone number.
     */
    getUsersByEmailOrPhone: a
      .query()
      .arguments({
        emails: a.string().array(),
        phoneNumbers: a.string().array(),
      })
[...]

Issue #1486, opened 2024-05-09
Detailed docs for: Go, Python Runtimes functions, how can we create and assign these functions


MarlonJD:
### Environment information

```plain text
System:
  OS: macOS 14.4.1
  CPU: (8) arm64 Apple M2
  Memory: 361.41 MB / 16.00 GB
  Shell: /bin/zsh
Binaries:
  Node: 18.20.2 - ~/.nvm/versions/node/v18.20.2/bin/node
  Yarn: 1.22.22 - /opt/homebrew/bin/yarn
  npm: 10.5.0 - ~/.nvm/versions/node/v18.20.2/bin/npm
  pnpm: undefined - undefined
NPM Packages:
  @aws-amplify/backend: 1.0.0
  @aws-amplify/backend-cli: 1.0.1
  aws-amplify: 6.3.0
  aws-cdk: 2.140.0
  aws-cdk-lib: 2.140.0
  typescript: 5.4.5
AWS environment variables:
  AWS_STS_REGIONAL_ENDPOINTS = regional
  AWS_NODEJS_CONNECTION_REUSE_ENABLED = 1
  AWS_SDK_LOAD_CONFIG = 1
No CDK environment variables
```

### Description

On AWS Amplify Gen 2 document, it says there is only nodejs runtimes for `defineFunction`

Currently, only Node runtimes are supported by defineFunction. However, you can change the Node version that is used by the function. The default is the oldest Node LTS version that is supported by AWS Lambda (currently Node 18). [Source](https://docs.amplify.aws/flutter/build-a-backend/functions/configure-functions/#runtime)

On the [Gen 1 vs. Gen 2 feature matrix](https://docs.amplify.aws/flutter/start/migrate-to-gen2/#gen-1-vs-gen-2-feature-matrix). It says, Function runtime: .NET 6, Go, Java, Javascript and Python is available with `Yes with CDK`.

How can we create for example Go Lambda function and use as a custom GraphQL query in gen 2. There was detailed document on gen 1 for custom query with lambda function.

We were creating go function with `amplify add function`, then just assigning this function name with this:

```
type Query {
  echo(msg: String): String @function(name: "echofunction", region: "us-east-1")
}
```

Now, how can we do this? Also we need to get current user's information with body, can we do this in gen 2 ? For example I'm using user's cognitoId to get user's detail and get permissions about custom job.

Which options we have? How can we connect our Golang function to custom query, can I get user's cognate id with this query?

[...]

PR #1602, opened 2024-06-03
Feature: custom runtimes to define function

MarlonJD:
## Problem

#1543, #1486

## Changes

~~I changed `runtime` parameter to string from int. It was only nodejs version. Now it can be:~~

~~`'GO_1_X_PROVIDED_AL2023' | 'NODEJS_16_X' | 'NODEJS_18_X' | 'NODEJS_20_X' | 'PYTHON_3_8' | 'PYTHON_3_9' | 'PYTHON_3_10' | 'PYTHON_3_11' | 'PYTHON_3_12'`~~

~~It's default `NODEJS_18_X ` same as old functions, people only need to specify `runtime: "NODEJS_16_X"` if they want another version, it's giving lint error, so it's easy to inform people if it's changed. If it's empty, it's still creating nodejs version, I also add required to `entry` parameter when it's go or python runtimes.~~

Edit: I modified `defineFunction` to props and callback function can be parameters at the same time: 

```
defineFunction({
  name: "my-regular-function",
})

// a go function
defineFunction((scope) => {
  return new GoFunction(scope, "GoFunction", {
    entry: "app/cmd/api",
  })
})
```

can be usable like proposal in RFC, go function will be usable out of box, python function will be require docker, so python function cannot be used on CI/CD, it can be deploy with [custom pipeline](https://docs.amplify.aws/react/deploy-and-host/fullstack-branching/custom-pipelines/) on local.

ie: disabling auto deploy, `npm ci && export CI=1 && npx ampx pipeline-deploy --branch BRANCH_NAME --app-id AMPLIFY_APP_ID` will do the trick for python.

## Checklist

- [X] If this PR requires a change to the [Project Architecture README](../PROJECT_ARCHITECTURE.md), I have included that update in this PR.
- [X] If this PR requires a docs update, I have linked to that docs PR above.


edwardfoyle (changes requested):
Hi @MarlonJD thanks for taking the time to open this. We will likely not merge this in its current form. There are a few reasons for this:
1. We cannot depend on CDK alpha packages in our stable release line. The approach outlined in the [RFC](https://github.com/aws-amplify/amplify-backend/issues/1543) you commented on moves the alpha CDK dependency to the customer's project so they are explicitly aware they are using alpha functionality.
2. Changing the `runtime` parameter from a number union to a string union is a breaking API change which we cannot do without a major version bump of our libraries
3. E2E tests that exercise building and calling python and go functions need to be added
4. The dependency on docker to build the functions needs more consideration of the implications (eg for backend deployments on Amplify hosting and getting started locally)

MarlonJD:
Hi @edwardfoyle.
1-) We need golang runtimes, what about another options to add, these are alpha packages, there are no another option.
2-) What about adding new parameter without modifying `runtime`?
3-) e2e tests will be added
4-)  python docker build is necessary on python build, if it's not possible ı can remove python in pr

josefaidt:
Hey @MarlonJD :wave: thanks for taking the time to file this! To add to @edwardfoyle 's note we'd like to avoid taking a dependency on Docker and the alpha distributions of L2 constructs for Go and Python. While those two runtimes are top of mind, the escape hatch is desirable to offer the capability of using `defineFunction` with any runtime, with an opt-out of TypeScript-based features like typed environment variables and automatic secrets resolution. Would you be interested in pivoting this PR to match the proposal? Happy to discuss any findings/thoughts you have along the way in the RFC!

MarlonJD:
Hello @josefaidt. I'm trying to change my PR, I'm trying to create defineFunctionWithAnyLanguage in @aws-amplify/backend. Now I'm struggling about creating callback function for Construct/scope and use this function, any help in here?

josefaidt:
Hey @MarlonJD you can define a union on the `defineFunction` argument to either take the existing props type or a callback like `(scope: Construct) => lambda.Function`, but this scope will be passed at buildtime from the `defineFunction` factory. 

the props here will need adjustments 
https://github.com/aws-amplify/amplify-backend/blob/main/packages/backend-function/src/factory.ts#L41
https://github.com/aws-amplify/amplify-backend/blob/main/packages/backend-function/src/factory.ts#L101

and the scope would be passed to the `AmplifyFunction` construct
https://github.com/aws-amplify/amplify-backend/blob/main/packages/backend-function/src/factory.ts#L235

We wouldn't want to necessarily change the scope but change how "props" (in this case, the callback) is called with the existing `scope` to initialize within the Amplify context

MarlonJD:
Hey @josefaidt I changed defineFunction to:

```
/**
 * Entry point for defining a function in the Amplify ecosystem
 */
export const defineFunction = (
  props?: any,
): ConstructFactory<
  ResourceProvider<FunctionResources> & ResourceAccessAcceptorFactory
> => {
  if (props === undefined || props === null) {
    return new FunctionFactory({}, new Error().stack);
  } else if (typeof props === 'function') {
    console.log("props is a function");
    return new FunctionFactory({}, new Error().stack, props);
  } else {
    return new FunctionFactory(props, new Error().stack);
  }
};
```

Is that a good idea?

MarlonJD:
@josefaidt I changed PR to this:

```
 const lambda = defineFunction({
  entry: './test-assets/default-lambda/handler.ts',
  runtime: 16,
});
```

or 

```
const lambda = defineFunction((scope) => {
  return new NodejsFunction(scope, 'customFunction', {
    entry:
      './packages/backend-function/src/test-assets/default-lambda/handler.ts',
  });
});
```

What are you thinking about this changes? If it's okay, I'll try to write e2e tests, can you guide me again?

MarlonJD:
I hope you can review this changes @edwardfoyle

.changeset/flat-chefs-refuse.md
```diff
@@ -0,0 +1,2 @@
+---
```
edwardfoyle, line 1: the changeset should include packages that are updated and a description of what's changing. See https://github.com/changesets/changesets/blob/main/docs/intro-to-using-changesets.md

packages/backend-function/src/factory.ts
```diff
@@ -31,17 +31,25 @@ import { FunctionEnvironmentTypeGenerator } from './function_env_type_generator.
 import { AttributionMetadataStorage } from '@aws-amplify/backend-output-storage';
 import { fileURLToPath } from 'node:url';
 import { AmplifyUserError, TagName } from '@aws-amplify/platform-core';
+import * as lambda from 'aws-cdk-lib/aws-lambda';
```
edwardfoyle, line 34: nit: prefer `import { Function } from 'aws-cdk-lib/aws-lambda'`

also, we should probably be using the `IFunction` interface here rather than the `Function` base class

packages/backend-function/src/factory.ts
```diff
@@ -31,17 +31,25 @@ import { FunctionEnvironmentTypeGenerator } from './function_env_type_generator.
...
  * Entry point for defining a function in the Amplify ecosystem
  */
 export const defineFunction = (
-  props: FunctionProps = {}
+  props?: FunctionProps | ((scope: Construct) => lambda.Function)
 ): ConstructFactory<
   ResourceProvider<FunctionResources> & ResourceAccessAcceptorFactory
-> => new FunctionFactory(props, new Error().stack);
+> => {
+  if (props === undefined || props === null) {
```
edwardfoyle, line 46: we should continue to pass props as-is into the FunctionFactory and move this if statement inside that class

packages/backend-function/src/factory.ts
```diff
@@ -298,20 +311,24 @@ class AmplifyFunction
...
-        timeout: Duration.seconds(props.timeoutSeconds),
-        memorySize: props.memoryMB,
-        runtime: nodeVersionMap[props.runtime],
-        bundling: {
-          format: OutputFormat.ESM,
-          banner: bannerCode,
-          inject: shims,
-          loader: {
-            '.node': 'file',
+      if (callback != null) {
```
edwardfoyle, line 314: we should check for `typeof callback === 'function'`

MarlonJD:
@edwardfoyle I did changes, if it's okay about code, can you help me how and where should write for integration and/or e2e tests?

MarlonJD:
@edwardfoyle Hey there, I hope you can review new changes

edwardfoyle:
Hi @MarlonJD, sorry for the delay here. While this PR is on the right track in terms of implementation, we are concerned about the implications when deploying using the Amplify hosting service. Specifically, CDK requires Docker to build some runtimes and currently Docker is not installed in the hosting build container. Docker also cannot be installed using a custom build script because it requires permissions that the execution environment doesn't have. This means that customers using custom runtimes would not be able to use Hosting deployments. We want to consider this feature holistically rather than only offer partial support across our product offering. As a result, we need to do some internal prerequisite work to ensure custom runtime deployments can work in Amplify hosting deployments.

MarlonJD:
Hey @edwardfoyle. I didn't have docker requirements for golang functions for custom functions for this PR, I didn't add python for this PR, I only added golang functions for now, so I hope it can be forward for now.

MarlonJD:
@edwardfoyle Hey there again, I'm just checking python-alpha package for bundling, docker is using for building on windows and macOS to linux right? So if we don't use docker for linux (it's not necessary on host machine is already linux x64), it should work, I'll looking for this, what's your idea about this?

MarlonJD:
@josefaidt @edwardfoyle Hey there! I have good news about python functions. There is already a `local` paramater in `BundlingOptions`. It gives us to bundle without docker. I just found a way to bypass docker build without a modify packages, here is the usage:

```
import { defineFunction } from "@aws-amplify/backend";
import { DockerImage, Duration } from "aws-cdk-lib";
import { Code, Function, Runtime } from "aws-cdk-lib/aws-lambda";
import { execSync } from "child_process";
import * as path from "path";

const functionDir = path.join(__dirname);

export const sayHelloPythonFunctionHandler = defineFunction(
  (scope) =>
    new Function(scope, "say-hello-python-xderawfe", {
      handler: "index.handler",
      runtime: Runtime.PYTHON_3_9,
      timeout: Duration.seconds(20),
      code: Code.fromAsset(functionDir, {
        bundling: {
          image: DockerImage.fromRegistry("dummy"),
          local: {
            tryBundle(outputDir: string) {
              // if you want to use another python version, make default this ie: pyvenv global 3.12
              // if you dont have python3.9 installed, you can use pyenv to install it from Amplify Hosting build settings
              // from amplify.yaml, you can add this command to install python3.9 in the build settings
              // ie: pyenv install 3.12, pyenv global 3.12

              // Add pyenv command here, if you have more than one python version installed in your functions

              execSync(
                `python3 -m pip install -r ${path.join(functionDir, "requirements.txt")} -t ${path.join(outputDir)} --platform manylinux2014_x86_64 --only-binary=:all:`
              );
              execSync(`rsync -rLv ${functionDir}/* ${path.join(outputDir)}`);
              return true;
            },
          },
        },
      }),
    })
);
```

I just tried any it's works. Now Go and Python functions ready to use, I hope we can move on to the PR and tests.

MarlonJD:
Any update here? I hope you can review @edwardfoyle

MarlonJD:
@edwardfoyle Any updates?

MarlonJD:
PTAL

Srakai:
Great work @MarlonJD! Hope this gets merged quickly as currently im trying to achieve same but in a really whacky way

MarlonJD:
Hey @josefaidt, I commented how we can use python functions without docker build, but there is no local parameter in `PythonFunction` in `@aws-cdk/aws-lambda-python-alpha`, so I just showed example on Pure `Function` in `aws-cdk-lib/aws-lambda` to create python function with local parameter, it's enough to use python function.

What are you thinking about this solution to using python functions like this, should we add local parameter to `@aws-cdk/aws-lambda-python-alpha` or is it okay to use like this?

mrubelmann:
Any updates on this PR?  @edwardfoyle, has @MarlonJD addressed your request?  It would be super, super helpful to get this merged!

guikcd:
> @josefaidt @edwardfoyle Hey there! I have good news about python functions. There is already a `local` paramater in `BundlingOptions`. It gives us to bundle without docker. I just found a way to bypass docker build without a modify packages, here is the usage:
> 
> ```
> import { defineFunction } from "@aws-amplify/backend";
> import { DockerImage, Duration } from "aws-cdk-lib";
> import { Code, Function, Runtime } from "aws-cdk-lib/aws-lambda";
> import { execSync } from "child_process";
> import * as path from "path";
> 
> const functionDir = path.join(__dirname);
> 
> export const sayHelloPythonFunctionHandler = defineFunction(
>   (scope) =>
>     new Function(scope, "say-hello-python-xderawfe", {
>       handler: "index.handler",
>       runtime: Runtime.PYTHON_3_9,
>       timeout: Duration.seconds(20),
>       code: Code.fromAsset(functionDir, {
>         bundling: {
>           image: DockerImage.fromRegistry("dummy"),
>           local: {
>             tryBundle(outputDir: string) {
>               // if you want to use another python version, make default this ie: pyvenv global 3.12
>               // if you dont have python3.9 installed, you can use pyenv to install it from Amplify Hosting build settings
>               // from amplify.yaml, you can add this command to install python3.9 in the build settings
>               // ie: pyenv install 3.12, pyenv global 3.12
> 
>               // Add pyenv command here, if you have more than one python version installed in your functions
> 
>               execSync(
>                 `python3 -m pip install -r ${path.join(functionDir, "requirements.txt")} -t ${path.join(outputDir)} --platform manylinux2014_x86_64 --only-binary=:all:`
>               );
>               execSync(`rsync -rLv ${functionDir}/* ${path.join(outputDir)}`);
>               return true;
>             },
>           },
>         },
>       }),
>     })
> );
> ```
> 
> I just tried any it's works. I used pure `Function`, for this reason we can use any language supported from Lambda, f you can bundle it correctly without any package use. We can add samples to docs how people can use functions. Now Go and Python functions ready to use, I hope we can move on to the PR and tests.

I was able to build my python custom function with my requirements. The only I have now is that this seems to be never updated even if I alter this line. Seems that a missing force/cache somewhere. Only workaround I've found now: remove the `.amplify` directory to force the re-generation of the bundle.

MarlonJD:
@guikcd oh it's a good review, let me check how nodejs changes triggers, maybe we can found fix about it

josefaidt:
Hey folks in the thread :wave: we're working towards a few things to smooth over the DX for this feature's prerequisites -- namely Docker builds in Amplify Hosting and custom synthesizer support.

@MarlonJD to your latest comment, the file watcher you will need to expand the watcher class to allow reacting to changes from files with extensions that are not `ts`, `tsx`

MarlonJD:
Hey @josefaidt, I wonder that why we we merge right now? Go functions are just ready for deploy, and also we can use python without docker (https://github.com/aws-amplify/amplify-backend/pull/1602#issuecomment-2209424964), go and python libraries are alpha so we can make some changes for default building for Linux in CI (https://github.com/aws-amplify/amplify-backend/pull/1602#issuecomment-2292149296) What's your ideas about these

MarlonJD:
Hey @josefaidt, I wonder that why we we can't merge right now? Go functions are just ready for deploy, and also we can use python without docker (https://github.com/aws-amplify/amplify-backend/pull/1602#issuecomment-2209424964), go and python libraries are alpha so we can make some changes for default building for Linux in CI (https://github.com/aws-amplify/amplify-backend/pull/1602#issuecomment-2292149296) What's your ideas about these

arash2060:
This is great and I was actually able to deploy a python lambda function using this. However, I don't see how this approach can be integrated with amplify secrets. Am I missing something? 

Context: within FunctionProps, we are able to do something like this:

```
environment: {
        SOURCE_ADDRESS: 'sourceAddress',
        OPEN_AI_KEY: secret('OPEN_AI_KEY'),
    },
    
```

MarlonJD:
> This is great and I was actually able to deploy a python lambda function using this. However, I don't see how this approach can be integrated with amplify secrets. Am I missing something?
> 
> Context: within FunctionProps, we are able to do something like this:
> 
> ```
> environment: {
>         SOURCE_ADDRESS: 'sourceAddress',
>         OPEN_AI_KEY: secret('OPEN_AI_KEY'),
>     },
>     
> ```

I'm not sure but as I understand from the code, secrets only usable from CI from amplify itself without manuel deploy,
Did you push with manual deploy?
I used .env files to use secret values as a workaround

packages/backend-function/src/factory.ts
arash2060, line 1: Following up on my overall comment, I'm not sure how the environment variable, I'm not sure how we're going to handle environment variables of type `Record<string, string | BackendSecret>;`. I see that you can add an environment variable via `addEnvironment(key: string, value: string, options?: EnvironmentOptions): this;` method of `PythonFunction`. However, the `value` cannot be of type `BackendSecret`

licarpen: @MarlonJD @arash2060 I am so excited to see this functionality so close!  I am new to Amplify and have been enjoying the ease of development...but got hung up royally today trying to figure out how to add a custom query to my graphql schema via a lambda with python 3_11 runtime.  This functionality will be well-used and loved as I build out slew of microservices that use LangChain.  I am wondering if you would consider deploying without supporting environment variables for now as a way to get this into our hands.  From my understanding, that would require adding a permission for the lambda to access the secret store with boto3 and a warning about the additional cost and latency while the full functionality is being built out.  Is this feasible?  Curious to hear your thoughts.

MarlonJD: > @MarlonJD @arash2060 I am so excited to see this functionality so close!  I am new to Amplify and have been enjoying the ease of development...but got hung up royally today trying to figure out how to add a custom query to my graphql schema via a lambda with python 3_11 runtime.  This functionality will be well-used and loved as I build out slew of microservices that use LangChain.  I am wondering if you would consider deploying without supporting environment variables for now as a way to get this into our hands.  From my understanding, that would require adding a permission for the lambda to access the secret store with boto3 and a warning about the additional cost and latency while the full functionality is being built out.  Is this feasible?  Curious to hear your thoughts.

Hey there! If you want to auto deploy with amplify we should use AWS Secret Manager, it's because `secret('foo')` using this, you can easily use .env file and manuel deploy without this I can give an example for this, if you have another solution or idea please tell me mayne we can build, I'm thinking publish my amplify fork and use this on amplify auto deploy because of limit issue and this python and golang functions, I'll share with you this examples here If you want to use custom functions on auto build in amplify console.

MarlonJD:
I updated my code

sobolk:
@MarlonJD We're planning to evaluate this change further, i.e. test proposed changes at runtime. We'll get back to you with further guidance.

tbolsh:
Hello!
When will we be able to use GoFunction with Amplify?
Thank you

MarlonJD:
I need to fix conflicts, but many things changed, now there is schedule, layers, bundling, resourceGroupName, logging more, I need to create this pr from the beginning again.

I'll create new commits with force push to my branch, maybe it's good to wait new release if you already doing some works in these files? What are you suggesting? @sobolk

sobolk:
@MarlonJD Please allow a bit more time for us to evaluate and provide guidance. Meanwhile, rebasing this PR with latest `main` is a good idea.

MarlonJD:
Hey @sobolk. Thanks for detailed review, I'm trying to understand changes and I'll try to prepare your requested changes. I'll let you know when I'm ready to share my changes.

MarlonJD:
Hey @sobolk 👋. I just made changes, add unit test and integration test, PTAL.

sobolk (commented):
The changes look good. Please take a look at pull request checks and make them passing.

I ran e2e tests here https://github.com/aws-amplify/amplify-backend/actions/runs/12435184989 . They are passing.

packages/integration-tests/src/test-projects/advanced-auth-and-functions/amplify/function.ts
```diff
@@ -24,6 +26,14 @@ export const funcNoMinify = defineFunction({
   },
 });
 
+export const funcProvided = defineFunction((scope) => {
+  return new NodejsFunction(scope, 'funcProvided', {
+    entry:
+      './packages/integration-tests/src/test-projects/advanced-auth-and-functions/amplify/func-src/handler_provided.ts',
```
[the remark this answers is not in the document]
MarlonJD: Actually I tested these on unit tests, this path should be like this, but I don't know about in integration test, if it's run without an error, that's cool

packages/backend-function/src/provided_function_factory.ts
```diff
@@ -0,0 +1,143 @@
...
+        (e.message.includes('docker exited with status 1') ||
+          e.message.includes('docker ENOENT'))
+      ) {
+        throw new AmplifyUserError(
+          'FunctionBundlingDockerError',
+          {
+            message: e.message,
+            // TODO better resolution
+            resolution:
+              'Ensure that docker is present and works correctly. See https://some.link.about.docker.',
```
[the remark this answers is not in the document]
MarlonJD: We can use go functions without and error, an also we can use python functions without docker. I tried and gave example before how we can use python without docker Any change to pass at least for golang?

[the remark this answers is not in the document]
MarlonJD: What about we make bundling without docker on Linux by default in python alpha package? On ci/cd it will use Amazon Linux, so we don't need docker in Linux for Python, it can be fast solution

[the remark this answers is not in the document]
sobolk: > What about we make bundling without docker on Linux by default in python alpha package? On ci/cd it will use Amazon Linux, so we don't need docker in Linux for Python, it can be fast solution

These kind of improvements can be added later outside of this PR. What we need to merge this PR is docs that clearly state the scope of support. Please allow us a bit more time to come up with that.

packages/integration-tests/src/test-projects/advanced-auth-and-functions/amplify/function.ts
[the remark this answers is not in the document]
sobolk: Ok. I can see what's missing.

We need to add `funcProvided` here https://github.com/aws-amplify/amplify-backend/blob/b574bce26ec2d55a4391aeceecc1d4957e79fba3/packages/integration-tests/src/test-projects/advanced-auth-and-functions/amplify/test_factories.ts#L10 . so that it actually becomes member of test project.

packages/backend-function/src/provided_function_factory.ts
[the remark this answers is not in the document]
MarlonJD: I can also create docs for creating new go and python runtimes with this new support, is this works or should I wait?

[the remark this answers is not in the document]
sobolk: > I can also create docs for creating new go and python runtimes with this new support, is this works or should I wait?

You could give it a stab. We should aim to have a sub page under https://docs.amplify.aws/react/build-a-backend/functions/ . Our docs are defined in https://github.com/aws-amplify/docs repo.

Couple of things to keep in mind:
1. The page should be called "Custom functions"
2. The page must call out lack of docker support.
3. We can use docker-less python build as an example.
4. I recommend to avoid GO as an example (it doesn't seem to have first class support, see https://aws.amazon.com/blogs/compute/migrating-aws-lambda-functions-from-the-go1-x-runtime-to-the-custom-runtime-on-amazon-linux-2/ )
5. We might need to wait for reviewers quorum until after Jan 6th due to holidays.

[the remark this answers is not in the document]
MarlonJD: Okay I'll try it, Golang is not anymore directly runtime but it still can be using directly just by giving runtime as provided.al2023, I personally using this without an error, I would be happy if I can add Golang there with Python.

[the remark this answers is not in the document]
MarlonJD: I created PR for documents [here](https://github.com/aws-amplify/docs/pull/8176). Happy holidays.

.changeset/long-berries-greet.md
```diff
@@ -0,0 +1,5 @@
+---
+'@aws-amplify/backend-function': minor
+---
```
sobolk, line 3:
```suggestion
---
'@aws-amplify/backend-function': minor
'@aws-amplify/backend': minor
---
```

This is to make changeset check passing.

packages/integration-tests/src/test-projects/advanced-auth-and-functions/amplify/function.ts
```diff
@@ -24,6 +26,13 @@ export const funcNoMinify = defineFunction({
   },
 });
 
+export const funcProvided = defineFunction((scope) => {
+  return new NodejsFunction(scope, 'funcProvided', {
+    entry: './func-src/handler_provided.ts',
+    runtime: Runtime.NODEJS_18_X,
+  });
+});
```
sobolk, line 34:
```suggestion
export const funcProvided = defineFunction((scope) => {
  return new NodejsFunction(scope, 'funcProvided', {
    entry: path.resolve(
      fileURLToPath(import.meta.url),
      '..',
      'func-src',
      'handler_provider.ts'
    ),
    runtime: Runtime.NODEJS_18_X,
  });
});
```


I've run e2e tests here https://github.com/aws-amplify/amplify-backend/actions/runs/12552245657 .

It seems that we need to resolve the entry path differently for provided function. The reason is that provided function entry resolution doesn't use (can't use) our implicit logic from [here](https://github.com/aws-amplify/amplify-backend/blob/22089b7a7321b8fc99e178a8dc777ccd77690e12/packages/backend-function/src/factory.ts#L273). Which is fine but we need to adjust test to accommodate it.

The reason this doesn't surface in "in memory tests" and only in e2e tests runs is that we don't include this project in https://github.com/aws-amplify/amplify-backend/tree/main/packages/integration-tests/src/test-in-memory (it's a gap in testing, so don't worry about this in this PR, I've manually ran proposed change to make sure it'll make e2e tests passing).

packages/backend-function/src/provided_function_factory.ts
```diff
@@ -0,0 +1,143 @@
...
+      } else {
+        throw new AmplifyUserError(
+          'FunctionProviderError',
+          {
+            message: e instanceof Error ? e.message : JSON.stringify(e),
+            // TODO better resolution
+            resolution: 'Ensure that function provider is working.',
+          },
+          e instanceof Error ? e : undefined
+        );
```
sobolk, line 103:
```suggestion
        throw new AmplifyUserError(
          'CustomFunctionProviderError',
          {
            message: e instanceof Error ? e.message : JSON.stringify(e),
            resolution: "Ensure that callback passed to 'defineFunction' executes without error. See https://docs.amplify.aws/react/build-a-backend/functions/custom-functions for more details.",
          },
          e instanceof Error ? e : undefined
        );
```

@josefaidt @Amplifiyer please take a look at proposed wording here.

packages/backend-function/src/provided_function_factory.ts
sobolk, line 90:
```suggestion
        throw new AmplifyUserError(
          'CustomFunctionProviderDockerRequiredError',
          {
            message: e.message,
            resolution:
              'See https://docs.amplify.aws/react/build-a-backend/functions/custom-functions for more details about current limitations and troubleshooting steps.',
```

@josefaidt @Amplifiyer please take a look at proposed wording here. The goal here is to:
- refrain from stating docker support in error message/resolution in case we find a way to support it (avoid hardcoding).
- point to web page that we can edit out of bound with most recent information about the state of support. (providing support should not involve code change in this repo, therefore need for decoupling).

sobolk (commented):
@MarlonJD 
I looked at latest diff and it looks good overall.

Couple of areas I left comments:
1. test project - I ran e2e tests here https://github.com/aws-amplify/amplify-backend/actions/runs/12552245657 - they failed, I left comments with changes to make them passing.
2. The error messages. For these please keep in mind that they need that docs PR to stabilize, so this might not be final iteration.

packages/backend-function/src/provided_function_factory.ts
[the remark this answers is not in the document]
sobolk: @josefaidt please take a look at docs PR ^ when you're back.

MarlonJD:
Hi @sobolk, I’ve implemented the changes based on your suggestions. Please let me know if there’s anything else that needs adjustment.

sobolk:
> Hi @sobolk, I’ve implemented the changes based on your suggestions. Please let me know if there’s anything else that needs adjustment.

Thank you. The change in test project needs imports, see https://github.com/aws-amplify/amplify-backend/actions/runs/12556641214/job/35018558350?pr=1602

MarlonJD:
Hi @sobolk, I've added the missing imports to address the issue.

sobolk:
> Hi @sobolk, I've added the missing imports to address the issue.

Thanks. The e2e look good here https://github.com/aws-amplify/amplify-backend/actions/runs/12584741475 .

Could you please address `lint` check? `npm run lint:fix` should do it.

MarlonJD:
Hi @sobolk,

I’ve resolved the linting errors and committed the changes. I ran `npm run lint`, and it now reports: `All matched files use Prettier code style!`

packages/backend-function/src/provided_function_factory.ts
josefaidt: We have that callout at the top of this page that calls out limitations with docker, but perhaps this warrants more clarity at the top of that page -- prereqs, etc.

There's a note I left on the docs PR to (separately) create a troubleshooting dedicated to docker errors which we can use to add more color for support

packages/backend-function/src/provided_function_factory.ts
josefaidt: what are a few examples of errors that would fall into this category? The wording seems fine but I'm concerned with whether it would be tough to debug

sobolk: Anything that function passed to `defineFunction` can throw.

Including:
1. Bundling errors,
2. Invalid properties/validation errors
3. Mistakes in function itself.

The underlying error will be printed in the console.
1. We copy it's message
2. We include actual root cause as `cause`, it is reported in std streams (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause ).

The idea here is that the troubleshooting section can be expanded on the docs page without changing this code. I.e. information about root cause and the docs page should be enough to investigate.

sobolk: Btw. technically docker errors could be handled this way as well, but for now in PR they get their own category as we expect it to be common problem.

Amplifiyer: * When we are wrapping the error, we should provide a different error message than the one being wrapped, specially as it is being passed in the cause. This error message should describe what happened at the call site e.g. "Failed to instantiate the custom function provider", or something like that.
* `Ensure that callback passed to 'defineFunction' executes without error`, are we conflating the concept of `callback` here?

sobolk:
@MarlonJD could you please merge latest `main` to your branch? it should fix remaining checks.

sobolk:
And please `npm run lint:fix` one time.

sobolk:
> Hi @sobolk, I'm getting this error after `npm run lint:fix`:
> 
> ```
> > eslint --max-warnings 0 . && tsx scripts/check_package_json.ts && prettier --check . packages/backend-function
> 
> 
> /Users/marlonjd/Developer/mobile/amplify/amplify-backend/packages/platform-core/src/cdk/enum_converters.ts
>   97:66  error  Invalid type "never" of template literal expression  @typescript-eslint/restrict-template-expressions
> 
> ✖ 1 problem (1 error, 0 warnings)
> ```

It's likely related to workspace being in inconsistent state after `main` merge.

You can try the following:
```
npm run clean && npm install && npm run build && npm run lint:fix
```

Or just run `npx prettier --write .` in the repository root.

sobolk (approved):

MarlonJD:
Hi @sobolk,

Thank you for pointing this out. I followed your recommendations, which resolved the inconsistency, and I’ve committed the necessary changes.

I also want to express my gratitude for your guidance and support—it’s much appreciated!

Amplifiyer (approved):

sobolk:
> Hi @sobolk,
> 
> Thank you for pointing this out. I followed your recommendations, which resolved the inconsistency, and I’ve committed the necessary changes.
> 
> I also want to express my gratitude for your guidance and support—it’s much appreciated!

Thank you for contributing this @MarlonJD .

sofferor:
Happy to hear it's finally possible to define a python function!
I also saw you add documentation about it in:
https://docs.amplify.aws/nextjs/build-a-backend/functions/custom-functions/#python

Can you also please add documentation about how a **Python** function can interact with the Amplify Data?
For example how to create/update/read/delete a table in Amplify Data.

Thanks!

MarlonJD:
Hey @sofferor, there are various ways to make CRUD operations such as GraphQL request to AppSync or directly manipulate DynamoDB. I prefer DynamoDB way, but AppSync is more robust. You can found examples for "boto3 DynamoDB", here is the some of examples:
- (article 1)[https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-dynamo-db.html]
- (article 2)[https://dev.classmethod.jp/articles/how-to-put-get-items-from-dynamodb-table-using-aws-lambda-boto3/]

You'll also need to add DynamoDB permissions to your lambda, you can do this directly in your backend.ts:

```
const dynamoDBPermission = new iam.PolicyStatement({
  sid: "AllowDynamoDBIdp4652",
  effect: iam.Effect.ALLOW,
  actions: ["dynamodb:*"],
  resources: ["*"],
});
;

backend.yourFunctionHandler.resources.lambda.addToRolePolicy(
  dynamoDBPermission
);
```

But if you still need assistance, I can provide you to full example later.

MarlonJD:
Hey @sofferor, there are various ways to make CRUD operations such as GraphQL request to AppSync or directly manipulate DynamoDB. I prefer DynamoDB way, but AppSync is more robust. You can found examples for "boto3 DynamoDB", here is the some of examples:
- [article 1](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-dynamo-db.html)
- [article 2](https://dev.classmethod.jp/articles/how-to-put-get-items-from-dynamodb-table-using-aws-lambda-boto3/)

You'll also need to add DynamoDB permissions to your lambda, you can do this directly in your backend.ts:

```
const dynamoDBPermission = new iam.PolicyStatement({
  sid: "AllowDynamoDBIdp4652",
  effect: iam.Effect.ALLOW,
  actions: ["dynamodb:*"],
  resources: ["*"],
});
;

backend.yourFunctionHandler.resources.lambda.addToRolePolicy(
  dynamoDBPermission
);
```

But if you still need assistance, I can provide you to full example later.

sofferor:
Hi @MarlonJD ,
First, thanks a lot for your contribution! much appreciated!

Regarding, how Python function can interact with the Amplify Data.
In a typescript function, according to this docs:
https://docs.amplify.aws/nextjs/build-a-backend/functions/examples/create-user-profile-record/
It's possible to interact with Amplify Data using the "client" object:
 
image
**Is it possible to do something similar in a python function? (if yes, can you please 🙏 provide example)**



If not, in a typescript function it's also possible to interact with the Amplify Data with GraphQL (as you mentioned) like this:
image
**Can you please 🙏 provide an example of how to do this in a python Amplify function?**

MarlonJD:
Hi @sofferor,

First of all, thank you for bringing this to my attention—it’s a very intriguing idea! I wasn’t aware that such an interaction was possible with Amplify Data in Gen2 using the client object. Adding support for other languages like Python in a similar fashion would indeed be a fantastic enhancement.

If you could allow me some time, I’d like to research this further and run a few tests. One potential approach that comes to mind is leveraging the pyexecjs library to execute the generateClient function temporarily within Python. This could provide a quick workaround for achieving this functionality until a more robust solution is in place.

Regarding your question about an example for a Python Amplify function: I personally transitioned many of my Python functions to Go due to its remarkable performance gains—up to 10x faster in some cases. However, I’ve previously worked on a similar use case with GraphQL in Python. I'll revisit my Git history to find that specific example, test it thoroughly to ensure it’s functional, and then provide you with a complete example tailored for this scenario.

Thank you for your patience, and I’ll get back to you shortly with a detailed solution!

sofferor:
@MarlonJD Thanks a lot, your effort is much appreciated🙏

sobolk:
Related docs update related to Docker bundling https://github.com/aws-amplify/docs/pull/8206 .

MarlonJD:
Thank you for the update @sobolk! The direct support for Docker is an excellent addition and a well-thought-out solution. I truly appreciate the effort and the detailed documentation update—it’s a great improvement.

Issue #1533, opened 2024-05-16
Explicit public surface definition


sobolk:
We should define and document our public contract. So that customers know how our public surface is evolved and for us to guide that evolution.

We should define and publish:
1. What API changes are allowed in what package.
2. How do we mark APIs as public, "public but for internal purpose", experimental etc.
3. Which stateful parts, like deployed resources are considered contract and which not. For example, CFN outputs.

In addition, we should implement automated checks to enforce that publicly facing elements are properly documented and annotated.

redjonzaci:
can i help?

PR #1613, opened 2024-06-04
chore: expose iam actions map without affecting public API

awsluja:
## Problem
We need to maintain a mapping of actions, meta actions, and their iam actions in the docs. If the mapping is exported we can use it in the docs site to render a table.

The goal of this change is to do so without having to modify the public API.

It can be imported by adding this package, @aws-amplify/backend-auth, to your package.json & importing it like:

`import { iamActionMap } from '@aws-amplify/backend-auth/internal/iam_actions';
`

## Changes

## Checklist


Amplifiyer (approved):

packages/backend-auth/package.json
```diff
@@ -10,6 +10,11 @@
       "types": "./lib/index.d.ts",
       "import": "./lib/index.js",
       "require": "./lib/index.js"
+    },
+    "./internal/iam_actions": {
+      "types": "./lib/internal/iam_actions.d.ts",
+      "import": "./lib/internal/iam_actions.js",
+      "require": "./lib/internal/iam_actions.js"
```
sobolk, line 17: will this show up in IDE syntax suggestions?

awsluja: You mean like this? or from the amplify DX? 
image

sobolk: > or from the amplify DX?

From amplify DX. Are customers going to see possibility to import these from `@aws-amplify/backend-auth/internal` ?

awsluja: Ok, so if the question is about doing this:
image

There is no intellisense and you must manually specify the /internal/iam_actions path to be able to import it

awsluja: https://github.com/aws-amplify/amplify-backend/assets/110861985/9ca997df-d782-4170-8ba7-4b96fe95045e

sobolk: I get this 
image


When trying to use similar export from here https://github.com/aws-amplify/amplify-backend/blob/4dd9d5a35c378895d1360c15a3b7ad1f09cc7653/packages/client-config/package.json#L14-L18

edwardfoyle: I'm okay with this as-is, but if we want an even stronger suggestion that customers shouldn't use this, we could namespace it `internal-only-do-not-use/iam-actions` or something to that effect

awsluja: If you want to try the same DX; try importing from  '@aws-amplify/backend/types/platform'

import * as types from '@aws-amplify/backend/types/platform';

sobolk: To be clear. The goal is to make it as hidden as possible.

We may say that `<root>/internal/*` is standard way of saying "public but don't use it". However, we haven't completed https://github.com/aws-amplify/amplify-backend/issues/1533 . So either we should define that first or treat this PR as precedent.

awsluja:
Closing this after further review, we have other ways of doing this that don't require custom exports.

PR #1636, opened 2024-06-11
remove username from username_attributes in outputs schema

josefaidt:
## Problem

in Cognito, username attributes are attributes that can be used in place of a username
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cognito.CfnUserPool.html#usernameattributes

## Changes

removes `username` from list of valid `username_attributes` in the outputs schema

## Checklist

### DIFF
--- a/.changeset/small-dodos-wash.md
+++ b/.changeset/small-dodos-wash.md
@@ -1,0 +1,5 @@
+---
+'@aws-amplify/client-config': patch
+---
+
+removes username from username_attributes
--- a/packages/client-config/src/client-config-schema/client_config_v1.ts
+++ b/packages/client-config/src/client-config-schema/client_config_v1.ts
@@ -142,7 +142,7 @@ export interface AWSAmplifyBackendOutputs {
      *
      * @minItems 1
      */
-    username_attributes?: ('email' | 'phone_number' | 'username')[];
+    username_attributes?: ('email' | 'phone_number')[];
     user_verification_types?: ('email' | 'phone_number')[];
     unauthenticated_identities_enabled?: boolean;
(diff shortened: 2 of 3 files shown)


edwardfoyle (commented):
need to update api extract (run `npm run update:api`). Also, isn't this a breaking change?

josefaidt:
@edwardfoyle that command just loops for me :/ I have to kill my terminal to stop it
CleanShot 2024-06-12 at 14 30 20@2x

I put this as a patch because I don't think it's being used, but I may have overlooked something (cc @Amplifiyer )

sobolk:
> need to update api extract (run `npm run update:api`). Also, isn't this a breaking change?

This is a breaking change. `username` should be deprecated, not removed. And recorded for removal next time we ship MV.

This is ugly, but leaving it as is won't hurt since we control the producer.

edwardfoyle:
@josefaidt
> that command just loops for me :/ I have to kill my terminal to stop it

I think you need to build locally first `npm run build && npm run update:api`

josefaidt:
I'll move forward with the change to a major bump in the changeset. Though this value does not get generated:

https://github.com/aws-amplify/amplify-backend/blob/main/packages/auth-construct/src/construct.ts#L952

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_cognito.CfnUserPool.html#usernameattributes

It is expected to be consumed
https://github.com/aws-amplify/amplify-js/blob/44696fb49130f3de213a2c53291f8afe3b53bb27/packages/core/src/parseAmplifyOutputs.ts#L144

Issue #1663, opened 2024-06-17
RFC: Deploying an EdgeFunction bound to a CloudFront distribution with a Storage bucket as origin (Pass props for root Stack)


renevatium:
### Environment information

```plain text
System:
  OS: Linux 5.15 Ubuntu 20.04.5 LTS (Focal Fossa)
  CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
  Memory: 12.83 GB / 15.51 GB
  Shell: /bin/bash
Binaries:
  Node: 18.20.3 - ~/.nvm/versions/node/v18.20.3/bin/node
  Yarn: undefined - undefined
  npm: 10.7.0 - ~/.nvm/versions/node/v18.20.3/bin/npm
  pnpm: undefined - undefined
NPM Packages:
  @aws-amplify/backend: 1.0.0
  @aws-amplify/backend-cli: 1.0.1
  aws-amplify: 6.3.5
  aws-cdk: 2.140.0
  aws-cdk-lib: 2.140.0
  typescript: 5.4.5
AWS environment variables:
  AWS_STS_REGIONAL_ENDPOINTS = regional
  AWS_NODEJS_CONNECTION_REUSE_ENABLED = 1
  AWS_SDK_LOAD_CONFIG = 1
No CDK environment variables
```

### Description

As the title says, I am trying to create and bind an EdgeFunction to a CloudFront distribution with a Storage S3 bucket (from defineStorage) as origin, through the Amplify backend (from defineBackend). 

I tried to use the root node (App) from one of the generated backend resources as the root Construct on a secondary Stack. This is necessary because the us-east-1 region must be explicitly set on the Stack for CDK to accept the EdgeFunction:

```
Caused By: Error: stacks which use EdgeFunctions must have an explicitly set region
    at EdgeFunction.createCrossRegionFunction (/home/path-to-project/node_modules/aws-cdk-lib/aws-cloudfront/lib/experimental/edge-function.js:1:7079)
```

Note: The above error would go away with something like [aws-cdk/issues/15620](https://github.com/aws/aws-cdk/issues/15620), but I don't think that would actually solve the underlying problem. Unless the EdgeFunction deployment would work without the explicit us-east-1 region set, and this is just a semantic check. 

Anyway, the EdgeFunction deploys correctly but I can't grant IAM Resource permissions to the function from graphqlApi:

```
[...]

PR #1675, opened 2024-06-24
pass valid props to root stack

renevatium:
## Problem

Currently, it is not possible to pass instantiation props to the Amplify root CDK Stack, and, by convention, any Nested Stack created through `backend.createStack`. This prevents certain resource deployments: An EdgeFunction for instance requires the region to be explicitly set. Furthermore, while Nested Stacks inherit their region from their parent stack, Amplify has a single root node convention, so multiple root stacks are not an option. Evaluation in CDK appears to be strict so the only way to do deploy an EdgeFunction from an Amplify project looks to be at instantiation of the root Stack.

Note: I used 'main' instead of 'root' because that is the naming convention used in the [current code](https://github.com/aws-amplify/amplify-backend/blob/1f4c6e0c298e8ee087000d0efd3eb25f18eec433/packages/backend/src/default_stack_factory.ts#L9). 
 
See **#1663** for more information on trying to deploy an EdgeFunction with current version. 

## Changes

This implementation specifically avoids breaking changes, but does require argument mutation. The relevant args are not typed as `const` so this may be fine, but if it goes against an established convention, let me know as I can update it easily enough. This was needed because the relevant calls are being made as default argument values in both [BackendFactory constructor](https://github.com/aws-amplify/amplify-backend/blob/1f4c6e0c298e8ee087000d0efd3eb25f18eec433/packages/backend/src/backend_factory.ts#L58) and [createDefaultStack](https://github.com/aws-amplify/amplify-backend/blob/1f4c6e0c298e8ee087000d0efd3eb25f18eec433/packages/backend/src/default_stack_factory.ts#L8C14-L8C33) respectively.

This PR introduces a `mainStackProps` argument to the `defineBackend` function. This gets passed down the callstack and used to define the root Stack in the `AmplifyStack` class.

### Props for root CDK Stack

[AWS documentation for aws-cdk-lib.StackProps](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.StackProps.html)

- Props are picked to ensure explicit addition of new StackProps is required.
  - description?: string
  - env?: Environment
  - tags?: { [key: string]: string }
  - analyticsReporting?: boolean
  - crossRegionReferences?: boolean
  - suppressTemplateIndentation?: boolean

- StackProps incompatible with Amplify's intended hierarchy, build or deployment processes are omitted:
  - stackName: Conflicts with dynamic resource naming.
  - synthesizer: Conflicts with managed deployments and resource references.
  - terminationProtection: Conflicts with sandbox/app delete.
  - permissionsBoundary: Conflicts with single root Stack ethos (i.e. Unable to create Role prior to `defineBackend`).

Props are passed down from `defineBackend`. e.g:

```
defineBackend({
  auth,
  storage,
}, {
  description: 'test-description',
  env: {
    region: 'us-east-1',
    account: '[YOUR_ACCOUNT_ID]'
  },
  tags: {
    'test-tag': 'test-tag-value',
  },
  analyticsReporting: true,
  crossRegionReferences: true,
  suppressTemplateIndentation: true,
})
```

## Validation

All existing unit, e2e and deployment tests pass. Coverage is unchanged. 

Using the npm proxy, I have fully deployed and tested the EdgeFunction case with various configurations.   

This change simply passes props tested in [aws-cdk-lib/core/test/stack.test.ts](https://github.com/aws/aws-cdk/blob/3b95777ee5dae32fd41481d81922d07c804a2c6b/packages/aws-cdk-lib/core/test/stack.test.ts). i.e It already follows defined convention. As such, I am not sure what automated testing would actually be useful. I did try though, so please let me know if you would like me to some tests back in?

I started out writing unit tests, but struggled to find a way to validate anything useful there. I could potentially check that `${AWS::Region}` does not exist in the generated Stack Template but that seems like a stretch. 

I wrote e2e and deployment tests, but they don't really do much to actually test anything in scope and they add a lot of unnecessary overhead. The deployment tests for instance, requires that I update the test stack generation process and/or write the tests without following the conventions established by the existing test structure. Again, if you would prefer these be in, I can put them back. 

## Checklist

A point on the below: The PROJECT_ARCHITECTURE.md readme doesn't exist. 

- [] If this PR includes a functional change to the runtime behavior of the code, I have added or updated automated test coverage for this change.
- [] If this PR requires a change to the [Project Architecture README](../PROJECT_ARCHITECTURE.md), I have included that update in this PR. 
- [] If this PR requires a docs update, I have linked to that docs PR above.
- [] If this PR modifies E2E tests, makes changes to resource provisioning, or makes SDK calls, I have run the PR checks with the `run-e2e` label set.

### DIFF
--- a/packages/backend/API.md
+++ b/packages/backend/API.md
@@ -31,6 +31,7 @@ import { ResourceProvider } from '@aws-amplify/plugin-types';
 import { SsmEnvironmentEntriesGenerator } from '@aws-amplify/plugin-types';
 import { SsmEnvironmentEntry } from '@aws-amplify/plugin-types';
 import { Stack } from 'aws-cdk-lib';
+import { StackProps } from 'aws-cdk-lib';
 
 export { a }
 
@@ -70,7 +71,7 @@ export { ConstructFactoryGetInstanceProps }
 export { defineAuth }
 
 // @public
-export const defineBackend: <T extends DefineBackendProps>(constructFactories: T) => Backend<T>;
+export const defineBackend: <T extends DefineBackendProps>(constructFactories: T, mainStackProps?: MainStackProps) => Backend<T>;
 
 // @public (undocumented)
 export type DefineBackendProps = Record<string, ConstructFactory<ResourceProvider & Partial<ResourceAccessAcceptorFactory<never>>>> & {
@@ -88,6 +89,9 @@ export { FunctionResources }
 export { GenerateContainerEntryProps }
 
 export { ImportPathVerifier }
+
+// @public
+export type MainStackProps = Pick<StackProps, 'description' | 'env' | 'tags' | 'analyticsReporting' | 'crossRegionReferences' | 'suppressTemplateIndentation'>;
 
 export { ResourceProvider }
 
--- a/packages/backend/src/backend_factory.ts
+++ b/packages/backend/src/backend_factory.ts
@@ -27,6 +27,7 @@ import {
 import { CustomOutputsAccumulator } from './engine/custom_outputs_accumulator.js';
 import { ObjectAccumulator } from '@aws-amplify/platform-core';
 import { DefaultResourceNameValidator } from './engine/validations/default_resource_name_validator.js';
+import { MainStackProps } from './engine/amplify_stack.js';
 
 // Be very careful editing this value. It is the value used in the BI metrics to attribute stacks as Amplify root stacks
 const rootStackTypeIdentifier = 'root';
@@ -55,7 +56,15 @@ export class BackendFactory<
    * Initialize an Amplify backend with the given construct factories and in the given CDK App.
    * If no CDK App is specified a new one is created
    */
-  constructor(constructFactories: T, stack: Stack = createDefaultStack()) {
+  constructor(
+    constructFactories: T,
+    stack?: Stack,
+    mainStackProps?: MainStackProps
+  ) {
+    if (stack === undefined) {
+      stack = createDefaultStack(undefined, mainStackProps);
+    }
+
     new AttributionMetadataStorage().storeAttributionMetadata(
       stack,
       rootStackTypeIdentifier,
@@ -150,9 +159,14 @@ export class BackendFactory<
  * @param constructFactories - list of backend factories such as those created by `defineAuth` or `defineData`
  */
 export const defineBackend = <T extends DefineBackendProps>(
-  constructFactories: T
+  constructFactories: T,
+  mainStackProps?: MainStackProps
 ): Backend<T> => {
-  const backend = new BackendFactory(constructFactories);
+  const backend = new BackendFactory(
+    constructFactories,
+    undefined,
+    mainStackProps
+  );
   return {
     ...backend.resources,
     createStack: backend.createStack,
--- a/.changeset/chatty-gifts-bathe.md
+++ b/.changeset/chatty-gifts-bathe.md
@@ -1,0 +1,5 @@
+---
+'@aws-amplify/backend': minor
+---
+
+Allow StackProps to be passed from defineBackend down to AmplifyStack for root Stack configuration
--- a/packages/backend/src/default_stack_factory.ts
+++ b/packages/backend/src/default_stack_factory.ts
@@ -1,14 +1,21 @@
 import { App, Stack } from 'aws-cdk-lib';
 import { ProjectEnvironmentMainStackCreator } from './project_environment_main_stack_creator.js';
 import { getBackendIdentifier } from './backend_identifier.js';
+import { MainStackProps } from './engine/amplify_stack.js';
 
 /**
  * Creates a default CDK scope for the Amplify backend to use if no scope is provided to the constructor
  */
-export const createDefaultStack = (app = new App()): Stack => {
+export const createDefaultStack = (
+  app?: App,
+  props?: MainStackProps
+): Stack => {
+  if (app === undefined) {
+    app = new App();
+  }
   const mainStackCreator = new ProjectEnvironmentMainStackCreator(
     app,
     getBackendIdentifier(app)
   );
-  return mainStackCreator.getOrCreateMainStack();
+  return mainStackCreator.getOrCreateMainStack(props);
 };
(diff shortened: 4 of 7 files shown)


packages/backend/API.md
```diff
@@ -70,7 +71,7 @@ export { ConstructFactoryGetInstanceProps }
 export { defineAuth }
 
 // @public
-export const defineBackend: <T extends DefineBackendProps>(constructFactories: T) => Backend<T>;
+export const defineBackend: <T extends DefineBackendProps>(constructFactories: T, mainStackProps?: MainStackProps) => Backend<T>;
```
sobolk, line 74: We should prepare this API for further expansion. I.e. main stack props might not be the only extension we do in the future.

Therefore, we should have `defineBackend(constructFactories T, props?: DefineBackendProps)`
and then `DefineBackendProps` should have `mainStackProps: MainStackProps` as optional property.

packages/backend/API.md
```diff
@@ -89,6 +90,9 @@ export { GenerateContainerEntryProps }
 
 export { ImportPathVerifier }
 
+// @public
+export type MainStackProps = Pick<StackProps, 'description' | 'env' | 'tags' | 'analyticsReporting' | 'crossRegionReferences' | 'suppressTemplateIndentation'>;
```
sobolk, line 94: Is this minimal set of picked properties to solve the problem? If not, please narrow it.

packages/backend/src/backend_factory.ts
```diff
@@ -55,7 +56,15 @@ export class BackendFactory<
    * Initialize an Amplify backend with the given construct factories and in the given CDK App.
    * If no CDK App is specified a new one is created
    */
-  constructor(constructFactories: T, stack: Stack = createDefaultStack()) {
+  constructor(
+    constructFactories: T,
+    stack?: Stack,
+    mainStackProps?: MainStackProps
```
sobolk, line 62: `stack` and `mainStackProps` should be mutually exclusive.

packages/backend/API.md
renevatium: Have updated in #[93b77dc](https://github.com/aws-amplify/amplify-backend/pull/1675/commits/93b77dcadb1e2d8531e8492fd91ccf2fa86f621e).

packages/backend/API.md
renevatium: Wasn't minimal for my specific issue. I have narrowed it in [e931b42](https://github.com/aws-amplify/amplify-backend/pull/1675/commits/e931b42cfb8eebc6a4023feff4fb41abd0e50560)

packages/backend/src/backend_factory.ts
renevatium: Have added constructor overloads to ensure this exclusivity in [3b24347](https://github.com/aws-amplify/amplify-backend/pull/1675/commits/3b24347294b12115cdb8ca80f822d89fd8706169)

renevatium:
@sobolk Thank you for reviewing my PR. 

Please let me know if I should rebase into a single commit at some point or if you prefer the history.

sobolk:
> @sobolk Thank you for reviewing my PR.
> 
> Please let me know if I should rebase into a single commit at some point or if you prefer the history.

We're using "squash merge" strategy for pull request so rebase is not necessary. It's actually not desired as it makes incremental reviews impossible to track in GitHub UI sometimes.

packages/backend/API.md
```diff
@@ -70,13 +71,18 @@ export { ConstructFactoryGetInstanceProps }
...
-export type DefineBackendProps = Record<string, ConstructFactory<ResourceProvider & Partial<ResourceAccessAcceptorFactory<never>>>> & {
+export type DefineBackendConstructFactories = Record<string, ConstructFactory<ResourceProvider & Partial<ResourceAccessAcceptorFactory<never>>>> & {
     [K in keyof BackendBase]?: never;
 };
 
+// @public (undocumented)
+export type DefineBackendProps = {
+    mainStackProps: MainStackProps;
+};
+
```
sobolk, line 85: I'm sorry I didn't realize that `DefineBackendProps` already exist as a type while providing initial feedback.

We should revert back to existing naming there (otherwise this is breaking change) and name new type something different.

Perhaps `DefineBackendOptions`. @Amplifiyer @edwardfoyle what do you think?

renevatium: To avoid any unnecessary delay, I have changed the new arg to DefineBackendOptions, but can change again easily enough, and DefineBackendConstructFactories has been reverted to DefineBackendProps.

packages/backend/src/backend_factory.ts
```diff
@@ -51,11 +58,37 @@ export class BackendFactory<
...
   private readonly stackResolver: StackResolver;
   private readonly customOutputsAccumulator: CustomOutputsAccumulator;
+
+  /**
+   * stack and props.mainStackProps are mutually exclusive
+   */
+  constructor(
+    constructFactories: T,
+    stack: Stack,
+    props?: ExplicitOmit<DefineBackendOptions, 'mainStackProps'>
```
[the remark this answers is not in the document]
renevatium: `never` would prevent passing anything in props, but only `mainStackProps` conflicts with a passed Stack. While there are currently no other props in the `DefineBackendOptions` type, @sobolk requested this implementation for the further expansion of the API and backend options could potentially be used for anything in the process tree, not only for the root Stack config. Semantically it also makes sense to allow `DefineBackendOptions` to be passed to `defineBackend` even if a Stack is passed, as long as those options aren't specific to the root Stack definition. 

FYI I did try with built-in Omit, but it's not explicit so doesn't prevent passing `mainStackProps`. Hence the `ExplicitOmit` implementation.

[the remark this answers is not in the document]
renevatium: We should maybe change the `props` key to `options` though, for consistency sake. Thoughts?

renevatium:
Hi @josefaidt 👋 No problem at all, big proponent of Amplify and keen to be more a part of the community. 

Considering it a point of conversation right now, I'll put my 2 cents in: Currently, this PR is a decent solution for my issue, but if I consider future flexibility, it does seem like exposing the explicit functionality necessary to define the App and Root Stack as independent objects and passing them into the (exposed) `BackendFactory` may be the most flexible state. I realize this conflicts with convention > config, but with CDK available in gen2, it feels like more custom config options/structure on the backend side will offer devs the most control with the least (necessary) limitation. Also, this would align with the current unit testing methodology, which doesn't use `defineBackend`, but rather explicitly defines App/Stack objects.

sisygoboom:
Amplify needs this desperately, without it I can't add a function created via `defineFunction` to a vpc imported into a `backend.createStack`. Are you any closer to a solution @josefaidt - its been 3 months now?
`Error: Cannot retrieve value from context provider vpc-provider since account/region are not specified at the stack level. Configure "env" with an account and region when you define your stack.`

PR #1678, opened 2024-06-25
feat: add imported tables to data

dpilch:
## Problem

Gen 1 -> Gen 2 table migration.

## Changes

Add new props:
* `migratedAmplifyGen1DynamoDbTableMap ` - mapping of model name to table name for imported tables.

Usage:

```typescript
const schema = `
  type Blog @model {
    title: String
    content: String
    authors: [String]
  }
  type ImportedModel @model {
    description: String
  }
`;
const dataFactory = defineData({
  schema,
  migratedAmplifyGen1DynamoDbTableMap: [{
    branchName: 'dev',
    modelTableNameMap: {
      ImportedModel: 'ImportedModel-1234-dev',
    },
  }],
});
```

## Validation

* Unit testing

## Checklist


packages/backend-data/API.md
```diff
@@ -12,6 +12,12 @@ import { DerivedModelSchema } from '@aws-amplify/data-schema-types';
 import { LogLevel } from '@aws-amplify/plugin-types';
 import { LogRetention } from '@aws-amplify/plugin-types';
 
+// @public
+export type AmplifyGen1DynamoDbTableMap = {
```
sobolk, line 16:
```suggestion
export type AmplifyGen1DynamoDbTableMapping = {

```

packages/backend-data/API.md
```diff
@@ -12,6 +12,12 @@ import { DerivedModelSchema } from '@aws-amplify/data-schema-types';
 import { LogLevel } from '@aws-amplify/plugin-types';
 import { LogRetention } from '@aws-amplify/plugin-types';
 
+// @public
+export type AmplifyGen1DynamoDbTableMap = {
+    branchName: string;
+    modelTableNameMap?: ModelTableNameMap;
```
sobolk, line 18:
```suggestion
    modelNameToTableNameMapping?: Record<string, string>;

```

or `tableNameByModelName` or `tableNameByModel`

i.e. some naming that indicates what is key vs what is value.

packages/backend-data/API.md
```diff
@@ -63,6 +70,9 @@ export type LambdaAuthorizationModeProps = {
     timeToLiveInSeconds?: number;
 };
 
+// @public
+export type ModelTableNameMap = Record<string, string>;
```
sobolk, line 74: I would inline this above and delete this type. It doesn't seem to bring much value and will show up in intellisense

packages/backend-data/src/factory.ts
```diff
@@ -149,7 +156,38 @@ class DataGenerator implements ConstructContainerEntryGenerator {
         ? this.props.schema.schemas
         : [this.props.schema];
 
-      schemas.forEach((schema) => {
+      // get the branch name and use the imported table map for that key
+      // use the sandbox key when the branch name is not available (e.g. in the sandbox deployment)
+      const amplifyBranchName =
+        scope.node.tryGetContext('amplifyEnvironmentName') ?? 'sandbox';
```
sobolk, line 162: could we use these?

https://github.com/aws-amplify/amplify-backend/blob/main/packages/platform-core/src/cdk_context_key.ts

Some examples:
https://github.com/aws-amplify/amplify-backend/blob/a6ef0f3afe87be8edf9c35d5352a35ada423c810/packages/backend-data/src/factory.ts#L245

https://github.com/aws-amplify/amplify-backend/blob/a6ef0f3afe87be8edf9c35d5352a35ada423c810/packages/backend/src/default_stack_factory.test.ts#L11-L13

packages/backend-data/src/factory.ts
```diff
@@ -149,7 +156,38 @@ class DataGenerator implements ConstructContainerEntryGenerator {
...
+      // ensure all branch names are unique
+      if (this.props.migratedAmplifyGen1DynamoDbTableMap) {
+        // Remove this warning for GA
+        Annotations.of(scope).addWarning(
+          'migratedAmplifyGen1DynamoDbTableMap is experimental and is not recommended for production use. This functionality may be changed or removed without warning.',
+        );
```
sobolk, line 168: we should annotate the property with `@experimental` then.

packages/backend-data/API.md
```diff
@@ -46,6 +52,7 @@ export type DataProps = {
     authorizationModes?: AuthorizationModes;
     functions?: Record<string, ConstructFactory<AmplifyFunction>>;
     logging?: DataLoggingOptions;
+    migratedAmplifyGen1DynamoDbTableMap?: AmplifyGen1DynamoDbTableMap[];
```
sobolk, line 55:
```suggestion
    migratedAmplifyGen1DynamoDbTableMappings?: AmplifyGen1DynamoDbTableMap[];

```

packages/backend-data/src/factory.ts
dpilch: This isn't a valid JSDoc tag according to eslint.

```
156:1  error  Invalid JSDoc tag name "experimental"  jsdoc/check-tag-names
```

Looks like it is a feature request, but has not been implemented. https://github.com/microsoft/TypeScript/issues/56808

sobolk (dismissed):

packages/backend-data/API.md
```diff
@@ -46,6 +52,7 @@ export type DataProps = {
     authorizationModes?: AuthorizationModes;
     functions?: Record<string, ConstructFactory<AmplifyFunction>>;
     logging?: DataLoggingOptions;
+    migratedAmplifyGen1DynamoDbTableMappings?: AmplifyGen1DynamoDbTableMapping[];
```
Amplifiyer, line 55: Do we expect more such props in the future? Should we premptively nest it under `gen1Migration` key to avoid polluting the top level API surface? In the future we can just take away the entire structure easily.

packages/backend-data/src/convert_schema.ts
```diff
@@ -225,3 +254,133 @@ const customSqlStatementsFromStrategies = (
...
+      importedDefinitionNodes as ObjectTypeDefinitionNode[];
+
+    importedModels.forEach((modelName) => {
+      if (
+        !importedObjectTypeDefinitionNodes.some(
+          (definitionNode) => definitionNode.name.value === modelName,
+        )
+      ) {
+        throw new Error(`Imported model not found in schema: ${modelName}`);
+      }
```
Amplifiyer, line 336: Use either `AmplifyUserError` or `AmplifyFault` in the backend factories.

packages/backend-data/src/factory.ts
```diff
@@ -149,7 +156,43 @@ class DataGenerator implements ConstructContainerEntryGenerator {
...
+      // use the sandbox key when in sandbox deployment
+      const amplifyBranchName = isSandboxDeployment
+        ? 'sandbox'
+        : scope.node.tryGetContext(CDKContextKey.BACKEND_NAME);
+      // ensure all branch names are unique
+      if (this.props.migratedAmplifyGen1DynamoDbTableMappings) {
+        // Remove this warning for GA
+        Annotations.of(scope).addWarning(
+          'migratedAmplifyGen1DynamoDbTableMappings is experimental and is not recommended for production use. This functionality may be changed or removed without warning.',
+        );
```
Amplifiyer, line 172: How does this warning work? Looking at the cdk docs, it seems the `addWarning` adds a un-acknowledgeable warning and fails in the `--strict` mode. Both sandbox and pipeline deployments are run in the `strict` mode.

Using `addWarningV2` also would fail as per the docs https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.Annotations.html#addwbrwarningwbrv2id-message

packages/backend-data/API.md
dpilch: I don't expect any more migrations related props in the future. The same concern came up in the API review and it was decided not to nest the property.

packages/backend-data/src/factory.ts
dpilch: The data construct is already using `Annotations.addWarning` and we haven't seen the deployments fail in strict mode. https://github.com/aws-amplify/amplify-category-api/blob/d6815d5fff632c648120e9cdbfe94beb696e7e44/packages/amplify-graphql-api-construct/src/amplify-graphql-api.ts#L181

Maybe the docs are inaccurate? I'll try testing this again.

Amplifiyer: This might be a bug in CDK if docs are correct such that if they fix we would fail. We should discuss this with CDK team.

dpilch: I confirmed that is deploying fine with `addWarning`. You are right that we could get the rug pulled though. I think we can change this to use `addInfo` for now. This will give us the same functionality of printing the message at synth time.

Then we can follow up with CDK to see why warning isn't working as expected.

packages/backend-data/src/convert_schema.ts
```diff
@@ -225,3 +255,136 @@ const customSqlStatementsFromStrategies = (
...
+        });
+      }
+    });
+
+    const importedSchemas = importedObjectTypeDefinitionNodes.map(
+      (definitionNode) => {
+        const importedTableName = (migratedAmplifyGen1DynamoDbTableMappings ??
+          {})[definitionNode.name.value];
+        if (!importedTableName) {
+          throw new Error(
```
Amplifiyer, line 348: One more instance needs `AmplifyError` treatment.

Amplifiyer (approved):

sobolk (approved):

PR #1695, opened 2024-06-28
chore: minify lambda function code and add source maps

Amplifiyer:
## Problem

## Changes

Enable minifying and adding sourcemaps to underlying lambda functions bundled code.

## Validation

Existing e2e passes and manually validated the minified and source maps in source code.

PS: I couldn't find a way to "inspect" the cdk assets so that I can observe the actual minification and source maps.

## Checklist


packages/backend-function/src/factory.ts
```diff
@@ -280,14 +280,14 @@ class AmplifyFunction
...
-     * This code concatenates the contents of the ssm resolver and invoker into a single line that can be used as the esbuild banner content
+     * This code concatenates the contents of the ssm resolver and invoker into a minified js that can be used as the esbuild banner content
      * This banner is responsible for resolving the customer's SSM parameters at runtime
      */
-    const bannerCode = readFileSync(ssmResolverFile, 'utf-8')
-      .concat(readFileSync(invokeSsmResolverFile, 'utf-8'))
-      .split(new RegExp(`${EOL}|\n|\r`, 'g'))
-      .map((line) => line.replace(/\/\/.*$/, '')) // strip out inline comments because the banner is going to be flattened into a single line
-      .join('');
+    const bannerCode = uglifyJs.minify(
```
sobolk, line 286: couple of questions here:
1. The banner isn't big, is it worth obscuring it?
2. Is `minify` knob below using same tool under the hood? If not, should we try to align that?


I'm asking because we're introducing new dependency and I'm not sure it's worth it.

Amplifiyer: 1. cc: @josefaidt as I got this as product req.
2. cdk uses esbuild's [minification](https://github.com/aws/aws-cdk/blob/06d2742b3fe9f9028d04e6660b654440ff9ca6e7/packages/aws-cdk-lib/aws-lambda-nodejs/lib/bundling.ts#L239) and I couldn't find if esbuild is using any external library for minification.

josefaidt: now that we're bundling the AWS SDK should we run this through esbuild with similar settings?

sobolk: We could but it seems to be an overkill.

The SSM snippet weighs 1877 bytes.
The Bundled AWS SDK for S3 with dependencies weighs 1390882 bytes (this is from SDK bundling prototype).

sobolk: I recommend to descope SSM snippet for now and seek esbuild based solution for it separately.

Amplifiyer: wouldn't the `aws-sdk` be bundled and hence minified by the `esbuild` ran by `aws-cdk`?

josefaidt: that seems fair

Amplifiyer: It won't minify the SSM banner that we include, that's why it was manually added here https://github.com/aws-amplify/amplify-backend/pull/1695/files#diff-e354c6c4617ce06e0e16948f4521eca5f9269445265b80ed16720372d336c886R286

edwardfoyle (approved):

rtpascual (approved):

Issue #1699, opened 2024-06-29
There is already a Construct with name 'storageRegion' in AmplifyStack


OperationalFallacy:
### Environment information

```plain text
npx ampx info
System:
  OS: macOS 14.5
  Shell: /bin/zsh
Binaries:
  Node: 20.15.0 - ~/.n/bin/node
  Yarn: 4.0.1 - ~/.n/bin/yarn
  npm: 10.7.0 - ~/.n/bin/npm
  pnpm: undefined - undefined
NPM Packages:
  @aws-amplify/backend: 1.0.4
  @aws-amplify/backend-cli: 1.1.0
  aws-amplify: 6.3.8
  aws-cdk: 2.147.2
  aws-cdk-lib: 2.147.2
  typescript: 5.5.2
AWS environment variables:
  AWS_STS_REGIONAL_ENDPOINTS = regional
  AWS_NODEJS_CONNECTION_REUSE_ENABLED = 1
  AWS_SDK_LOAD_CONFIG = 1
No CDK environment variables
```

### Description

If I add multiple defineStorage, the synth fails with the above error.

This is a use case for a lambda function in the backend that has access to data and needs to access storage buckets that the external stack creates. 

If I create buckets in the external (nested) stacks and try to grant access to lambda from the backend, there is a circular dependency error.

There are name conflicts if I try to create a bucket as storage in the backend.

josefaidt:
Hey @OperationalFallacy :wave: thanks for raising this! We do not currently support defining multiple buckets with `defineStorage` but it is top of mind to address! Marking this as a feature request

OperationalFallacy:
No problem

There is a simple workaround if anybody needs it.
It adds buckets to the storage stack (without auth, of course).
 
```
// Add extra buckets
const env = backend.data.node.tryGetContext("amplify-backend-name")
const outputBucket = new Bucket(backend.storage.resources.cfnResources.cfnBucket.stack, "output", {
  bucketName: `output-${Aws.ACCOUNT_ID}-${Aws.REGION}-${env}`
})
```

PR #1742, opened 2024-07-15
feat: Multi bucket support

0618:
## Problem

This PR allow users to create more than one S3 bucket via Amplify Storage.

**User Story**

- In amplify/storage/resource.ts...
```ts
import { defineStorage } from '@aws-amplify/backend';

export const storageOne = defineStorage({
  name: 'amplifyTeamDrive-one'
  isDefault: true
    access: (allow) => ({
    'profile-pictures/{entity_id}/*': [
      allow.guest.to(['read']),
      allow.entity('identity').to(['read', 'write', 'delete'])
    ],
    'picture-submissions/*': [
      allow.authenticated.to(['read','write']),
      allow.guest.to(['read', 'write'])
    ],
  })
});

export const storageTwo = defineStorage({
  name: 'amplifyTeamDrive-two'
});
export const storageThree = defineStorage({
  name: 'amplifyTeamDrive-three'
});
```

- In amplify/backend.ts...
```ts
import { storageOne, storageTwo, StorageThree } from './storage/resource';

defineBackend({
  auth,
  storageOne,
  storageTwo,
  storageThree,
});
```

- throw error if no default and multiple default buckets

Screenshot 2024-07-15 at 21 38 17

Screenshot 2024-07-15 at 21 43 23

fixes https://github.com/aws-amplify/amplify-backend/issues/1699
## Changes

- update schema_v1 and cliend_config_v1 to v1.1
- modify `BackendOutput` type
- rename policy to add more policies
- allow more buckets to be added

## Checklist


packages/backend-storage/src/construct.ts
```diff
@@ -146,15 +151,22 @@ export class AmplifyStorage
...
+     * The default bucket takes the `storageRegion` and `bucketName` name without a number post-fix.
+     */
+    const num = isDefault ? '' : Math.floor(Math.random() * 100);
     outputStorageStrategy.addBackendOutputEntry(storageOutputKey, {
       version: '1',
       payload: {
-        storageRegion: Stack.of(this).region,
-        bucketName: this.resources.bucket.bucketName,
+        [`storageRegion${num}`]: Stack.of(this).region,
+        [`bucketName${num}`]: this.resources.bucket.bucketName,
```
0618, line 169: cdk would throw an error saying "There's already a Construct name 'storageRegion'...", so we need to give it different names if it's not the default bucket.

packages/backend-storage/src/storage_access_policy_factory.ts
```diff
@@ -63,9 +63,15 @@ export class StorageAccessPolicyFactory {
       });
     }
 
-    return new Policy(this.stack, `${this.namePrefix}${this.policyCount++}`, {
```
0618, line 66: The previous name always started from 1,2,3 for each bucket. We need to give it different names.

I can try if the `num` in the construct.ts is available in a new commit.

packages/deployed-backend-client/src/backend_output_client.ts
```diff
@@ -22,6 +23,50 @@ export class DefaultBackendOutputClient implements BackendOutputClient {
       this.amplifyClient
     ).getStrategy(backendIdentifier);
     const output = await outputFetcher.fetchBackendOutput();
+
+    if (
+      output['AWS::Amplify::Storage'] &&
+      Object.keys(output['AWS::Amplify::Storage'].payload).length >= 4
```
0618, line 29: - [ ] need to double check if `Object.keys(output['AWS::Amplify::Storage'].payload).length >= 4` or `>=2`

packages/deployed-backend-client/src/backend_output_client.ts
```diff
@@ -22,6 +23,50 @@ export class DefaultBackendOutputClient implements BackendOutputClient {
...
+          message: 'No default bucket set in the Amplify project.',
+          resolution:
+            'Add `isDefault: true` to one of the buckets `defineStorage` in the Amplify project.',
+        });
+      }
+      bucketNames.forEach((key) => {
+        if (!key.startsWith('bucketName')) {
+          return;
+        }
+        const postfix = key.replace('bucketName', '');
```
0618, line 51: the `postfix` here are the `num` added in the `Construct`.

packages/backend-output-schemas/API.md
```diff
@@ -275,24 +275,29 @@ export const unifiedBackendOutputSchema: z.ZodObject<{
         payload: z.ZodObject<{
             bucketName: z.ZodString;
             storageRegion: z.ZodString;
+            buckets: z.ZodOptional<z.ZodArray<z.ZodRecord<z.ZodString, z.ZodString>, "many">>;
```
sobolk, line 278: why typing for `buckets` is different here than in other places ?

additionally. should we use records here or should be come up with subtype with named properties? (this applies to all `buckets` introduced in this file)

packages/backend-output-schemas/src/storage/v1.ts
```diff
@@ -1,9 +1,12 @@
 import { z } from 'zod';
 
+const bucketSchema = z.record(z.string(), z.string());
```
sobolk, line 3: should we type this stronger as well?

packages/backend-storage/API.md
```diff
@@ -23,9 +22,13 @@ export type AmplifyStorageFactoryProps = Omit<AmplifyStorageProps, 'outputStorag
...
 // @public (undocumented)
 export type AmplifyStorageProps = {
+    isDefault?: boolean;
     name: string;
     versioned?: boolean;
-    outputStorageStrategy?: BackendOutputStorageStrategy<StorageOutput>;
+    outputStorageStrategy?: BackendOutputStorageStrategy<{
+        version: '1';
+        payload: Record<string, string>;
+    }>;
```
sobolk, line 31: why this change?
Shouldn't we change `StorageOutput` type ?

packages/backend-output-storage/src/stack_metadata_output_storage_strategy.ts
```diff
@@ -36,9 +36,19 @@ export class StackMetadataBackendOutputStorageStrategy
...
+        : { stackOutputs: [] };
+
     this.stack.addMetadata(keyName, {
+      ...metadataValue,
       version: backendOutputEntry.version,
-      stackOutputs: Object.keys(backendOutputEntry.payload),
+      stackOutputs: [
+        ...metadataValue.stackOutputs,
+        ...Object.keys(backendOutputEntry.payload),
+      ],
```
sobolk, line 51: Could you explain this change?

Edit: after reading rest of the code I'm guessing this is to solve a problem of "multiple defineStorage contributing to same stack metadata/output key". 
I was facing similar problem in the POC I was building for custom client config. This wasn't merged as final approach was different but you might find it useful for this PR, see https://github.com/aws-amplify/amplify-backend/pull/946#discussion_r1466988928 (i.e. `addBackendOutputEntry` and logic around it).

packages/backend-storage/src/construct.ts
```diff
@@ -122,7 +127,7 @@ export class AmplifyStorage
       },
     };
 
-    this.storeOutput(props.outputStorageStrategy);
+    this.storeOutput(props.outputStorageStrategy, props.isDefault || false);
```
sobolk, line 130:
```suggestion
    this.storeOutput(props.outputStorageStrategy, props.isDefault ?? false);
```

packages/backend-storage/src/construct.ts
```diff
@@ -146,15 +151,22 @@ export class AmplifyStorage
...
+      version: '1';
+      payload: Record<string, string>;
+    }> = new StackMetadataBackendOutputStorageStrategy(Stack.of(this)),
+    isDefault: boolean = false
   ): void => {
+    /* The following code can guarantee there's only one `isDefault`
+     * because if we can only create one construct with the same `storageRegion` and `bucketName` name.
+     * The default bucket takes the `storageRegion` and `bucketName` name without a number post-fix.
+     */
+    const num = isDefault ? '' : Math.floor(Math.random() * 100);
```
sobolk, line 164: We shouldn't use random generator here. This will cause re-deployments of storage as outputs are going to change every time impacting performance (e.g. this will cause sandbox never hotswapping).
Instead of this we should either apply consistent hashing OR have global counter for storage instances.

packages/backend-storage/src/storage_access_policy_factory.ts
```diff
@@ -63,9 +63,15 @@ export class StorageAccessPolicyFactory {
...
-      statements,
-    });
+    return new Policy(
+      this.stack,
+      `${this.bucket.bucketName.slice(0, 6)}${Math.floor(
```
sobolk, line 68: same here, this needs consistent hashing or global counter.

packages/deployed-backend-client/src/backend_output_client.ts
```diff
@@ -22,6 +27,64 @@ export class DefaultBackendOutputClient implements BackendOutputClient {
       this.amplifyClient
     ).getStrategy(backendIdentifier);
     const output = await outputFetcher.fetchBackendOutput();
-    return unifiedBackendOutputSchema.parse(output);
+
+    return unifiedBackendOutputSchema.parse({
+      ...output,
+      ...(output[storageOutputKey] && {
+        [storageOutputKey]: this.parseStorageOutput(output[storageOutputKey]),
```
sobolk, line 34: it is odd to have storage output parsing being so specific.
can't this be solved by modeling schema properly?

packages/deployed-backend-client/src/backend_output_client.ts
```diff
@@ -22,6 +27,64 @@ export class DefaultBackendOutputClient implements BackendOutputClient {
...
+      key.startsWith('bucketName')
+    );
+    const hasDefaultBucket = bucketNames.includes('bucketName');
+    if (!hasDefaultBucket && bucketNames.length > 1) {
+      throw new AmplifyUserError('NoDefaultBucketError', {
+        message: 'No default bucket set in the Amplify project.',
+        resolution:
+          'Add `isDefault: true` to one of the buckets `defineStorage` in the Amplify project.',
+      });
+    }
```
sobolk, line 62: this validation should happen at synthesis time. not here.

packages/plugin-types/API.md
```diff
@@ -74,10 +74,10 @@ export type BackendIdentifier = {
 };
 
 // @public (undocumented)
-export type BackendOutput = Record<string, BackendOutputEntry>;
+export type BackendOutput<T = Record<string, string>> = Record<string, BackendOutputEntry<T>>;
```
sobolk, line 77: why is this needed?

packages/backend-storage/src/construct.ts
0618: `isDefault` can't be `0` or `""` since it's set to be `boolean` type, so I think `||` is appropriate here. Even if it was `0` or `""`, I think we can assume it means `false`.

packages/backend-storage/src/construct.ts
0618: Thanks. Yea, I'm looking for some suggestions to replace the random generator. I'll try those.

packages/deployed-backend-client/src/backend_output_client.ts
0618: Thanks. This is the comment I've been waiting for :p

packages/plugin-types/API.md
0618: Because `buckets` is `Record<string, string>[]`, but I can't set it in the type since it breaks other categories.

Parameter `T` was introduced so that we can restrict it only for Storage.

sobolk: I think by adapting the concept suggested here https://github.com/aws-amplify/amplify-backend/pull/1742#discussion_r1681550455 we could nest buckets under single storage output and avoid this.

packages/backend-output-storage/src/stack_metadata_output_storage_strategy.ts
rtpascual: Could we maybe use `appendToBackendOutputList` (see [here](https://github.com/aws-amplify/amplify-backend/blob/main/packages/backend-output-storage/src/stack_metadata_output_storage_strategy.ts#L48)) to lazily append multiple buckets to the same output key similar to what is done for functions? See [here](https://github.com/aws-amplify/amplify-backend/blob/main/packages/backend-function/src/factory.ts#L390).

0618: Good point @rtpascual! I somehow didn't see `appendToBackendOutputList` in this file.

packages/backend-storage/src/construct.ts
```diff
@@ -32,6 +29,11 @@ const storageStackType = 'storage-S3';
 export type AmplifyStorageTriggerEvent = 'onDelete' | 'onUpload';
 
 export type AmplifyStorageProps = {
+  /**
+   * Whether this storage resource is the default storage resource for the backend
+   * @default false
```
edwardfoyle, line 34: isn't the default true? ie if I just have one storage instance and I don't define this field, then it's the default

packages/backend-storage/src/construct.ts
```diff
@@ -42,7 +44,10 @@ export type AmplifyStorageProps = {
    * @default false
    */
   versioned?: boolean;
-  outputStorageStrategy?: BackendOutputStorageStrategy<StorageOutput>;
+  outputStorageStrategy?: BackendOutputStorageStrategy<{
+    version: '1';
+    payload: Record<string, string>;
```
edwardfoyle, line 49: I think @sobolk had a similar comment elsewhere, but we should keep this strongly typed, even if that means updating the type

packages/backend-storage/src/construct.ts
edwardfoyle: We should do the same thing that we do for auth and data: https://github.com/aws-amplify/amplify-backend/blob/7361ff4ed9da10f9b48e327c96eff25eb81a9e10/packages/backend-auth/src/factory.ts#L68

keep a static counter in the factory class

packages/deployed-backend-client/src/backend_output_client.ts
edwardfoyle: +1 this should all be handled by zod if we model the type properly

edwardfoyle: Coming back to this, I think you may be conflating reading backend output and formatting client config. This class should only be responsible for reading stack output and metadata and reconstituting it into a validated and typed object. Then the client config contributors take that object and format it however is needed. That happens here for storage: https://github.com/aws-amplify/amplify-backend/blob/7361ff4ed9da10f9b48e327c96eff25eb81a9e10/packages/client-config/src/client-config-contributor/client_config_contributor_v1.ts#L254

packages/backend-output-storage/src/stack_metadata_output_storage_strategy.ts
0618: Done [e8548cf](https://github.com/aws-amplify/amplify-backend/pull/1742/commits/e8548cf39936a5a284ce458e2390f7eb01cc833e)

packages/backend-storage/src/construct.ts
0618: Good point. I don't know if the default should be `true` or `false` because it's conditional. If there's only one bucket, then it's `true`. If multiple buckets, it's `false`. Maybe I should update the JSdoc to clarify the conditions

0618: Done 
bc4d412

packages/backend-storage/src/construct.ts
```diff
@@ -32,6 +29,11 @@ const storageStackType = 'storage-S3';
 export type AmplifyStorageTriggerEvent = 'onDelete' | 'onUpload';
 
 export type AmplifyStorageProps = {
+  /**
+   * Whether this storage resource is the default storage resource for the backend
```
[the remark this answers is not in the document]
0618: Done 
bc4d412

packages/plugin-types/API.md
0618: The doesn't affect here because the buckets array will be under single storage, which is not acceptable for `BackendOutput` before the current change with `T`. I saw the following error:

```
Type '{ name: string; bucket_name: string; aws_region: string; }[]' is not assignable to type 'string'.
```

sobolk: Perhaps we should do something like this https://github.com/aws-amplify/amplify-backend/blob/7361ff4ed9da10f9b48e327c96eff25eb81a9e10/packages/backend-output-schemas/src/function/v1.ts#L6 .

packages/backend-storage/API.md
0618: StorageOutput is from schema, which is the final output, but the initial output we stored in stack metadata is different, so I introduced a new type, `StorageOutputToStore` in a new commit [f6a1c12](https://github.com/aws-amplify/amplify-backend/pull/1742/commits/f6a1c121a2238c021c9c09c29281a18257090af7)

0618: See how the new commit [f6a1c12](https://github.com/aws-amplify/amplify-backend/pull/1742/commits/f6a1c121a2238c021c9c09c29281a18257090af7) looks?

sobolk: I don't understand why do we need type `StorageOutputToStore` .

Auth and Functions are just using their respective outputs, see:
https://github.com/aws-amplify/amplify-backend/blob/7361ff4ed9da10f9b48e327c96eff25eb81a9e10/packages/auth-construct/src/construct.ts#L990
and
https://github.com/aws-amplify/amplify-backend/blob/7361ff4ed9da10f9b48e327c96eff25eb81a9e10/packages/backend-function/src/factory.ts#L238

sobolk: > StorageOutput is from schema

`StorageOutput` is a shape that we store in CFN metadata/outputs. This is not related to client config schemas - that translation happens when client config package reads stacks and transforms it to amplify outputs.

packages/backend-storage/src/construct.ts
0618: Done cddcd97830e605666c3d3c8be81e4f2fedbdb5b5

packages/deployed-backend-client/src/backend_output_client.ts
0618: Done 1b08c09c20565483c9d856e04dc39407c3aecda6

packages/backend-storage/src/storage_access_policy_factory.ts
0618: done 2f932ec1fd42f741a56d36e2c04096d81dae0711

packages/backend-storage/API.md
0618: There would be some mis-match after `CfnOutput` sending the output ([code](https://github.com/aws-amplify/amplify-backend/blob/main/packages/backend-output-storage/src/stack_metadata_output_storage_strategy.ts#L86)), then the [output](https://github.com/aws-amplify/amplify-backend/blob/main/packages/deployed-backend-client/src/backend_output_client.ts#L24) we get from CFN would be like the following

```
{
 version: "1",
 payload: {
   name: [..., ..., ...],
   bucketName: [..., ..., ...],
   storageRegion: [..., ..., ...],
 }
}
```

So we need to change the type, and modify the output when parsing to `amplify_outputs.json` ([code](https://github.com/aws-amplify/amplify-backend/blob/main/packages/deployed-backend-client/src/backend_output_client.ts#L25C45-L25C51)).

Is there a way to avoid the change?

0618: tl;dr -- When we store output, we store the individual bucket ([code](https://github.com/aws-amplify/amplify-backend/blob/main/packages/backend-storage/src/construct.ts#L153)) (which has `name`, `bucketName`, and `storageRegion`), but when generate the amplify_outputs.json, we also need to add `buckets`, and remove `name` etc. So, there are two different `storageOutput` . 

There might be a way to avoid the complex, and I missed it.

sobolk: The way we should evolve this is
Change
```
export const storageOutputSchema = z.object({
  version: z.literal('1'),
  payload: z.object({
    bucketName: z.string(),
    storageRegion: z.string(),
  }),
});
```

To

```
export const storageOutputSchema = z.object({
  version: z.literal('1'),
  payload: z.object({
    bucketName: z.string(),
    storageRegion: z.string(),
    buckets: z.string(), // JSON array as string (serialized buckets, like in functions output), this should likely be optional.
  }),
});

```

And then in order for this to be backwards compatible we should populate bucketName and storageRegion with the one that's 'isDefault = true'.

sobolk: And then use this trick https://github.com/aws-amplify/amplify-backend/blob/890ce865cd9988d504878ea115104e01cfdfbc47/packages/backend-function/src/factory.ts#L393-L398 to append to `buckets`.

packages/backend-output-schemas/src/storage/v1.ts
```diff
@@ -1,9 +1,16 @@
 import { z } from 'zod';
 
+const bucketSchema = z.object({
+  name: z.string(),
+  bucketName: z.string(),
+  storageRegion: z.string(),
```
Amplifiyer, line 6: Why do we need the `name` here?

packages/backend/src/backend_factory.ts
```diff
@@ -56,6 +60,16 @@ export class BackendFactory<
...
+    if (
+      !AmplifyStorageFactory.hasDefault &&
+      AmplifyStorageFactory.factoryCounter > 1
+    ) {
+      throw new AmplifyUserError('NoDefaultBucketError', {
+        message: 'No default bucket set in the Amplify project.',
+        resolution:
+          'Add `isDefault: true` to one of the buckets `defineStorage` in the Amplify project.',
+      });
+    }
```
Amplifiyer, line 72: Is there no other place for this validation than using static in Storage and using in the `backend_factory`? Perhaps using CDK.Lazy like here in function to delay this validation just before synth

https://github.com/aws-amplify/amplify-backend/blob/ec12ac83d730a37c539ff6ae2ee282341d4c5342/packages/backend-function/src/function_env_translator.ts#L43

packages/backend-output-schemas/src/storage/v1.ts
0618: `name` needs to be included in amplify_outputs.json and schema. If we don't pass it here, we can't collect the info...or, can we?

packages/backend-output-schemas/API.md
0618: Done d67854ec809441ddd20bc88407be944aa6950d42

packages/backend-output-schemas/src/storage/v1.ts
0618: Done d67854ec809441ddd20bc88407be944aa6950d42

packages/backend-storage/API.md
0618: Done d67854ec809441ddd20bc88407be944aa6950d42

packages/backend-storage/src/storage_access_policy_factory.ts
0618: - [ ] need to revisit

packages/backend/src/backend_factory.ts
0618: - [ ] TODO

packages/plugin-types/API.md
0618: - [ ] need to figure out why API.md doesn't update properly

packages/backend-storage/src/construct.ts
```diff
@@ -148,13 +155,23 @@ export class AmplifyStorage
...
       payload: {
-        storageRegion: Stack.of(this).region,
-        bucketName: this.resources.bucket.bucketName,
+        storageRegion: isDefault ? Stack.of(this).region : '',
+        bucketName: isDefault ? this.resources.bucket.bucketName : '',
+        buckets: JSON.stringify({
+          name,
+          bucketName: this.resources.bucket.bucketName,
+          storageRegion: Stack.of(this).region,
+        }),
```
sobolk, line 174: We should make sure that:
1. `storageRegion` and `bucketName` stay the same (not become array) and point to default bucket.
2. We append only to `buckets`. 

This may require some changes to outputStorageStrategy.
I think we should make `appendToBackendOutputList` to take `Partial<BackendOutputEntry>` to achieve that.

package.json
```diff
@@ -64,7 +64,7 @@
     "@aws-sdk/client-ssm": "^3.465.0",
     "@changesets/cli": "^2.26.1",
     "@changesets/get-release-plan": "^4.0.0",
-    "@microsoft/api-extractor": "^7.34.4",
+    "@microsoft/api-extractor": "7.40.0",
```
0618, line 67: ping the version for https://github.com/microsoft/rushstack/issues/4842

packages/backend/src/backend_factory.ts
0618: Deferring a calculation until before synths makes sense, but for validation, don't we want to fail fast?

packages/backend-storage/src/storage_access_policy_factory.ts
0618: Done. Using global counter ec113d0077

packages/backend-storage/src/construct.ts
0618: Done 
3380bd9

`Partial` doesn't do "deep partial", so I had to introduce `PickPayload` and `TransformType`.

.changeset/calm-otters-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+'@aws-amplify/backend-output-schemas': patch
```
sobolk, line 2:
```suggestion
'@aws-amplify/backend-output-schemas': minor
```

packages/backend-storage/API.md
```diff
@@ -7,13 +7,36 @@
...
+
+// @public
+export class AmplifyStorageFactory implements ConstructFactory<ResourceProvider<StorageResources>> {
+    constructor(props: AmplifyStorageFactoryProps, importStack?: string | undefined);
+    // (undocumented)
+    static factoryCounter: number;
+    getInstance: (getInstanceProps: ConstructFactoryGetInstanceProps) => AmplifyStorage;
+    // (undocumented)
+    static hasDefault: boolean;
+}
```
sobolk, line 39: Why are we exporting these types ?

packages/backend-storage/src/construct.ts
```diff
@@ -153,20 +155,26 @@ export class AmplifyStorage
    * Store storage outputs using provided strategy
    */
   private storeOutput = (
-    outputStorageStrategy: BackendOutputStorageStrategy<StorageOutput> = new StackMetadataBackendOutputStorageStrategy(
-      Stack.of(this)
-    ),
+    outputStorageStrategy: BackendOutputStorageStrategy<
+      StorageOutput | StorageBucketsPayload
```
sobolk, line 159: we shouldn't create `StorageBucketsPayload`. I.e. we shouldn't introduce storage specific view of its output schema. 

Instead

We should add type like this one https://www.geodev.me/blog/deeppartial-in-typescript to `plugin-types` and then make `appendToBackendOutputList` take `DeepPartial<BackendOutput>` . I.e. loosen abstraction.

packages/backend-storage/API.md
0618: Because we need to use use `AmplifyStorageFactory.isDefault` and counter in packages/backend/src/backend_factory.ts

packages/backend-storage/src/construct.ts
0618: I was not sure if to add the new type (or `DeepPartial<BackendOutput>`) to storage schema or to `plugin-types` because `StorageOutput` is in storage schema. How do we decide where to add?

sobolk: This API should be modified to accept partials https://github.com/aws-amplify/amplify-backend/blob/ec12ac83d730a37c539ff6ae2ee282341d4c5342/packages/plugin-types/src/output_storage_strategy.ts#L8 .
Hence type should be defined to be accessible to it.

packages/backend-storage/API.md
sobolk: Information about storage and buckets must not leak into `packages/backend/src/backend_factory.ts` . 
We'd be exposing implementation detail of vertical. We can't do that.
There should be some other way to achieve that that is scoped to `backend-storage`. 

Please check if you could either:
1. Hook up into validation callbacks of CDK constructs
2. Add CDK aspect to storage stack that inspects child nodes.

The goals are:
1. Throw error at synthesis
2. Do not leak storage information to core components.

packages/backend-storage/src/construct.ts
0618: Done [88ea17d](https://github.com/aws-amplify/amplify-backend/pull/1742/commits/88ea17da1f11c5eccfa655be9d00ec87515df95f) Thanks for help!

.changeset/calm-otters-speak.md
0618: Removed.

packages/backend-storage/API.md
0618: Done [50ebda5](https://github.com/aws-amplify/amplify-backend/pull/1742/commits/50ebda51550a3839e566bcbe4bb884e478d63c3d)

Used `Aspects` to validate.

packages/backend-output-storage/src/stack_metadata_output_storage_strategy.ts
```diff
@@ -36,18 +37,34 @@ export class StackMetadataBackendOutputStorageStrategy
...
+      const excludedStackOutputs = existingMetadataEntry.stackOutputs.filter(
+        (output: string) =>
+          !Object.keys(backendOutputEntry.payload).includes(output)
+      );
+      this.stack.addMetadata(keyName, {
+        version: backendOutputEntry.version,
+        stackOutputs: Object.keys(backendOutputEntry.payload).concat(
+          excludedStackOutputs
+        ),
+      });
```
sobolk, line 53: Can this be simplified by using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set ?.

If I understand this. This piece of code is trying to merge two arrays without duplicates right?

packages/backend-output-storage/src/stack_metadata_output_storage_strategy.ts
```diff
@@ -61,32 +78,46 @@ export class StackMetadataBackendOutputStorageStrategy
...
+
+    if (existingMetadataEntry) {
+      this.stack.addMetadata(keyName, {
+        version,
+        stackOutputs: Array.from(listsMap ? listsMap.keys() : []).concat(
+          existingMetadataEntry.stackOutputs.filter(
+            (output: string) => !listsMap?.has(output)
+          )
+        ),
+      });
```
sobolk, line 115: it seems this could also benefit from same simplification with `Set` .

And. This block could be lifted up to be closer to `existingMetadataEntry` definition above, so operations on metadata are toghether.

Also consider creating `private createOrAppendToMetadata()` method that is used in both `addBackendOutputEntry` and `append...` functions.

packages/backend-storage/API.md
```diff
@@ -7,22 +7,34 @@
...
 import { FunctionResources } from '@aws-amplify/plugin-types';
 import { IBucket } from 'aws-cdk-lib/aws-s3';
+import { IFunction } from 'aws-cdk-lib/aws-lambda';
 import { ResourceAccessAcceptor } from '@aws-amplify/plugin-types';
 import { ResourceAccessAcceptorFactory } from '@aws-amplify/plugin-types';
 import { ResourceProvider } from '@aws-amplify/plugin-types';
 import { StorageOutput } from '@aws-amplify/backend-output-schemas';
 
+// @public
+export class AmplifyStorage extends Construct implements ResourceProvider<StorageResources> {
```
sobolk, line 23: this should not be exported from package.

0618: Done [c8a36f7](https://github.com/aws-amplify/amplify-backend/pull/1742/commits/c8a36f7d015bff1d8908332a89286e9799fe8b00)

packages/backend-output-storage/src/stack_metadata_output_storage_strategy.ts
0618: Good catch. Done [9843705](https://github.com/aws-amplify/amplify-backend/pull/1742/commits/9843705aeb2cc5ba2894938175d691752ebae988)

packages/backend-output-storage/src/stack_metadata_output_storage_strategy.ts
0618: Done [9843705](https://github.com/aws-amplify/amplify-backend/pull/1742/commits/9843705aeb2cc5ba2894938175d691752ebae988)

packages/backend-storage/src/storage_access_policy_factory.ts
```diff
@@ -17,11 +17,9 @@ export type Permission = {
  * Generates IAM policies scoped to a single bucket
  */
 export class StorageAccessPolicyFactory {
-  private readonly namePrefix = 'storageAccess';
+  static policyCount = 1;
```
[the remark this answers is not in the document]
0618: because we need to track if we have a `isDefault` amoung all the buckets if there's more than 1 bucket.

packages/backend-storage/src/storage_access_policy_factory.ts
```diff
@@ -63,9 +61,16 @@ export class StorageAccessPolicyFactory {
...
 
-    return new Policy(this.stack, `${this.namePrefix}${this.policyCount++}`, {
-      statements,
-    });
+    return new Policy(
+      this.stack,
+      `${this.bucket.bucketName.slice(
+        0,
+        6
+      )}${StorageAccessPolicyFactory.policyCount++}`,
```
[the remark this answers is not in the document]
0618: because the policy has a length limit, we have to trim it. But `6` is just an arbitrary number.

packages/backend-storage/src/factory.ts
```diff
@@ -46,10 +63,34 @@ class AmplifyStorageFactory
...
 
+class StorageValidator implements IAspect {
+  public visit(node: IConstruct): void {
+    if (!(node instanceof AmplifyStorage) || AmplifyStorageFactory.validated) {
+      return;
+    }
+    if (
+      !AmplifyStorageFactory.hasDefault &&
+      AmplifyStorageFactory.factoryCounter > 1
+    ) {
```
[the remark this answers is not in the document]
0618: I think our stack and nestedStacks are all created in the `@aws-amplify/backend` package, right? We don't want to expost storage there.

packages/integration-tests/src/test-project-setup/test_project_base.ts
```diff
@@ -183,6 +183,12 @@ export abstract class TestProjectBase {
       await currentCodebaseBackendOutputClient.getOutput(backendId);
     const npmOutputs = await npmBackendOutputClient.getOutput(backendId);
 
+    console.log(
+      '🪵 currentCodebaseOutputs: ',
+      JSON.stringify(currentCodebaseOutputs, null, 2)
+    );
+    console.log('🪵 npmOutputs: ', JSON.stringify(npmOutputs, null, 2));
+
```
0618, line 191: adding the log to make it easier to compare if the output updates are intented. Otherwise, it's hard to see what the differences are. e.g. https://github.com/aws-amplify/amplify-backend/actions/runs/10120959378/job/28002086625

packages/backend-storage/src/storage_access_policy_factory.ts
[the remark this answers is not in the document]
Amplifiyer: Isn't this being passed for a CDK construct Id? https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_iam.Policy.html#initializer

We have length restrictions for this?

packages/backend-storage/src/storage_access_policy_factory.ts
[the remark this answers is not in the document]
Amplifiyer: I was asking why static for `policyCount` and I suppose you'd want to number all the policies across constructs. But do we even need to? Wouldn't CDK ensures a unique Id for the policyName?

packages/backend-storage/src/storage_access_policy_factory.ts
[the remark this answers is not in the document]
0618: - [ ] test with a long bucketName

packages/backend-storage/src/storage_access_policy_factory.ts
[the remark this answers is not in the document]
0618: Yep, CDK is unfortunately not that smart 🤦‍♀️   We had to use a `this.policyCount` for single bucket as well.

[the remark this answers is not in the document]
0618: - [ ] try hashing the statement

packages/backend-storage/src/factory.ts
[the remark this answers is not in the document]
0618: Done!

packages/backend-storage/src/storage_access_policy_factory.ts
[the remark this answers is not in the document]
0618: Checked. It has a limitation for the length of the id. 

People has some workaround though https://github.com/aws/aws-cdk/issues/3402

packages/backend-storage/src/storage_access_policy_factory.ts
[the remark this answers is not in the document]
0618: I just realized that we could use `this.stack.node.children.length`, which is different every time when creating the policy.

packages/backend-storage/src/factory.ts
```diff
@@ -46,10 +49,68 @@ class AmplifyStorageFactory
         getInstanceProps
       );
     }
-    return constructContainer.getOrCompute(this.generator) as AmplifyStorage;
+    const amplifyStorage = constructContainer.getOrCompute(
+      this.generator
+    ) as AmplifyStorage;
+    Aspects.of(amplifyStorage).add(
+      new StorageValidator(Stack.of(amplifyStorage))
```
Amplifiyer, line 56: Instead of creating an aspect for the individual `AmplifyStorage`, you can just create one for the nested stack which will get executed just once and you wouldn't have to pass the nested stack in the constructor

```suggestion
    Aspects.of(Stack.of(amplifyStorage)).add(
      new StorageValidator()
```

Amplifiyer (commented):
Looking good overall.

packages/backend-output-storage/src/stack_metadata_output_storage_strategy.ts
```diff
@@ -61,32 +66,75 @@ export class StackMetadataBackendOutputStorageStrategy
...
+     * If there's only one bucket and not the bucket is not set as default, then it only has 'buckets' in the stackOutputs.
+     * So we need to manually add the bucketName and storageRegion to the stackOutputs.
+     * As long as there's a bucket, the bucketName, storageRegion, and buckets will always be present in the stackOutputs.
+     */
+    const stackOutputs = Object.keys(backendOutputEntry.payload).includes(
+      'buckets'
+    )
+      ? ['buckets', 'bucketName', 'storageRegion']
+      : Object.keys(backendOutputEntry.payload);
+
```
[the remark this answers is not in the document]
0618: I may need to re-write the comment. The idea is...

- not default bucket -> `appendToBackendOutputList` -> `metaData` has `[buckets]`
- default bucket -> `addBackendOutputEntry` and `appendToBackendOutputList` -> `metaData` has `['buckets', 'bucketName', 'storageRegion']`

There's no problem if there's a default bucket, but if there's no default bucket (We allow to have no default bucket if there's only one bucket), the `metaData` is lack of `[bucketName', 'storageRegion']`

packages/backend-output-schemas/src/storage/v1.ts
0618: `name` is in the outputs. The output looks like this

```
...
  "storage": {
    "aws_region": "us-east-1",
    "bucket_name": "amplify-viteproject-zhamu-amplifyteamdrivebucket28-nriywgwjhjnt",
    "buckets": [
      {
        "name": "amplifyTeamDrive",
        "bucket_name": "amplify-viteproject-zhamu-amplifyteamdrivebucket28-nriywgwjhjnt",
        "aws_region": "us-east-1"
      },
      {
        "name": "amplifyPoCTwo",
        "bucket_name": "amplify-viteproject-zhamu-amplifypoctwobucket0466f-caqijvcd79nc",
        "aws_region": "us-east-1"
      }
    ]
  },
  ...
```

packages/backend-output-storage/src/stack_metadata_output_storage_strategy.ts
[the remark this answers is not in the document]
Amplifiyer: I think I understand the problem now. You don't know the best time to add the top level `bucketName` and `storageRegion` to the metadata as they exist today, since if a bucket is marked default, that should be added to the metadata, but if not (and it's the only bucket) you'd have to delay the addition until you are sure that there exists only one bucket.

I think a little less hacky solution for this would be to get hold of `outputStorageStrategy` from the `getInstanceProps` in the storage factory. You can then use the `outputStorageStrategy.addBackendOutputEntry()` in the aspect that you are creating at the storage bucket nested stack level, at which point you have the complete picture of all the buckets. This way you can also avoid [creating the CFNOutputs manually in the aspect ](https://github.com/aws-amplify/amplify-backend/pull/1742/files/db34c500c02e0994a7e62d8b043b44bce83affd5#diff-27a0769d9fccd4e1743973fa17e8ef386c1ac148f35c41fce90bbfe11d0837afR98-R103)

@sobolk thoughts?

packages/backend-storage/src/factory.ts
```diff
@@ -52,7 +58,22 @@ export class AmplifyStorageFactory
...
+    const firstStorage = Stack.of(amplifyStorage).node.children.filter(
+      (el) => el instanceof AmplifyStorage
+    )[0] as AmplifyStorage;
+
+    /*
+     * only call Aspects once,
+     * otherwise there will be the an error -
+     * "there is already a construct with name 'storageRegion'"
+     */
+    if (firstStorage.name === this.props.name) {
```
0618, line 71: Had to handle it in this hacky way to make sure the Aspect only runs once. 

Would it be better to handle it on the top stack level? But that'll be out of the storage package?

Any suggestions?

packages/backend-storage/src/factory.ts
```diff
@@ -73,38 +100,95 @@ export class StorageValidator implements IAspect {
...
     }
-    let storageCount = 0;
+    const storageInstances = Stack.of(node).node.children.filter(
+      (el) => el instanceof AmplifyStorage
+    );
+    const storageCount = storageInstances.length;
+    const firstStorage = Stack.of(node).node.children.filter(
+      (el) => el instanceof AmplifyStorage
+    )[0];
+    if (node !== firstStorage) {
```
0618, line 110: Same here

packages/backend-storage/src/construct.ts
```diff
@@ -25,13 +25,20 @@ import {
 import { fileURLToPath } from 'node:url';
 import { IFunction } from 'aws-cdk-lib/aws-lambda';
 import { S3EventSourceV2 } from 'aws-cdk-lib/aws-lambda-event-sources';
+import { AmplifyUserError } from '@aws-amplify/platform-core';
```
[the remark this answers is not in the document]
0618: Done.

packages/backend-output-storage/src/stack_metadata_output_storage_strategy.ts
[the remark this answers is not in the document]
0618: Refactored in 179a1bb9c8e48b89f1e449a16fba3c57a1fcca47, but I didn't figure out an elegant way to run Aspect only once. Any suggestions? @edwardfoyle @sobolk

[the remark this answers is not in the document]
0618: oh, I was trying to @Amplifiyer instead 😂

packages/backend-storage/src/factory.ts
sobolk: you can try something like 

```
const aspects = Aspects.of(Stack.of(amplifyStorage));
if (!aspects.all.find(aspect => aspect instanceof StorageValidator) { // or any other predicate that would work here.
  aspects.add(new StorageValidator(...));
}
```

packages/backend-storage/src/factory.ts
```diff
@@ -46,7 +55,139 @@ class AmplifyStorageFactory
...
+      (el) => el instanceof AmplifyStorage
+    )[0] as AmplifyStorage;
+
+    /*
+     * only call Aspects once,
+     * otherwise there will be the an error -
+     * "there is already a construct with name 'storageRegion'"
+     */
+    if (firstStorage.name === this.props.name) {
+      Aspects.of(Stack.of(amplifyStorage)).add(
```
sobolk, line 72: Regarding. `Stack.of(amplifyStorage)` .

I was chatting with @josefaidt about necessity to place every `defineFunction` in separate stack.
It may happen that we'll want same behavior for `defineStorage` as opt in feature.
If this happens. Will this algorithm work if we do `Stack.of(amplifyStorage)`, then traverse parent nodes to find root and place aspect there?

packages/backend-storage/src/factory.ts
```diff
@@ -46,7 +55,139 @@ class AmplifyStorageFactory
...
+  /**
+   * Constructs a new instance of the StorageValidator class.
+   */
+  constructor(
+    outputStorageStrategy: BackendOutputStorageStrategy<StorageOutput>
+  ) {
+    this.outputStorageStrategy = outputStorageStrategy;
+  }
+  /**
+   * Visit method to perform validation on the given node.
```
sobolk, line 96: please update comments to reflect new behaviors in this class.

packages/backend-storage/src/factory.ts
```diff
@@ -46,7 +55,139 @@ class AmplifyStorageFactory
...
+    );
+    const storageCount = storageInstances.length;
+    const firstStorage = Stack.of(node).node.children.filter(
+      (el) => el instanceof AmplifyStorage
+    )[0];
+    if (node !== firstStorage) {
+      return;
+    }
+    this.node = node;
+    let hasDefault = false;
```
sobolk, line 114:
```suggestion
    let defaultStorageFound = false;
```

packages/backend-storage/src/factory.ts
```diff
@@ -46,7 +55,139 @@ class AmplifyStorageFactory
...
+
+    Stack.of(node).node.children.forEach((child) => {
+      if (!(child instanceof AmplifyStorage)) {
+        return;
+      }
+      if (child.isDefault && !hasDefault) {
+        hasDefault = true;
+      } else if (child.isDefault && hasDefault) {
+        throw new AmplifyUserError('MultipleDefaultBucketError', {
```
sobolk, line 123:
```suggestion
        throw new AmplifyUserError('MultipleDefaultStoragesError', {
```

packages/backend-storage/src/factory.ts
```diff
@@ -46,7 +55,139 @@ class AmplifyStorageFactory
...
+          message: `More than one default buckets set in the Amplify project.`,
```
sobolk, line 124:
```suggestion
          message: `More than one default storages set in the Amplify project.`,
```

packages/integration-tests/src/test-project-setup/test_project_base.ts
sobolk: These may include API_KEY.
We should not log them.

We may want to introduce a flag to have this locally, but we shouldn't do this in CI.

sobolk (commented):
Looks good overall.

packages/backend-storage/src/factory.ts
0618: That's a good catch. I need to investigate on that

- [ ]

packages/backend-storage/src/factory.ts
0618: Done.

packages/backend-storage/src/factory.ts
sobolk: Consider:

```
export class StorageOutputsAspect implements IAspect {
  ...
  private areStoragesProcessed = false;
  
  
  visit() {
  
      if (this.areStoragesProcessed) {
         return;
      }
      this.areStoragesProcessed = true;
      /// processing logic

```

0618: Good point. Updated [cb7cb4f](https://github.com/aws-amplify/amplify-backend/pull/1742/commits/cb7cb4f8a061cb0b018ca3de54e1ee568ddc2b52)

packages/backend-storage/src/factory.ts
0618: Yea, we should be able to use the parent stack

image

sobolk (commented):
LGTM.
Please make checks happy.

sobolk:
Could you change
https://github.com/aws-amplify/amplify-backend/blob/39175772301cd9db58fd07ff6ed3e2c0e20fb876/packages/integration-tests/src/test-project-setup/test_project_base.ts#L186

To use `isMatch` from https://lodash.com/docs/4.17.15#isMatch . Or something equivalent.
I.e. test shouldn't be sensitive to non-breaking additions.

sobolk (dismissed):
Please fix outputs test before merging. (i.e. make it ready for next time it fails that way)

packages/backend-storage/src/factory.ts
```diff
@@ -46,7 +55,143 @@ class AmplifyStorageFactory
...
+          child.isDefault,
+          child.name,
+          child.resources.bucket.bucketName
+        );
+      });
+    }
+  }
+
+  private storeOutput = (
+    outputStorageStrategy: BackendOutputStorageStrategy<StorageOutput> = new StackMetadataBackendOutputStorageStrategy(
```
Amplifiyer, line 167: Do we need this default initialization?

packages/backend-storage/src/factory.ts
```diff
@@ -46,7 +55,143 @@ class AmplifyStorageFactory
...
+    /**
+     * only store traverse the siblings once to store the outputs,
+     * otherwise there will be the an error -
+     * "there is already a construct with name 'storageRegion'"
+     */
+    if (this.isStorageProcessed) {
+      return;
+    }
+    this.isStorageProcessed = true;
+    this.node = node;
```
Amplifiyer, line 117: If this is just needed to find the region in `storeOutputs` just pass it in the method itself

packages/backend-storage/src/factory.ts
```diff
@@ -46,7 +55,143 @@ class AmplifyStorageFactory
...
+    const storageCount = storageInstances.length;
+
+    /**
+     * only store traverse the siblings once to store the outputs,
+     * otherwise there will be the an error -
+     * "there is already a construct with name 'storageRegion'"
+     */
+    if (this.isStorageProcessed) {
+      return;
+    }
```
Amplifiyer, line 115: move this in the beginning for early return

packages/backend-storage/src/factory.ts
```diff
@@ -46,7 +55,143 @@ class AmplifyStorageFactory
...
+    const storageInstances = Stack.of(node).node.children.filter(
+      (el) => el instanceof AmplifyStorage
+    );
+    const storageCount = storageInstances.length;
+
+    /**
+     * only store traverse the siblings once to store the outputs,
+     * otherwise there will be the an error -
+     * "there is already a construct with name 'storageRegion'"
+     */
```
Amplifiyer, line 112:
```suggestion
    /**
     * only traverse the siblings once to store the outputs,
     * storing the same outputs multiple times result in error
     */
```

Amplifiyer (dismissed):
LGTM, some minor comments

packages/client-config/src/client-config-contributor/client_config_contributor_v1.ts
```diff
@@ -258,10 +258,17 @@ export class StorageClientConfigContributor implements ClientConfigContributor {
...
+      storageOutput.payload.buckets ?? '[]'
+    );
     config.storage = {
       aws_region: storageOutput.payload.storageRegion,
       bucket_name: storageOutput.payload.bucketName,
+      buckets: bucketsStringArray.map((b: string) => ({
+        name: JSON.parse(b).name,
+        bucket_name: JSON.parse(b).bucketName,
+        aws_region: JSON.parse(b).storageRegion,
+      })),
```
edwardfoyle, line 271: we should validate that the parsed elements are actually the expected shape. We can also simplify a bit to only parse once:

```suggestion
      buckets: bucketsStringArray
        .map((bucketJsonString) => zodSchema.parse(JSON.parse(bucketJsonString))
        .map(({name, bucketName, storageRegion}) => ({
          name,
          bucket_name: bucketName,
          aws_region: storageRegion
        })
```

packages/backend-storage/src/storage_access_policy_factory.ts
```diff
@@ -63,9 +60,13 @@ export class StorageAccessPolicyFactory {
       });
     }
 
-    return new Policy(this.stack, `${this.namePrefix}${this.policyCount++}`, {
-      statements,
-    });
+    return new Policy(
+      this.stack,
+      `storageAccess${this.stack.node.children.length}`,
```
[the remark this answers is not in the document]
0618: Yep, that's right.

packages/client-config/src/client-config-contributor/client_config_contributor_v1.ts
```diff
@@ -12,7 +12,7 @@ import {
   clientConfigTypesV1,
 } from '../client-config-types/client_config.js';
 import { ModelIntrospectionSchemaAdapter } from '../model_introspection_schema_adapter.js';
-import { AwsAppsyncAuthorizationType } from '../client-config-schema/client_config_v1.js';
+import { AwsAppsyncAuthorizationType } from '../client-config-schema/client_config_v1.1.js';
```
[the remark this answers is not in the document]
0618: We are changing the schema version from v1 to v1.1 in this PR as well. 

The test file name client_config_contributor_v1.test.ts needs to be updated.

packages/backend-storage/src/factory.ts
0618: We also need it for the default `outputStorageStrategy` , i.e. `new StackMetadataBackendOutputStorageStrategy(
      Stack.of(this.node).nestedStackParent || Stack.of(this.node)
    )`

packages/backend-storage/src/factory.ts
0618: If it needed the default initialization when it was in the construct.ts, why don't it need it when it's moved here?

packages/backend-storage/src/factory.ts
Amplifiyer: because now it's private and all the call sites are passing a fully formed object.

packages/backend-storage/src/factory.ts
Amplifiyer: because now it's private and all the call sites are passing a fully formed object.

packages/backend-storage/src/factory.ts
```diff
@@ -46,7 +55,143 @@ class AmplifyStorageFactory
...
+    }
+
+    return amplifyStorage;
+  };
+}
+
+/**
+ * StorageValidator class implements the IAspect interface.
+ */
+export class StorageOutputsAspect implements IAspect {
```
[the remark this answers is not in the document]
0618: Done. 0940ed6856

packages/backend-storage/src/factory.ts
0618: Done!

packages/backend-storage/src/factory.ts
0618: Just checked. The only place we call it is here https://github.com/aws-amplify/amplify-backend/blob/9d28f1ac43e2bbbcbc9e0d2e2912e2094e9087c9/packages/backend-storage/src/storage_container_entry_generator.ts#L37, and it's not optional, so yea, we don't need the default value.

0618: Done 1a7f0950eb

packages/backend-storage/src/factory.ts
0618: Done 1a7f0950eb

packages/backend-storage/src/storage_outputs_aspect.ts
```diff
@@ -0,0 +1,111 @@
...
+        },
+      });
+    }
+
+    // both default and non-default buckets should have the name, bucket name, and storage region stored in `buckets` field
+    outputStorageStrategy.appendToBackendOutputList(storageOutputKey, {
+      version: '1',
+      payload: {
+        buckets: JSON.stringify({
+          node: node.name,
```
Amplifiyer, line 104: This also signifies no unit testing for this piece?
```suggestion
          name: node.name,
```

0618:
> [storage_outputs_aspect.ts](/aws-amplify/amplify-backend/pull/1742/files/9ade76aba59af26ae98a3ed13a4b1d848f2316bb..1a7f0950eb2c1a3cb26a3a8ca3721b7c19dce539#diff-6763b3a3f63802c541bcfd4e003dd5177a04fac159d3617a32cf53284733a3d4)

image

It's happy now 🎉 https://github.com/aws-amplify/amplify-backend/actions/runs/10276327564/job/28436722264?pr=1742

Amplifiyer (approved):

sobolk (approved):

PR #1785, opened 2024-07-25
Revert "pin api extractor"

0618:
Reverts aws-amplify/amplify-backend#1780

After the change, `ajv-draft-04` was removed from node_modules.

image

Pasted_Image_2024-07-25__09_32


sobolk (changes requested):
The check that re-generates api passed here https://github.com/aws-amplify/amplify-backend/actions/runs/10080970693/job/27872235290?pr=1780 .

Are you sure this isn't local problem ?

PR #1807, opened 2024-08-01
fix: error mapping regex for host specific newline characters

Amplifiyer:
## Problem

Current error mapping regexes were using hard coded line feeds `\n` which doesn't work on some OS.

## Changes

Parameterize those regex to include os specific line feed.
1. Change explicit `\n` to `os.EOL`
2. Change implicit new line to `os.EOL`  and simple line concatenation in tests.

## Checklist

### DIFF
--- a/packages/backend-deployer/src/cdk_error_mapper.ts
+++ b/packages/backend-deployer/src/cdk_error_mapper.ts
@@ -5,6 +5,7 @@ import {
   AmplifyUserError,
 } from '@aws-amplify/platform-core';
 import { BackendDeployerOutputFormatter } from './types.js';
+import { EOL } from 'os';
 
 /**
  * Transforms CDK error messages to human readable ones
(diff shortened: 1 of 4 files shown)


packages/backend-deployer/src/cdk_error_mapper.ts
```diff
@@ -109,7 +110,9 @@ export class CdkErrorMapper {
       classification: 'ERROR',
     },
     {
-      errorRegex: /(SyntaxError|ReferenceError|TypeError):((?:.|\n)*?at .*)/,
+      errorRegex: new RegExp(
+        `(SyntaxError|ReferenceError|TypeError):((?:.|${EOL})*?at .*)`
```
sobolk, line 114: an option to consider here https://stackoverflow.com/questions/1331815/regular-expression-to-match-cross-platform-newline-characters .

in case CDK starts to emit just `\n` on windows some day.

sobolk (approved):

rtpascual (approved):

packages/backend-deployer/src/cdk_error_mapper.ts
Amplifiyer: Fair point, though that would be a disaster of unimaginable proportion. I'd keep OS specific handling for now, if cdk does something unimaginable, worse we would see an increase in our telemetry at which point we can revisit this.

Amplifiyer: Looks like this is already a reality, CDK doesn't seem to generate OS specific line endings already 🤦‍♂️ 🥹 👊

sobolk: lol :-)

PR #1822, opened 2024-08-05
Add `ai-constructs` and default implementation of conversation handler

sobolk:
## Problem / Changes

This PR adds `ai-constructs` package which will host CDK components that support AI based use cases.

This PR also adds initial implementation of "conversation handler lambda". I.e.:
1. It adds construct which will be consumed by conversation directive transformer (and defineConversationHandler in the future).
2. Defines types for conversation turn event
3. Defines top level API to `handleConversationTurnEvent` (i.e. public facing facade to later allow building handlers with programmatically defined tools).
4. Adds e2e test.

Out of scope (in this PR):
1. Tools
2. `defineConversationHandler` (and `backend-ai` package)

## Validation

Added tests. 

## Checklist


packages/backend-ai/src/conversation/runtime/bedrock_converse_adapter.ts
```diff
@@ -0,0 +1,47 @@
...
+        maxTokens: 2000,
+        temperature: 0,
+      },
+    };
+    const bedrockResponse = await this.bedrockClient.send(
+      new ConverseCommand(converseCommandInput)
+    );
+    if (bedrockResponse.output?.message) {
+      messages.push(bedrockResponse.output?.message);
+    }
```
Amplifiyer, line 43: Is this for a subsequent `askBedRock` for the same turnevent?

packages/backend-ai/src/conversation/runtime/conversation_turn_response_sender.ts
```diff
@@ -0,0 +1,75 @@
...
+    private readonly _fetch = fetch
+  ) {}
+
+  sendResponse = async (message: ContentBlock[]) => {
+    const request = this.createMutationRequest(message);
+    const res = await this._fetch(request);
+    if (!res.ok) {
+      const body = await res.text();
+      throw new Error(
+        `Assistant response mutation request was not successful response headers=${JSON.stringify(
```
Amplifiyer, line 30: nit 
```suggestion
        `Assistant response mutation request was not successful, response headers=${JSON.stringify(
```

packages/backend-ai/src/conversation/runtime/conversation_turn_response_sender.ts
```diff
@@ -0,0 +1,75 @@
...
+   * TODO docs
+   */
+  constructor(
+    private readonly event: ConversationTurnEvent,
+    private readonly _fetch = fetch
+  ) {}
+
+  sendResponse = async (message: ContentBlock[]) => {
+    const request = this.createMutationRequest(message);
+    const res = await this._fetch(request);
```
Amplifiyer, line 26: Shouldn't there be a try/catch as well since fetch can throw?

packages/backend-ai/API.md
```diff
@@ -0,0 +1,87 @@
...
+        systemPrompt: string;
+        region?: string;
+    };
+    request: {
+        headers: {
+            authorization: string;
+        };
+    };
+    messages: Array<ConversationMessage>;
+};
```
Amplifiyer, line 71: I'm guessing this is defined by AppSync? How will we manage changes in this type?

Amplifiyer (commented):
Looks good overall. Should add unit tests as well.

packages/backend-ai/src/conversation/runtime/bedrock_converse_adapter.ts
sobolk: yeah, this is a remnant from transplanting poc code. 
I.e. yes, it will be needed in subsequent PRs when we bring tools. But in this one is not needed, I'll remove this for now.

packages/backend-ai/API.md
sobolk: AppSync will be sending these events.

However, conversation-transformer that will deploy lambda in default scenario is going to see this type (through dependency on new construct) and can use this definition to deploy correct event producer.
In addition.  conversation-transformer and handler are going to be part of same deps graph and deploy together, so risk of them going out of sync is zero (or near zero).

@atierian please correct me if I'm wrong here.

packages/backend-ai/src/conversation/runtime/conversation_turn_response_sender.ts
sobolk: I refrained from aggressively catching exceptions, following this philosophy.  https://softwareengineering.stackexchange.com/questions/365427/try-catch-log-rethrow-is-anti-pattern .

From practical point of view (that I observed when working on this)
1. When `fetch` throws:
   1. Exception bubbles up all the way to lambda runtime
   2. Exception is logged into CW as is.
   3. Lambda metrics show failed executions.
2. Statements below were added to catch "silent failures":
   1. When request comes back with non-2xx status
   2. When GQL error is in payload with 2xx status (this is horrible btw...)
   3. In both cases fetch was happy and there was no trace of failure.

packages/backend-ai/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@aws-amplify/backend-ai",
```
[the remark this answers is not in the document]
sobolk: We were (still are?) struggling with vision around packaging "ai features".

Couple of thoughts:
1. The coming "ai features" seem to be independent islands of functionality. I.e. `defineAi`, `ai-construct` (singular) kind of abstraction don't seem to be correct. E.g. conversational bot is different set of components than generating images from description. These seem to be clusters of functionality under single "theme" (ai).
2. Because of point 1. If we were to follow `auth` and `data` we'll end up with multitude of packages like `ai-conversation-handler-construct`, `ai-image-generation-construct` etc. This seems a bit too aggressive and fragmented approach. Especially that we're introducing lower level public components like "conversation handler" (and there doesn't seem to be alternative for that).
3. The reason why `data` and `auth` constructs are separated is that there was hard requirement for them to be usable in raw CDK. On the other hand test I wrote seems to indicate that data + conversation construct could be used in raw CDK. (and a good signal to reconsider dumping everything to `backend-ai`).
4. While prototyping it felt odd that "construct package" would host 99% of the code now and "backend-ai" would be rather a facade. But, having that split eagerly might not be a bad idea to show next set of features a trajectory to follow.
5. The "conversation handler" construct is exposed publicly, so that both conversation transformer AND defineConversationHandler can see and use it. There didn't seem to be other need for construct to be public. But then `defineConversationHandler` translates to `new Construct` in raw CDK world. But we don't have strict requirement for raw CDK, so conversation construct being public is a necessity rather than desire right now.


I think we should take your suggestion, with small modification. I.e. do this:
1. Introduce a commonjs package called `ai-constructs` (plural!) . That will host any raw CDK components we need for ai "theme". (Today it's gonna be construct + lambda runtime components)
2. `backend-ai` to adapt raw CDK primitives to "Gen2 backend style". I.e. today it'll have `defineConversationHandler + re-exports of lambda runtime apis).

Additional supporting thought for such split is that data construct depending on ai-constructs seems to smell much better than data construct depending on backend-ai.

@Amplifiyer @atierian @josefaidt @dbanksdesign any objections to this ?

packages/backend-ai/API.md
atierian: Correct.

> can use this definition to deploy correct event producer.

We're not currently doing this in the transformer, but this sounds like a good idea. Made a note to explore this.

packages/backend-ai/src/conversation/runtime/conversation_turn_response_sender.ts
Amplifiyer: I agree with just re-throwing is bad, but if a better error message can be included I think it's worth it. Since this is a data plane call and going through AppSync, what will customers see when this fails?

sobolk: > if a better error message can be included I think it's worth it

We can do that. But right place for that is going to be in top level component. I.e. add try-catch to this top level messaging https://github.com/aws-amplify/amplify-backend/blob/ae3fbbafbc8e454acddafd990f3057b750ca30fc/packages/backend-ai/src/conversation/runtime/conversation_turn_executor.ts#L27-L30 . I.e explicitly print that conversation handler failed. (following this https://stackoverflow.com/questions/13042975/exception-handling-practices-top-level-vs-every-function )

> Since this is a data plane call and going through AppSync, what will customers see when this fails?

This is a questionable part. As lambda invocation is asynchronous from appsync pov (fire and forget). I.e. frontend will see bot going silent.
This is a tradeoff made in upstream (appsync part) design and there doesn't seem to be 100% good solution visible right now.

packages/backend-ai/package.json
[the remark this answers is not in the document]
sobolk: Discussed offline. We'll create `ai-constructs` package to host CDK stuff.

packages/ai-constructs/src/conversation/conversation_handler_construct.ts
```diff
@@ -0,0 +1,104 @@
...
+  constructor(
+    scope: Construct,
+    id: string,
+    private readonly props: ConversationHandlerFunctionProps
+  ) {
+    super(scope, id);
+
+    if (this.props.entry && !path.isAbsolute(this.props.entry)) {
+      throw new Error('Entry must be absolute path');
+    }
```
sobolk, line 54: I added this restriction for now. We need some smarter resolution logic which is similar to what `defineFunction` and `NodeJSFunction` construct is doing. But I didn't want to inflate this PR with that.
We'll address this later (when `defineConversationHandler`) happens.

Amplifiyer (approved):

packages/ai-constructs/src/conversation/conversation_handler_construct.ts
```diff
@@ -0,0 +1,104 @@
...
+            ],
+          }),
+        }),
+      }
+    );
+
+    if (this.props.models && this.props.models.length > 0) {
+      const resources = this.props.models.map(
+        (model) =>
+          `arn:aws:bedrock:${model.region}::foundation-model/${model.modelId}`
```
rtpascual, line 84: What happens if a customer's account is in one region and they use a model in a different region? Or is that not possible/not a concern?

packages/ai-constructs/src/conversation/conversation_handler_construct.ts
```diff
@@ -0,0 +1,104 @@
...
+    if (this.props.entry && !path.isAbsolute(this.props.entry)) {
+      throw new Error('Entry must be absolute path');
+    }
+
+    const conversationHandler = new NodejsFunction(
+      this,
+      `conversationHandlerFunction`,
+      {
+        runtime: LambdaRuntime.NODEJS_18_X,
+        timeout: Duration.seconds(60),
```
rtpascual, line 61: Do we have plans in the future to allow customers to pass in runtime and timeout like with `defineFunction`?

packages/ai-constructs/src/conversation/conversation_handler_construct.ts
sobolk: > if a customer's account

I guess you mean main stack resources. Account has access to many regions.

> they use a model in a different region?

That is perfectly legal. It just ends up being cross region call which comes with extra latency. The API is created in such a way that they have to opt in to do this and upstream layers (data construct?) will do some defaulting.
It is possible to allow usage of same model in multiple regions, one just adds another entry with same modelid and different region.
The reason to expose region at all as an option is that number of models available per region varies and is quite constrained comparing to other resources we provision (like cognito).

packages/ai-constructs/src/conversation/conversation_handler_construct.ts
sobolk: Timeout is likely going to be surfaced at some point. These transactions have inherently quite high latency. (hence different default here than `defineFunction`.
Runtime less likely - we are not planing to ship runtime components for other runtimes than Node.

rtpascual (approved):

packages/ai-constructs/src/conversation/conversation_handler_construct.ts
rtpascual: Not in scope of this PR but maybe we can add a warning that there will be extra latency. Might be confusing if you switch models to one in another region and all of a sudden your conversation handler consistently times out.

sobolk: That latency penalty is not that high comparing to LLM execution time.

Take a look at this post https://repost.aws/questions/QUl7IRSjpMQVe2mMOFojP6qA/a-question-about-inter-region-latency to get a raw estimate of how impactful this can be. I.e. in case of S3 1MB transfers it seem to range between 81ms and 755ms depending on regions selections. Which is still order of magnitude less than LLM execution alone.

I'd say we should hold with such a warning as it's not something magical we'd be doing. Customer explicitly specifies that, so the fact that cross region communication may be costly should be rather obvious.

packages/integration-tests/src/test-projects/conversation-handler/amplify/backend.ts
```diff
@@ -0,0 +1,17 @@
+import { defineBackend } from '@aws-amplify/backend';
+import { ConversationHandlerFunction } from '@aws-amplify/ai-constructs/conversation';
+import { auth } from './auth/resource.js';
+import { data } from './data/resource.js';
+
+const backend = defineBackend({ auth, data });
+
+const stack = backend.createStack('conversationHandlerStack');
+
+new ConversationHandlerFunction(stack, 'conversationHandlerFunction', {
```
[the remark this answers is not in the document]
sobolk: In real life it will be - to appsync resolver.

in the test, we're mocking the upstream appsync component so that we can (kind of) e2e test this in isolation while other components are still built.
I.e.
The test sends event instead of appsync resolver [here](https://github.com/aws-amplify/amplify-backend/blob/cf7778fa3877f3650743b793283b3f4cae362407/packages/integration-tests/src/test-project-setup/conversation_handler_project.ts#L145-L176) and we assert that response was published [here](https://github.com/aws-amplify/amplify-backend/blob/cf7778fa3877f3650743b793283b3f4cae362407/packages/integration-tests/src/test-project-setup/conversation_handler_project.ts#L220-L224)


This might get replaced with simpler approach later when other components (related to schema builder changes) mature. But for now this is what can be done.

packages/ai-constructs/src/conversation/conversation_handler_construct.ts
rtpascual: Got it, wasn't sure if the extra latency would be substantial or not compared to the total execution time.

awsluja (approved):

PR #1830, opened 2024-08-07
Add Bedrock tools to conversation handler

sobolk:
## Problem/Changes

This PR builds on top of https://github.com/aws-amplify/amplify-backend/pull/1822 and adds:
1. Executable tool abstraction
2. Ability to define programmatic tools in custom lambda handler
3. Ability to use "data/graphql tools" definitions from incoming event to query graphql.
4. Enhancement of bedrock converse adapter to handle bedrock's requests to use tools.

For what Bedrock tool is and how it works please refer to:
1. https://docs.aws.amazon.com/bedrock/latest/userguide/tool-use-examples.html
2. https://docs.aws.amazon.com/bedrock/latest/userguide/tool-use.html

## Out of scope

1. I attempted to introduce tool invocation limit, but I ran into edge case when bedrock would throw validation error if tools are removed from request after initial usage. This needs further toughts.
2. The logic mapping event tool definition to grapqh query may require further refinement. I took it as is from prototype and just split into smaller classes. I did create test harness, so this should be easier in separate PRs.
3. "UI tools" are not included here, will be added later.

## Validation

Added unit tests.
Added e2e tests.

## Checklist


packages/ai-constructs/src/conversation/runtime/bedrock_converse_adapter.ts
```diff
@@ -3,43 +3,156 @@ import {
...
-  ) {}
+    ),
+    eventToolsProvider = new ConversationTurnEventToolsProvider(event)
+  ) {
+    this.allTools = [...eventToolsProvider.getEventTools(), ...additionalTools];
+    this.allTools.forEach((t) => {
+      if (this.toolByName.has(t.name)) {
+        throw new Error(
+          `Tools must have unique names. Duplicate tools with '${t.name}' name detected.`
+        );
```
rtpascual, line 38: Nit: I think it would be a better experience to collect all duplicate names and then throw

rtpascual (dismissed):

packages/ai-constructs/src/conversation/runtime/bedrock_converse_adapter.ts
sobolk: Done.

packages/ai-constructs/src/conversation/runtime/types.ts
```diff
@@ -24,4 +30,23 @@ export type ConversationTurnEvent = {
     };
   };
   messages: Array<ConversationMessage>;
+  toolsConfiguration?: {
+    tools: Array<{
```
awsluja, line 34: nit: complex array type should be defined like above in line 32

packages/ai-constructs/src/conversation/runtime/types.ts
```diff
@@ -24,4 +30,23 @@ export type ConversationTurnEvent = {
...
+      graphqlRequestInputDescriptor: {
+        queryName: string;
+        selectionSet: string[];
+        propertyTypes: Record<string, string>;
+      };
+    }>;
+  };
+};
+
+export type ExecutableTool = {
```
awsluja, line 47: ExecutableTool & Tool array above seems to share common interface { name, description, input schema }, should we define this shared interface?

packages/integration-tests/src/test-project-setup/conversation_handler_project.ts
```diff
@@ -158,7 +210,130 @@ class ConversationHandlerTestProject extends TestProjectBase {
...
+        },
+      ],
+      request: {
+        headers: { authorization: accessToken },
+      },
+      responseMutationInputTypeName:
+        'CreateConversationMessageAssistantResponseInput',
+      responseMutationName: 'createConversationMessageAssistantResponse',
+      modelConfiguration: {
+        modelId: 'anthropic.claude-3-haiku-20240307-v1:0',
```
awsluja, line 266: are these magic ids?

packages/integration-tests/src/test-projects/conversation-handler/amplify/backend.ts
```diff
@@ -2,12 +2,28 @@ import { defineBackend } from '@aws-amplify/backend';
...
 
 const backend = defineBackend({ auth, data });
 
 const stack = backend.createStack('conversationHandlerStack');
 
-new ConversationHandlerFunction(stack, 'conversationHandlerFunction', {
+new ConversationHandlerFunction(stack, 'defaultConversationHandlerFunction', {
+  models: [
+    {
+      modelId: 'anthropic.claude-3-haiku-20240307-v1:0',
```
awsluja, line 15: if we are reusing these IDs lets pull them out

packages/integration-tests/src/test-project-setup/conversation_handler_project.ts
```diff
@@ -158,7 +210,130 @@ class ConversationHandlerTestProject extends TestProjectBase {
...
       ],
       request: {
-        headers: { authorization: authenticatedUserCredentials.accessToken },
+        headers: { authorization: accessToken },
+      },
+      responseMutationInputTypeName:
+        'CreateConversationMessageAssistantResponseInput',
+      responseMutationName: 'createConversationMessageAssistantResponse',
+      modelConfiguration: {
+        modelId: 'anthropic.claude-3-haiku-20240307-v1:0',
```
awsluja, line 219: another one

packages/integration-tests/src/test-project-setup/conversation_handler_project.ts
```diff
@@ -158,7 +210,130 @@ class ConversationHandlerTestProject extends TestProjectBase {
...
+        ],
+      },
+    };
+    const response = await this.executeConversationTurn(
+      event,
+      defaultConversationHandlerFunction,
+      apolloClient
+    );
+    // Assert that tool was used.
+    assert.match(response.content, /85/);
```
awsluja, line 303: 85 is the expected temperature?

packages/integration-tests/src/test-project-setup/conversation_handler_project.ts
```diff
@@ -168,37 +343,32 @@ class ConversationHandlerTestProject extends TestProjectBase {
         systemPrompt: 'You are helpful bot.',
       },
     };
+    const response = await this.executeConversationTurn(
+      event,
+      customConversationHandlerFunction,
+      apolloClient
+    );
+    // Assert that tool was used.
+    assert.match(response.content, /75/);
```
awsluja, line 352: why 75 here?

packages/ai-constructs/src/conversation/runtime/types.ts
sobolk: I refrained from that because we don't expect customer to create this and we'd rather not eagerly add public types that are useless to customer and pollute intellisense. I.e. we'd rather have `API.md` less verbose than absolutely necessary.

fwiw for internal consumption I did extract this type here https://github.com/aws-amplify/amplify-backend/blob/c04decbe0e8067bd2b800ecd1235c52aa886334e/packages/ai-constructs/src/conversation/runtime/event-tools-provider/types.ts#L3-L5 .

packages/ai-constructs/src/conversation/runtime/types.ts
sobolk: Regarding this and https://github.com/aws-amplify/amplify-backend/pull/1830#discussion_r1712019497.

I have intentionally refrained from aggressively extracting named types in public api in order to not pollute IDE syntax highlighting. 

The reasoning behind this is that customer is not expected to create events themselves, yet these events are inputs  to lambda and one of public APIs so they have to have public type.


`ExecutableTool` and `Tool array` do share that commonality, but for the sake of simple `API.md` I did not seek common abstraction.
I was experimenting with `ToolSpecification` from bedrock as common type here https://github.com/aws-amplify/amplify-backend/pull/1830/commits/436a0a2a557bf444db231a5539130baff8ba321a#diff-a6be72429e1e423a62ed353b71d2e55bf9563168f5bb3e9f1cecb6be1eb9d2e2 , but it came with `string | undefined` typings which seemed unwanted for now.

packages/integration-tests/src/test-project-setup/conversation_handler_project.ts
awsluja: should we also match the unit here? for unit in the schema 'F'

packages/integration-tests/src/test-project-setup/conversation_handler_project.ts
sobolk: They're from https://docs.aws.amazon.com/bedrock/latest/userguide/model-ids.html .

packages/integration-tests/src/test-projects/conversation-handler/amplify/backend.ts
sobolk: we can do that.

packages/integration-tests/src/test-project-setup/conversation_handler_project.ts
sobolk: correct.

packages/integration-tests/src/test-project-setup/conversation_handler_project.ts
sobolk: It's separate scenario, so I wanted to use different value to make sure that they don't accidentally cross wire.

packages/integration-tests/src/test-project-setup/conversation_handler_project.ts
sobolk: > should we also match the unit here? for unit in the schema 'F'

I was trying to but LLM converted it to some sentence with `Fahrenheit` .
The point of this test is to assure that tool was ran and ran successfully. The numerical value seems to be good enough signal that this has happened. But given how LLM work too strict assertion here will be problematic.

packages/integration-tests/src/test-project-setup/conversation_handler_project.ts
```diff
@@ -158,7 +210,130 @@ class ConversationHandlerTestProject extends TestProjectBase {
...
+      modelConfiguration: {
+        modelId: 'anthropic.claude-3-haiku-20240307-v1:0',
+        systemPrompt: 'You are helpful bot.',
+      },
+    };
+    await this.executeConversationTurn(
+      event,
+      defaultConversationHandlerFunction,
+      apolloClient
+    );
```
awsluja, line 227: we don't make assertions on the response here, is this because we are just passing the request to bedrock (which might produce inconsistent result?)

packages/integration-tests/src/test-project-setup/conversation_handler_project.ts
awsluja: Nvm i see.. we are just matching the response json?

packages/integration-tests/src/test-projects/conversation-handler/amplify/custom_handler.ts
```diff
@@ -0,0 +1,27 @@
...
+} from '@aws-amplify/ai-constructs/conversation/runtime';
+import { ToolResultContentBlock } from '@aws-sdk/client-bedrock-runtime';
+
+const thermometer: ExecutableTool = {
+  name: 'thermometer',
+  description: 'Returns current temperature in Seattle',
+  execute: (): Promise<ToolResultContentBlock> => {
+    return Promise.resolve({
+      // We use this value in test assertion.
+      text: '75F',
```
awsluja, line 14: does this need to be the same shape as the defined schema?

awsluja: nvm i see we aren't using custom handler in that schema

packages/integration-tests/src/test-project-setup/conversation_handler_project.ts
sobolk: Correct.
There is a basic assertion inside `this.executeConversationTurn` which checks that some response was emitted. But we do not inspect the answer as it might not be very stable. This could perhaps be changed if we use more concrete question. I'll give it a shot.

packages/integration-tests/src/test-project-setup/conversation_handler_project.ts
sobolk: correct, and we use mocked appsync schema to capture the response so `response.content` contains serialized json. at least for now while other components (schema builder) are in works (but we may choose to retain mocked schema to be independent from these components).

the goal of e2e test at this point is to assert that roundtrip data exchange happened and details of emitted. responses are captured in unit tests mostly.

packages/integration-tests/src/test-projects/conversation-handler/amplify/custom_handler.ts
sobolk: > does this need to be the same shape as the defined schema?

No, the input schema is something we give to LLM to let it know what is the shape of json that tool expects. (we do not test that particular aspect well, but I expect to make this up later when these components mature more).
The output from tool can be anything really. LLM is handling it internally somehow.

packages/ai-constructs/src/conversation/runtime/types.ts
awsluja: I see

packages/ai-constructs/src/conversation/runtime/types.ts
awsluja: Got it, how are we tracking _intentionally refraining from_ exposing this type? should we leave a note if a type should not be refactored for that reason?

sobolk: > how are we tracking intentionally refraining from exposing this type?

We don't. Other than relying on natural instinct to doge naming types we can avoid naming (public api needs good names).

> should we leave a note if a type should not be refactored for that reason?

I will leave a note. Btw. this might change later if there's legitimate need to name more parts of the event, but this is going to be on demand activity later.

packages/integration-tests/src/test-project-setup/conversation_handler_project.ts
awsluja: Oh, that is interesting. So we are asserting on what can be guaranteed, i'll keep that in mind; is the LLM always going to respond with numeric 75 - versus seventy-five ?

sobolk: > is the LLM always going to respond with numeric 75 - versus seventy-five ?

Hard to say. But for now it seems to be best bet. Hopefully behavior is somewhat fixed by hardcoding modelid (it is versioned). However, while playing with other questions I did see some variations of answers.

packages/integration-tests/src/test-project-setup/conversation_handler_project.ts
sobolk: I found something that should be stable.

packages/ai-constructs/src/conversation/runtime/types.ts
sobolk: added note

packages/integration-tests/src/test-projects/conversation-handler/amplify/backend.ts
sobolk: done.

rtpascual (approved):
Nice using value of pi for a more consistent response!

awsluja (approved):

Issue #1549, opened 2024-05-23
RFC: Use existing Lambda Layers with `referenceFunctionLayer`


josefaidt:
Hey folks :wave: we’re looking to introduce a mechanism to bind and consume existing Lambda Layers with Functions defined with `defineFunction`. This will not facilitate the creation and management of Lambda Layers, however provides you with a way to consume Node.js packages that require OS-specific addons without requiring a dependency on Docker to build Functions locally. This is a follow-up to the following issues:

* https://github.com/aws-amplify/amplify-backend/issues/1281
* https://github.com/aws-amplify/amplify-backend/issues/1432

The primary use case we’re seeking to address with this feature is the ability to use dependencies like `sharp` that require OS-specific addons to function as expected in the Node.js Lambda runtime environment, which requires linux addons. When building Functions locally (i.e. on deploy), Amplify uses [esbuild](https://esbuild.github.io/), which will bundle modules that are imported in your source code. With layers, you need to prevent esbuild from bundling your layer code. Using CDK it is possible to force a Function build using Docker, however we would like to avoid requiring this dependency.

In addition to the bundler complications, you also need a way to provide a similar type-safe experience that you get when authoring your Function’s handler. For example, if you’re using a layer that provides access to specific npm packages like `sharp`, you should expect to see the same APIs and type-safety you’d see when installing and using the package normally. This is typically accomplished by marking certain dependencies as “external”.

```ts
import { defineFunction, referenceFunctionLayer } from "@aws-amplify/backend"

defineFunction({
  name: "my-function-with-layers",
  layers: {
    // keyed by module name
    sharp: referenceFunctionLayer("arn:.../my-sharp-layer"),
  },
})
```

The introduction of `referenceFunctionLayer` comes with an added `layers` prop on `defineFunction` . This object is keyed by the module name hosted on your existing layer. In the example above, we’re binding a layer that simply hosts the `sharp` dependency to the `sharp` key. When esbuild executes, `sharp` will be externalized and available via your layer at runtime.

For layers hosting multiple dependencies, you can create multiple key/value pairs with the same layer reference:

```ts
import { defineFunction, referenceFunctionLayer } from "@aws-amplify/backend"

defineFunction({
  name: "my-function-with-layers",
  layers: {
    // keyed by module name
    sharp: referenceFunctionLayer("arn:.../my-sharp-layer"),
    "some-other-dependency": referenceFunctionLayer("arn:.../my-sharp-layer"),
  },
})
```

[...]

PR #1835, opened 2024-08-08
adds layers to function

ykethan:
## Problem

RFC: https://github.com/aws-amplify/amplify-backend/issues/1549

## Changes

adds ability to reference Lambda layers and attach to a function. 
Can attach up to 5 layers on a function. Auto sets the layer keys as external modules on the function.

example:

resource.ts
```
import { defineFunction } from "@aws-amplify/backend";

export const sayHello = defineFunction({
  name: "say-hello1",
  entry: "./handler.ts",
  layers: {
   "@aws-lambda-powertools/logger":
      "arn:aws:lambda:us-east-1:094274105915:layer:AWSLambdaPowertoolsTypeScriptV2:11",
  },
});

```

handler.ts. keep using your dependancy without having to use `/opt` as the `NODE_PATH` would point to `/opt`: https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime

```
import { Logger } from "@aws-lambda-powertools/logger";

const logger = new Logger({ serviceName: "serverlessAirline" });

export const handler = async (): Promise<void> => {
  logger.info("Hello World");
};

```

## Checklist


packages/backend-function/src/factory.ts
```diff
@@ -121,6 +123,20 @@ export type FunctionProps = {
...
+   * import { referenceFunctionLayer } from "@aws-amplify/backend";
+   *
+   * layers: {
+   *    myModule: referenceFunctionLayer("arn:aws:lambda:<current-region>:<account-id>:layer:myLayer:1")
+   * },
+   *
+   * The object is keyed by the module name hosted on your existing layer and a value that references to an existing layer using an ARN. The keys will be externalized and available via your layer at runtime
+   * Maximum of 5 layers can be attached to a function. Layers must be in the same region as the function.
+   */
+  layers?: FunctionLayerReferences;
```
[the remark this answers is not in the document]
ykethan: this was built according to the RFC requirements
https://github.com/aws-amplify/amplify-backend/issues/1549

packages/backend-function/src/factory.ts
```diff
@@ -121,6 +123,20 @@ export type FunctionProps = {
...
   schedule?: FunctionSchedule | FunctionSchedule[];
+
+  /**
+   * Attach Lambda layers to a function
+   * @example
+   * import { referenceFunctionLayer } from "@aws-amplify/backend";
+   *
+   * layers: {
+   *    myModule: referenceFunctionLayer("arn:aws:lambda:<current-region>:<account-id>:layer:myLayer:1")
+   * },
```
josefaidt, line 134: does this render okay? I remember some wonkiness with multiline examples and spacing. Also can we use the powertools example?

packages/backend-function/src/factory.ts
```diff
@@ -121,6 +123,20 @@ export type FunctionProps = {
    * schedule: "0 9 * * 2" // every Monday at 9am
    */
   schedule?: FunctionSchedule | FunctionSchedule[];
+
+  /**
+   * Attach Lambda layers to a function
```
josefaidt, line 128: can we add a `@see` with a stubbed docs link?

packages/backend-function/src/factory.ts
```diff
@@ -121,6 +123,20 @@ export type FunctionProps = {
...
+   * @example
+   * import { referenceFunctionLayer } from "@aws-amplify/backend";
+   *
+   * layers: {
+   *    myModule: referenceFunctionLayer("arn:aws:lambda:<current-region>:<account-id>:layer:myLayer:1")
+   * },
+   *
+   * The object is keyed by the module name hosted on your existing layer and a value that references to an existing layer using an ARN. The keys will be externalized and available via your layer at runtime
+   * Maximum of 5 layers can be attached to a function. Layers must be in the same region as the function.
```
josefaidt, line 137: can we split this into two `@note`'s?

packages/backend-function/src/factory.ts
ykethan: image

it appears to render it correctly

packages/backend-function/src/factory.ts
```diff
@@ -121,6 +123,18 @@ export type FunctionProps = {
    * schedule: "0 9 * * 2" // every Monday at 9am
    */
   schedule?: FunctionSchedule | FunctionSchedule[];
+
+  /**
+   * Attach Lambda layers to a function
+   * @see [Lambda layers](https://docs.aws.amazon.com/lambda/latest/dg/chapter-layers.html)
```
josefaidt, line 129:
```suggestion
   * @see https://docs.amplify.aws/react/build-a-backend/functions/lambda-layers
   * @see [AWS documentation for Lambda layers](https://docs.aws.amazon.com/lambda/latest/dg/chapter-layers.html)
```

adds a placeholder for a new docs page in the amplify docs and makes the anchor text a more descriptive in it describes where we're navigating to

packages/backend-function/src/factory.ts
```diff
@@ -121,6 +123,18 @@ export type FunctionProps = {
...
+   * @example
+   * layers: {
+   *    "@aws-lambda-powertools/logger": "arn:aws:lambda:<current-region>:094274105915:layer:AWSLambdaPowertoolsTypeScriptV2:11"
+   * },
+   * @description The object is keyed by the module name hosted on your existing layer and a value that references to an existing layer using an ARN. The keys will be externalized and available via your layer at runtime
+   * @description Maximum of 5 layers can be attached to a function and must be in the same region as the function.
```
josefaidt, line 135: `@description` is typically used to describe a symbol/type, is there another we could use? 

https://jsdoc.app/
https://tsdoc.org/

ykethan: switched to bullet points as the only other tags i could use are 
[@remarks](https://tsdoc.org/pages/tags/remarks/)
[@summary](https://jsdoc.app/tags-summary)

packages/backend-function/src/factory.ts
ykethan: updated both with anchor text

packages/backend-function/src/factory.ts
```diff
@@ -265,7 +281,9 @@ class FunctionFactory implements ConstructFactory<AmplifyFunction> {
   };
 }
 
-type HydratedFunctionProps = Required<FunctionProps>;
+type HydratedFunctionProps = Required<FunctionProps> & {
+  layers: FunctionLayerReferences;
+};
```
Amplifiyer, line 286: why do we need this? Wouldn't `layers` be automatically be present and required in `HydratedFunctionProps`

packages/backend-function/src/factory.ts
```diff
@@ -279,10 +297,15 @@ class FunctionGenerator implements ConstructContainerEntryGenerator {
     scope,
     backendSecretResolver,
   }: GenerateContainerEntryProps) => {
+    const resolvedLayers = resolveLayers(
+      this.props.layers,
+      scope,
+      this.props.name
+    );
```
Amplifiyer, line 304: should this be part of `hydrateDefaults`?

packages/backend-function/src/reference_layer.ts
```diff
@@ -0,0 +1,92 @@
...
+export const arnPattern = new RegExp(
+  '^(arn:(aws[a-zA-Z-]*)?:lambda:[a-z]{2}((-gov)|(-iso([a-z]?)))?-[a-z]+-\\d{1}:\\d{12}:layer:[a-zA-Z0-9-_]+:[0-9]+)|(arn:[a-zA-Z0-9-]+:lambda:::awslayer:[a-zA-Z0-9-_]+)$'
+);
+
+/**
+ * Checks if the provided ARN is valid.
+ */
+export const isValidLayerArn = (arn: string): boolean => {
+  return arnPattern.test(arn);
+};
```
Amplifiyer, line 18: do we need these methods exported or used outside of this file? If not, can we bring them into the class as privates.

packages/backend-function/src/reference_layer.ts
```diff
@@ -0,0 +1,92 @@
...
+import { ILayerVersion, LayerVersion } from 'aws-cdk-lib/aws-lambda';
+import { Arn } from 'aws-cdk-lib/core';
+import { Construct } from 'constructs';
+
+/**
+ * Defines the ARN regex pattern for a layer.
+ */
+export const arnPattern = new RegExp(
+  '^(arn:(aws[a-zA-Z-]*)?:lambda:[a-z]{2}((-gov)|(-iso([a-z]?)))?-[a-z]+-\\d{1}:\\d{12}:layer:[a-zA-Z0-9-_]+:[0-9]+)|(arn:[a-zA-Z0-9-]+:lambda:::awslayer:[a-zA-Z0-9-_]+)$'
+);
```
Amplifiyer, line 11: Can you add the source of this ARN in the documentation.

packages/backend-function/src/reference_layer.ts
```diff
@@ -0,0 +1,92 @@
...
+  if (uniqueArns.size > 5) {
+    throw new AmplifyUserError('MaximumLayersReachedError', {
+      message: 'A maximum of 5 unique layers can be attached to a function.',
+      resolution: 'Remove unused layers in your function',
+    });
+  }
+
+  return new Set<FunctionLayerArn>(
+    Array.from(uniqueArns).map(parseFunctionLayerArn)
+  );
```
Amplifiyer, line 72: If the method is called `validate` it shouldn't do anything else but validate what it's supposed to do. This parsing/mapping shouldn't be part of it.

packages/backend-function/src/reference_layer.ts
```diff
@@ -0,0 +1,92 @@
...
+ * Checks if the provided ARN is valid.
+ */
+export const isValidLayerArn = (arn: string): boolean => {
+  return arnPattern.test(arn);
+};
+
+/**
+ * Class to represent and handle Lambda Layer ARNs.
+ */
+export class FunctionLayerArn {
```
Amplifiyer, line 23: Why do we need a class to represent the arn? It seems like it's just doing validation.

Recommendation: Create a class `FunctionLayerParser`/`FunctionLayerReferenceAdapter` (or something like that) with one public method `parse`/`adapt` that takes in the user input, i.e. `Record<String, String>` and does several validations and finally produces `ILayerVersion[]`

Amplifiyer (changes requested):

packages/backend-function/src/factory.ts
```diff
@@ -121,6 +123,19 @@ export type FunctionProps = {
    * schedule: "0 9 * * 2" // every Monday at 9am
    */
   schedule?: FunctionSchedule | FunctionSchedule[];
+
+  /**
+   * Attach Lambda layers to a function
+   * - The object is keyed by the module name hosted on your existing layer and a value that references to an existing layer using an ARN. The keys will be externalized and available via your layer at runtime
```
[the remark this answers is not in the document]
ykethan: updated the jsdoc to be more verbose

packages/backend-function/src/factory.ts
ykethan: removed this

packages/backend-function/src/reference_layer.ts
ykethan: separated the logic and added them as private.

packages/backend-function/src/reference_layer.ts
ykethan: got this regex from CFN error, but modified this now to be similar to the docs now and linked the documentation.

packages/backend-function/src/reference_layer.ts
ykethan: recreated this as a FunctionLayerParser class with parse as suggested

packages/backend-function/API.md
```diff
@@ -21,6 +21,9 @@ export type CronSchedule = `${string} ${string} ${string} ${string} ${string}` |
 // @public
 export const defineFunction: (props?: FunctionProps) => ConstructFactory<ResourceProvider<FunctionResources> & ResourceAccessAcceptorFactory & AddEnvironmentFactory>;
 
+// @public (undocumented)
+export type FunctionLayerReferences = Record<string, string>;
```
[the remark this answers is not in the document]
ykethan: inlined this. But to note there is a way we can add a template string literal this but would require a workaround wasnt comfortable using this as unsure if it will be removed: https://github.com/Microsoft/TypeScript/issues/29729#issuecomment-567871939

https://www.typescriptlang.org/play/?#code/C4TwDgpgBAYghgGwQIzgYwNYGVgCcCWAdgOZQC8UAznkaQGRQDeUcuhAMvhhAPwBcLQiCgBfANwBYAFChIsAK6E0wfAHsOcEBFwBBNuSgADVoT5wA7pT4I4AW2QATOHwAkjagRIjX7ml+ua2j4etN5uIV6GUAA+sIgo6Nh+xJIy4NAwispqGlq6bADyBMREiAbGbGaWAfZOwclhvp7E3jZ59c2NES2GqdJo6tRQwBDUAplKKursgfmERfglhGUUAOQmVVY2tc4jY3vAAIwB7QcATHwHAMyrqQOEQ+fjWVO52nqEBuuVFlt2jrtRsBLkDjm0gk9rrdpGBcKo0KNKAA6CCEABuSIOAFF0QYDndBsBhkCrs9JjkZnkPgslisoLD4YiUejMUCcWiWJQqMkCQ8iQcACxk7LTWYfAwMhGUZGojHY3FwLndPpSaTSIA

.changeset/serious-dragons-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+'@aws-amplify/backend-function': minor
+---
+
+adds layers to defineFunction
```
Amplifiyer, line 5:
```suggestion
adds support to reference existing layers in defineFunction
```

packages/backend-function/src/layer_parser.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    for (const [key, arn] of Object.entries(layers)) {
+      if (!this.isValidLayerArn(arn)) {
+        throw new AmplifyUserError('InvalidLayerArnFormatError', {
+          message: `Invalid ARN format for layer: ${arn}`,
+          resolution: `Update the layer ARN with the expected format: arn:aws:lambda:<current-region>:<account-id>:layer:<layer-name>:<version> for function: ${functionName}`,
+        });
+      }
+      validLayers[key] = arn;
+    }
```
Amplifiyer, line 36: We are validating on the count of uniqueArns not the number of entries in the map. But later we are not deduping it later on. So for example if I provide a map of
```
{
  key1: arn1,
  key2: arn1,
  key3: arn1,
  key4: arn1,
  key5: arn1,
  key6: arn1,
}
```
our validation will succeed, but we will be passing/creating 6 reference layers to the `NodeJSFunction` constructor which may fail later.

Ideally we should dedupe here on the ARN and provide any of the key here since the key is only used for generating the name of the layer.

ykethan: oops thanks for the catch, updated this and also added a test to ensure unique Arn's are passed.

Amplifiyer (approved):

Amplifiyer (approved):

sobolk (approved):

PR #1853, opened 2024-08-13
UI tools

sobolk:
## Problem/Changes

This PR adds "client tools" concept to conversation handler.

They way this works is:
1. Client side can send "client tools" configuration in the event. This propagates through AppSync to conversation handler.
2. "Client tool" purpose is to grab inputs from LLM and execute processing in UI (or whatever "client side" is).
3. Client tool use interrupts a tool use loop in bedrock adapter to yield control back to UI.

## Out of scope

Bedrock API may respond with a tool use response that includes multiple tools.
It is unclear what exactly should we do if mixed tool kinds are requested, so for simplicity, for now if any of the tools is client tool we interrupt, ignore other tools and respond with client side tool usage blocks.
This is to enable exploration of real use cases with client side tools.

## Validation

Added e2e tests and unit tests.

## Checklist

### DIFF
--- a/packages/ai-constructs/API.md
+++ b/packages/ai-constructs/API.md
@@ -83,10 +83,7 @@ type ConversationTurnEvent = {
 };
 
 // @public (undocumented)
-type ExecutableTool = {
-    name: string;
-    description: string;
-    inputSchema: ToolInputSchema;
+type ExecutableTool = ToolDefinition & {
     execute: (input: DocumentType | undefined) => Promise<ToolResultContentBlock>;
 };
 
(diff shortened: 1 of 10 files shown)


packages/ai-constructs/API.md
```diff
@@ -103,10 +98,18 @@ declare namespace runtime {
...
+        ToolDefinition
     }
 }
 
+// @public (undocumented)
+type ToolDefinition = {
+    name: string;
+    description: string;
+    inputSchema: ToolInputSchema;
+};
```
sobolk, line 111: This abstraction appeared in third place already in public API. Therefore, extracting a type.

packages/integration-tests/src/test-project-setup/conversation_handler_project.ts
```diff
@@ -320,6 +327,71 @@ class ConversationHandlerTestProject extends TestProjectBase {
...
+      },
+      ...commonEventProperties,
+    };
+    const response = await this.executeConversationTurn(
+      event,
+      defaultConversationHandlerFunction,
+      apolloClient
+    );
+    // Assert that tool was used. I.e. that LLM used value returned by the tool.
+    assert.match(response.content, /Seattle/);
```
rtpascual, line 390: Am I understanding correctly that the event messages are included in the response? If not, how are we sure the response will always include `Seattle` instead of something like `The temperature is x`?

rtpascual (approved):

awsluja (approved):

PR #1858, opened 2024-08-13
Add `defineConversationHandlerFunction` and `@aws-amplify/backend-ai` package.

sobolk:
## Changes

This PR adds `defineConversationHandlerFunction` abstraction that can be used to define custom conversation handler in Gen 2 way.

For example
```
import { defineConversationHandlerFunction } from '@aws-amplify/backend-ai/conversation';
import { bedrockModelId } from '../constants.js';

export const customConversationHandler = defineConversationHandlerFunction({
  name: 'customConversationHandlerFunction',
  entry: './custom_handler.ts',
  models: [
    {
      modelId: bedrockModelId,
    },
  ],
});

```

```

const backend = defineBackend({ customConversationHandler });
```

## Validation

1. Added unit tests
2. Modified custom handler e2e tests to use new abstraction

## Checklist


packages/platform-core/src/caller_directory_extractor.ts
```diff
@@ -0,0 +1,42 @@
...
+      throw unresolvedImportLocationError;
+    }
+    const stackTraceImportLine = stacktraceLines[1]; // the first entry is the file where the error was initialized (our code). The second entry is where the customer called our code which is what we are interested in
+
+    const filePath = new FilePathExtractor(stackTraceImportLine).extract();
+    if (filePath) {
+      return path.dirname(filePath);
+    }
+    throw unresolvedImportLocationError;
+  };
```
sobolk, line 41: This logic is moved as is from `packages/backend-function/src/get_caller_directory.ts`

rtpascual: Curious why we switched this to become a class now, could we have kept this as a function?

sobolk: So that it looks consistent next to `FilePathExtractor` in https://github.com/aws-amplify/amplify-backend/blob/main/packages/platform-core/API.md .

packages/backend-ai/src/conversation/factory.ts
```diff
@@ -0,0 +1,121 @@
...
+    private readonly props: DefineConversationHandlerFunctionProps,
+    private readonly callerStack: string | undefined
+  ) {}
+
+  getInstance = ({
+    constructContainer,
+    outputStorageStrategy,
+    resourceNameValidator,
+  }: ConstructFactoryGetInstanceProps): ConversationHandlerFunction => {
+    resourceNameValidator?.validate(this.props.name);
```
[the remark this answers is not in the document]
sobolk: I made the `name` mandatory for now to avoid this. (and force customer to give it a name they can find later).
Defining custom handler is expected to be rather niche scenario, so that friction might be ok and we can improve it later.

packages/backend-ai/src/conversation/factory.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    });
+  });
+
+  void it('uses entry that is explicitly specified', () => {
+    const factory = customEntryHandler;
+    const lambda = factory.getInstance(getInstanceProps);
+    const template = Template.fromStack(Stack.of(lambda.resources.lambda));
+    template.resourceCountIs('AWS::Lambda::Function', 1);
+    template.hasResourceProperties('AWS::Lambda::Function', {
+      Handler: 'index.handler',
```
awsluja, line 76: shouldn't this be something like index.custom_handler from https://github.com/aws-amplify/amplify-backend/pull/1858/files#diff-bc30defca6e6547d73100e297149be6f8eb1940b1ec0a5ecaf67b311e8511624R9

sobolk: `Handler: 'index.handler',` is not related to file name.

`index` is there because esbuild bundles everything into single `index.js` . This happens no matter how input sources are defined.

`handler` is expected export from `index.js`. I.e. relates to `export const handler = async () => {` .

packages/backend-ai/src/conversation/factory.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    });
+  });
+
+  void it('throws on invalid name', () => {
+    mock
+      .method(resourceNameValidator, 'validate')
+      .mock.mockImplementationOnce(() => {
+        throw new Error('test validation error');
```
awsluja, line 84: why mock this if this is what you're testing? we should just be looking for the error containing ```Resource name contains invalid characters, found ```

packages/backend-ai/src/conversation/factory.test.ts
```diff
@@ -0,0 +1,125 @@
...
+      const factory = defineConversationHandlerFunction({
+        entry: './test-assets/with-default-entry/handler.ts',
+        name: 'testHandlerName',
+        models: [],
+      });
+      factory.getInstance(getInstanceProps);
+      const template = Template.fromStack(rootStack);
+      const outputValue =
+        template.findOutputs('definedFunctions').definedFunctions.Value;
+      assert.deepStrictEqual(outputValue, {
```
awsluja, line 106: assert.deepStrictEqual(outputValue, <something in the stack>.ref) ?

packages/backend-ai/src/conversation/factory.test.ts
sobolk: I borrowed this from functions.

It seems to be written there this way because the unit under test is `factory` not `resourceNameValidator`. Which is correct approach if the goal is "true unit test". `resourceNameValidator` has it's own unit tests tests.

We could use real `resourceNameValidator` give it's simplicity, but we'd make this test less "unit-ish".

awsluja: nit: test name should be more clear like "throws if resourceNameValidator detects an invalid name"

packages/backend-ai/src/conversation/factory.test.ts
awsluja: gotcha, then how is this test validating that the correct entry is used? is there any other attribute that can be matched

sobolk: We can't do that at this level of testing, so this test is kind of best effort. It will surface some bugs when entry resolution points to non existing files.
Otherwise this aspect is also covered in e2e tests where outcomes from lambdas are used in assertions.

See https://github.com/aws-amplify/amplify-backend/blob/4fb161cef9fae2cad5d2ae2e271eefd27470ba29/packages/backend-function/src/factory.test.ts#L141-L143

packages/backend-ai/src/conversation/factory.test.ts
sobolk: That seems to be not doable.
```
      console.log(lambda.resources.lambda.functionName);
      console.log(lambda.resources.cfnResources.cfnFunction.ref);
```
prints
```
${Token[TOKEN.208]}
${Token[TOKEN.207]}
```

I also looked at APIs in `Template` but they don't return desired values either.

It seems this is all we can do here.
Similar places https://github.com/search?q=repo%3Aaws-amplify%2Famplify-backend+%27Fn%3A%3AGetAtt%27&type=code

rtpascual (approved):

awsluja (approved):

PR #1864, opened 2024-08-14
Handle namespaces better in api check

sobolk:
## Problem

This PR solves two problems:
1. The existing solution to add namespace to symbols doesn't scale well. I.e. there's a need to cover every place a symbol is used in usage snippets.
2. API extractor does not support multiple entry point in package.json (exports section). We workaround this with `index.internal.ts` but this workaround needs more refinement to handle more advanced usage on `feature/conversation-handler` branch.

I.e.
image

## Changes

1. Refactor namespacing solution to use match-replace after snippet generation. This is cheap and covers everything we need without complicating generators to process every possible type usage (which would otherwise require unpacking generics in every place, which is a lot of work).
2. Add special convention with `__export__` prefix in `index.internal.ts`. So that we can leverage most of namespacing logic to handle extra exports from package.json (see modified test project). The idea is to use magic prefix to instruct api validator to "star import extra export as namespace" and use rest of namespacing logic to handle this.

## Out of scope

This PR prepares check to deal with namespaces better, but it does not change any of packages yet, neither on main nor in branches (that work will be done separately).

## Validation

1. This PRs validation is green (no regressions on main)
2. Added tests.
3. Verified locally with `feature/conversation-handler` (this PR is to `main` since it's independent from feature work).

image
image

## Checklist

### DIFF
--- a/scripts/components/api-changes-validator/api_usage_statements_generators.ts
+++ b/scripts/components/api-changes-validator/api_usage_statements_generators.ts
@@ -1,7 +1,6 @@
 import {
-  NamespaceDefinitions,
-  UsageStatements,
   UsageStatementsGenerator,
+  UsageStatementsGeneratorOutput,
 } from './types.js';
 import ts from 'typescript';
 import { EOL } from 'os';
@@ -30,7 +29,7 @@ export class GenericTypeParameterDeclarationUsageStatementsGenerator
   /**
    * @inheritDoc
    */
-  generate = (): UsageStatements => {
+  generate = (): UsageStatementsGeneratorOutput => {
     if (!this.typeParameters || this.typeParameters.length === 0) {
       return {};
     }
@@ -70,7 +69,7 @@ export class GenericTypeParameterUsageStatementsGenerator
   /**
    * @inheritDoc
    */
-  generate = (): UsageStatements => {
+  generate = (): UsageStatementsGeneratorOutput => {
     if (!this.typeParameters || this.typeParameters.length === 0) {
       return {};
     }
@@ -95,7 +94,7 @@ export class ImportUsageStatementsGenerator
    * @inheritDoc
    */
   constructor(private readonly node: ts.ImportDeclaration) {}
-  generate = (): UsageStatements => {
+  generate = (): UsageStatementsGeneratorOutput => {
     return {
       importStatement: this.node.getText(),
     };
@@ -123,10 +122,9 @@ export class TypeUsageStatementsGenerator implements UsageStatementsGenerator {
    */
   constructor(
     private readonly typeAliasDeclaration: ts.TypeAliasDeclaration,
-    private readonly packageName: string,
-    private readonly namespaceDefinitions: NamespaceDefinitions
+    private readonly packageName: string
   ) {}
-  generate = (): UsageStatements => {
+  generate = (): UsageStatementsGeneratorOutput => {
     const typeName = this.typeAliasDeclaration.name.getText();
     const constName = toLowerCamelCase(typeName);
     const genericTypeParametersDeclaration =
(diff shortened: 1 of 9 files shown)


rtpascual (approved):

scripts/components/api-changes-validator/usage_statemets_renderer.ts
```diff
@@ -0,0 +1,79 @@
+import {
+  NamespaceDefinitions,
+  UsageStatementsGeneratorOutput,
+} from './types.js';
+import { EOL } from 'os';
+
+/**
+ * Renders usage statements.
+ */
+export class UsageStatementsRenderer {
```
0618, line 10: Do we need to add unit tests for `usage_statements_runderer` and `import_statements_renderer`?

scripts/components/api-changes-validator/api_usage_statements_generators.ts
0618, line 1: This file seems not have unit test either

scripts/components/api-changes-validator/usage_statemets_renderer.ts
```diff
@@ -0,0 +1,79 @@
...
+          namespaceHierarchy.unshift(currentSymbolName);
+        }
+      } while (currentSymbolName);
+      const symbolWithNamespace = `${namespaceHierarchy.join(
+        '.'
+      )}.${symbolName}`;
+
+      // characters that can be found before or after symbol
+      // this is to prevent partial matches in case one symbol's characters are subset of longer one
+      const symbolTerminators = '[\\s\\,\\(\\)<>]';
```
0618, line 66: Do we try to match any charactors in `\s,()<>`?

Would it be helpful to make a comment about what exactly we try to match here with the Regex?

scripts/components/api-changes-validator/import_statements_renderer.ts
```diff
@@ -0,0 +1,58 @@
...
+        ) {
+          importStatements.push(
+            `import { ${generatorOutput.symbolDescriptor.symbolName} } from '${this.packageName}';`
+          );
+        }
+      }
+    }
+
+    for (const namespaceName of this.namespaceDefinitions.topLevelNamespaces) {
+      if (namespaceName.startsWith('__export__')) {
```
0618, line 38: Could you help me fine where we add `__export__`?

scripts/components/api-changes-validator/usage_statemets_renderer.ts
sobolk: I don't think so.
We have a pattern of using higher level (integration) testing strategies in `scripts` folder. Which seems to be sufficient strategy for now.

Combination of 
https://github.com/aws-amplify/amplify-backend/blob/main/scripts/components/api-changes-validator/api_usage_generator.test.ts
and
https://github.com/aws-amplify/amplify-backend/tree/main/scripts/components/api-changes-validator/test-resources/test-projects ran by https://github.com/aws-amplify/amplify-backend/blob/main/scripts/components/api-changes-validator/api_changes_validator.test.ts
provide enough coverage.

The reason for this is that components in `scripts` often rely on external tools and proving that those integrations work is more important than pedantic unit testing (which in this case would be an art to check a box rather than something useful). Also, higher level tests in `scripts` are fast and stable enough to not worry about their volume.

scripts/components/api-changes-validator/api_usage_statements_generators.ts
sobolk: They're all tested by https://github.com/aws-amplify/amplify-backend/blob/main/scripts/components/api-changes-validator/api_usage_generator.test.ts .

See [my comment above.](https://github.com/aws-amplify/amplify-backend/pull/1864#discussion_r1718953405)

scripts/components/api-changes-validator/usage_statemets_renderer.ts
sobolk: > Do we try to match any charactors in \s,()<>?

Correct.

> Would it be helpful to make a comment about what exactly we try to match here with the Regex?

"characters that can be found before or after symbol" in the comment above is indicating that. In TS/JS there's finite set of characters that can be found around symbols, i.e. whitespace, comma and bracket. 
This set might not be complete right now, but can be expanded when new patterns are added to API.md files.

scripts/components/api-changes-validator/usage_statemets_renderer.ts
sobolk: We had simialar discussion here https://github.com/aws-amplify/amplify-backend/pull/800#discussion_r1423269820 .

scripts/components/api-changes-validator/import_statements_renderer.ts
sobolk: See `Out of scope` section in description.

The change that will follow looks like that https://github.com/aws-amplify/amplify-backend/pull/1869 .

We also included this case in tests here https://github.com/aws-amplify/amplify-backend/blob/bf9de828fa68c29ef516e9c19362057fb0305572/scripts/components/api-changes-validator/test-resources/test-projects/without-breaks/project-with-namespace/API.md
( the API.md in that test project is simulating how `index.internal.ts` would be translated into api extract)

0618 (approved):

PR #1894, opened 2024-08-21
chore: refactor error handling for secrets client

Amplifiyer:
## Problem

Secrets client is used between a lambda to fetch secrets at runtime as well as by the cli commands for CRUD operations. The error handling for cli commands was either happening at the call site or none at all.

## Changes

This PR introduces another wrapper over `SSMSecretClient` to perform error handling for amplify cli commands use case.

## Checklist

### DIFF
--- a/packages/backend-secret/API.md
+++ b/packages/backend-secret/API.md
@@ -12,6 +12,9 @@ import { BackendIdentifier } from '@aws-amplify/plugin-types';
 export const getSecretClient: (secretClientOptions?: SecretClientOptions) => SecretClient;
 
 // @public
+export const getSecretClientWithAmplifyErrorHandling: (secretClientOptions?: SecretClientOptions) => SecretClient;
+
+// @public
 export type Secret = SecretIdentifier & {
     value: string;
     lastUpdated?: Date;
--- a/packages/backend-secret/src/secret.ts
+++ b/packages/backend-secret/src/secret.ts
@@ -2,6 +2,7 @@ import { SSMSecretClient } from './ssm_secret.js';
 import { AwsCredentialIdentityProvider } from '@aws-sdk/types';
 import { SSM } from '@aws-sdk/client-ssm';
 import { AppId, BackendIdentifier } from '@aws-amplify/plugin-types';
+import { SSMSecretClientWithAmplifyErrorHandling } from './ssm_secret_with_amplify_error_handling.js';
 
 /**
  * The unique identifier of the secret.
@@ -72,7 +73,7 @@ export type SecretClientOptions = {
 };
 
 /**
- * Creates an Amplify secret client.
+ * Creates an Amplify secret client. Used in the backend lambda for fetching secrets
  */
 export const getSecretClient = (
   secretClientOptions?: SecretClientOptions
@@ -82,5 +83,17 @@ export const getSecretClient = (
       credentials: secretClientOptions?.credentials,
       region: secretClientOptions?.region,
     })
+  );
+};
+
+/**
+ * Creates an Amplify secret client with inbuilt amplify error handling
+ * Used in the amplify CLI commands such as sandbox and pipeline-deploy
+ */
+export const getSecretClientWithAmplifyErrorHandling = (
+  secretClientOptions?: SecretClientOptions
+): SecretClient => {
+  return new SSMSecretClientWithAmplifyErrorHandling(
+    getSecretClient(secretClientOptions)
   );
 };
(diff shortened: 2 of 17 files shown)


packages/backend-secret/API.md
```diff
@@ -11,6 +11,9 @@ import { BackendIdentifier } from '@aws-amplify/plugin-types';
 // @public
 export const getSecretClient: (secretClientOptions?: SecretClientOptions) => SecretClient;
 
+// @public
+export const getSecretClientWithAmplifyErrorHandling: (secretClientOptions?: SecretClientOptions) => SecretClient;
+
```
sobolk, line 16: Would `cli-core` be better place for this? 

Do we do any "amplify error handling" outside of "cli-ish" packages?

sobolk: On the other hand. We'd need to put dependency on backend-secret in cli-core which seems wrong. So let's keep it here.

One last question. Should we introduce `getSecretClientWithAmplifyErrorHandling` or add `enableAmplifyErrorHandling` to `secretClientOptions` ?

sobolk (dismissed):

packages/backend-secret/src/ssm_secret_with_amplify_error_handling.ts
```diff
@@ -0,0 +1,110 @@
...
+        [
+          'UnrecognizedClientException',
+          'AccessDeniedException',
+          'NotAuthorized',
+          'ExpiredTokenException',
+          'CredentialsProviderError',
+        ].includes(error.cause.name)
+      ) {
+        return new AmplifyUserError('SSMCredentialsError', {
+          message: `Failed to ${apiName} secrets. ${error.cause.name}: ${error.cause?.message}`,
```
rtpascual, line 88: Nit: With `apiName`s passed in this would be "Failed to Get|List|Set|Remove secrets", do we want these to be lowercase?

rtpascual (approved):
LGTM, just have one nit

packages/backend-secret/src/ssm_secret_with_amplify_error_handling.ts
Amplifiyer: Updated!

rtpascual (dismissed):

sobolk (dismissed):

sobolk (approved):

packages/backend-secret/src/secret.ts
```diff
@@ -72,7 +73,7 @@ export type SecretClientOptions = {
 };
 
 /**
- * Creates an Amplify secret client.
+ * Creates an Amplify secret client. Used in the backend lambda for fetching secrets
  */
 export const getSecretClient = (
```
awsluja, line 78: If this is being used specifically for backend lambda - should we make that more clear and rename this to getSecretClientForBackendLambda?

sobolk: We can't rename existing API without MV bump.

packages/backend-secret/src/secret.ts
```diff
@@ -84,3 +85,15 @@ export const getSecretClient = (
     })
   );
 };
+
+/**
+ * Creates an Amplify secret client with inbuilt amplify error handling
+ * Used in the amplify CLI commands such as sandbox and pipeline-deploy
+ */
+export const getSecretClientWithAmplifyErrorHandling = (
```
awsluja, line 93: withAmplifyErrorHandling is a code smell; if AmplifyErrorHandling is the default behavior that we want, this should just be getSecretClient. The other one is the exception to this rule, and that's why I think getSecretClient -> getSecretClientForBackendLambda/getSecretClientWithoutAmplifyErrorHandling ; and this one becomes just 'getSecretClient'

packages/backend-secret/src/secret.ts
awsluja: we can update the implementation for this, and introduce getSecretClientForBackendLambda instead

packages/backend-secret/src/secret.ts
sobolk: > withAmplifyErrorHandling is a code smell

Agreed.

> that's why I think getSecretClient -> getSecretClientForBackendLambda/getSecretClientWithoutAmplifyErrorHandling ; and this one becomes just 'getSecretClient'

Changing type of exception thrown could. constitute runtime breaking change (think of people who wrote catch clauses). 
It would be safer to log work item to revise this when we ship MV of this package. And perhaps have only one variant with amplify errors.

packages/backend-secret/src/secret.ts
Amplifiyer: it does change the behavior of the existing API (i.e. from `throws SecretError` to `throws AmplifyError`) which in effect changes the API contract though not contractual in TS. I'm worried if console is already using this method and hence wanted to not touch that.

awsluja (approved):

PR #1898, opened 2024-08-21
Allow SQL Lambda to be hotswapped in sandbox

bobbyu99:
## Problem

Provide hotswap friendly resources in sandbox for faster developments. Backend will determine if we are doing a sandbox deployment, then passes a flag to data construct to enable hotswap friendly resources. 

## Changes

This PR adds `getDeploymentType` to determine deployment type. If it's a sandbox deployment, we set the new flag `provideHotswapFriendlyResources` in `translationBehavior` to be true.

## Validation
Added a unit test to test `getDeploymentType`.

## Checklist

### DIFF
--- a/.changeset/gorgeous-suits-camp.md
+++ b/.changeset/gorgeous-suits-camp.md
@@ -1,7 +1,5 @@
 ---
-'@aws-amplify/backend-data': minor
-'@aws-amplify/plugin-types': minor
-'@aws-amplify/backend': minor
+'@aws-amplify/backend-data': patch
 ---
 
 added provisionHotswapFriendlyResources and getDeploymentType to determine if we are in sandbox and would like to enable hotswap
--- a/.changeset/gorgeous-suits-camp.md
+++ b/.changeset/gorgeous-suits-camp.md
@@ -1,0 +1,7 @@
+---
+'@aws-amplify/backend-data': minor
+'@aws-amplify/plugin-types': minor
+'@aws-amplify/backend': minor
+---
+
+added provisionHotswapFriendlyResources and getDeploymentType to determine if we are in sandbox and would like to enable hotswap
--- a/packages/backend-data/src/factory.ts
+++ b/packages/backend-data/src/factory.ts
@@ -232,8 +232,8 @@ class DataGenerator implements ConstructContainerEntryGenerator {
     });
     let amplifyApi = undefined;
 
-    const shouldProvisionHotswapFriendlyResources =
-      stableBackendIdentifiers?.getDeploymentType?.() === 'sandbox';
+    const isSandboxDeployment =
+      scope.node.tryGetContext(CDKContextKey.DEPLOYMENT_TYPE) === 'sandbox';
 
     try {
       amplifyApi = new AmplifyData(scope, this.name, {
@@ -249,8 +249,7 @@ class DataGenerator implements ConstructContainerEntryGenerator {
            * The CI/CD check should take the responsibility to validate if any tables are being replaced and determine whether to execute the changeset
            */
           allowDestructiveGraphqlSchemaUpdates: true,
-          _provisionHotswapFriendlyResources:
-            shouldProvisionHotswapFriendlyResources,
+          _provisionHotswapFriendlyResources: isSandboxDeployment,
         },
       });
     } catch (error) {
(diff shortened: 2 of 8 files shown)


packages/backend-data/src/factory.ts
```diff
@@ -232,6 +232,9 @@ class DataGenerator implements ConstructContainerEntryGenerator {
     });
     let amplifyApi = undefined;
 
+    const shouldProvisionHotswapFriendlyResources =
+      stableBackendIdentifiers?.getDeploymentType?.() === 'sandbox';
```
sobolk, line 236: Can you please use this instead of introducing this new API (which doesn't seem to belong to stable identifiers) ?

https://github.com/aws-amplify/amplify-backend/blob/67feca033741c8382149b26c6634fe3b28c94b8f/packages/backend-data/src/factory.ts#L268-L269

sobolk (changes requested):

packages/backend-data/src/factory.ts
sobolk: Additionally. Can you please add unit test sfor change in `factory.ts` ?

bobbyu99: Thanks for your comments, Kamil. I have switched to use `isSandboxDeployment`, currently trying to figure out how to uni test this change in `factory.test.ts`. 

The image 1 below shows how I unit test this change in `amplify-category-api`, where I inspect the synthesized stack template or `CloudFormationStackArtifact` to be exact and verify through the physical resource id generated. I am struggling to do something similar in `factory.test.ts`. 

I tried `const amplifyTableStackTemplate = Template.fromStack(
      Stack.of(dataConstruct.resources.nestedStacks['Todo']));`, it looks like image 2 below. 

And `const amplifyTableStackTemplate = Template.fromStack(
      Stack.of(dataConstruct.resources.graphqlApi));` looks like image 3. 

Both are not quite what I am looking for. Also not quite sure which method of `class Template` could come in handy in this case. Could you advise on how I can unit test this change?

Screenshot 2024-08-22 at 9 32 43 AM

Screenshot 2024-08-22 at 9 53 10 AM

Screenshot 2024-08-22 at 9 54 59 AM

sobolk: Take a look at https://github.com/aws-amplify/amplify-backend/blob/97207363ec928bdc6de1ac505e98ee2e90acb941/packages/backend-data/src/factory.test.ts#L817-L828 .

Which asserts other aspect of sandbox deployment.

In `factory.test.ts` you don't have to thoroughly test data construct behaviors (this is covered by data construct tests). What you should be looking at is a single simplest signal that new flag was applied, if such signal is available you can write assertion using it.

bobbyu99: im not able to find it somehow. 
image


> a single simplest signal that new flag was applied

if i get it right, the unit test should have picked up the flag set here at line 255 right?
Screenshot 2024-08-22 at 10 21 25 AM

sobolk: You won't find `_provisionHotswapFriendlyResources` in template. You should be looking for resources and/or their properties that are affected by that flag.

sobolk: Okay. I took a closer look at the change introduced in https://github.com/aws-amplify/amplify-category-api/pull/2772 and played with debugger a bit.

Unfortunately, it turns out that that physical resource id change can't be accessed by neither inspecting construct tree nor via template inspection. I.e. it doesn't seem to be possible.

There's yet one option to assert this. I.e. inspect physical ids in e2e tests where they can be read from CFN stacks. But it seems overkill, so let's leave this untested for now.

.changeset/gorgeous-suits-camp.md
```diff
@@ -0,0 +1,7 @@
+---
+'@aws-amplify/backend-data': minor
+'@aws-amplify/plugin-types': minor
+'@aws-amplify/backend': minor
```
sobolk, line 4:
```suggestion
'@aws-amplify/backend-data': patch
```

1. We're not shipping new apis nor features, just performance optimization
2. Plugin types are no longer affected.
3. We don't need to bump `backend`. package managers will resolve new backend-data when they upgrade deps graph or do new installation.

sobolk (approved):

sobolk (dismissed):

sobolk (approved):

rtpascual (approved):

PR #1900, opened 2024-08-21
Pass profile explicitly to CDK CLI when using `npx ampx sandbox --profile <profile_name>`

sobolk:
## Problem

Passing profile via environment variables (see [here](https://github.com/aws-amplify/amplify-backend/blob/b2464ab051a1e4b8380fda8ebd911a610d64ac6e/packages/cli/src/command_middleware.ts#L35-L37)) is not enough to assure consistency.

Customer using `npx ampx sandbox --profile <profile_name>` might run into edge case scenarios where local or user home scoped cdk configuration might override environemnt.
See [code in CDK](https://github.com/aws/aws-cdk/blob/c7980a2e068b5336914364050f72e41259aebc90/packages/aws-cdk/lib/settings.ts#L134-L139). I.e. config is resolved in such a way that env variables have the least priority (implicitly).

## Changes

Pass profile explicitly when calling CDK commands. So that it has the same priority as in ampx cli.

## Validation

1. Added tests.
2. Manual validations with debugger.

## Checklist

### DIFF
--- a/packages/cli/src/commands/sandbox/sandbox_command.test.ts
+++ b/packages/cli/src/commands/sandbox/sandbox_command.test.ts
@@ -239,10 +239,10 @@ void describe('sandbox command', () => {
     await new Promise((resolve) => setTimeout(resolve, 0));
     assert.equal(sandboxStartMock.mock.callCount(), 1);
     assert.equal(sandboxDeleteMock.mock.callCount(), 1);
-    assert.strictEqual(
-      sandboxDeleteMock.mock.calls[0].arguments[0].profile,
-      profile
-    );
+    assert.deepStrictEqual(sandboxDeleteMock.mock.calls[0].arguments[0], {
+      identifier: undefined,
+      profile,
+    });
   });
 
   void it('asks to delete the sandbox environment when users send ctrl-C and say no to delete', async (contextual) => {
--- a/packages/cli/src/commands/sandbox/sandbox_command.test.ts
+++ b/packages/cli/src/commands/sandbox/sandbox_command.test.ts
@@ -205,6 +205,46 @@ void describe('sandbox command', () => {
     assert.equal(sandboxDeleteMock.mock.callCount(), 1);
   });
 
+  void it('asks to delete the sandbox environment when users send ctrl-C and say yes to delete with profile', async (contextual) => {
+    // Mock process and extract the sigint handler after calling the sandbox command
+    const processSignal = contextual.mock.method(process, 'on', () => {
+      /* no op */
+    });
+    const sandboxStartMock = contextual.mock.method(
+      sandbox,
+      'start',
+      async () => Promise.resolve()
+    );
+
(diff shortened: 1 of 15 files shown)


packages/sandbox/API.md
```diff
@@ -25,6 +25,7 @@ export type Sandbox = {
 // @public (undocumented)
 export type SandboxDeleteOptions = {
     identifier?: string;
+    profile?: string;
```
sobolk, line 28: surprisingly `SandboxOptions` below already have `profile` but was unused.

rtpascual: Looks like on our docs both `sandbox` and `sandbox delete` are mentioned to have `profile` which leads to confusion

rtpascual (approved):

packages/sandbox/API.md
sobolk: Which is fine, they both are supposed to handle that (except edge cases I described in PR description).
I was just surprised that `SandboxOptions.profile` already exists but noting in the code was reading (using) it to do anything useful.

packages/cli/src/commands/sandbox/sandbox_command.test.ts
```diff
@@ -205,6 +205,46 @@ void describe('sandbox command', () => {
...
+    // User said yes to delete
+    contextual.mock.method(AmplifyPrompter, 'yesOrNo', () =>
+      Promise.resolve(true)
+    );
+
+    const profile = 'test_profile';
+    await commandRunner.runCommand(`sandbox --profile ${profile}`);
+
+    // Similar to the later 0ms timeout. Without this tests in github action are failing
+    // but working locally
```
awsluja, line 232: is it possible something is missing in the code? I don't see this required in line [360](https://github.com/aws-amplify/amplify-backend/pull/1900/files#diff-fe3062a0618a08be0e78b23acc94e8621cd2d865ad1b63da87b19278b8fe06c7R360)

awsluja: possibly a CI flag in the yargs tool?

sobolk: The other test you mentioned is not testing SIGINT.

I took this from test above. I.e. line 197.

I wanted to cover a change where profile is passed to sandbox delete that happens when you say "yes" after CTRL+C.

awsluja: do we need to re-import yargs in our beforeEach? https://github.com/yargs/yargs/blob/c36c571e4e15dfe26be1d919e4991fb6ab6ed9fd/test/yargs.js#L20

sobolk: Afaik we can't re-import yargs while using ESM imports. 

This solution was not invented in this PR.
See https://github.com/aws-amplify/amplify-backend/pull/100#discussion_r1280702279

awsluja (approved):

PR #1938, opened 2024-08-30
Api adjustments for conversation handler

sobolk:
## Problem

Feedback from API review:
1. We should accept `a.ai.model.anthropic.claude3Haiku()` as input for model Id
2. We should re-export types from `backend-ai` that take part in custom conversation handler implementation.

## Changes

1. Re-export types from `backend-ai`
2. Use these types in integration tests
3. Change inputs to `defineConversationHandlerFunction`.

## Validation

This PR's checks

## Checklist

### DIFF
--- a/packages/backend-ai/src/conversation/factory.ts
+++ b/packages/backend-ai/src/conversation/factory.ts
@@ -29,10 +29,25 @@ class ConversationHandlerFunctionGenerator
   ) {}
 
   generateContainerEntry = ({ scope }: GenerateContainerEntryProps) => {
+    const constructProps: ConversationHandlerFunctionProps = {
+      entry: this.props.entry,
+      models: this.props.models.map((model) => {
+        let modelId;
+        if (typeof model.modelId === 'string') {
+          modelId = model.modelId;
+        } else {
+          modelId = model.modelId.resourcePath;
+        }
+        return {
+          modelId,
+          region: model.region,
+        };
+      }),
+    };
     const conversationHandlerFunction = new ConversationHandlerFunction(
       scope,
       this.props.name,
-      this.props
+      constructProps
     );
     this.storeOutput(this.outputStorageStrategy, conversationHandlerFunction);
     return conversationHandlerFunction;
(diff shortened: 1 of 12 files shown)


packages/backend-ai/src/conversation/factory.ts
```diff
@@ -29,10 +29,25 @@ class ConversationHandlerFunctionGenerator
...
   generateContainerEntry = ({ scope }: GenerateContainerEntryProps) => {
+    const constructProps: ConversationHandlerFunctionProps = {
+      entry: this.props.entry,
+      models: this.props.models.map((model) => {
+        let modelId;
+        if (typeof model.modelId === 'string') {
+          modelId = model.modelId;
+        } else {
+          modelId = model.modelId.resourcePath;
+        }
```
Amplifiyer, line 40: should we validate that this is indeed in the right format and perhaps valid? What if customers provide `{}` or `{ resourcePath: 'rubbish', },`

Amplifiyer (approved):
Looks good, left one suggestion

packages/backend-ai/src/conversation/factory.ts
sobolk: Compiler is going to bark at `{}`. `resourcePath` is mandatory.
The `{ resourcePath: 'rubbish', }` is trickier to guard against, it's typed as `string` because we don't want to constrain it on the client side, given that this list is rather volatile https://docs.aws.amazon.com/bedrock/latest/userguide/model-ids.html .

This additional typing is to make types defined [here](https://github.com/aws-amplify/amplify-api-next/blob/b5b0760cff505e824174ab0b41a00baa735b8744/packages/data-schema/src/ai/AiModelType.ts#L4-L30) flow into this API. But without declaring explicit dependency (just make them equivalent type algebra wise). This is somewhat similar to this solution for function handlers (see [here](https://github.com/aws-amplify/amplify-api-next/blob/b5b0760cff505e824174ab0b41a00baa735b8744/packages/data-schema-types/src/builder/types.ts#L61-L64)).

One possible solution to this might be to lift `AiModel` to `data-schema-types` and use it here. But that's something we should consider for GA, after types on schema mature.

rtpascual (approved):

PR #1960, opened 2024-09-04
Added new lint rules for AmplifyUserError and AmplifyFault to restrict where they should be used

ShadowCat567:
## Problem

**Issue number, if available:** #1293

## Changes

- adds 2 new custom lint rules -- `no-amplify-errors` (for packages that should not use AmplifyUserError or AmplifyFault) and `prefer-amplify-errors` (for packages that should avoid using Error base class)
- adds mechanism to turn on the new lint rules in certain packages
- implements ability for `prefer-amplify-errors` to ignore tests
- **NOTE**: for the time being, `prefer-amplify-errors` is turned off for the entire project (including in the `cli` package where it will eventually be active) because it throws a lot of errors related to the use of Error base class and fixing these errors would be too much to be included in this PR

## Validation

- added tests for the new lint rules that ensure they perform expected behaviors

## Checklist


.changeset/great-drinks-burn.md
```diff
@@ -0,0 +1,5 @@
+---
+'eslint-plugin-amplify-backend-rules': minor
+---
+
+Adding two new custom rules and overrides to control how they are applied
```
sobolk, line 5:
```suggestion
---
---

```

For [private packages](https://github.com/aws-amplify/amplify-backend/blob/150541422f6bcb13fc5b1312cff01791a07cac5c/packages/eslint-rules/package.json#L3) empty changeset should be good enough .

The proposed change is equivalent to running `npx changeset --empty` .

packages/eslint-rules/src/rules/prefer_amplify_errors.ts
```diff
@@ -0,0 +1,46 @@
...
+      NewExpression(node) {
+        const checkNode = (errorType: string, ignorePatterns: string) => {
+          const fileNameWithPath = context.physicalFilename!.replace(
+            join(context.cwd, sep),
+            ''
+          );
+
+          const ignoreMatch = [
+            ...fileNameWithPath.matchAll(new RegExp(ignorePatterns, 'g')),
+          ];
```
sobolk, line 18: I thought we abandon the regex route.

packages/eslint-rules/src/rules/prefer_amplify_errors.ts
```diff
@@ -0,0 +1,46 @@
...
+            node.callee.name === errorType &&
+            !ignoreMatch.length
+          ) {
+            context.report({
+              messageId: 'useOfErrorDetected',
+              node,
+            });
+          }
+        };
+        checkNode('Error', '.test.ts');
```
sobolk, line 30: I guess this was added for testing. should we remove it?

Btw. with these overrides in the other file - have you found a solution to filter out test files ?

sobolk (commented):
Looks good overall.

packages/eslint-rules/src/rules/prefer_amplify_errors.ts
ShadowCat567: I left it for ignoring tests, since I would otherwise have to include a long list of ignores for every directory level tests could be on. I can experiment with removing regex and moving everything to the index.ts file

packages/eslint-rules/src/rules/prefer_amplify_errors.ts
ShadowCat567: There is a way to filter for tests, I discussed a bit in the previous comment, I want to see if I can find a way to do it where I only have to have one list item that will filter out all tests. In my previous attempt I had to have list item for every directory level.

packages/eslint-rules/src/rules/prefer_amplify_errors.ts
sobolk: I see. It would be best if configuration of what files are included and excluded is in one place.
I.e. in that `overrides` section in the other file.

Things to try:
1. The `files` in `overrides` takes [glob](https://en.wikipedia.org/wiki/Glob_(programming)) . Perhaps a pattern like `packages/cli/src/**/*!(.test.ts)` could work (I made the pattern up, but there might be some option to negate group of characters.
2. If above doesn't work (and maybe it's more explicit) the `overrides` should also take `ignores ` (see [docs](https://eslint.org/docs/latest/use/configure/configuration-files)) which is also a "glob" but is used to filter out unwanted files.

If these two options don't yield anything good, we can keep regex.

packages/eslint-rules/src/rules/prefer_amplify_errors.ts
sobolk: Ok. I didn't realize how this parameter is used. If we keep regex this is fine.

ShadowCat567: committing an update that filters out tests using the same method as filtering for packages we want the rules to be applied on

packages/eslint-rules/src/rules/prefer_amplify_errors.ts
```diff
@@ -0,0 +1,36 @@
...
+            node.callee.type === 'Identifier' &&
+            node.callee.name === errorType
+          ) {
+            context.report({
+              messageId: 'useOfErrorDetected',
+              node,
+            });
+          }
+        };
+        checkNode('Error');
```
sobolk, line 20: nit. (feel free to ignore).
we could remove `checkNode` and inline this logic. 
In other places we have it because we're dealing with multiple symbols and messages.

sobolk (dismissed):

packages/eslint-rules/src/rules/prefer_amplify_errors.ts
ShadowCat567: the `excludedFiles` field in `overrides` actually takes care of this now, turns out if you use `**/*.test.ts` it takes care of ignoring test files as all levels (which I did not realize previously). So the regex is now gone.

packages/eslint-rules/src/rules/prefer_amplify_errors.ts
ShadowCat567: if it's unnecessary, I'll get rid of it

sobolk (approved):

rtpascual (approved):
LGTM, just have one nit.

Let's create an issue to track the follow up work of turning the rules on and fixing the new lint errors.

PR #1967, opened 2024-09-06
fix: surface synth errors if the function declaration files have not been written yet

### LINKED ISSUE #1722: Missing '$amplify/env/xy' upon deploying
### Environment information

```plain text
System:
  OS: macOS 14.5
  CPU: (8) arm64 Apple M3
  Memory: 124.56 MB / 24.00 GB
  Shell: /bin/zsh
Binaries:
  Node: 20.14.0 - /opt/homebrew/opt/node@20/bin/node
  Yarn: undefined - undefined
  npm: 10.7.0 - /opt/homebrew/opt/node@20/bin/npm
  pnpm: undefined - undefined
NPM Packages:
  @aws-amplify/backend: 1.0.3
  @aws-amplify/backend-cli: 1.0.4
  aws-amplify: 6.3.7
  aws-cdk: 2.147.0
  aws-cdk-lib: 2.147.0
  typescript: 5.5.2
AWS environment variables:
  AWS_STS_REGIONAL_ENDPOINTS = regional
  AWS_NODEJS_CONNECTION_REUSE_ENABLED = 1
  AWS_SDK_LOAD_CONFIG = 1
No CDK environment variables
```

## Simplified structure
```
amplify
├───data
│      ├───resource.ts
│      └───test-mutation-handler
│                    └───testMutationHandler.ts
├───functions
│      └───daily-function
│                    ├───handler.ts
│                    └───resource.ts
├───backend.ts
└───tsconfig.json
```
## Code
_amplify/data/resource.ts_
```ts
import { type ClientSchema, a, defineData, defineFunction } from '@aws-amplify/backend';
...
import { dailyFunction } from '../functions/daily-function/resource';

const testMutationHandler = defineFunction({
  entry: './test-mutation-handler/testMutationHandler.ts'
});

const schema = a.schema({
...
  testMutation: a.mutation()
  .arguments({
    arg1: a.string().required(),
    arg2: a.string().required()
  })
  .returns(a.string())
  .authorization(allow => allow.authenticated())
  .handler(a.handler.function(testMutationHandler)),
...
})
.authorization(allow => [
  allow.resource(testMutationHandler),
  allow.resource(dailyFunction)
...
]);
```
_amplify/data/test-mutation-handler/testMutationHandler.ts_
```ts
import type { Schema } from '../resource';
import { env } from "$amplify/env/testMutationHandler";
...

Amplify.configure({...})
const client = generateClient<Schema>({
  authMode: "iam"
});
export const handler: Schema["testMutation"]["functionHandler"] = async (event, context) => {
...
};
```
_amplify/functions/daily-function/handler.ts_
```ts
import type { Schema } from '../../data/resource';
import { env } from "$amplify/env/daily-function";
...

Amplify.configure({...})
const client = generateClient<Schema>({
  authMode: "iam"
});
export const handler: Handler = async (event, context) => {
...
};
```
_amplify/functions/daily-function/resource.ts_
```ts
import { defineFunction } from '@aws-amplify/backend';

export const dailyFunction = defineFunction({
  name: 'daily-function',
  environment: {
    test: "2"
  }
});
```
_amplify/backend.ts_
```ts
import * as events from 'aws-cdk-lib/aws-events';
import * as targets from 'aws-cdk-lib/aws-events-targets';
import { data } from './data/resource';
import { storage } from './storage/resource';
import { dailyFunction } from './functions/daily-function/resource';

const backend = defineBackend({
  auth,
  data,
  storage,
  dailyFunction
});

const rule = new events.Rule(
  backend.dailyFunction.resources.lambda,
  'Rule',
  {
    description: 'Rule that triggers the lambda function every day at 00:00',
    schedule: events.Schedule.cron({minute: "0", hour: "0"}),
  },
);

rule.addTarget(
  new targets.LambdaFunction(backend.dailyFunction.resources.lambda),
);
```
_amplify/tsconfig.json_
```json
{
  "compilerOptions": {
    "target": "es2022",
    "module": "es2022",
    "moduleResolution": "bundler",
    "resolveJsonModule": true,
    "esModuleInterop": true,
    "forceConsistentCasingInFileNames": true,
    "strict": true,
    "skipLibCheck": true,
    "paths": {
      "$amplify/*": [
        "../.amplify/generated/*"
      ]
    }
  }
}
```
### What is expected to happen
The necessary environment files are correctly generated using the sandbox for local development as well as in the cloud-backend upon deploying.
### What actually happens
Using the sandbox everything works as expected.
However, upon deploying, the build-process fails with the following statements:
```sh
2024-07-06T12:32:57.614Z [INFO]: # Executing command: npx ampx pipeline-deploy --branch $AWS_BRANCH --app-id $AWS_APP_ID
2024-07-06T12:33:08.889Z [INFO]:
2024-07-06T12:33:15.365Z [INFO]: amplify/data/test-mutation-handler/testMutationHandler.ts(2,21): error TS2307: Cannot find module '$amplify/env/testMutationHandler' or its corresponding type declarations.
amplify/functions/daily-function/handler.ts(3,21): error TS2307: Cannot find module '$amplify/env/daily-function' or its corresponding type declarations.
```

Although there have already been similar issues (e.g. #1374), those seem to be connected to developing a Next.js app, as where I am developing a swift app (hence there is no tsconfig.json file at the root of the project).

One last side note: I have also implemented a PostConfirmationTriggerHandler according to the [documentation](https://docs.amplify.aws/swift/build-a-backend/functions/examples/create-user-profile-record/) which also uses `import { env } from "$amplify/env/post-confirmation";`. Strangely, this seems to be the only case for which there is no error in the build-console after deploying.

Amplifiyer:
## Problem

fixes #1722

When customers are deploying their apps for the first time, i.e. `.amplify/generated/env` doesn’t yet exist, if an issue happens during the ESBuild step as part of cdk synth, the env files fail to get auto-generated. TypeScript check that run after then complain

```console
amplify/func-src/handler.ts(1,21): error TS2307: Cannot find module '$amplify/env/defaultNodeFunction' or its corresponding type declarations.
TypeScript validation check failed.
```

The root cause is a cyclic dependency issue between Synth process and TS checks.

### Typescript dependency on Synth:

Synth generates the .env files which are required to make TS happy. Without that TS will complain about the above error.

### Synth  dependency on TypeScript:

The coding errors that are caught by TS and ESBuild overlap by a lot. However, ESBuild is run by cdk in a child process and TS is run by us, meaning we have a better handle of error handling and error display to the customers for TS errors and not so much for the ESBuild errors. Moreover, some ESBuild errors are sometimes cryptic and TS errors are lot more helpful

To break this cycle, the following was implemented:

```ts
let synthError;
try {
  runSynth();
} catch (e) {
  synthError = e;
}

runTSC();

if (synthError) {
  throw synthError;
}
```

## Changes

The change in this PR is to evaluate the TS error and see if it's because of missing function env declaration files. If that's the case *and* synth had failed, we throw synth error instead.

Another accompanying change is to route the TS's `stdout` to `stderr` as TS errors (from TS perspective user errors) are printed on `stdout`. This way we correctly populate the TS error in the `AmplifyError`'s downstream cause.

## Validation

Automated and manual testing

## Checklist

### DIFF
--- a/packages/backend-deployer/src/cdk_deployer.ts
+++ b/packages/backend-deployer/src/cdk_deployer.ts
@@ -13,6 +13,7 @@ import {
   type PackageManagerController,
 } from '@aws-amplify/plugin-types';
 import {
+  AmplifyError,
   AmplifyUserError,
   BackendLocator,
   CDKContextKey,
@@ -80,7 +81,23 @@ export class CDKDeployer implements BackendDeployer {
     const synthTimeSeconds = Math.floor((Date.now() - startTime) / 10) / 100;
 
     // then run type checks
-    await this.invokeTsc(deployProps);
+    try {
+      await this.invokeTsc(deployProps);
+    } catch (typeError: unknown) {
+      if (
+        synthError &&
+        typeError instanceof AmplifyError &&
+        typeError.cause?.message.match(
+          /Cannot find module '\$amplify\/env\/.*' or its corresponding type declarations/
+        )
+      ) {
+        // synth have failed and we don't have auto generated function environment definition files. This will
+        // result in the exception caught here which is not very useful for the customers.
+        // We instead throw the synth error for customers to fix what caused the synth to fail.
+        throw synthError;
+      }
+      throw typeError;
+    }
 
     // If somehow TSC was successful but synth wasn't, we now throw to surface the synth error
     if (synthError) {
@@ -129,7 +146,9 @@ export class CDKDeployer implements BackendDeployer {
    */
   executeCommand = async (
     commandArgs: string[],
-    options: { printStdout: boolean } = { printStdout: true }
+    options: { redirectStdoutToStderr: boolean } = {
+      redirectStdoutToStderr: false,
+    }
   ) => {
     // We let the stdout and stdin inherit and streamed to parent process but pipe
     // the stderr and use it to throw on failure. This is to prevent actual
@@ -157,7 +176,9 @@ export class CDKDeployer implements BackendDeployer {
 
     childProcess.stderr?.pipe(aggregatorStderrStream);
 
-    if (options?.printStdout) {
+    if (options?.redirectStdoutToStderr) {
+      childProcess.stdout?.pipe(aggregatorStderrStream);
+    } else {
       childProcess.stdout?.pipe(process.stdout);
     }
 
@@ -203,21 +224,24 @@ export class CDKDeployer implements BackendDeployer {
           '--project',
           dirname(this.backendLocator.locate()),
         ],
-        { printStdout: false }
+        { redirectStdoutToStderr: true } // TSC prints errors to stdout by default
       );
     } catch (error) {
       // If we cannot load ts config, turn off type checking
       return;
     }
     try {
-      await this.executeCommand([
-        'tsc',
-        '--noEmit',
-        '--skipLibCheck',
-        // pointing the project arg to the amplify backend directory will use the tsconfig present in that directory
-        '--project',
-        dirname(this.backendLocator.locate()),
-      ]);
+      await this.executeCommand(
+        [
+          'tsc',
+          '--noEmit',
+          '--skipLibCheck',
+          // pointing the project arg to the amplify backend directory will use the tsconfig present in that directory
+          '--project',
+          dirname(this.backendLocator.locate()),
+        ],
+        { redirectStdoutToStderr: true } // TSC prints errors to stdout by default
+      );
     } catch (err) {
       throw new AmplifyUserError<CDKDeploymentError>(
         'SyntaxError',
--- a/packages/backend-deployer/src/cdk_deployer.ts
+++ b/packages/backend-deployer/src/cdk_deployer.ts
@@ -91,8 +91,8 @@ export class CDKDeployer implements BackendDeployer {
           /Cannot find module '\$amplify\/env\/.*' or its corresponding type declarations/
         )
       ) {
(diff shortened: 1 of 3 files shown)


packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -80,7 +81,23 @@ export class CDKDeployer implements BackendDeployer {
...
+        typeError.cause?.message.match(
+          /Cannot find module '\$amplify\/env\/.*' or its corresponding type declarations/
+        )
+      ) {
+        // synth has failed and we don't have auto generated function environment definition files. This
+        // resulted in the exception caught here, which is not very useful for the customers.
+        // We instead throw the synth error for customers to fix what caused the synth to fail.
+        throw synthError;
+      }
+      throw typeError;
```
sobolk, line 99: I do wonder if we should just always throw synthError if that happens and refrain from running TSC.

Or perhaps, if both yielded errors maybe we should throw https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError and somewhat display both messages together.

sobolk (approved):

packages/backend-deployer/src/cdk_deployer.ts
Amplifiyer: The reason we run TS is to produce better errors for customers as I mentioned in the overview. I thought about aggregating it, but in most cases when it's just duplicate with different wording which gets very confusing.

rtpascual (approved):

allxie:
Woooo!! Thank you!!

PR #1962, opened 2024-09-05
Remove runtime from main exports in ai-constructs

sobolk:
## Problem

Exporting `runtime` via `<root_entry_point>` -> `conversation` -> `runtime` 
is causing module loader eagerly load bedrock sdk in IaC (CDK) context.
I.e. attempt to use just construct triggers it.

The runtime sub namespace is only needed for esbuild bundling of function and building custom handlers.

I.e.
```
"Error: Cannot find module '@aws-sdk/middleware-host-header'\r
Require stack:\r
- /private/var/folders/rd/hr92pbz54kx78lddndlp9jg40000gr/T/amplify-e2e-tests/generation-model_7149fdda9_20dc2c2f/node_modules/@aws-amplify/graphql-api-construct/node_modules/@aws-sdk/client-bedrock-runtime/dist-cjs/index.js\r
- /private/var/folders/rd/hr92pbz54kx78lddndlp9jg40000gr/T/amplify-e2e-tests/generation-model_7149fdda9_20dc2c2f/node_modules/@aws-amplify/graphql-api-construct/node_modules/@aws-amplify/ai-constructs/lib/conversation/runtime/bedrock_converse_adapter.js\r
- /private/var/folders/rd/hr92pbz54kx78lddndlp9jg40000gr/T/amplify-e2e-tests/generation-model_7149fdda9_20dc2c2f/node_modules/@aws-amplify/graphql-api-construct/node_modules/@aws-amplify/ai-constructs/lib/conversation/runtime/conversation_turn_executor.js\r
- /private/var/folders/rd/hr92pbz54kx78lddndlp9jg40000gr/T/amplify-e2e-tests/generation-model_7149fdda9_20dc2c2f/node_modules/@aws-amplify/graphql-api-construct/node_modules/@aws-amplify/ai-constructs/lib/conve"]
[11.935,"o","rsation/runtime/index.js\r
- /private/var/folders/rd/hr92pbz54kx78lddndlp9jg40000gr/T/amplify-e2e-tests/generation-model_7149fdda9_20dc2c2f/node_modules/@aws-amplify/graphql-api-construct/node_modules/@aws-amplify/ai-constructs/lib/conversation/index.js\r
- /private/var/folders/rd/hr92pbz54kx78lddndlp9jg40000gr/T/amplify-e2e-tests/generation-model_7149fdda9_20dc2c2f/node_modules/@aws-amplify/graphql-api-construct/node_modules/@aws-amplify/ai-constructs/lib/index.js\r
[...]

PR #1969, opened 2024-09-06
Do not bundle AWS SDK if default implementation is used.

sobolk:
## Problem

Bundling Bedrock SDK is problematic in default implementation case. I.e. when data construct uses conversation handler construct under the hood. 

That behavior demands that data construct bundles Bedrock SDK with it's dependencies as well. Which is not optimal.

## Changes

We started conversation handler with bundling of Bedrock SDK because it was not available in Lambda runtime yet.

Since then however. Lambda has updated SDKs and they now include Bedrock. Which is also compatible with our default implementation.
Therefore, we disable bundling if default handler is used to simplify situation on data construct side.

## Note
This is another solution like [this one](https://github.com/aws-amplify/amplify-backend/pull/1962) that aims to simplify caller (upstream) situation and smoothen DX for preview purposes. We do have an action item to revisit bundling on data side.

## Validation

This change is not unit-testable as there's no good way to assert bundling behavior in unit tests.

However, the logical branch introduced here is already covered by e2e tests which deploy both default and custom implementation of conversation handler.

Hence no new tests are added in this PR.

## Checklist

### DIFF
--- a/.changeset/happy-starfishes-trade.md
+++ b/.changeset/happy-starfishes-trade.md
@@ -1,0 +1,5 @@
+---
+'@aws-amplify/ai-constructs': patch
+---
+
+Do not bundle AWS SDK if default implementation is used
--- a/packages/ai-constructs/src/conversation/conversation_handler_construct.ts
+++ b/packages/ai-constructs/src/conversation/conversation_handler_construct.ts
@@ -62,7 +62,10 @@ export class ConversationHandlerFunction
         entry: this.props.entry ?? defaultHandlerFilePath,
         handler: 'handler',
         bundling: {
-          bundleAwsSDK: true,
+          // Do not bundle SDK if conversation handler is using our default implementation which is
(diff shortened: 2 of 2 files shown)


atierian (approved):
Can confirm that we're able to drop `@aws-sdk/client-bedrock-runtime` in the graphql-api-construct E2E tests for the conversation transformer ([ref](https://github.com/aws-amplify/amplify-category-api/compare/cc52849ed494851b66162c8a4a340caa952eda4c...5b4b7085219a2766d9992ad557f886b93d5f1a91#diff-455f8729f27403b483bf84b6e8bdcc5afe0ea696ba56209ffcbc7141507cebb3L28-R28)).

Also manually confirmed by successfully deploying a sandbox with a tagged release of the Data / API construct targeting this change.

🚢

Amplifiyer (approved):

PR #1973, opened 2024-09-09
Implemented no-amplify-errors in packages where it should be active

ShadowCat567:
## Problem

`no-amplify-errors` should be active in more packages than just `auth-construct`

## Changes

Implemented `no-amplify-errors` lint rule in most packages where it should be active.
Only packages that are excluded which will eventually have `no-amplify-errors` activated in are `platform-core` and `client-config` -- has too many errors to be included in this PR.
`model-generator` and `schema-generator` had some violations, but all their violations seem to be fine to leave as AmplifyUserError (added in line ignores for them).

## Validation

linter is happy with the updates

## Checklist


packages/model-generator/src/get_backend_output_with_error_handling.ts
```diff
@@ -20,6 +20,7 @@ export const getBackendOutputWithErrorHandling = async (
       error instanceof BackendOutputClientError &&
       error.code === BackendOutputClientErrorType.DEPLOYMENT_IN_PROGRESS
     ) {
+      // eslint-disable-next-line amplify-backend-rules/no-amplify-errors
```
sobolk, line 23: Now that I see the code around the suppression I do wonder what we should do about this long term.

1. Should we be moving the error handling like this up? I.e. move this to relevant CLI commands instead of having this handling here?
2. Should we re-classify this package and apply `prefer-amplify-errors` instead and keep local mapping at this level?

I would say 1. seems to be more correct. And I think lifting all these mappings to cli would be applicable to any command that pulls outputs. These don't seem to be specific to model generation.

@Amplifiyer what do you think?

rtpascual: I'm leaning towards 1 as well since this is already done for the `ampx generate form` command https://github.com/aws-amplify/amplify-backend/blob/main/packages/cli/src/commands/generate/forms/generate_forms_command.ts#L75-L137.

I believe moving this file to cli will get rid of the need for the duplicate code for forms as a side effect.

sobolk: Btw. if we align on number 1 then this PR should go as is and work item should be opened to move this handling up separately.

sobolk (approved):

rtpascual (approved):

Issue #908, opened 2024-01-11
linter should catch inaccessible packages


0618:
### Environment information

```plain text
n/a
```

### Description

Error like mentioned in https://github.com/aws-amplify/amplify-backend/pull/906 should be caught by linter.

PR #1997, opened 2024-09-12
Updates to packages unresolvable by no-extraneous-dependencies

ShadowCat567:
## Problem

Linter previously was unable to tell whether these packages were included as dependencies and would not throw an error if they were excluded in a package's dependency list but used as a dependency by a file within the package.

**Issue number, if available:** #908 

## Changes

Added main field to `package.json` of 8 packages that previously lacked the field (this makes packages resolvable to `no-extraneous-dependencies`). 
**Note about addition of main field**: there is precedent in this repo for using both exports and main, `platform-core` and `backend-output-schemas` already do this. Additionally when exports and main are used together, exports takes precedence over main (https://www.cs.unb.ca/~bremner/teaching/cs2613/books/nodejs-api/packages/)

## Validation

Verified that packages that were previously unresolvable are able to be resolved while executing no-extraneous-dependencies rule.
## Checklist

### DIFF
--- a/.changeset/yellow-jokes-kick.md
+++ b/.changeset/yellow-jokes-kick.md
@@ -1,0 +1,12 @@
+---
+'@aws-amplify/deployed-backend-client': patch
+'@aws-amplify/backend-deployer': patch
+'@aws-amplify/schema-generator': patch
+'@aws-amplify/model-generator': patch
+'@aws-amplify/backend-secret': patch
+'@aws-amplify/form-generator': patch
+'@aws-amplify/client-config': patch
+'@aws-amplify/sandbox': patch
+---
+
+added main field to packages known to lack one
--- a/packages/backend-deployer/package.json
+++ b/packages/backend-deployer/package.json
@@ -12,6 +12,7 @@
       "require": "./lib/index.js"
     }
   },
+  "main": "lib/index.js",
   "types": "lib/index.d.ts",
   "scripts": {
     "update:api": "api-extractor run --local"
--- a/packages/backend-secret/package.json
+++ b/packages/backend-secret/package.json
@@ -12,6 +12,7 @@
       "require": "./lib/index.js"
     }
   },
+  "main": "lib/index.js",
   "types": "lib/index.d.ts",
   "scripts": {
     "update:api": "api-extractor run --local"
--- a/packages/client-config/package.json
+++ b/packages/client-config/package.json
@@ -17,6 +17,7 @@
       "require": "./lib/paths/index.js"
     }
   },
+  "main": "lib/index.js",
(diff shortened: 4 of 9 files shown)


scripts/check_for_package_main.ts
```diff
@@ -0,0 +1,22 @@
+import { glob } from 'glob';
```
sobolk, line 1: I'd name this script `scripts/check_package_json.ts` or `scripts/check_package_jsons.ts`

If we discover a need for more custom checks like that for package.json files this will be the place to add them.

scripts/check_for_package_main.ts
```diff
@@ -0,0 +1,22 @@
...
+import { readPackageJson } from './components/package-json/package_json.js';
+
+const packagePaths = await glob('./packages/*');
+
+const errors: string[] = [];
+for (const packagePath of packagePaths) {
+  const { main, name, isPrivate } = await readPackageJson(packagePath);
+  if (isPrivate || name === '@aws-amplify/integration-tests') {
+    continue;
+  }
```
sobolk, line 11:
```suggestion
  const { main, name, private } = await readPackageJson(packagePath);
  if (private) {
    continue;
  }
```

We can just skip private packages. Their correctness is not that relevant. There seems to be a bug in this logic though. See below.

See https://github.com/aws-amplify/amplify-backend/blob/fe6c23d543b80cff33de9ee8896e05b30afdb4e9/scripts/components/package-json/package_json.ts#L7 .

and

https://github.com/aws-amplify/amplify-backend/blob/fe6c23d543b80cff33de9ee8896e05b30afdb4e9/packages/integration-tests/package.json#L3

scripts/check_package_json.ts
```diff
@@ -0,0 +1,30 @@
+import { glob } from 'glob';
+import { readPackageJson } from './components/package-json/package_json.js';
+
+const packagePaths = await glob('./packages/*');
+
+const tempExclude: string[] = [
+  '@aws-amplify/plugin-types',
+  '@aws-amplify/cli-core',
+];
```
sobolk, line 9:
```suggestion

// Excluding packages that violate 'import/no-extraneous-dependencies' rule until they're fixed.
const tempExclude: string[] = [
  '@aws-amplify/plugin-types',
  '@aws-amplify/cli-core',
];
```

Please also cut GH issue to follow up on this (and perhaps include link to it in the comment).

ShadowCat567: issue created: https://github.com/aws-amplify/amplify-backend/issues/2001

sobolk (dismissed):

sobolk (approved):

Amplifiyer (approved):

Issue #1825, opened 2024-07-31
Deployment Pipeline failing randomly TimeoutError: {"state":"TIMEOUT","reason":"Waiter has timed out"}


renatofrias26:
### Environment information

```plain text
System:
  OS: macOS 14.5
  CPU: (8) arm64 Apple M1 Pro
  Memory: 205.28 MB / 16.00 GB
  Shell: /bin/zsh
Binaries:
  Node: 18.20.1 - /usr/local/bin/node
  Yarn: 1.22.19 - /usr/local/bin/yarn
  npm: 10.8.2 - /usr/local/bin/npm
  pnpm: undefined - undefined
NPM Packages:
  @aws-amplify/backend: 1.0.4
  @aws-amplify/backend-cli: 1.2.0
  aws-amplify: 6.4.0
  aws-cdk: 2.148.0
  aws-cdk-lib: 2.148.0
  typescript: 5.4.5
AWS environment variables:
  AWS_STS_REGIONAL_ENDPOINTS = regional
  AWS_NODEJS_CONNECTION_REUSE_ENABLED = 1
  AWS_SDK_LOAD_CONFIG = 1
No CDK environment variables
```

### Data packages

```plain text
├─┬ @aws-amplify/backend-cli@1.2.0
│ └─┬ @aws-amplify/schema-generator@1.1.0
│   └── @aws-amplify/graphql-schema-generator@0.9.2
└─┬ @aws-amplify/backend@1.0.4
  └─┬ @aws-amplify/backend-data@1.0.3
    └── @aws-amplify/data-construct@1.9.1
```

### Description

My pipeline is failing multiple times without much explanation, re-triggering the pipeline seems to solve the issue eventually.
The issue stated happening after adding Google & Apple social sign in.

Error:

> amplify-xxxx-xxxx-xxxx | 9:11:07 AM | [31mUPDATE_FAILED [39m | Custom::SecretFetcherResource | [31m[1mauth/SIWA_CLIENT_IDSecretFetcherResource/Default[22m[39m (SIWACLIENTIDSecretFetcherResource) [31m[1mReceived response status [FAILED] from custom resource. Message returned: TimeoutError: {"state":"TIMEOUT","reason":"Waiter has timed out"}

Auth resource.ts:

```typescript
import { defineAuth, secret } from "@aws-amplify/backend";
import { preSignUp } from "./link-user-provider/resource";

export const auth = defineAuth({
  loginWith: {
    email: {
[...]

PR #2011, opened 2024-09-17
fix: consolidate backend secret custom resources

awsluja:
## Problem
This consolidates secret fetcher custom resources into one.

A single custom resource will be created, instead of a resource for each secret.

https://github.com/aws-amplify/amplify-backend/issues/1797
https://github.com/aws-amplify/amplify-backend/issues/1825
## Changes

## Checklist


packages/backend/src/engine/backend-secret/backend_secret.ts
```diff
@@ -16,9 +16,11 @@ export class CfnTokenBackendSecret implements BackendSecret {
    * The name of the secret to fetch.
    */
   constructor(
-    private readonly name: string,
+    private readonly secretName: string,
     private readonly secretResourceFactory: BackendSecretFetcherFactory
-  ) {}
+  ) {
+    BackendSecretFetcherFactory.registerSecret(secretName);
+  }
```
sobolk, line 23: This doesn't look like a healthy data flow.

Could we move this detail into factory (since we do receive reference to factory in this ctor already)?

For example.
Change scope of these variables https://github.com/aws-amplify/amplify-backend/blob/8dd7286bf1a870e69766963e872fad4add68bc84/packages/backend/src/secret.ts#L30-L33 and promote them to global scope (to make factory a singleton)
Use cdk.Lazy to accumulate secret names inside factory?

packages/backend/src/engine/backend-secret/backend_secret_fetcher_factory.test.ts
```diff
@@ -25,34 +25,31 @@ void describe('getOrCreate', () => {
   const providerFactory = new BackendSecretFetcherProviderFactory();
   const resourceFactory = new BackendSecretFetcherFactory(providerFactory);
 
+  beforeEach(() => {
+    BackendSecretFetcherFactory.clearRegisteredSecrets();
```
sobolk, line 29: Instead of using static methods, we should rather control scope of variables and make the factory an effective singleton in the code, but not in tests.

packages/backend/src/engine/backend-secret/backend_secret.ts
awsluja: These are singleton factories, except they do not use DI system like other factories, they use the scope.node.tryFindChild('<token>')

sobolk: can `registerSecret` become internal detail of the factory then ?

sobolk: @awsluja this is what I have in mind https://github.com/aws-amplify/amplify-backend/commit/9816f1714dfdfe0d72881d3a4de5a9f5c4999f54 .

packages/backend/src/engine/backend-secret/backend_secret_fetcher_factory.ts
```diff
@@ -33,15 +52,18 @@ export class BackendSecretFetcherFactory {
     scope: Construct,
     secretName: string,
     backendIdentifier: BackendIdentifier
-  ): CustomResource => {
-    const secretResourceId = `${secretName}SecretFetcherResource`;
+  ): SecretFetcherCustomResource => {
+    const secretResourceId = `SecretFetcherResource`;
```
sobolk, line 56:
```suggestion
    const secretResourceId = `AmplifySecretFetcherResource`;
```

just in case.

packages/backend/src/engine/backend-secret/lambda/backend_secret_fetcher.ts
```diff
@@ -47,54 +47,59 @@ export const handler = async (
...
   event: CloudFormationCustomResourceEvent
-): Promise<string> => {
+): Promise<Record<string, string>> => {
   const props = event.ResourceProperties as unknown as SecretResourceProps;
-  let secret: string | undefined;
+  const secretMap: Record<string, string> = {};
+  for (const secretName of props.secretNames) {
+    let secretValue: string | undefined = undefined;
+    try {
+      const resp = await secretClient.getSecret(
```
sobolk, line 56: An optimization idea:
https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/secrets-manager/command/BatchGetSecretValueCommand/

since we're fetching multiple secrets.

sobolk (approved):

packages/backend/src/engine/backend-secret/backend_secret_fetcher_factory.ts
awsluja: updated (AmplifySecretFetcherCustomResource)

sobolk: I meant resource id not the class name.

packages/backend/src/engine/backend-secret/lambda/backend_secret_fetcher.ts
awsluja: may be a good optimization as separate item, it would be https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/ssm/command/GetParametersCommand/ since we don't use secrets manager ; but we have custom try-catch logic for each secret here, so it would be something like batching all secrets with one command, then batch the retries, and aggregate errors

sobolk (approved):

blakedunson (approved):

sobolk (approved):

Amplifiyer (approved):

PR #2013, opened 2024-09-17
update data construct

sobolk:
## Problem

Canaries started to fail https://github.com/aws-amplify/amplify-backend/actions/runs/10893420780/job/30228299757 .

The change is not breaking, but we should catch up.

## Changes

1. Update data-construct dependency
2. Adjust tests.

## Validation

Canary run: https://github.com/aws-amplify/amplify-backend/actions/runs/10905264373

## Checklist

### DIFF
--- a/packages/backend-data/src/factory.test.ts
+++ b/packages/backend-data/src/factory.test.ts
@@ -535,25 +535,8 @@ void describe('DataFactory', () => {
                   'Fn::Join': [
                     '',
                     [
-                      'arn:',
                       {
-                        Ref: 'AWS::Partition',
-                      },
-                      ':appsync:',
-                      {
-                        Ref: 'AWS::Region',
-                      },
-                      ':',
-                      {
-                        Ref: 'AWS::AccountId',
-                      },
-                      // eslint-disable-next-line spellcheck/spell-checker
-                      ':apis/',
-                      {
-                        'Fn::GetAtt': [
-                          'amplifyDataGraphQLAPI42A6FA33',
-                          'ApiId',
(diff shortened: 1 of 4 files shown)


packages/backend-data/src/factory.test.ts
```diff
@@ -757,22 +692,8 @@ void describe('DataFactory', () => {
...
-                    },
-                    ':',
-                    {
-                      Ref: 'AWS::AccountId',
-                    },
-                    // eslint-disable-next-line spellcheck/spell-checker
-                    ':apis/',
                     {
-                      'Fn::GetAtt': ['amplifyDataGraphQLAPI42A6FA33', 'ApiId'],
+                      'Fn::GetAtt': ['amplifyDataGraphQLAPI42A6FA33', 'Arn'],
```
Amplifiyer, line 696: so `ApiId` is no longer available from the nested stack? Why is this not breaking change?

sobolk: It's not that it's not available.

Data construct just changed how do they obtain arn for policy statement.

I.e. instead of formatting arn themselves from other attributes they read this field now https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_appsync.CfnGraphQLApi.html#attrarn .

sobolk: which in turn affected the assertion. runtime behavior should be same.

Amplifiyer (approved):

ShadowCat567 (approved):

PR #2018, opened 2024-09-17
Handle CDK version mismatch

sobolk:
## Problem

When installed `aws-cdk` and `aws-cdk-lib` are not compatible a fault is raised.

**Repro.**

Create project like and use mismatched CDK versions.
```
{
  "name": "testapp048",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "@aws-amplify/backend": "^1.2.1",
    "@aws-amplify/backend-cli": "^1.2.6",
    "aws-cdk": "2.130.0",
    "aws-cdk-lib": "2.158.0",
    "constructs": "^10.3.0",
    "esbuild": "^0.23.1",
    "tsx": "^4.19.1",
    "typescript": "^5.6.2"
  },
  "dependencies": {
    "aws-amplify": "^6.6.1"
  }
}

```

Observe
image

## Changes

Add mapping and classify as user error.

## Validation

1. Added test.
2. Manual verification

image

## Checklist

### DIFF
--- a/.changeset/spicy-monkeys-guess.md
+++ b/.changeset/spicy-monkeys-guess.md
@@ -1,0 +1,5 @@
+---
+'@aws-amplify/backend-deployer': patch
+---
+
+Handle CDK version mismatch
--- a/.eslint_dictionary.json
+++ b/.eslint_dictionary.json
@@ -14,6 +14,7 @@
   "argv",
   "arn",
   "arns",
+  "aws",
   "backends",
   "birthdate",
   "bundler",
--- a/packages/backend-deployer/src/cdk_error_mapper.test.ts
+++ b/packages/backend-deployer/src/cdk_error_mapper.test.ts
@@ -259,6 +259,15 @@ const testErrorMappings = [
     errorName: 'FilePermissionsError',
     expectedDownstreamErrorMessage: `EACCES: permission denied, unlink '.amplify/artifacts/cdk.out/synth.lock'`,
   },
+  {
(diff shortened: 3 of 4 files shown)


rtpascual (approved):
LGTM. One question, so the only way customers run into this is if they update their `package.json`? We would have to keep a close eye to make sure this isn't unintentionally changed to have mismatched versions https://github.com/aws-amplify/amplify-backend/blob/5af7781d27f90518432d8140d3edae8a9136ade5/packages/create-amplify/src/amplify_project_creator.ts#L18-L19

sobolk:
> LGTM. One question, so the only way customers run into this is if they update their `package.json`? We would have to keep a close eye to make sure this isn't unintentionally changed to have mismatched versions
> 
> https://github.com/aws-amplify/amplify-backend/blob/5af7781d27f90518432d8140d3edae8a9136ade5/packages/create-amplify/src/amplify_project_creator.ts#L18-L19

The other way would be to surgically `npm update aws-cdk-lib` without the other one. I.e. they won't get into this situation by following our create flow. The mismatch happens after create flow, by playing with package.json or npm commands.


As for us changing create flow to mismatched versions. This one is tricky to assert because as long as we `^` both versions the npm will install both at latest - which in theory is correct state.
We would grossly mess this up if we skip `^`, but this would surface in e2e tests or canaries.

Issue #1535, opened 2024-05-17
Define Resource Tags for AWS Resources


fullstackcreatives:
### Environment information

```plain text
System:
  OS: macOS 14.4.1
  CPU: (8) x64 Apple M1
  Memory: 74.03 MB / 16.00 GB
  Shell: /bin/zsh
Binaries:
  Node: 20.11.1 - ~/.nvm/versions/node/v20.11.1/bin/node
  Yarn: 1.22.4 - /usr/local/bin/yarn
  npm: 10.2.4 - ~/.nvm/versions/node/v20.11.1/bin/npm
  pnpm: undefined - undefined
NPM Packages:
  @aws-amplify/backend: 1.0.2
  @aws-amplify/backend-cli: 1.0.3
  aws-amplify: 6.3.2
  aws-cdk: 2.142.0
  aws-cdk-lib: 2.142.0
  typescript: 5.4.5
```

### Description

In Amplify Gen1 it was possible to define tags for AWS resources by adding a tags.json to the project. It seems that this does not work in Gen2. Would be great to have a feature like this in Gen2 also.

josefaidt:
Hey @fullstackcreatives :wave: thanks for raising this! While it's possible to apply tags to individual resource stacks using CDK, we currently do not support applying tags to all resources within the root stack. Marking as a feature request

Sanuki-073:
Is there any update about this issue?

OOCAZ:
Would love any updates to this and curious when this might get prioritized. Any updates?

viaferreirj:
This is required functionality for working with Amplify in an enterprise environment. Any updates on this feature, or a workaround for accessing the Amplify root stack scope?

Using CDK scoping, tagging can be applied to the data backend resources, however that's not exposed for the auth backend, nor the root Amplify stack.

viaferreirj:
For anyone watching this request, I was able to tag all the resources with the following hack:
```ts
backend.data.node.scopes.forEach((scope) => {
    for (const [key, value] of Object.entries(tags)) {
        Tags.of(scope).add(key, value);
    }
})
```
Where `tags` is a key-value map.

Issue #1242, opened 2024-04-03
add a stack reference on backend resources


josefaidt:
### Environment information

```plain text
System:
  OS: macOS 14.3.1
  CPU: (10) arm64 Apple M1 Pro
  Memory: 103.66 MB / 32.00 GB
  Shell: /opt/homebrew/bin/fish
Binaries:
  Node: 20.11.1 - ~/Library/Caches/fnm_multishells/47118_1712176426928/bin/node
  Yarn: 1.22.19 - ~/Library/Caches/fnm_multishells/47118_1712176426928/bin/yarn
  npm: 10.2.4 - ~/Library/Caches/fnm_multishells/47118_1712176426928/bin/npm
  pnpm: 8.15.5 - ~/Library/Caches/fnm_multishells/47118_1712176426928/bin/pnpm
NPM Packages:
  @aws-amplify/backend: 0.13.0-beta.15
  @aws-amplify/backend-cli: 0.12.0-beta.17
  aws-amplify: 6.0.27
  aws-cdk: 2.135.0
  aws-cdk-lib: 2.135.0
  typescript: 5.4.3
AWS environment variables:
  AWS_PROFILE = josef-gen2
  AWS_STS_REGIONAL_ENDPOINTS = regional
  AWS_NODEJS_CONNECTION_REUSE_ENABLED = 1
  AWS_SDK_LOAD_CONFIG = 1
No CDK environment variables
```

### Description

It can be cumbersome to import `Stack` to get a stack reference on a backend resource created with a `define*` function. This is typically reached for when I add a custom policy or some additional resource with CDK that requires a `scope`. 

To remove the need for importing `Stack` and referencing the stack of resources, add a property/method to reference the stack

from
```ts
import { defineBackend } from "@aws-amplify/backend"
import { Stack } from "aws-cdk-lib/core"
import { auth } from "./auth/resource"
import { data } from "./data/resource"

const backend = defineBackend({
  auth,
  data,
})

// pick an arbitrary resource to resolve the stack
const authStack = Stack.of(backend.auth.resources.userPool)
```

to
```ts
import { defineBackend } from "@aws-amplify/backend"
import { auth } from "./auth/resource"
import { data } from "./data/resource"

const backend = defineBackend({
  auth,
  data,
})

const authStack = backend.auth.stack
```

ykethan:
Marking as feature request to support to referencing stack as a method or property.

PR #2021, opened 2024-09-18
Exposing stack parameter at higher level for backend and related resources

ShadowCat567:
## Problem

Users would like to be able to easily access the root stack and the stacks of their resources. Helpful for adding tags. 

**Issue number, if available:** https://github.com/aws-amplify/amplify-backend/issues/1535  https://github.com/aws-amplify/amplify-backend/issues/1242

## Changes

- lifted stack parameter for backend and auth, function, and storage resources

## Validation

- In test project able to utilize new stack parameters without error 
image

- added tests to ensure that expected values were coming from each of the stack parameters

## Checklist


packages/backend-auth/src/factory.ts
```diff
@@ -195,6 +203,7 @@ class AmplifyAuthGenerator implements ConstructContainerEntryGenerator {
           policy.attachToRole(role);
         },
       }),
+      stack: Stack.of(this.generateContainerEntry.prototype),
```
ShadowCat567, line 206: having trouble with this, if I use t`his.generateContainerEntry.prototype.scope`, tests fail because `scope` could be undefined. If I use `Stack.of(this.generateContainerEntry.prototype)`, tests fail because `Stack.of()` is undefined/cannot be read. I can't get rid of this or make it undefined because this is how the `stack` parameter is able to be called in `backend.auth.stack`

packages/backend/src/backend.ts
```diff
@@ -12,6 +12,7 @@ export type BackendBase = {
   addOutput: (
     clientConfigPart: DeepPartialAmplifyGeneratedConfigs<ClientConfig>
   ) => void;
+  stack: () => Stack;
```
ShadowCat567, line 15: The type of this has also been giving me trouble. I have tried changing the type of `stack` to `Stack` instead of `()=>Stack`, but when I do that I get errors in `backend_factory.ts`. So far the only way I have been able to set `stack` to the root stack is through `this.stackResolver.getStackfor('root')`. But if I change `stack`'s type to `Stack`, I cannot use this method since `stackResolver` is marked as being called before it is set. I've also tried keeping `rootStack`'s type as `()=>Stack` and changing `stack`'s type to `Stack`, but run into an error where these variables are unable to be assigned to each other because of their types.

.changeset/chatty-beans-begin.md
```diff
@@ -0,0 +1,7 @@
+---
+'@aws-amplify/backend-function': minor
+'@aws-amplify/backend-auth': minor
+'@aws-amplify/backend': minor
+---
+
+expose stack parameter for backend, function resource, and auth resource
```
sobolk, line 7:
```suggestion
expose stack property for backend, function resource, and auth resource
```

.changeset/four-drinks-yell.md
```diff
@@ -0,0 +1,7 @@
+---
+'@aws-amplify/integration-tests': minor
+'@aws-amplify/backend-storage': minor
+'@aws-amplify/plugin-types': minor
+---
+
+add new type to handle exposing stack, expose stack in storage
```
sobolk, line 7: We can remove integration-tests from changesets.
I'd move storage from this changeset to the other one. and keep this one for plugin types only.
It'll generate more coherent changelogs.

packages/backend/src/backend_factory.ts
```diff
@@ -157,5 +159,6 @@ export const defineBackend = <T extends DefineBackendProps>(
     ...backend.resources,
     createStack: backend.createStack,
     addOutput: backend.addOutput,
+    stack: backend.mainStack,
```
sobolk, line 162:
```suggestion
    stack: backend.stack,
```

if we're going with just `stack` let's keep naming consistent on backend factory as well.

sobolk (commented):
Looks good.

sobolk (approved):

packages/backend-function/src/factory.test.ts
```diff
@@ -504,7 +510,7 @@ void describe('AmplifyFunctionFactory', () => {
       name: 'anotherName',
     });
     const functionStack = Stack.of(
-      functionFactory.getInstance(getInstanceProps).resources.lambda
+      functionFactory.getInstance(getInstanceProps).stack
     );
```
Amplifiyer, line 514: nit: do we need `Stack.of` here?

Amplifiyer (approved):

packages/backend-function/src/factory.test.ts
ShadowCat567: no, we don't, that would be a quick thing for me to fix since that looks like the only instance where I didn't properly remove `Stack.of`

Amplifiyer (approved):

sobolk (approved):

sobolk:
> Any reason the data category was not included in this PR for lifting the stack parameter?

@Amplifiyer It's coming . See https://github.com/aws-amplify/amplify-category-api/pull/2899 

We typed the factory with full `AmplifyData` instead of interfaces [here](https://github.com/aws-amplify/amplify-backend/blob/b7ac6a303a85a306d375b4ef55a375e4aac45ab6/packages/backend-data/src/factory.ts#L325). Therefore, it was not possible to handle that property in factory alone.

PR #2028, opened 2024-09-19
correctly handle stack argument for generate schema command

rtpascual:
## Problem

When passing stack option for `generate schema-from-database` command, secret client `getSecret` is expecting backend identifier to be like https://github.com/aws-amplify/amplify-backend/blob/main/packages/plugin-types/src/backend_identifier.ts#L17 but it is resolved to `{ stackName: '<stack>' }`. So we fail to convert the backend identifier to parameter path strings for connection URI and SSL Cert secrets.

## Changes

If stack is passed, we correctly convert to `BackendIdentifier` before getting secrets.

## Validation

Updated test.

## Checklist

### DIFF
--- a/packages/cli/src/backend-identifier/backend_identifier_resolver.ts
+++ b/packages/cli/src/backend-identifier/backend_identifier_resolver.ts
@@ -1,6 +1,6 @@
 import { DeployedBackendIdentifier } from '@aws-amplify/deployed-backend-client';
 import { NamespaceResolver } from './local_namespace_resolver.js';
-import { BackendIdentifier } from '@aws-amplify/plugin-types';
+import { BackendIdentifier, DeploymentType } from '@aws-amplify/plugin-types';
 import { BackendIdentifierConversions } from '@aws-amplify/platform-core';
 
 export type BackendIdentifierParameters = {
@@ -61,7 +61,7 @@ export class AppBackendIdentifierResolver implements BackendIdentifierResolver {
       return {
         namespace: deployedBackendId.appName,
         name: deployedBackendId.branchName,
-        type: 'branch' as 'sandbox' | 'branch',
+        type: 'branch' as DeploymentType,
       };
     }
 
--- a/packages/cli/src/backend-identifier/backend_identifier_resolver.ts
+++ b/packages/cli/src/backend-identifier/backend_identifier_resolver.ts
@@ -1,5 +1,7 @@
 import { DeployedBackendIdentifier } from '@aws-amplify/deployed-backend-client';
 import { NamespaceResolver } from './local_namespace_resolver.js';
+import { BackendIdentifier } from '@aws-amplify/plugin-types';
+import { BackendIdentifierConversions } from '@aws-amplify/platform-core';
 
 export type BackendIdentifierParameters = {
   stack?: string;
@@ -11,6 +13,9 @@ export type BackendIdentifierResolver = {
   resolve: (
     args: BackendIdentifierParameters
   ) => Promise<DeployedBackendIdentifier | undefined>;
+  resolveDeployedBackendIdToBackendId: (
+    deployedBackendId?: DeployedBackendIdentifier
+  ) => Promise<BackendIdentifier | undefined>;
 };
 
 /**
@@ -40,5 +45,26 @@ export class AppBackendIdentifierResolver implements BackendIdentifierResolver {
       };
     }
     return undefined;
+  };
+  resolveDeployedBackendIdToBackendId = async (
+    deployedBackendId?: DeployedBackendIdentifier
+  ): Promise<BackendIdentifier | undefined> => {
+    if (!deployedBackendId) {
+      return undefined;
+    }
+
+    if ('stackName' in deployedBackendId) {
+      return BackendIdentifierConversions.fromStackName(
+        deployedBackendId.stackName
+      );
+    } else if ('appName' in deployedBackendId) {
+      return {
+        namespace: deployedBackendId.appName,
+        name: deployedBackendId.branchName,
+        type: 'branch' as 'sandbox' | 'branch',
+      };
+    }
+
+    return deployedBackendId;
   };
 }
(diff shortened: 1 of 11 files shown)


packages/cli/src/commands/generate/schema-from-database/generate_schema_command.ts
```diff
@@ -67,8 +70,13 @@ export class GenerateSchemaCommand
     const connectionUriSecretName = args.connectionUriSecret as string;
     const outputFile = args.out as string;
 
+    const backendIdentifierIsStack = 'stackName' in backendIdentifier;
+    const sanitizedBackendId = backendIdentifierIsStack
+      ? BackendIdentifierConversions.fromStackName(backendIdentifier.stackName)
+      : backendIdentifier;
+
```
Amplifiyer, line 77: Can this be turned into a `convertDeployedBackendIdentifierToBackendIdentifier`? Because we have two types `DeployedBackendIdentifier` and `BackendIdentifier` this confusion or issue may happen else where as well.

packages/cli/src/commands/generate/schema-from-database/generate_schema_command.ts
```diff
@@ -67,8 +70,13 @@ export class GenerateSchemaCommand
...
     const connectionUriSecret = await this.secretClient.getSecret(
-      backendIdentifier as BackendIdentifier,
+      sanitizedBackendId as BackendIdentifier,
```
Amplifiyer, line 79: We should remove this force cast once we have an instance of `BackendIdentifier`

sobolk: Looks like type system was trying to warn about the problem in original implementation.

packages/cli/src/commands/generate/schema-from-database/generate_schema_command.ts
sobolk: Or perhaps we should refactor this and have.
`BackendIdentifierResolver.resolveDeployedBackendIdentifier`
and
`BackendIdentifierResolver.resolveBackendIdentifier` 
so that this conversion logic stays in one place.

these are all internal types to cli package so we could do that.

rtpascual: I'm leaning towards keeping it all in `BackendIdentifierResolver` for now, we can determine later if there is a need to separate this logic.

packages/cli/src/backend-identifier/backend_identifier_resolver.ts
```diff
@@ -11,6 +13,9 @@ export type BackendIdentifierResolver = {
   resolve: (
     args: BackendIdentifierParameters
   ) => Promise<DeployedBackendIdentifier | undefined>;
+  resolveDeployedBackendIdToBackendId: (
+    deployedBackendId?: DeployedBackendIdentifier
+  ) => Promise<BackendIdentifier | undefined>;
```
sobolk, line 18:
```suggestion
  resolveDeployedBackendIdToBackendId: (
    deployedBackendId: DeployedBackendIdentifier
  ) => Promise<BackendIdentifier>;
```

Should we have caller handle `undefined` ?

Or perhaps should we have 
```
resolveBackendIdentifier: (
    args: BackendIdentifierParameters
  ) => Promise<BackendIdentifier | undefined>;
```
as additional API ?

The thing is that if callers give this type `BackendIdentifierParameters` they should be getting final product instead of passing return value from one API to the other one - that solution is not much beneficial over just calling "Conversions". 

I like the later more as I type this.

rtpascual: I'm all for the latter, it just makes me question the use of the original `resolve` for other commands but looks like those are different and do require `DeployedBackendIdentifier`.

sobolk: > for other commands but looks like those are different and do require DeployedBackendIdentifier.

Yep. that's why that api has to stay. Consider renaming it though to make it less confusing.

sobolk (approved):

packages/cli/src/commands/generate/schema-from-database/generate_schema_command.ts
```diff
@@ -54,9 +53,8 @@ export class GenerateSchemaCommand
...
     args: ArgumentsCamelCase<GenerateSchemaCommandOptions>
   ): Promise<void> => {
-    const backendIdentifier = await this.backendIdentifierResolver.resolve(
-      args
-    );
+    const backendIdentifier =
+      await this.backendIdentifierResolver.resolveBackendIdentifier(args);
 
     if (!backendIdentifier) {
       throw new AmplifyFault('BackendIdentifierFault', {
```
Amplifiyer, line 60: Should this be a fault if the input is coming from the user?

rtpascual: I think this should stay as a fault because if we can't resolve using input from the user we fallback to resolving to sandbox backend ID. The only thing really controlled by the user at that point is if they have the `name` field in their `package.json` which is covered here https://github.com/aws-amplify/amplify-backend/blob/main/packages/cli/src/backend-identifier/local_namespace_resolver.ts#L27.

Amplifiyer: This is concerning. if a user provided a `--stack` argument that doesn't result in a valid identifier, why would we fallback to sandbox backend identifier? This seems like a wrong fallback when customer intention is clear.

rtpascual: True... this fallback is there for all generate commands unfortunately

rtpascual: I can follow up with a PR to not fallback if user provided stack or app ID and branch then change this to an error

Amplifiyer: Sounds good. I think in some context it might makes sense to have this fallback but maybe not in the cli commands.

PR #2031, opened 2024-09-19
fix: clearing the .amplify/generated/env/ before synthesis

vigy02:
## Problem
The `.amplify/generated/env/` directory is not cleared before deployments, causing outdated/renamed environment files to remain. This can lead to confusion and potential errors when the function's handler code does not match the current environment settings.

## Changes
- Added a method in `backend_factory.ts` to clear the `.amplify/generated/env/` directory at the start of each deployment.
- Updated existing tests to verify that the directory is cleared during deployments.

## Validation
- Manually verified that the directory is indeed cleared before and after deployments.

## Checklist
- [x] Automated test coverage for the change has been added or updated.


packages/backend/src/backend_factory.ts
```diff
@@ -98,6 +99,11 @@ export class BackendFactory<
 
     const resourceNameValidator = new DefaultResourceNameValidator();
 
+    const pathToDelete = `${process.cwd()}/.amplify/generated/env/`;
+
+    // Clear any existing files and subdirectories in the directory
+    fs.rmSync(pathToDelete, { recursive: true, force: true });
```
sobolk, line 105: Two things.

1. The path should be concatenated with `path.join` api . This means we also have this problem [here](https://github.com/aws-amplify/amplify-backend/blob/dce0518c35b21b0ba7539990e0cc983ba55c0cf4/packages/backend-function/src/function_env_type_generator.ts#L24) - could you include fix for this part in the PR ?
2. This is a function specific code. The logic should somehow go to `backend-function`. Possible solutions for that:
    1. Have this as static initializer somewhere, perhaps `FunctionEnvironmentTypeGenerator` .
    2. Expose lifycycle hooks from [ConstructFactory](https://github.com/aws-amplify/amplify-backend/blob/dce0518c35b21b0ba7539990e0cc983ba55c0cf4/packages/plugin-types/src/construct_factory.ts#L18). For example `ConstructFactory.beforeAll` (having in mind that we might add `ConstructFactory.beforeEach` some time in the future).
    
 @rtpascual @Amplifiyer wdyt ?

rtpascual: I was initially thinking 2i but we would have to be careful this won't delete the generated directory during synth of multiple functions right?

I'm wondering if we will ever see more use cases similar to this in other packages to warrant 2ii.

sobolk: > won't delete the generated directory during synth of multiple functions right?

If we make it run once in static context, before any FunctionEnvironmentTypeGenerator ctor runs then in theory this should work. But won't be super testable.

2ii seems like a proper way to hook this up into framework and control control flow explicitly.
Would we need this else where, hard to say, but we could ride on 2i until second candidate appears.

rtpascual: > If we make it run once in static context, before any FunctionEnvironmentTypeGenerator ctor runs then in theory this should work. But won't be super testable.

Oh I see what you mean now. Yeah I can only think a good way to test this is in e2e, for example we create app with function -> run sandbox-> update function name -> check for old generated file after hotswap.

Amplifiyer: I'm not sure if I understand the problem correctly and if this is the right solution for the root cause.
1. What are the confusion and potential errors when there are left-over type files?
2. Should this be handled in the shim generator to clear the files to start again https://github.com/aws-amplify/amplify-backend/blob/f33ff6b875470e8cf0135aff9082b5c18e6a59e1/packages/backend-function/src/function_env_type_generator.ts#L79?

vigy02: > I'm not sure if I understand the problem correctly and if this is the right solution for the root cause.
> 
> 1. What are the confusion and potential errors when there are left-over type files?
> 2. Should this be handled in the shim generator to clear the files to start again https://github.com/aws-amplify/amplify-backend/blob/f33ff6b875470e8cf0135aff9082b5c18e6a59e1/packages/backend-function/src/function_env_type_generator.ts#L79
>    ?

While leaving the files might not cause immediate errors, there are several potential issues that can arise:

Stale Type Definitions: If a function is removed from the project or its environment variables are changed, the old type definition file would remain. This could lead to confusion and potential errors if developers reference these outdated types.

Type Conflicts: In scenarios where function names or environment variable structures change, you might end up with conflicting type definitions. This can cause TypeScript compilation errors

Build Inconsistencies: If the build process relies on the presence or absence of these files, inconsistent clearing could lead to inconsistent builds across different environments or developers' machines.



> I'm not sure if I understand the problem correctly and if this is the right solution for the root cause.
> 
> 1. What are the confusion and potential errors when there are left-over type files?
> 2. Should this be handled in the shim generator to clear the files to start again https://github.com/aws-amplify/amplify-backend/blob/f33ff6b875470e8cf0135aff9082b5c18e6a59e1/packages/backend-function/src/function_env_type_generator.ts#L79
>    ?

1. While leaving the files might not cause immediate errors, there are several potential issues:

-  If a function is removed from the project or its environment variables are changed, the old type definition file would remain. This could lead to confusion and potential errors if developers reference these outdated types.

- Type Conflicts: In scenarios where function names or environment variable structures change, you might end up with conflicting type definitions. This can cause TypeScript compilation errors or misleading IntelliSense suggestions in your IDE.

- Accumulation of Unused Files

- If the build process relies on the presence or absence of these files, inconsistent clearing could lead to inconsistent builds across different environments or developers' machines.

2. Handling in shim generator would clear the directory for every functions thats being created.

.changeset/lovely-cars-carry.md
```diff
@@ -0,0 +1,6 @@
+---
```
sobolk, line 1: please combine all these changesets added in PR into single file.

packages/backend/src/backend_factory.test.ts
```diff
@@ -9,8 +9,8 @@ import { Bucket } from 'aws-cdk-lib/aws-s3';
 import { BackendFactory } from './backend_factory.js';
 import { App, Stack } from 'aws-cdk-lib';
 import { Template } from 'aws-cdk-lib/assertions';
-import assert from 'node:assert';
 import { ClientConfig } from '@aws-amplify/client-config';
+import assert from 'node:assert';
```
sobolk, line 13: since we're not changing `backend` package please revert this and remove `backend` from changesets.

packages/backend-function/src/function_env_type_generator.ts
```diff
@@ -84,6 +87,21 @@ export class FunctionEnvironmentTypeGenerator {
     this.writeShimFile(content);
   };
 
+  /**
+   * Clear existing files and subdirectories in the generated env directory
+   */
+  clearGeneratedEnvDirectory(): void {
```
sobolk, line 93:
```suggestion
  clearGeneratedEnvDirectory = (): void => {
```

this is the preferred notation in the repo (see other functions around).
we do have a rule for this here https://github.com/aws-amplify/amplify-backend/blob/f6b1943b42d2e1669109f3411ff3c19bae14a2c0/.eslintrc.cjs#L76C6-L76C47 . I'm not sure why it didn't flag this.

packages/backend-function/src/factory.ts
```diff
@@ -354,6 +354,9 @@ class AmplifyFunction
     const functionEnvironmentTypeGenerator =
       new FunctionEnvironmentTypeGenerator(id);
 
+    // Clear the shim generation directory
+    functionEnvironmentTypeGenerator.clearGeneratedEnvDirectory();
```
sobolk, line 358: I'm afraid this won't work for project with multiple functions.
I.e. multiple `defineFunctions` will erase others outputs.

That's why I left this feedback here https://github.com/aws-amplify/amplify-backend/pull/2031#discussion_r1767497928 .

It has to be "static initializer" somewhere. (if we go with solution 2 ii).

sobolk: and. we should work on e2e tests to assert that this works as expected in multiple functions scenario.

PR #2044, opened 2024-09-23
fix: added static context to clear the .amplify/generated/env Directory before synthesis

vigy02:
## Problem
The `.amplify/generated/env/` directory is not cleared before deployments, causing outdated/renamed environment files to remain. This can lead to confusion and potential errors when the function's handler code does not match the current environment settings.

PR #2054, opened 2024-09-25
fix: added static context to clear the .amplify/generated/env Directory before synthesis

vigy02:
Follow up PR: https://github.com/aws-amplify/amplify-backend/pull/2044

Using this PR to run the e2e test 

## Problem
The `.amplify/generated/env/` directory is not cleared before deployments, causing outdated/renamed environment files to remain. This can lead to confusion and potential errors when the function's handler code does not match the current environment settings.


packages/backend-function/src/function_env_type_generator.test.ts
```diff
@@ -69,4 +73,54 @@ void describe('FunctionEnvironmentTypeGenerator', () => {
 
     await fsp.rm(targetDirectory, { recursive: true, force: true });
   });
+  void it('clears the generated env directory', async () => {
```
awsluja, line 76: this test name should explain or comment on why a 2nd call to new FunctionEnvironmentTypeGenerator is needed here (to showcase that it handles multiple function calls correctly)

awsluja (changes requested):
Looks good just a small nit regarding test naming

packages/backend-function/src/function_env_type_generator.test.ts
vigy02: Done. Updated it

awsluja (approved):

sobolk (dismissed):

awsluja (approved):

sobolk (dismissed):

sobolk (approved):

rtpascual (approved):

PR #2070, opened 2024-10-01
Infer executable tool input type from input schema

sobolk:
## Problem

1. The current type of `ExecutableTool.inputSchema` doesn't hint how to author valid JSON schema.
2. The type we're using for `input` into `ExecutableTool.execute` is overly restrictive and doesn't leverage fact that in some cases it could be inferred from schema.

## Changes

This PR changes type for `ExecutableTool.inputSchema` as well as adds type inference for `ExecutableTool.execute`'s `input` when possible.

So that the following DX can be achieved:
image

Notes about the solution.
1. We're using OSS package https://github.com/ThomasAribart/json-schema-to-ts .
    1. It provides decent typing for JSON schema content.
    2. It implements type inference from JSON schema.
    3. It appears to be the only available solution to this problem.
    4. It is used only at compile time, not runtime. Therefore, taking dependency on it is an acceptable risk.
    5. We re-export two top level types `JSONSchema` and `FromSchema` so that these can be replaced by our own implementation if needed in the future.
2. We type `ExecutableTool` as `ExecutableTool<TJSONSchema extends JSONSchema = JSONSchema, TToolInput = FromJSONSchema<TJSONSchema>>` so that:
    1. It is possible for customer to override `input` type if necessary.
        1. When they can't define schema as TS constant. But have other means to define typings.
        2. They don't care about input typing at all. (i.e. it's opaque value, like in our graphql tools).
3. We define `createExecutableTool` factory method. So that customer can write a code where both schema and input typings are discovered by compiler dynamically. (This is impossible with just types). Customer can either use type or factory method, but we recommend factory method as an entry point.

## Limitations

Type inference is only possible when schema is defined in TS code. See https://github.com/ThomasAribart/json-schema-to-ts/blob/main/documentation/FAQs/does-json-schema-to-ts-work-on-json-file-schemas.md for detailed explanation.

Customer can escape this by providing their own input type or ignoring failed inference (which falls back to `unknown` in worst case).

## Validation

Added tests.

## Checklist


rtpascual (approved):
LGTM, any questions I had around `json-schema-to-ts` were answered in the PR description.

Amplifiyer (approved):

Issue #2058, opened 2024-09-26
support configuring function log settings


josefaidt:
### Environment information

```plain text
n/a
```

### Describe the feature

support configuring function log settings on `defineFunction`, with a pattern that can extend to other resources like Auth

```ts
import { defineFunction } from "@aws-amplify/backend"

defineFunction({
  name: "my-function",
  logging: {
	// https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda_nodejs.NodejsFunction.html#applicationloglevelv2
    level: "INFO",
	// https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda_nodejs.NodejsFunction.html#logretention
    retention: "1 week",
  },
})
```

### Use case

to customize log settings

ykethan:
Marking as feature request to support logging customization.

lukebellamy053:
I raised this MR yesterday to allow the specification of the log retention period: https://github.com/aws-amplify/amplify-backend/pull/2055 
I'd be fairly happy to extend it to support more functionality like this if desired

sobolk:
I addition to this should we have `npx ampx sandbox --stream-function-logs --log-level debug` ?

PR #2073, opened 2024-10-02
Stream conversation logs in sandbox

sobolk:
## Problem

Conversation handler logs should show up when executing `npx ampx sandbox --stream-function-logs`.

## Changes

This PR makes the following changes to stream and improve logs in conversation handler.

1. Adds logs with log level in conversation handler runtime componenets.
    1. Conversation handler lambda is configured to emit structured logs. So that it supports log levels, see https://docs.aws.amazon.com/lambda/latest/dg/nodejs-logging.html for reference. I.e. we set log format to `json` as opposed to `text` (which doesn't support application log levels).
    2. We add more `info` and `debug` logs around incoming event, bedrock invocations, tools invocation, emitting response. Debug logs dump request/response payloads.
2. We change Sandbox component to support arbitrary log group attached to lambda function.
    1. Previously Sandbox assumed that log group is `/aws/lambda/<functionName>`. This however won't work for conversation handler (and later if we expose knobs related to logging configuration.
    2. Conversation handler requires custom log group. I attempted to use `<functionName>` to mimic default convention, but the problem is that log group is input to function ctor, therefore it's impossible to use function's properties to create log group (as this happens first). The only alternative I found was to generate function name ourselves, but that wasn't great either - components used by function ctor internally are not available to us, so if we generate name it would look differently than other functions. Hence I figured that it's better to just make sandbox support arbitrary log group.
    3. The `ListTags` call in Sandbox is replaced by `GetFunction` which gives both information that we need - `tags` and `logGroupName`. The `AmplifyBackendDeployFullAccess ` already has necessary permissions to call `GetFunction`.
3. I moved outputs generation from `backend-ai` to `ai-construct`. So that data construct can pass `outputStrategy` to it in default scenario.

## Out of scope

1. Setting log level for function at runtime. This should be covered separately with https://github.com/aws-amplify/amplify-backend/issues/2058 . Meanwhile customers can adjust the level on AWS Console after deployment.
2. This PR exposes ability to pass `outputStrategy` to ai construct. The necessary plumbing on data side will be pursued separately. The idea was validated by a temporary change (reverted [here](https://github.com/aws-amplify/amplify-backend/pull/2073/commits/3e8b590a7a9d3957d0bb2a31ab5dcfa5b0c25e1a))

## Validation

1. Added/modified unit tests.
5. Manual check

Manual check with default `info` level:
image

Manual check with log level set to `debug` via AWS Console:
image
image

## Checklist

### DIFF
--- a/packages/ai-constructs/API.md
+++ b/packages/ai-constructs/API.md
@@ -6,8 +6,10 @@
 
 /// <reference types="node" />
 
+import { BackendOutputStorageStrategy } from '@aws-amplify/plugin-types';
 import * as bedrock from '@aws-sdk/client-bedrock-runtime';
 import { Construct } from 'constructs';
+import { FunctionOutput } from '@aws-amplify/backend-output-schemas';
 import { FunctionResources } from '@aws-amplify/plugin-types';
 import { ResourceProvider } from '@aws-amplify/plugin-types';
 import * as smithy from '@smithy/types';
@@ -49,6 +51,7 @@ type ConversationHandlerFunctionProps = {
         modelId: string;
         region?: string;
     }>;
+    outputStorageStrategy?: BackendOutputStorageStrategy<FunctionOutput>;
 };
 
 // @public (undocumented)
--- a/packages/ai-constructs/API.md
+++ b/packages/ai-constructs/API.md
@@ -6,10 +6,10 @@
 
 /// <reference types="node" />
 
+import { AiConversationOutput } from '@aws-amplify/backend-output-schemas';
 import { BackendOutputStorageStrategy } from '@aws-amplify/plugin-types';
 import * as bedrock from '@aws-sdk/client-bedrock-runtime';
 import { Construct } from 'constructs';
-import { FunctionOutput } from '@aws-amplify/backend-output-schemas';
 import { FunctionResources } from '@aws-amplify/plugin-types';
 import { ResourceProvider } from '@aws-amplify/plugin-types';
 import * as smithy from '@smithy/types';
@@ -51,7 +51,7 @@ type ConversationHandlerFunctionProps = {
         modelId: string;
         region?: string;
     }>;
-    outputStorageStrategy?: BackendOutputStorageStrategy<FunctionOutput>;
+    outputStorageStrategy?: BackendOutputStorageStrategy<AiConversationOutput>;
 };
 
 // @public (undocumented)
--- a/packages/ai-constructs/API.md
+++ b/packages/ai-constructs/API.md
@@ -6,7 +6,7 @@
 
 /// <reference types="node" />
 
-import { AiConversationOutput } from '@aws-amplify/backend-output-schemas';
+import { AIConversationOutput } from '@aws-amplify/backend-output-schemas';
 import { BackendOutputStorageStrategy } from '@aws-amplify/plugin-types';
 import * as bedrock from '@aws-sdk/client-bedrock-runtime';
 import { Construct } from 'constructs';
@@ -51,7 +51,7 @@ type ConversationHandlerFunctionProps = {
         modelId: string;
         region?: string;
     }>;
-    outputStorageStrategy?: BackendOutputStorageStrategy<AiConversationOutput>;
+    outputStorageStrategy?: BackendOutputStorageStrategy<AIConversationOutput>;
 };
 
 // @public (undocumented)
--- a/packages/ai-constructs/src/conversation/conversation_handler_construct.ts
+++ b/packages/ai-constructs/src/conversation/conversation_handler_construct.ts
@@ -21,7 +21,7 @@ import { Construct } from 'constructs';
 import path from 'path';
 import { TagName } from '@aws-amplify/platform-core';
 import {
-  AiConversationOutput,
+  AIConversationOutput,
   aiConversationOutputKey,
 } from '@aws-amplify/backend-output-schemas';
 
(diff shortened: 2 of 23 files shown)


packages/ai-constructs/API.md
```diff
@@ -49,6 +51,7 @@ type ConversationHandlerFunctionProps = {
         modelId: string;
         region?: string;
     }>;
+    outputStorageStrategy?: BackendOutputStorageStrategy<FunctionOutput>;
```
sobolk, line 54: This is similar to what data and auth construct do.

https://github.com/aws-amplify/amplify-backend/blob/27202b41269d029e7efc392d9f6340da36f12389/packages/auth-construct/src/types.ts#L472-L475

packages/sandbox/src/lambda_function_log_streamer.ts
```diff
@@ -54,33 +46,26 @@ export class LambdaFunctionLogStreamer {
...
-    }
-
-    for (const entry of deployedFunctionNameToArnMap) {
-      const listTagsResponse = await this.lambda.send(
-        new ListTagsCommand({
-          Resource: entry.arn,
+    for (const functionName of deployedFunctionNames) {
+      const getFunctionResponse = await this.lambda.send(
+        new GetFunctionCommand({
+          FunctionName: functionName,
```
sobolk, line 52: We already do have permissions to call `GetFunction` in `AmplifyBackendDeployFullAccess`

packages/ai-constructs/src/conversation/conversation_handler_construct.ts
```diff
@@ -105,5 +125,23 @@ export class ConversationHandlerFunction
...
+   */
+  private storeOutput = (
+    outputStorageStrategy:
+      | BackendOutputStorageStrategy<FunctionOutput>
+      | undefined
+  ): void => {
+    outputStorageStrategy?.appendToBackendOutputList(functionOutputKey, {
+      version: '1',
+      payload: {
+        definedFunctions: this.resources.lambda.functionName,
```
Amplifiyer, line 143: But this is not a `definedFunction`. We should store this under it's own AIConstructOutputKey.

packages/ai-constructs/src/conversation/conversation_handler_construct.ts
```diff
@@ -67,6 +86,7 @@ export class ConversationHandlerFunction
           // For custom entry we do bundle SDK as we can't control version customer is coding against.
           bundleAwsSDK: !!this.props.entry,
         },
+        loggingFormat: LoggingFormat.JSON,
```
Amplifiyer, line 89: To be able to actually see the debug logs, don't we need to also expose the `ApplicationLogLevel` property so customers can configure it? if I recall cdk sets the default as `INFO`?

packages/sandbox/src/lambda_function_log_streamer.ts
```diff
@@ -54,33 +46,26 @@ export class LambdaFunctionLogStreamer {
...
         })
       );
+      const logGroupName =
+        getFunctionResponse.Configuration?.LoggingConfig?.LogGroup;
+      if (!logGroupName) {
+        this.printer.log(
+          `[Sandbox] Could not log group for lambda function ${functionName}. Logs will not be streamed for this function.`,
```
Amplifiyer, line 59:
```suggestion
          `[Sandbox] Could not find logGroup for lambda function ${functionName}. Logs will not be streamed for this function.`,
```

packages/sandbox/src/lambda_function_log_streamer.ts
Amplifiyer: The reason we went with `ListTags` earlier was for performance reasons, since it will be called after each sandbox deployment.

packages/ai-constructs/src/conversation/conversation_handler_construct.ts
sobolk: Correct. I called this out in the "Out of scope" section in the PR description.

packages/sandbox/src/lambda_function_log_streamer.ts
sobolk: Was `ListTags` significantly faster than `GetFunction` ?

Some considerations:
1. The number of requests to lambda CP doesn't change. We were calling `ListTags` per each function.
2. We eliminated extra CloudFormation calls to resolve ARNs.
3. The alternative would be to call `GetFunctionConfiguration` in addition to `ListTags`. But, making two requests is not great quota and latency wise. 
4. It seems that default quota for `GetFunction` is much more generous than other CP APIs. See picture below.

image

sobolk: Synced offline. We'll keep `GetFunction` given that alternatives are not better.

packages/backend-output-schemas/src/index.ts
```diff
@@ -84,6 +85,20 @@ export * from './function/index.js';
...
+
+/**
+ * re-export the AI conversation output schema
+ */
+export * from './ai/conversation/index.js';
+
+/**
+ * Expected key that AI conversation output is stored under
+ */
+export const aiConversationOutputKey = 'AWS::Amplify::AI::Conversation';
```
sobolk, line 100: the naming here follows `ai-constructs/<sub_usecase>` 
I.e. if we add another AI feature not related to conversation we'd have `AWS::Amplify::AI::SomethingNew`

Amplifiyer (approved):

rtpascual (approved):

PR #2076, opened 2024-10-03
Reference auth

awsluja:
## Problem

## Changes

## Checklist


packages/backend-auth/package.json
```diff
@@ -20,6 +20,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@aws-amplify/auth-construct": "^1.3.1",
+    "@aws-amplify/reference-auth-construct": "^0.1.0",
```
sobolk, line 23: Do we have to create `@aws-amplify/reference-auth-construct` ? I.e. is there a desire to use it in raw CDK apps?

It could be hosted internally in `backend-auth` instead of creating new package.

So far in our repo we have:
1. AuthConstruct - public because there was requirement to support raw CDK.
2. ConversationHandler - public because it needs to be visible on data repo side.
3. AmplifyFunction - private/internal, because there was no need to make it public
4. AmplifyStorage - private/internal, because there was no need to make it public

I suggest we make ReferenceAuthConstruct private initially. We can move it later to separate package if needed. Less public API is better.

sobolk (commented):
Overall is looking good.
Left a comment about packaging.

packages/backend-auth/package.json
awsluja: We could make it private; I don't think we have a use case for making it public unless customers want to take advantage of the custom resource logic

packages/reference-auth-construct/src/lambda/reference_auth_initializer.ts
```diff
@@ -0,0 +1,239 @@
...
+  | CloudFormationCustomResourceFailedResponse
+> => {
+  console.info(`Received '${event.RequestType}' event`);
+  const physicalId =
+    event.RequestType === 'Create' ? randomUUID() : event.PhysicalResourceId;
+
+  let data;
+  if (!(event.RequestType === 'Update' || event.RequestType === 'Create')) {
+    data = await initialize(event);
+  }
```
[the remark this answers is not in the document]
awsluja: I forgot to undo this if clause because I was testing something else; going to push the correct settings here shortly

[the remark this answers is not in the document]
awsluja: in my local testing, switching userpools/identitypools in sandbox mode works fine

packages/reference-auth-construct/src/lambda/reference_auth_initializer.ts
```diff
@@ -0,0 +1,239 @@
...
+import { randomUUID } from 'node:crypto';
+import { ReferenceAuthInitializerProps } from './reference_auth_initializer_types.js';
+import { AuthOutput } from '@aws-amplify/backend-output-schemas';
+
+/**
+ * Entry point for the lambda-backend custom resource to retrieve a backend secret.
+ */
+export const handler = async (
+  event: CloudFormationCustomResourceEvent
+): Promise<
```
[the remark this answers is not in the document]
awsluja: just added coverage for this, PTAL

packages/reference-auth-construct/src/lambda/reference_auth_initializer.ts
```diff
@@ -0,0 +1,239 @@
...
+  const userPoolResponse = await userPoolClient.send(userPoolCommand);
+  const userPool = userPoolResponse.UserPool;
+  if (!userPool) {
+    return undefined;
+  }
+  const mfaCommand = new GetUserPoolMfaConfigCommand({
+    UserPoolId: userPoolId,
+  });
+  const mfaResponse = await userPoolClient.send(mfaCommand);
+  // TBD - what to do with userpool containing lots of providers? ie, pagination
```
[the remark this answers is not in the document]
awsluja: updated to support pagination

packages/backend-auth/package.json
awsluja: Updated , reference-auth-construct is removed, code lives within auth-construct

packages/auth-construct/src/reference_construct.test.ts
```diff
@@ -0,0 +1,186 @@
...
+    'arn:aws:iam::000000000000:role/amplify-sample-unauth-role-name',
+  identityPoolId: 'us-east-1:identityPoolId',
+  userPoolClientId: 'userPoolClientId',
+  userPoolId: 'us-east-1_userPoolId',
+};
+
+void describe('AmplifyConstruct', () => {
+  // beforeEach(() => {
+
+  // })
```
sobolk, line 27: please remove commented out code.

packages/backend-auth/package.json
sobolk: > code lives within auth-construct

This doesn't make it private, i.e. it's still exported somewhere. But definitely makes situation better.

To make this private we'd need to move the construct to `backend-auth` and not export it there. It's fine to have constructs in `backend-X` if they're supposed to be private (see functions, storage).

However, perhaps there's no harm in having this in `auth-construct` and saying that we support reference auth in raw CDK context as well. It's just that we sign up to support extra public API.

@Amplifiyer @josefaidt wdyt ?

PR #2083, opened 2024-10-07
Fix case when multiple tools are requested by Bedrock.

sobolk:
## Problem

Currently. We append a new message if bedrock emits multiple tool use blocks in single bedrock response.
However, bedrock then validates that we return it a single array with tool responses. NOT multiple messages with tool responses.

I.e. we want to fix this error.
image

## Changes

Accumulate tool responses into single "user message" instead of multiple messages.

## Validation

Enhanced both unit and e2e test to be sensitive to this scenario.

## Checklist

### DIFF
--- a/packages/integration-tests/src/test-projects/conversation-handler/amplify/custom-conversation-handler/custom_handler.ts
+++ b/packages/integration-tests/src/test-projects/conversation-handler/amplify/custom-conversation-handler/custom_handler.ts
@@ -3,7 +3,7 @@ import {
   createExecutableTool,
   handleConversationTurnEvent,
 } from '@aws-amplify/backend-ai/conversation/runtime';
-import { expectedTemperatureInProgrammaticToolScenario } from '../constants.js';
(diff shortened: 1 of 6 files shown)


packages/integration-tests/src/test-projects/conversation-handler/amplify/custom-conversation-handler/custom_handler.ts
```diff
@@ -20,12 +20,13 @@ const thermometer = createExecutableTool(
     json: thermometerInputSchema,
   },
   (input) => {
-    if (input.city === 'Seattle') {
+    const city = input.city;
+    if (city === 'Seattle' || city === 'Boston' || city === 'Miami') {
```
sobolk, line 24: had to do this so that compiler doesn't complain about indexing the record.

awsluja (approved):

rtpascual (approved):

PR #2084, opened 2024-10-07
force esm via tsconfig, remove package.json creation

josefaidt:
## Problem

The purpose of the existing `amplify/package.json` file is to force the backend project to be an ESM project, and treat all backend project files as ES modules. The existence of this file and the one created at the project root (i.e. above `amplify/`) can be confusing as it is often used to define dependencies. There have been a few reports of the confusing experience as modules will resolve as they are installed during sandbox, but are not resolved during branch builds using the default configuration, and customer branch deployments fail.

This change is largely from the guidance from the [official tsx docs's recommendation](https://tsx.is/typescript#recommendation)

```jsonc
{
	"compilerOptions": {

		// Treat files as modules even if it doesn't use import/export
		"moduleDetection": "force",

		// Ignore module structure
		"module": "Preserve",

		// Allow JSON modules to be imported
		"resolveJsonModule": true,

		// Allow JS files to be imported from TS and vice versa
		"allowJs": true,

		// Use correct ESM import behavior
		"esModuleInterop": true,

		// Disallow features that require cross-file awareness
		"isolatedModules": true,
	},
}
```

This change:
- changes `module` from `es2022` to `preserve`
- removes `moduleResolution`
- adds [`moduleDetection`](https://www.typescriptlang.org/tsconfig/#moduleDetection) with `force`
- removes `package.json` creation

Existing customers will not be impacted, and no changes are required on their part, however later it will be noticed that `create-amplify` no longer creates the same files or configuration. 

Moving forward, would it benefit us from creating the `tsconfig.json` inside `.amplify/` (creating this file preemptively), and `extend`-ing it from the customer's `amplify/` directory? This would allow us to make further changes to this file and create on demand since it is gitignored.

## Changes

## Validation

I created a sample project and modified it in place, 
https://github.com/josefaidt/repro-amplify-gen2-moduledetection/commit/076d870a59d3e52cd1605ce65b93b1265623ab31

## Checklist


packages/cli-core/src/package-manager-controller/package_manager_controller_base.ts
```diff
@@ -96,8 +96,8 @@ export abstract class PackageManagerControllerBase
     const tsConfigTemplate = {
       compilerOptions: {
         target: 'es2022',
-        module: 'es2022',
-        moduleResolution: 'bundler',
+        module: 'preserve',
```
sobolk, line 99: Few things here.

1. TS docs mention `esnext` next to `preserve`: `You very likely want "nodenext" for modern Node.js projects and preserve or esnext for code that will be bundled` . Why would we choose `preserve` over `esnext` ? (unless bundling doesn't matter here anymore).
2. `preserve` was added in Typescript 5.4 . If we do this we have to bump Typescript to `^5.4.0`. Including here https://github.com/aws-amplify/amplify-backend/blob/7bf0c64ea43a01e220109d30e98a353864457bf7/packages/create-amplify/src/amplify_project_creator.ts#L21

sobolk (commented):
Can you run e2e tests on this PR?
And make sure all checks are passing?

packages/cli-core/src/package-manager-controller/package_manager_controller_base.ts
josefaidt: for 1, bundling doesn't matter here aside from Functions. `nodenext` is the one that'll force you to specify `.js` file extensions (though I think there's a new compiler option coming that'll support `.ts` extensions and transform). `preserve` will not transform your module syntax, which should yield the same results when writing with ESM syntax using import/export instead of require/module.exports

for 2, bumped the minimum version

PR #2087, opened 2024-10-07
feat: Add support for custom Lambda function email senders in Auth construct

vigy02:
**Problem**

This PR addresses the need for more flexible email sender configuration in the Auth construct, specifically allowing the use of custom Lambda functions for email sending capabilities.

**Changes**

* Added support for using Lambda functions as email senders in the Auth construct
* Implemented KMS key management for custom email senders
* Updated types and interfaces to accommodate new email sender options
* Added new test case for custom email sender functionality
* Modified existing code to handle Lambda function email senders, including permission setup
* Introduced functionality for attaching a user's KMS key to the customEmailSender for message encryption and decryption.  
* Added integration tests to verify customEmailSender functionality.

**Validation**

A new test case was added  to verify the behavior when a function is provided as an email sender. This test ensures that the custom email sender is properly set up when a Lambda function is provided.

** Manual Validation for KMS-ARN pass check **
Screenshot 2024-10-24 at 6 19 55 PM
Screenshot 2024-10-24 at 6 19 45 PM
Screenshot 2024-10-24 at 6 19 36 PM

**Checklist**

* [x] If this PR includes a functional change to the runtime behavior of the code, I have added or updated automated test coverage for this change.


packages/auth-construct/src/types.ts
```diff
@@ -421,7 +425,10 @@ export type AuthProps = {
      * SES configurations enable the use of customized email sender addresses and names
      * @see https://docs.amplify.aws/react/build-a-backend/auth/moving-to-production/#email
```
ShadowCat567, line 426: You will probably want to update this docs section with the parts that I left out when initially implementing senders -- 
this was the full docs message:
     Configure Cognito to send emails from SES or a custom message trigger
     SES configurations enable the use of customized email sender addresses and names
     Custom message triggers enable the use of third-party email providers when sending email notifications to users
     @see https://docs.amplify.aws/react/build-a-backend/auth/moving-to-production/#email
     @see https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-custom-email-sender.html

packages/auth-construct/API.md
```diff
@@ -47,7 +49,7 @@ export type AuthProps = {
         externalProviders?: ExternalProviderOptions;
     };
     senders?: {
-        email: Pick<UserPoolSESOptions, 'fromEmail' | 'fromName' | 'replyTo'>;
+        email: Pick<UserPoolSESOptions, 'fromEmail' | 'fromName' | 'replyTo'> | IFunction | AmplifyFunction;
```
sobolk, line 52: `AmplifyFunction` should be an input to `defineAuth` not to the construct.

There are couple of examples is `backend-data` that may be looked at.
For example, take a look how these two are handled 
1. https://github.com/aws-amplify/amplify-backend/blob/6e4a62f1d0fe98e4a96cdc44c6d6b40cd9269397/packages/backend-data/src/types.ts#L41
2. https://github.com/aws-amplify/amplify-backend/blob/6e4a62f1d0fe98e4a96cdc44c6d6b40cd9269397/packages/backend-data/src/types.ts#L141


It should be that

```
const someFunction = defineFunction();
const auth = defineAuth({ senders: { email: someFunction });
```


under the hood. the auth construct factory should resolve the reference to IFunction (see how data does this under the hood) and pass it to auth construct as `IFunction`.

packages/auth-construct/src/construct.ts
```diff
@@ -141,24 +149,44 @@ export class AmplifyAuth
...
+
+    const userPoolProps = {
+      ...this.computedUserPoolProps,
+    };
+
+    if (
+      props.senders?.email &&
+      props.senders.email instanceof lambda.Function
+    ) {
+      if (!this.customSenderKmsKey) {
```
sobolk, line 165: does this have to be an instance variable?

packages/auth-construct/src/construct.ts
```diff
@@ -141,24 +149,44 @@ export class AmplifyAuth
...
+        this.customSenderKmsKey = new Key(this, 'CustomSenderKey', {
+          description: 'KMS key for Cognito custom sender',
```
sobolk, line 167: please include user pool name (or some other identifier) in the description.

.changeset/unlucky-chairs-train.md
```diff
@@ -0,0 +1,6 @@
+---
+'@aws-amplify/auth-construct': patch
+'@aws-amplify/backend-auth': patch
+---
+
+chore: added new changesets
```
sobolk, line 6: This should describe functionality being added.

.changeset/unlucky-chairs-train.md
```diff
@@ -0,0 +1,6 @@
+---
+'@aws-amplify/auth-construct': patch
+'@aws-amplify/backend-auth': patch
```
sobolk, line 3:
```suggestion
'@aws-amplify/auth-construct': minor
'@aws-amplify/backend-auth': minor
```

See https://semver.org/

packages/auth-construct/src/construct.ts
```diff
@@ -141,24 +146,41 @@ export class AmplifyAuth
...
+      props.senders?.email &&
+      props.senders.email instanceof lambda.Function
+    ) {
+      if (!customSenderKmsKey) {
+        customSenderKmsKey = new Key(this, 'CustomSenderKey', {
+          enableKeyRotation: true,
+        });
+      }
+      customSenderKmsKey.grantDecrypt(props.senders.email);
+      props.senders.email.addPermission('CognitoInvokeEmail', {
```
sobolk, line 168:
```suggestion
      props.senders.email.addPermission('CognitoInvokeEmailSenderHandler', {
```

or something more descriptive.

packages/auth-construct/src/construct.ts
```diff
@@ -141,24 +146,41 @@ export class AmplifyAuth
...
+        principal: new ServicePrincipal('cognito-idp.amazonaws.com'),
+        action: 'lambda:InvokeFunction',
```
sobolk, line 170: we should narrow down this policy to have conditions on account and user pool arn.

See https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pools-working-with-lambda-triggers.html how to do this.

if we don't this lambda will be callable from any user pool.

sobolk (commented):
This is looking good.

Few things to look at before we can merge:
1. I left a comment about narrowing down permissions, this one is important.
2. We should see if this can be e2e tested. please take a stab at it.
3. Changesets - this is important as well.

packages/auth-construct/src/construct.ts
```diff
@@ -141,24 +146,47 @@ export class AmplifyAuth
...
     this.name = props.name ?? '';
     this.domainPrefix = props.loginWith.externalProviders?.domainPrefix;
-
     // UserPool
     this.computedUserPoolProps = this.getUserPoolProps(props);
+
+    const userPoolProps = {
+      ...this.computedUserPoolProps,
+    };
+    let customSenderKmsKey: Key | undefined;
```
awsluja, line 157: where is this being used?

packages/auth-construct/src/construct.ts
```diff
@@ -141,24 +146,47 @@ export class AmplifyAuth
...
+      const accountId = stack.account;
+      const userPoolArn = this.userPool.userPoolArn;
+
+      props.senders.email.addPermission('CognitoInvokeEmailSenderHandler', {
+        principal: new ServicePrincipal('cognito-idp.amazonaws.com'),
+        action: 'lambda:InvokeFunction',
+        sourceAccount: accountId,
+        sourceArn: userPoolArn,
+      });
+      userPoolProps.customSenderKmsKey = customSenderKmsKey;
```
awsluja, line 178: oh i see.. the code above should probably just live inside of getUserPoolProps, and you should update the types there so we don't clutter the constructor here

packages/auth-construct/src/construct.ts
```diff
@@ -504,13 +531,23 @@ export class AmplifyAuth
...
       },
       email: props.senders
-        ? cognito.UserPoolEmail.withSES({
-            fromEmail: props.senders.email.fromEmail,
-            fromName: props.senders.email.fromName,
-            replyTo: props.senders.email.replyTo,
-            sesRegion: Stack.of(this).region,
-          })
+        ? (() => {
+            if ('fromEmail' in props.senders.email) {
```
awsluja, line 535:
```suggestion
            email: (props.senders && 'fromEmail' in props.senders.email) ? <ses stuff> : undefined
```

packages/auth-construct/src/construct.ts
```diff
@@ -504,13 +531,23 @@ export class AmplifyAuth
...
+                fromEmail: props.senders.email.fromEmail,
+                fromName: props.senders.email.fromName,
+                replyTo: props.senders.email.replyTo,
+                sesRegion: Stack.of(this).region,
+              });
+            }
+            return undefined;
+          })()
         : undefined,
+      lambdaTriggers: {
```
awsluja, line 546: we should have a test to make sure that setting lambdaTriggers to {} does not remove any triggers that get added elsewhere

packages/backend-auth/src/factory.test.ts
```diff
@@ -355,6 +355,85 @@ void describe('AmplifyAuthFactory', () => {
...
+        };
+      },
+    };
+
+    resetFactoryCount();
+
+    const authWithTriggerFactory = defineAuth({
+      loginWith: { email: true },
+      //if senders aren't passed its essentially lambdaTriggers={}
+      triggers: { preSignUp: funcStub },
```
awsluja, line 423: this looks good, but it should also include email senders function here. this is so we can test there are no conflicts between the two

awsluja: ie, make sure the trigger exists for presignup, and also the custom sender function

packages/auth-construct/src/construct.ts
```diff
@@ -141,24 +148,43 @@ export class AmplifyAuth
...
+    /**
+     * Configure custom email sender for Cognito User Pool
+     * Grant necessary permissions for Lambda function to decrypt emails
+     * and allow Cognito to invoke the Lambda function
+     */
+    if (
+      props.senders?.email &&
+      props.senders.email instanceof lambda.Function &&
+      this.customEmailSenderKmsKey
+    ) {
```
awsluja, line 170: I don't know if all of this code is necessary if we use .addTrigger on this.userPool, for example:

this.userPool.addTrigger(UserPoolOperation.of('customEmailSender'), props.senders.email)

awsluja: You can check to see if the template includes the necessary IAM policies and KMS key resource. 
Its essentially what we do today for the other triggers

packages/backend-ai/API.md
```diff
@@ -1,74 +1,74 @@
...
+const handleConversationTurnEvent: (event: ConversationTurnEvent, props?: {
+    tools?: Array<ExecutableTool<runtime.JSONSchema, any>>;
+}) => Promise<void>;
+
+// @public (undocumented)
+type ToolResultContentBlock = runtime.ToolResultContentBlock;
+
+// (No @packageDocumentation comment for this package)
+
+```
```
awsluja, line 74: im not sure why this file is here, perhaps you have to pull from main again

packages/backend-auth/src/factory.test.ts
```diff
@@ -355,6 +355,85 @@ void describe('AmplifyAuthFactory', () => {
...
+
+    const template = Template.fromStack(backendAuth.stack);
+    template.hasResourceProperties('AWS::Cognito::UserPool', {
+      LambdaConfig: {
+        PreSignUp: {
+          Ref: Match.stringLikeRegexp('testFunc'),
+        },
+      },
+    });
+  });
```
awsluja, line 436: can you also verify that the KMS key resource exists ?

packages/backend-ai/API.md
sobolk: yep, this file shouldn't be here.
and it seems it has whitespace changes.

for the purpose of this pr just remove this from diff please.

to prevent this from happening you can try the following:
1. What @awsluja said above - merge latest main
2. Clean - rebuild
3. Check these settings https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings (the diff seems to be whitespace changes).

.eslint_dictionary.json
```diff
@@ -193,5 +193,7 @@
   "xlarge",
   "yaml",
   "yargs",
-  "zoneinfo"
+  "zoneinfo",
+  "SKey",
+  "decrypt"
```
sobolk, line 198: please insert those in alphabetical order.

packages/ai-constructs/API.md
```diff
@@ -54,6 +54,7 @@ type ConversationHandlerFunctionProps = {
         modelId: string;
         region?: string;
     }>;
+    memoryMB?: number;
```
sobolk, line 57: you might need to merge latest main. this change shouldn't be on the diff.

packages/auth-construct/API.md
```diff
@@ -47,7 +48,8 @@ export type AuthProps = {
         externalProviders?: ExternalProviderOptions;
     };
     senders?: {
-        email: Pick<UserPoolSESOptions, 'fromEmail' | 'fromName' | 'replyTo'>;
+        email: Pick<UserPoolSESOptions, 'fromEmail' | 'fromName' | 'replyTo'> | IFunction;
+        kmsKeyArn?: string;
```
sobolk, line 52: Is `kmsKeyArn` applicable to any of `Pick<UserPoolSESOptions, 'fromEmail' | 'fromName' | 'replyTo'>` ?

If not then we should introduce type for custom sender.
Something like:
`type CustomEmailSender = { handler: IFunction, kmsKeyArn?: string }` and then use that instead of `IFunction` in the union.

Or perhaps have `email: Pick<UserPoolSESOptions, 'fromEmail' | 'fromName' | 'replyTo'> | IFunction | CustomEmailSender` so that passing function without kmskey is still easy.

The bottom line is that we should find typing that doesn't let you specify `kmsKeyArn` if `IFunction` is not used.

@josefaidt wdyt ?

packages/backend-auth/API.md
```diff
@@ -36,10 +38,14 @@ export type AmazonProviderFactoryProps = Omit<AmazonProviderProps, 'clientId' |
...
 
 // @public (undocumented)
-export type AmplifyAuthProps = Expand<Omit<AuthProps, 'outputStorageStrategy' | 'loginWith'> & {
+export type AmplifyAuthProps = Expand<Omit<AuthProps, 'outputStorageStrategy' | 'loginWith' | 'senders'> & {
     loginWith: Expand<AuthLoginWithFactoryProps>;
     triggers?: Partial<Record<TriggerEvent, ConstructFactory<ResourceProvider<FunctionResources>>>>;
     access?: AuthAccessGenerator;
+    senders?: {
+        email: Pick<UserPoolSESOptions, 'fromEmail' | 'fromName' | 'replyTo'> | ConstructFactory<AmplifyFunction>;
+        kmsKeyArn?: string;
```
sobolk, line 47: Similar feedback here. We should find better typing.

packages/integration-tests/src/test-project-setup/data_storage_auth_with_triggers.ts
```diff
@@ -263,11 +263,18 @@ class DataStorageAuthWithTriggerTestProject extends TestProjectBase {
       (name) => name.includes('funcNoMinify')
     );
 
+    const funcCustomSender = await this.resourceFinder.findByBackendIdentifier(
+      backendId,
+      'AWS::Lambda::Function',
+      (name) => name.includes('say-hello')
+    ); // TODO - add func name
```
sobolk, line 270: yes, we should name this function something different.

packages/integration-tests/src/test-project-setup/data_storage_auth_with_triggers.ts
```diff
@@ -280,6 +287,7 @@ class DataStorageAuthWithTriggerTestProject extends TestProjectBase {
     await this.checkLambdaResponse(node16Lambda[0], expectedResponse);
     await this.checkLambdaResponse(funcWithSsm[0], 'It is working');
     await this.checkLambdaResponse(funcWithAwsSdk[0], 'It is working');
+    await this.checkLambdaResponse(funcCustomSender[0], 'Hello, World!');
```
sobolk, line 290: This is not good assertion for this feature. It just sends request directly to lambda. We should be asserting that communication between cognito and lambda works.

How to make this happen?:
1. We need to execute an operation on user pool that triggers email sending.
2. We need a function that can signal a fact that it was called by something and await that.
    1. Please take a look at solution we have for "scheduled functions". We use SQS queue to capture the fact that function was called.
        1. https://github.com/aws-amplify/amplify-backend/blob/10ef35d2fe00025fefd12603bfb2e3dce696be9a/packages/integration-tests/src/test-projects/data-storage-auth-with-triggers-ts/amplify/func-src/handler_with_aws_sqs.ts#L20-L25
        2. https://github.com/aws-amplify/amplify-backend/blob/10ef35d2fe00025fefd12603bfb2e3dce696be9a/packages/integration-tests/src/test-project-setup/data_storage_auth_with_triggers.ts#L614
        3. And fail with timeout (if sqs message doesn't arrive),

packages/auth-construct/API.md
awsluja: this is a good point, i think this would be enough:
email: Pick<UserPoolSESOptions, 'fromEmail' | 'fromName' | 'replyTo'> | CustomEmailSender

josefaidt: feedback makes sense. I like this take
```
email: Pick<UserPoolSESOptions, 'fromEmail' | 'fromName' | 'replyTo'> | IFunction | CustomEmailSender
```
this is also simple, and easier to understand that you only have two valid configuration options (ses or a sender function with/without a custom kms key)
```
email: Pick<UserPoolSESOptions, 'fromEmail' | 'fromName' | 'replyTo'> | CustomEmailSender
```

packages/auth-construct/API.md
```diff
@@ -84,6 +85,12 @@ export type CustomAttributeString = CustomAttributeBase & StringAttributeConstra
     dataType: 'String';
 };
 
+// @public
+export type CustomEmailSenderConstruct = {
```
sobolk, line 89:
```suggestion
export type CustomEmailSender = {

```

or `CustomEmailSenderOptions`

This is not a construct. It has a reference to IFunction inside but this type is not implementing `Construct` super class.

packages/integration-tests/src/test-project-setup/data_storage_auth_with_triggers.ts
```diff
@@ -642,4 +653,84 @@ class DataStorageAuthWithTriggerTestProject extends TestProjectBase {
...
+  private assertCustomEmailSenderWorks = async (
+    backendId: BackendIdentifier
+  ) => {
+    const TIMEOUT_MS = 1000 * 60 * 2; // 2 minutes
+    const startTime = Date.now();
+    const queue = await this.resourceFinder.findByBackendIdentifier(
+      backendId,
+      'AWS::SQS::Queue',
+      (name) => name.includes('testFuncQueue')
+    );
```
sobolk, line 666: We should add separate queue for this scenario.

Otherwise messages from email and scheduled functions will land in the same queue and we can't tell which one is which.

packages/integration-tests/src/test-project-setup/data_storage_auth_with_triggers.ts
```diff
@@ -642,4 +653,84 @@ class DataStorageAuthWithTriggerTestProject extends TestProjectBase {
...
+
+    const username = `testuser_${Date.now()}@example.com`;
+    const password = 'TestPassword123!';
+
+    await this.cognitoClient.send(
+      new AdminCreateUserCommand({
+        UserPoolId: userPoolId[0],
+        Username: username,
+        TemporaryPassword: password,
+        MessageAction: 'SUPPRESS',
```
sobolk, line 727: Does Cognito try to send email if we suppress messaging ?

packages/integration-tests/src/test-projects/data-storage-auth-with-triggers-ts/amplify/auth/resource.ts
```diff
@@ -1,5 +1,9 @@
 import { defineAuth, secret } from '@aws-amplify/backend';
-import { defaultNodeFunc } from '../function.js';
+import { defaultNodeFunc, funcWithSchedule } from '../function.js';
+
+const customEmailSenderFunction = {
+  handler: funcWithSchedule,
+};
```
sobolk, line 6: Related to my previous comments.

We shouldn't reuse `funcWithSchedule` here. Because we'll be blending together two independent scenarios.

We should create a function similar to `funcWithSchedule` with it's own associated queue. I.e. create similar but isolated setup.

packages/integration-tests/src/test-project-setup/data_storage_auth_with_triggers.ts
sobolk: It seems that sequencing of post deployment assertions help. But it's very implicit assumption. Using separate resources for both scenarios will remove that trap.

packages/auth-construct/src/types.ts
```diff
@@ -380,6 +381,14 @@ export type CustomAttribute =
 export type UserAttributes = StandardAttributes &
   Record<`custom:${string}`, CustomAttribute>;
 
+/**
+ * CustomEmailSender type for configuring a custom Lambda function for email sending
+ */
+export type CustomEmailSenderConstruct = {
```
awsluja, line 387: same as above, this should not end with *Construct in the name

packages/integration-tests/src/test-projects/data-storage-auth-with-triggers-ts/amplify/auth/resource.ts
```diff
@@ -1,5 +1,9 @@
 import { defineAuth, secret } from '@aws-amplify/backend';
-import { defaultNodeFunc } from '../function.js';
+import { defaultNodeFunc, funcWithSchedule } from '../function.js';
+
+const customEmailSenderFunction = {
+  handler: funcWithSchedule,
```
rtpascual, line 5: I think we should use another function as `funcWithSchedule` has a cron job to invoke it every minute which may make testing this inaccurate.

rtpascual: We would probably want a new/different function with same SQS setup for a different queue and no schedule set. That way only message in the queue will be from when you create a new user in `triggerEmailSending` and not from scheduled invocations.

sobolk (approved):

awsluja (approved):

PR #2122, opened 2024-10-17
chore: add working setup for e2e resources in ref auth

awsluja:
## Problem
Deployment Test
image

Sandbox Test
image

## Changes

## Checklist


packages/integration-tests/src/test-project-setup/reference_auth_project.ts
```diff
@@ -0,0 +1,329 @@
...
+
+    // generate resources
+    const {
+      userPool,
+      userPoolClient,
+      identityPool,
+      authRole,
+      unauthRole,
+      adminGroup,
+    } = await project.setupTestResources();
```
sobolk, line 61: Doing this before `deploy` is good idea.

But to make it fit into existing abstraction better we should pursue one of these options:
1. Override `deploy` to add extra processing. Like here https://github.com/aws-amplify/amplify-backend/blob/22aa7a659168c6e04a39343e4a52b712b6372e9c/packages/integration-tests/src/test-project-setup/data_storage_auth_with_triggers.ts#L171-L172 .
2. Add `TestProjectBase.setUp` (no-op in base class, override in this project) in addition to existing `tearDown` and hook it up into lifecycle

packages/integration-tests/src/test-project-setup/reference_auth_project.ts
```diff
@@ -0,0 +1,329 @@
...
+        domain,
+        identityPool,
+        authRole: roles.authRole,
+        unauthRole: roles.unauthRole,
+        adminGroup,
+      };
+    } catch (e) {
+      await this.authResourceCreator.cleanupResources();
+      throw e;
+    }
```
sobolk, line 319: if we make sure that resources are cleanup-script-friendly (which we should anyway) we can remove this catch block

sobolk (commented):
The overall shape of changes is very good.

Please see comments.

packages/integration-tests/src/resource-creation/auth_resource_creator.ts
```diff
@@ -0,0 +1,369 @@
...
+        console.error(`Failed to delete resource: ${task.arn ?? task.id}`, e);
+      }
+    }
+    console.error(
+      'Failed tasks:',
+      failedTasks.map((t) => t.arn ?? t.id)
+    );
+  };
+
+  createUserPoolBase = async (props: CreateUserPoolCommandInput) => {
```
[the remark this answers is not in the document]
awsluja: ah nothing like that , it is just an arbitrary suffix i chose here because I wanted to make sure all 'base' methods follow the same pattern ; and to avoid repeated the '-${suffix}' modification to resource name - only 'base' methods should do that. I am considering keeping only the base methods in this class, renaming them to remove 'base' and then making separate location for those 'non-base' methods / helpers below

packages/integration-tests/src/test-e2e/reference.test.ts
```diff
@@ -0,0 +1,292 @@
...
+      ],
+    }).toString(),
+    PolicyName: `ref-auth-idp-policy-${shortUuid()}`,
+    RoleName: createUnauthRoleResponse.Role?.RoleName,
+  })
+);
+
+void describe('reference tests', () => {
+  beforeEach(async () => {
+    // create resources
```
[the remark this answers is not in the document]
awsluja: yup i will remove this , left it in on accident

packages/integration-tests/src/test-project-setup/reference_auth_project.ts
awsluja: my thinking here was originally 1, but if setup fails - that is not the same as deployment failure; so failure is in project creator. perhaps if we had a 'beforeEach' that called testProjectCreator.beforeEach or similar. I also wasn't sure if we were timing the deployment (haven't looked too far into this, didn't want that to be included in deploy time)- on that note - setup is almost instant/very fast. Opt 2. is good

packages/integration-tests/src/test-project-setup/reference_auth_project.ts
awsluja: TIL, going to look at cleanup script and will update

packages/integration-tests/src/test-project-setup/reference_auth_project.ts
sobolk: Some more context. The original idea behind those abstractions was :
`TestProjectCreator` - takes "project template" from `packages/integration-tests/src/test-projects` and renders it in temporary folder (perhaps it should be called `TestProjectCreatorOnDisk` or something like that.
Then classes implementing `TestProjectBase` are representing "rendered test project in temp directory" and provide operations for it.
I'd keep `TestProjectCreator` responsibility end at rendering files on disk and develop new hook in `TestProjectBase` for pre-deployment activities.

packages/integration-tests/src/resource-creation/auth_resource_creator.ts
```diff
@@ -0,0 +1,369 @@
...
+      e2eToolingClientConfig
+    ),
+    private cognitoIdentityClient: CognitoIdentityClient = new CognitoIdentityClient(
+      e2eToolingClientConfig
+    ),
+    private iamClient: IAMClient = new IAMClient(e2eToolingClientConfig),
+    private createResourceNameSuffix: () => string = shortUuid
+  ) {}
+
+  cleanupResources = async () => {
```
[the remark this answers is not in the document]
awsluja: updated to use prefix; also added waiter to check if stack is deleted (since roles can't be deleted due to attached policies until the stack policies are removed)

packages/integration-tests/src/test-project-setup/reference_auth_project.ts
awsluja: i updated the resource names to be cleanup script friendly

sobolk (commented):
LGTM

packages/integration-tests/src/test-project-setup/test_project_base.ts
```diff
@@ -100,21 +102,89 @@ export abstract class TestProjectBase {
...
+        ) {
+          deleted = true;
+          console.log(`Stack ${stackName} was deleted successfully.`);
+          break;
+        } else {
+          console.error(
+            `Could not describe stack ${stackName} while waiting for deletion.`,
+            e
+          );
+          break;
```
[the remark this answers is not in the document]
awsluja: updated

PR #2147, opened 2024-10-24
feat: Add support for lambda data client

stocaaro:
## Problem

Users would like to use the Gen2 data client interface to access their APIs from application functions (lambda's).

Our [docs offer instructions on this experience](https://docs.amplify.aws/react/build-a-backend/data/customize-authz/grant-lambda-function-access-to-api/#access-the-api-using-aws-amplify), which stop short of indicating how the customer should provide the model introspection schema needed to make the whole experience work.

This gap is because the model introspection schema hasn't been available at the time the lambda's get built, which forces customers into suboptimal work-arounds.

This change introduces a new target experience where customer handlers can be configured as follow.

#### Example: `amplify/functions/todo-count/handler.ts`
```ts
import { Amplify } from "aws-amplify";
import { generateClient } from "aws-amplify/data";
import type { Schema } from "../../data/resource";
import { resourceConfig, libraryOptions } from "$amplify/client-config/todo-count"

Amplify.configure(resourceConfig, libraryOptions);

const client = generateClient<Schema>();

  const todos = (await client.models.Todo.list()).data;
  return todos.length;
};
```

## Changes
1. Make the data factory a provider by adding `provides = 'DataResources'` so it can be consumed by the function factory
2. Build the model introspection schema (MIS) into the data container entry by using the `generateModelsSync` behavior
3. Consume the MIS in the functions factory and build generated `model-introspection-schema` and `client-config` files for each function that consumes them in its handler entry file.

1. TODO

## Validation

1. Manually validated using a test-project
2. TODO Added complete unit testing
3. TODO Added integration testing

## Checklist


packages/backend-function/src/function_client_config_generator.ts
```diff
@@ -0,0 +1,128 @@
...
+    },
+  };
+};
+const { resourceConfig, libraryOptions } = lambdaConfig(
+  env,
+  modelIntrospection
+);
+
+export { resourceConfig, libraryOptions };
+`;
```
stocaaro, line 78: This is a lot of code to be pushing into the generated files. Most of this logic should be able to live in either `Amplify.configure` or a close variant which would leave this templace <10 lines long.

packages/backend-function/src/function_client_config_generator.ts
```diff
@@ -0,0 +1,128 @@
...
+  /**
+   * Check if the script content has a client-config import
+   * @param scriptContent script content to check
+   * @returns true if the script content has a client-config import, false otherwise
+   */
+  usedBy(scriptContent: string) {
+    // Matches an import of client-config
+    const clientConfigImportRegex = /import.*from.*client-config/g;
+    return clientConfigImportRegex.test(scriptContent);
+  }
```
stocaaro, line 117: We've talked about only generating these files if the customer is using them. This does that, but leaves me with concerns.
1. What happens if the customer has multiple files where the handler doesn't import the file we're looking for?
2. What happens if the customer us using a different code consumption pattern such as require?

I considered doing something more complicated like loading the script AST, but settled on starting with this and following up with a discussion.

packages/backend-function/src/function_client_config_generator.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  private writeFile = (content: string) => {
+    const clientConfigFileDirname = path.dirname(this.clientConfigFilePath);
```
stocaaro, line 120: This pattern is copied from the env shim generation. Is it worth abstracting out the shared parts to keep the code cleaner and ensure that any updates to the write pattern only need be performed in one place?

packages/backend-function/src/function_mis_generator.ts
```diff
@@ -0,0 +1,50 @@
+import fs from 'fs';
+import path from 'path';
+
+/**
+ * Generates a model introspection schema
+ */
+export class FunctionModelIntrospectionSchemaGenerator {
+  private readonly header =
+    '// This file is auto-generated by Amplify. Edits will be overwritten.';
```
stocaaro, line 9: I don't think we're using this. Since this is a json file, this shouldn't be added to the top of the file. Should we either:
1. Move to typescript and export the JSON in a variable so we can include this comment. --or--
2. Include this in the JSON in an unobtrusive way that will appear at the top of the file. --or--
3. Something else I haven't thought of

packages/client-config/API.md
```diff
@@ -627,7 +627,7 @@ export type CustomClientConfig = {
 export const DEFAULT_CLIENT_CONFIG_VERSION: ClientConfigVersion;
 
 // @public
-export const generateClientConfig: <T extends "1" | "1.1" | "1.2" | "1.3" | "0">(backendIdentifier: DeployedBackendIdentifier, version: T, awsClientProvider?: AWSClientProvider<{
+export const generateClientConfig: <T extends "1.3" | "1.2" | "1.1" | "1" | "0">(backendIdentifier: DeployedBackendIdentifier, version: T, awsClientProvider?: AWSClientProvider<{
```
stocaaro, line 630: I don't know what happened here.

packages/backend-function/src/factory.ts
```diff
@@ -394,6 +416,8 @@ class AmplifyFunction
 
     const require = createRequire(import.meta.url);
 
+    const handlerEntryContent = readFileSync(props.entry, 'utf-8');
```
stocaaro, line 419: Loads the content of the entry file, which we then pattern match to see if the imports are being used.  Is there a better way to detect use? Commented on elsewhere.

packages/backend-function/src/factory.ts
```diff
@@ -182,10 +185,26 @@ class FunctionFactory implements ConstructFactory<AmplifyFunction> {
     outputStorageStrategy,
     resourceNameValidator,
   }: ConstructFactoryGetInstanceProps): AmplifyFunction => {
+    // TODO this should be conditional on some signal,
+    // for example that any access to data was granted for this function.
     if (!this.generator) {
+      const dataResources = constructContainer
+        .getConstructFactory<AmplifyData>('DataResources')
```
stocaaro, line 192: Passing the generic AmplifyData into this correctly types the instance returned, but I don't believe we get any guarantees from this. I am inclined to type narrow with a type guard instead, but wanted input before I introduce that.

packages/backend-function/src/factory.ts
```diff
@@ -182,10 +185,26 @@ class FunctionFactory implements ConstructFactory<AmplifyFunction> {
...
+        ?.getInstance({
+          constructContainer,
+          outputStorageStrategy,
+          resourceNameValidator,
+        });
+      let modelIntrospectionSchema: string | undefined;
+
+      if (dataResources && dataResources.modelIntrospectionSchema) {
+        modelIntrospectionSchema = dataResources.modelIntrospectionSchema;
+      }
```
stocaaro, line 202: This can now be consolidated down to:
`const modelIntrospectionSchema = dataResources?.modelIntrospectionSchema`
once the type is added to `AmplifyData`

packages/backend-function/src/function_mis_generator.ts
```diff
@@ -0,0 +1,50 @@
...
+  };
+
+  /**
+   * Check if the script content has a model-introspection-schema import
+   * @param scriptContent script content to check
+   * @returns true if the script content has a model-introspection-schema import, false otherwise
+   */
+  usedBy(scriptContent: string) {
+    // Matches an import of model-introspection-schema
+    const misImportRegex = /import.*from.*model-introspection-schema/g;
```
stocaaro, line 37: Maybe `Pattern` instead of `Regex`?

stocaaro (commented):
Open questions:
- Is there a way to share code so I don't have to repeat the client-code over and over again?
- Any concerns about pulling the file into memory?
- Do we support multi-file handlers? If so, could a non-entry file be the import point for client-config?
- I am considering adding a `defineFunction` param for force mis / client-config, to ensure the mutli-file case has a way to just work without using these imports at the top level. Thoughts?

packages/backend-data/package.json
```diff
@@ -32,6 +32,8 @@
     "@aws-amplify/backend-output-schemas": "^1.1.0",
     "@aws-amplify/data-construct": "^1.10.1",
     "@aws-amplify/plugin-types": "^1.2.2",
-    "@aws-amplify/data-schema-types": "^1.2.0"
+    "@aws-amplify/graphql-generator": "^0.4.0",
+    "@aws-amplify/data-schema-types": "^1.2.0",
+    "@graphql-codegen/plugin-helpers": "^3.1.1"
```
stocaaro, line 37: This shouldn't be necessary/helpful once `@aws-amplify/graphql-generator` is released (a hack to align version in my local build)

package.json
```diff
@@ -96,7 +96,8 @@
     "tsx": "^4.6.1",
     "typedoc": "^0.25.3",
     "typescript": "~5.2.0",
-    "verdaccio": "^5.24.1"
+    "verdaccio": "^5.24.1",
+    "@graphql-codegen/plugin-helpers": "^3.1.1"
```
sobolk, line 100: nit: please put this in alphabetical order (or closest possible if we don't have strict ordering yet in this file).

packages/backend-data/src/factory.ts
```diff
@@ -45,6 +45,8 @@ import {
   FunctionSchemaAccess,
   JsResolver,
 } from '@aws-amplify/data-schema-types';
+// eslint-disable-next-line import/no-extraneous-dependencies
```
sobolk, line 48: This particular lint rule must not be disabled.

packages/backend-function/src/factory.ts
```diff
@@ -23,6 +23,7 @@ import {
   SsmEnvironmentEntry,
   StackProvider,
 } from '@aws-amplify/plugin-types';
+import type { AmplifyData } from '@aws-amplify/data-construct';
```
sobolk, line 26: We should not take dependency on `@aws-amplify/data-construct` here.

We should introduce type for `DataResources` like this one https://github.com/aws-amplify/amplify-backend/blob/main/packages/plugin-types/src/auth_resources.ts .

And then use resource provider interface like here https://github.com/aws-amplify/amplify-backend/blob/10ef35d2fe00025fefd12603bfb2e3dce696be9a/packages/backend-data/src/factory.ts#L100-L102 .

I.e. the compile time dependencies should be narrowed down types, not classes.

packages/backend-function/src/factory.ts
sobolk: This is fine. This typing is for us and we know what `DataResources` are. No need to double check that.

But this should be `.getConstructFactory<ResourceProvider<DataResources>>('DataResources')` not full construct class.

josefaidt:
> Do we support multi-file handlers? If so, could a non-entry file be the import point for client-config?

yes, and non-entrypoint files will be bundled. Are we guarding imports for the `env` object today?

> I am considering adding a defineFunction param for force mis / client-config, to ensure the mutli-file case has a way to just work without using these imports at the top level. Thoughts?

imports should work fine without the need for an additional prop across handler files

> We've talked about only generating these files if the customer is using them.

I think it's okay to always generate if the function is granted access to the data schema or a specific model (TBD). It adds to the footprint of gitignored files on the local filesystem, but if it is not imported it will not needlessly bloat the bundle

packages/backend-function/src/factory.ts
```diff
@@ -182,10 +185,26 @@ class FunctionFactory implements ConstructFactory<AmplifyFunction> {
     outputStorageStrategy,
     resourceNameValidator,
   }: ConstructFactoryGetInstanceProps): AmplifyFunction => {
+    // TODO this should be conditional on some signal,
+    // for example that any access to data was granted for this function.
```
sobolk, line 189: This needs to be solved eagerly.

Otherwise our "circular dependency problem" between auth-data-functions is going to be amplified by this. (We've been focusing on this problem since original poc of MIS in functions). @rtpascual please correct me if I'm wrong here.

We need some mechanism (implicit is best) to detect which function really needs MIS at runtime.

packages/backend-function/src/factory.ts
sobolk: In the POC I was thinking that granting access to data would be a better signal than reading the file.

The code running in lambda needs access do call appsyncapis and such grant should be explicit in backend definition.

The code won't work anyway without lambda execution role having proper access to data.


Then. considering the "circular dependency problem". It might be quite surprising if lambda suddenly depends on data just by editing it's code and adding import. I.e. "I added import to lambda and now my CFN deployment fails with circular dep problem". I think that having access grants drive that dependency is easier to reason about. I.e. I added access from function to data. now I see circular dependency problem I need to solve.

packages/backend-function/src/function_client_config_generator.ts
```diff
@@ -0,0 +1,128 @@
+import fs from 'fs';
+import { EOL } from 'os';
+import path from 'path';
+
+const lambdaConfigHelper = (
```
sobolk, line 5: can we find better name ?

packages/backend-function/src/function_client_config_generator.ts
```diff
@@ -0,0 +1,128 @@
...
+  functionName: string
+) => `import { env } from "../env/${functionName}";
+import modelIntrospection from "../model-introspection-schema/${functionName}.json";
+
+import {
+  CredentialsAndIdentityIdProvider,
+  CredentialsAndIdentityId,
+} from "aws-amplify/auth";
```
sobolk, line 13: if we're injecting this into customer's project we should scope as much as we can to avoid collisions with customer symbols.

for example https://github.com/aws-amplify/amplify-backend/blob/10ef35d2fe00025fefd12603bfb2e3dce696be9a/packages/backend-function/src/lambda-shims/resolve_ssm_params.ts#L12 .


Additionally, it would be great to extract content of this snippet to normal `ts` file and parametrize `functionNAme`.

For example:

```
internalInitializeAmplifyDataClient = async (functionName: string) => {
  await import('../env/${functionName}');
  ... rest of this code.
}

```

this will both scope it and get this under compiler checks. and. make it testable perhaps.

packages/backend-function/src/function_client_config_generator.ts
sobolk: it's fine if it's 100 lines log as long as all symbols are scoped and invisible to customer code.
but yes shifting this to ampliy configure is best.

packages/backend-function/src/function_client_config_generator.ts
```diff
@@ -0,0 +1,128 @@
...
+    this.clientConfigFilePath = `${process.cwd()}/.amplify/generated/client-config/${
+      this.functionName
+    }.ts`;
+  }
+  /**
+   * Generate a client configuration
+   */
+  generateClientConfig = () => {
+    this.writeFile(
+      [this.header, lambdaConfigHelper(this.functionName)].join(EOL)
```
sobolk, line 104: https://www.google.com/search?q=helper+code+smell

can this be `generateAmplifyClientInitializationCode` ??

packages/backend-function/src/function_client_config_generator.ts
```diff
@@ -0,0 +1,128 @@
...
+   */
+  constructor(private readonly functionName: string) {
+    this.clientConfigFilePath = `${process.cwd()}/.amplify/generated/client-config/${
+      this.functionName
+    }.ts`;
+  }
+  /**
+   * Generate a client configuration
+   */
+  generateClientConfig = () => {
```
sobolk, line 102: nit: we should find better name for this. Perhaps `generateAmplifyClientInitializationSnippet` or something like that. I.e. more descriptive.

The reason is: we already have this name used here https://github.com/aws-amplify/amplify-backend/blob/main/packages/client-config/src/generate_client_config.ts .

packages/backend-function/src/function_client_config_generator.ts
sobolk: And they can also use good old `require` :( .

I left comment somewhere else to use access grants for this purpose.

If we keep this approach one option is to run bundling twice and inspect the bundled file. then regenerate.

packages/backend-function/src/function_client_config_generator.ts
sobolk: I'd follow the rule of three here and wait for more examples.

It's hard to say if this is true or incidental duplication.
https://anthonysciamanna.com/2018/07/28/the-dry-principle-and-incidental-duplication.html

packages/backend-function/src/function_mis_generator.ts
sobolk: I think putting this next to `index.mjs` as extra resource and `await import`ing it in relevant place would be fine.

There was some way to achieve this with bundling/function options.

packages/backend-function/src/factory.ts
```diff
@@ -425,6 +449,26 @@ class AmplifyFunction
...
+        functionClientConfigGenerator.usedBy(handlerEntryContent)
+      )
+        functionMISGenerator.generateModelIntrospectionSchema(
+          modelIntrospectionSchema
+        );
+
+      if (functionClientConfigGenerator.usedBy(handlerEntryContent)) {
+        functionClientConfigGenerator.generateClientConfig();
+      }
+    }
```
sobolk, line 470: we should get rid of these `usedBy` somehow.

They seem to be used to wire together two code fragments that could be combined?

Could we try something like this:

```
if (modeIntrospectionIsAvailable && functionIsUsingData) {
   amplifyClientsInitializerGenerator.generateSingleAwesomeSnippetThatIsBlockScopedAndProducesRequiredState();
}
```

packages/client-config/API.md
sobolk: this happens sometimes, clean-build usually solves it.
or just remove this from diff please. Whatever makes GH actions happy.

packages/backend-function/src/factory.ts
rtpascual: Yup I agree with this, the existence of the graphql endpoint env var for the lambda will be better than reading handler code and solves a couple problems pointed out in the comments elsewhere.

packages/backend-function/src/function_mis_generator.ts
rtpascual: Yup that would be `banner` in bundling options https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda_nodejs.BundlingOptions.html

packages/backend-function/src/function_client_config_generator.ts
stocaaro: Yep. I assumed this would go away and didn't proactively refactor it. With the refactor this morning, it's ~30 loc.

I have a message out to @HuiSF to discuss this, but I am expecting that the preferred approach is to keep these 30 loc and the related MIS built to this file instead of pushing it across into the js repo.

packages/backend-function/src/function_client_config_generator.ts
stocaaro: Talking with @josefaidt yesterday, the impact of generating these files for every function is minimal. I will experiment with resource permission based filtering this morning. This is discussed in a couple places and I'll use this thread for followups on this topic.

packages/backend-function/src/function_client_config_generator.ts
stocaaro: We have 3 if we keep the MIS json separate, but having removed that file just now, we only have 2 instances so I will skip this refactor for now.

packages/backend-function/src/function_mis_generator.ts
stocaaro: I killed the separate json file, so this is a non-issue with the update I will post shortly.

packages/backend-function/src/factory.ts
stocaaro: As [noted above](https://github.com/aws-amplify/amplify-backend/pull/2147#discussion_r1815331212), I've removed these checks and am looking into resource permission based filtering. Good thought on looking at the graphql endpoint env var as I think this is the logic I need.

packages/backend-function/src/function_client_config_generator.ts
sobolk: > the impact of generating these files for every function is minimal

performance wise perhaps yes. But it creates unwanted link in resource graph making the "cycling dependency problem" more likely to happen. This is not "minimal".

packages/backend-function/src/factory.ts
```diff
@@ -161,6 +162,10 @@ export type FunctionBundlingOptions = {
   minify?: boolean;
 };
 
+type DataResources = ResourceProvider<Record<string, unknown>> & {
+  modelIntrospectionSchema?: string;
+};
```
stocaaro, line 167: I'm testing this against a modified version of `amplify-category-api`. Its unclear to me whether the MIS should be passed through at the root object level or within `.resources`.  Open to both, I left it at the top level to avoid additional plumbing, but I'm not sure what is best practice in this object.

packages/backend-function/src/function_client_config_generator.ts
stocaaro: I don't have context on the cycling dep problem. It may be related to the error we're chasing right now. I expect to followup on this in another update.

I think the goal is to use the resource auth rules to limit which functions we build files for, however these permissions are causing the current problem, so we need to overcome that before I can test any solution that would constrain based on these auth rules.

packages/backend-function/src/factory.ts
stocaaro: I made a fix to this. A bit unclear on how this object is meant to be structured and I've made a comment on my [new type](https://github.com/aws-amplify/amplify-backend/pull/2147/files#r1821316875) to continue this conversation

packages/backend-function/src/factory.ts
stocaaro: Removed / updated

packages/backend-function/src/function_client_config_generator.ts
stocaaro: Renamed

packages/backend-function/src/function_client_config_generator.ts
```diff
@@ -0,0 +1,73 @@
+import fs from 'fs';
+import path from 'path';
+
+const lambdaConfigTemplate = (
+  functionName: string,
+  modelIntrospectionSchema: string
+) => `// This file is auto-generated by Amplify. Edits will be overwritten.
+import { env } from "../env/${functionName}";
+
+export const libraryOptions = {
```
stocaaro, line 10: I seem to have lost a comment from @sobolk suggesting this content be pushed down into a file to be imported.

Which package would this file be in?  I had been advocating for amplify-js, however, that doesn't seem like its going to be possible.

Since this is essentially customer code, could this behavior be usable/consumable from the backend package? It seems safe to assume this will be present during the build.

packages/backend-function/src/factory.ts
stocaaro: I haven't experienced this issue yet (unless this is the auth issue we're troubleshooting right now), do you have instructions/examples of what is happening in these cases?

stocaaro (commented):
I have posted updates and resolved a bunch of conversations. The current active threads I'm tracking/following up on are:
- Resource authorization causes build errors (maybe related to the import cycles problem between auth/data)
- Confirm the resource interface and ensure we get the latest version from @aws-amplify/data-construct
- We should not build client config / mis for every function as this will cause more interactions resulting in import cycles
- Should we relocated the lambdaConfigTemplate code to somewhere we can import from?

packages/backend-function/src/function_client_config_generator.ts
sobolk: It does not have to be "imported" in literal sense. See https://github.com/aws-amplify/amplify-backend/blob/9a5d8e3f985e7b4ecf46babe709ca482d1d03b4f/packages/backend-function/src/factory.ts#L411-L419 .

The point is to get it under compiler and be able to write unit tests for it.

packages/backend/src/internal/index.ts
```diff
@@ -0,0 +1 @@
+export { dataClientConfiguration } from '@aws-amplify/backend-function/internal';
```
stocaaro, line 1: I'm expecting this to be controversial, but I think its the right pattern to support our customers for the following reasons.

- The data client config needs the sdk to access s3 in the customers runtime
- We know very little about the customers runtime dependencies, except that they must be using `@aws-amplify/backend` in order to have `defineFunction` to configure their function backend.
- Lambda esbuild isn't treeshaking, so we need to import from a surface that doesn't have any additional behavior exports `@aws-amplify/backend` => 22MB lambda vs `@aws-amplify/backend/internal` => 1.5MB lambda vs ~500k in the version with no S3 access (embedding the MIS directly in the function)

Questions:
- Is `@aws-amplify/backend/internal` an acceptable pattern?
  - Internal becomes a bit overloaded with this (internal is used to describe the API surface that is tested but not exported)
  - Is there a better name than internal?
- To keep the code surface of backend concise, this is implemented under `@aws-amplify/backend-function/internal` and re-exported in the backend. Anything about this pattern that we should change?
- What do we think about the `dataClientConfiguration` function name?
  - This is specific to configuring the data client and internal for our use. Should internal be added to the name as a prefix?

packages/backend-function/src/factory.ts
```diff
@@ -203,6 +213,7 @@ class FunctionFactory implements ConstructFactory<AmplifyFunction> {
       schedule: this.resolveSchedule(),
       bundling: this.resolveBundling(),
       layers,
+      generateDataClientConfig: this.props.generateDataClientConfig ?? false,
```
stocaaro, line 216: I did this because the function factory can't communicate with the data factory, but it looks like there is some crosstalk between them at build using the SSM variables.

TODO: Revisit this and see if we can remove this param by using the presence of the SSM vars to trigger the file to be built.

packages/backend-data/src/factory.ts
```diff
@@ -263,6 +277,21 @@ class DataGenerator implements ConstructContainerEntryGenerator {
       );
     }
 
+    // TODO Any risk that this throws?
```
sobolk, line 280: no.

packages/backend-function/src/factory.ts
sobolk: > by using the presence of the SSM vars to trigger the file to be built.

I'd rather always embed snippet from "data client config generator" and just make it no-op if env vars in function runtime are not populated.

This optimization should be deferred (seems premature) until:
1. Generated snippet is large enough to cause visible impact on cold start.
2. OR we pivot to "solution 2" where wiring between data and func factory becomes a must.

packages/backend-function/src/internal/index.ts
```diff
@@ -0,0 +1,61 @@
+/* eslint-disable @typescript-eslint/naming-convention */
```
sobolk, line 1: this should not be disabled in whole file. please be more specific.

packages/backend/package.json
```diff
@@ -13,6 +13,11 @@
     },
     "./types/platform": {
       "types": "./lib/types/platform.d.ts"
+    },
+    "./internal": {
+      "types": "./lib/internal/index.d.ts",
+      "import": "./lib/internal/index.js",
+      "require": "./lib/internal/index.js"
     }
```
sobolk, line 21: We shouldn't be exporting this from `backend`. Customers won't be using this directly?
These are not necessary as long as `backend-function` is visible to bundler. In other words, having this in `backend-function` should be enough.

packages/backend/src/internal/index.ts
sobolk: `runtime` is better wording, (see https://github.com/aws-amplify/amplify-backend/blob/f08abe44c0a02d80d3985f718e610140ed6cca3c/packages/ai-constructs/package.json#L17 ).

This should not be exported from `backend`.

`backend-function/runtime` should be enough. I wonder if we should get `backend-function-runtime` package (But this might be premature).

> What do we think about the dataClientConfiguration function name?

To future proof it we should call it. `getAmplifyClientsConfiguration` or something like that. 
I.e. it's purpose is to 1. read env 2. spit out partials of what goes into `Amplify.configure`.

packages/backend/package.json
stocaaro: Customers apps will be using it directly via the generated files built to `.amplify/generated/client-config/NAME.ts`. I did this because of a concern @josefaidt raised about some build tools not supporting transitive dependencies, which might force some customers to depend on `@aws-amplify/backend-function` directly in their package.json to make things happy.

I can revert to consuming this from `@aws-amplify/backend-function/internal` for now if your confident this won't cause issues.

packages/backend-function/src/factory.ts
stocaaro: Understood. I have 2 concerns with always generating the files.
1. I think it might cause a 1MB build size diff, even if unused. I will test this today.
2. If this is built conditionally, when it is build and permissions are missing, type feedback tells the developer things aren't working, where if we always built it with permissive types feedback for misconfiguration will be at runtime and the error message shows up in the lambda logs, which makes debug/troubleshooting cycle larger.

packages/backend/package.json
sobolk: > Customers apps will be using it directly via the generated files built to .amplify/generated/client-config/NAME.ts

that's not direct usage.

> I did this because of a concern @josefaidt raised about some build tools not supporting transitive dependencies

we need a proof (poc app or user complaint) not speculation. until then we should defer.

Also, if this (https://github.com/aws-amplify/amplify-backend/pull/2147/files#r1833214679)  is by any chance possible then perhaps we won't need exports at all.

sobolk: Sorry wrong link https://github.com/aws-amplify/amplify-backend/pull/2147#discussion_r1833214679 .
It's leading to comment I left in a thread above.

packages/backend-function/src/factory.ts
stocaaro: I think I found a happy middleground. I'm using a [typeGuard](https://github.com/aws-amplify/amplify-backend/pull/2147/files#diff-402c679c31e276ed80382af706cdde8817f744890aa6e5b077ad928e5310483aR27-R45) to return valid configuration inputs if all the env vars are present and returns `unknown` configuration inputs when they aren't.

I verified that a user who doesn't import the file doesn't see any additional lambda size, so I have no concern exporting these for every function.

I would like the invalid `unknown` configuration inputs to give a clear error message telling the dev to add resource authorization to get it to work, but I couldn't find a way to include this in the tooltip error message here.

I'm open to exploring the precompilation path if you think that lands somewhere better. Do you have an example of how that would work?

packages/backend/src/internal/index.ts
stocaaro: Changes made

packages/backend-function/src/factory.ts
sobolk: Btw. we were considering this for SSM snippet in the past but we didn't pursue this because:
1. SSM snippet can use lambda provided SSM SDK for its logic.
2. SSM snippet is small. so this would be micro optimization.

Having said that however:
Have we considered including this logic in a same way we include SSM snippet? I.e. have this become part of banner and rely on Lambda provided SDK to make s3 call ? Instead of bundling ?

stocaaro:
Closing this change out, splitting it into two changes:
1. [Data factory shares the MIS via S3/SSM](https://github.com/aws-amplify/amplify-backend/pull/2223)
2. [Function factory generates data config shim making use trivial](https://github.com/aws-amplify/amplify-backend/pull/2224)

PR #2167, opened 2024-10-30
Option 1: Inject MIS via SSM (possibly S3 url)

sobolk:
I was looking at old PR that explains cyclic dependency problem https://github.com/aws-amplify/amplify-backend/pull/969 .
See that PR description.

This POC is using a fact that access rules between data and function expose a way to inject anything from data into function env vars.
We seem to be already injecting GQL url in existing implementation so we'd be following an established pattern here.

The downside is that MIS read in function requires network call.

### DIFF
--- a/test-projects/lambda-client/amplify/functions/todo-count/handler.ts
+++ b/test-projects/lambda-client/amplify/functions/todo-count/handler.ts
@@ -1,14 +1,17 @@
 import type { Handler } from "aws-lambda";
-import { Amplify } from "aws-amplify";
-import { generateClient } from "aws-amplify/data";
-import type { Schema } from "../../data/resource";
-import { resourceConfig, libraryOptions } from "$amplify/client-config/todo-count"
+// import { Amplify } from "aws-amplify";
+// import { generateClient } from "aws-amplify/data";
+// import type { Schema } from "../../data/resource";
+// import { resourceConfig, libraryOptions } from "$amplify/client-config/todo-count"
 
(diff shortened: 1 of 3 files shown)


test-projects/lambda-client/amplify/functions/todo-count/handler.ts
```diff
@@ -1,14 +1,17 @@
...
-const client = generateClient<Schema>();
+// Amplify.configure(resourceConfig, libraryOptions);
+//
+// const client = generateClient<Schema>();
 
 export const handler: Handler = async () => {
-  const todos = (await client.models.Todo.list()).data;
-  return todos.length;
+
+  console.log(JSON.stringify(process.env, null, 2));
```
sobolk, line 13: Added this to see if plumbing works.

image

test-projects/lambda-client/amplify/functions/todo-count/handler.ts
```diff
@@ -1,14 +1,28 @@
...
+  const s3Client = new S3Client();
+  const response = await s3Client.send(
+    new GetObjectCommand({
+      Bucket: process.env.AMPLIFY_DATA_MODEL_INTROSPECTION_SCHEMA_BUCKET_NAME,
+      Key: process.env.AMPLIFY_DATA_MODEL_INTROSPECTION_SCHEMA_KEY,
+    })
+  );
+
+  const content = await response.Body?.transformToString();
+  console.log(content);
```
sobolk, line 24: image

PR #2169, opened 2024-10-30
Option 2: Split out schema generation.

sobolk:
This is a (preferred?) alternative to https://github.com/aws-amplify/amplify-backend/pull/2167 .

The idea here is to split out schema computations from data construct creation. So that schema computation can become both input to function and data construct.

Elements:
1. Split schema generation logic.
2. Add new concept in construct container to register provider/factory method that can provide arbitrary piece of data.

The upside of this solution is that schema is can be bundling time asset without network calls from lambda.

The downside (if this is a downside at all?) is that we're inventing new way of how verticals can communicate and construct container becomes closer to be generic DI container (like syringe) where modules may register arbitrary logic. (which can be abused if contract between verticals established this way is not guarded with proper types/interfaces which become public api of a vertical).

### DIFF
--- a/test-projects/lambda-client/amplify/functions/todo-count/handler.ts
+++ b/test-projects/lambda-client/amplify/functions/todo-count/handler.ts
@@ -1,14 +1,16 @@
 import type { Handler } from "aws-lambda";
-import { Amplify } from "aws-amplify";
-import { generateClient } from "aws-amplify/data";
-import type { Schema } from "../../data/resource";
-import { resourceConfig, libraryOptions } from "$amplify/client-config/todo-count"
-
-Amplify.configure(resourceConfig, libraryOptions);
-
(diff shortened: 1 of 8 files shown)


test-projects/lambda-client/amplify/functions/todo-count/handler.ts
```diff
@@ -1,14 +1,16 @@
...
+//
+// const client = generateClient<Schema>();
 
 export const handler: Handler = async () => {
-  const todos = (await client.models.Todo.list()).data;
-  return todos.length;
+  // const todos = (await client.models.Todo.list()).data;
+  // return todos.length;
+
+  console.log(process.env.MIS);
```
sobolk, line 15: it works as well.

image

PR #2181, opened 2024-11-01
Add metadata to user agent in conversation handler runtime

sobolk:
## Problem

We're not emitting basic data from conversation handler.

## Changes

Enrich user agent.
1. Add library name and version (ai-constructs)
2. Add turn response mutation type

## Validation

Added tests

## Checklist

### DIFF
--- a/packages/ai-constructs/src/conversation/runtime/user_agent_provider.test.ts
+++ b/packages/ai-constructs/src/conversation/runtime/user_agent_provider.test.ts
@@ -54,41 +54,12 @@ void describe('User Agent provider', () => {
     } as unknown as ConversationTurnEvent);
 
     const userAgent = userAgentProvider.getUserAgent({
-      metadata1: 'value1',
-      metadata2: 'value2',
+      'turn-response-type': 'streaming',
     });
 
     assert.strictEqual(
       userAgent,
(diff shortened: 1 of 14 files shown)


packages/ai-constructs/src/conversation/runtime/user_agent_provider.test.ts
```diff
@@ -0,0 +1,65 @@
...
+import { ConversationTurnEvent } from './types';
+
+void describe('User Agent provider', () => {
+  // Read package json from disk (i.e., in a different way than actual implementation does).
+  const packageVersion = JSON.parse(
+    fs.readFileSync(
+      path.resolve(__dirname, '..', '..', '..', 'package.json'),
+      'utf-8'
+    )
+  ).version;
```
rtpascual, line 15: Curious why we are testing with getting package version in a different way than the implementation?

sobolk: To avoid tautology. I.e. using feeding data for assertion using copy of implementation would be that.

Otherwise, the assertion would need to be - there's some version in UA, it matches semver regex. (actual value is not constant).

rtpascual (approved):

ShadowCat567 (approved):

PR #2183, opened 2024-11-01
feat: share mis via getArtifacts

stocaaro:
## Problem

## Changes

## Checklist


packages/backend-data/src/factory.ts
```diff
@@ -124,12 +194,26 @@ class DataGenerator implements ConstructContainerEntryGenerator {
...
     ssmEnvironmentEntriesGenerator,
     backendSecretResolver,
     stableBackendIdentifiers,
   }: GenerateContainerEntryProps) => {
+    const schemaDefinition = new GraphqlSchemaDefinition({
+      props: this.props,
+      backendSecretResolver,
+      stableBackendIdentifiers,
+    });
+    this.schemaDefinition = schemaDefinition;
```
stocaaro, line 215: This assignment has to happen inside of the `generateContainerEntry` because thats where the secretResolver and backend identifiers are introduced. This leaves me worrying about the possability of calling for artifacts before the assignment would happen, could result in `undefined` responses that occur because the order of operations has changed.

packages/backend-data/src/factory.ts
```diff
@@ -124,12 +194,26 @@ class DataGenerator implements ConstructContainerEntryGenerator {
     this.name = props.name ?? 'amplifyData';
   }
 
+  getArtifact(key: string): unknown | undefined {
+    if (key === 'mis') {
+      return this.schemaDefinition?.modelIntrospectionSchema;
```
stocaaro, line 199: Hypothetically, this could be called before schemaDefinition is set on the factory. I couldn't see a way to make that happen and would want testing to assert that it isn't possible if this solution is selected to proceed with.

PR #2209, opened 2024-11-07
Added retry mechanism to catch known errors for health checks

vigy02:
## Problem

The integration tests were failing due to intermittent errors, specifically unexpected token errors and command failures. These errors were causing the tests to fail even when they might have succeeded on a retry.

## Changes

- Added a retry mechanism to the 'creates new project and deploy them without an error' test.
- Implemented a 3-minute timeout for the entire test.
- Added specific error handling for two known issues:
  1. Unexpected token '<', "<!DOCTYPE "... is not valid JSON
  2. Command failed with exit code 1: yarn add @aws-amplify/backend
- Implemented a 5-second delay between retries.
- Added logging for retry attempts and error messages.

## Checklist


packages/integration-tests/src/package_manager_sanity_checks.test.ts
```diff
@@ -81,77 +81,142 @@ void describe('getting started happy path', async () => {
...
+
+        // If we reach here without errors, break the loop
+        return;
+      } catch (error) {
+        console.error('Error occurred:', error);
+
+        const errorMessage = error instanceof Error ? error.message : '';
 
-    assert.ok(clientConfigStats.isFile());
+        const isCommandFailedError = errorMessage.includes('exit code 1');
```
rtpascual, line 155:
```suggestion
        const isFailedInstallDepsError = errorMessage.includes('exit code 1') && (errorMessage.includes('@aws-amplify/backend') || errorMessage.includes('aws-amplify'));
```
I think we should check if error message has this and one of the dependencies we install during create amplify. `exit code 1` may show up in other errors so we should be more explicit here while keeping it package manager agnostic.

packages/integration-tests/src/package_manager_sanity_checks.test.ts
```diff
@@ -81,77 +81,145 @@ void describe('getting started happy path', async () => {
...
-        branchBackendIdentifier.namespace,
-      ],
-      tempDir,
-      { env: { CI: 'true' } }
-    ).run();
-
-    const clientConfigStats = await fsp.stat(
-      await getClientConfigPath(ClientConfigFileBaseName.DEFAULT, tempDir)
-    );
+    while (Date.now() - startTime < TIMEOUT_MS) {
```
sobolk, line 88: Could you please create `retry.ts` file and create generalized function for this.

Something like

```
export const runWithRetry = async <T>(
    callable: () => Promise<T>, // code under retry
    retryPredicate: (error: Error) => boolean, // predicate to decide if retry or not
    maxAttempts = 3,
    timeout,
    ): Promise<T> => {...};
```


Similarly to this https://github.com/aws-amplify/amplify-backend/blob/06bf37e746254d49c7a6a6074449be63505f1f08/packages/integration-tests/src/test-project-setup/conversation_handler_project.ts#L967-L989 

This ^ can be then replaced with new generalized function.

packages/integration-tests/src/package_manager_sanity_checks.test.ts
```diff
@@ -81,77 +81,145 @@ void describe('getting started happy path', async () => {
...
+              return true;
+            }
+          );
+        }
+      } catch (error) {
+        console.error('Error occurred:', error);
+        const errorMessage = error instanceof Error ? error.message : '';
+        const isUnexpectedTokenError = errorMessage.includes(
+          'Unexpected token \'<\', "<!DOCTYPE "... is not valid JSON'
+        );
```
sobolk, line 205: this retry should happen inside `amplify_app_pool.ts`

.changeset/nine-toes-dress.md
```diff
@@ -0,0 +1,5 @@
+---
+'@aws-amplify/integration-tests': patch
+---
+
+Added retry mechanism to catch known errors in package manager sanity checks
```
sobolk, line 5:
```suggestion
---
---

```


for integration test changeses empty changeset is enough.

packages/integration-tests/src/amplify_app_pool.ts
```diff
@@ -46,114 +47,124 @@ class DefaultAmplifyAppPool implements AmplifyAppPool {
...
-        );
-      nextToken = listAppsCommandOutput.nextToken;
-      listAppsCommandOutput.apps
-        ?.filter((app: App) => app.name?.startsWith(this.testAppPrefix))
-        .forEach((app: App) => {
-          apps.push(app);
-        });
-    } while (nextToken);
-    return apps;
+    return this.retryableOperation(async () => {
```
sobolk, line 93: it is enough to wrap top level public methods in retry.
We shouldn't be applying this aggressively. 

please remove this from private methods for now.

packages/integration-tests/src/amplify_app_pool.ts
```diff
@@ -173,6 +184,24 @@ class DefaultAmplifyAppPool implements AmplifyAppPool {
...
+      operation,
+      (error) => {
+        // Add specific error conditions here that warrant a retry
+        return (
+          error.message.includes("Unexpected token '<'") ||
+          error.name === 'ThrottlingException' ||
+          (error.message.includes('exit code 1') &&
+            error.message.includes('@aws-amplify/backend')) ||
+          error.message.includes('aws-amplify')
+        );
```
sobolk, line 199: When I made this comment https://github.com/aws-amplify/amplify-backend/pull/2209#discussion_r1833502663 .

i meant that errors related to Unexpected token should be handled here in app pool.

But handling errors related to yarn installation should be happening in that test you modified originally.
This code here does not do anything with package installations.

Btw I'd just check if message includes `Unexpected token` (this error comes in couple of flavors so loose matching should be fine here).
I'd remove `ThrottlingException` from here. We handled that case by configuring retries in SDK clients elsewhere. therefore this would be redundant.

packages/integration-tests/src/retry.ts
```diff
@@ -0,0 +1,44 @@
...
+  timeout?: number
+): Promise<T> => {
+  const startTime = Date.now();
+  const collectedErrors: Error[] = [];
+
+  for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+    try {
+      if (timeout && Date.now() - startTime >= timeout) {
+        throw new Error(`Operation timed out after ${timeout}ms`);
+      }
```
sobolk, line 20: I'd remove timeouts from here for now.
We configure timeouts in heath_checks.yml, so this is redundant.

packages/integration-tests/src/retry.ts
```diff
@@ -0,0 +1,44 @@
...
+      return result;
+    } catch (error) {
+      if (error instanceof Error) {
+        collectedErrors.push(error);
+
+        if (!retryPredicate(error) || attempt === maxAttempts) {
+          break;
+        }
+
+        console.log(`Attempt ${attempt} failed. Retrying...`);
```
sobolk, line 32: I'd remove this log.
It does not say which operation is being retried so logging that is going to be noise rather than useful information. We won't care much if operation succeeds within max attempts.

Throwing aggregate error at the end gives information about number of retires and how they failed.

packages/integration-tests/src/retry.ts
```diff
@@ -0,0 +1,44 @@
...
+        await new Promise((resolve) => setTimeout(resolve, 5000)); // 5 second delay between retries
```
sobolk, line 33: I'd remove delay for now.

we don't seem to have data that would say it's really needed. we should add this later if we see that operations still fail under retry and that errors are related to throttling.

packages/integration-tests/src/package_manager_sanity_checks.test.ts
```diff
@@ -81,76 +82,109 @@ void describe('getting started happy path', async () => {
...
+
+        assert.ok(clientConfigStats.isFile());
+      },
+      (error) => {
+        // Retry on network-related errors or command failures
+        return (
+          error.message.includes('exit code 1') &&
+            (error.message.includes('@aws-amplify/backend') ||
+              error.message.includes('aws-amplify')),
```
sobolk, line 151:
```suggestion
            (error.message.includes('aws-amplify'),
```

there's some redundancy in this check.

packages/integration-tests/src/package_manager_sanity_checks.test.ts
```diff
@@ -81,76 +82,109 @@ void describe('getting started happy path', async () => {
...
-        }),
-        (error) => {
-          const errorMessage = error instanceof Error ? error.message : '';
-          assert.match(
-            errorMessage,
-            /Amplify does not support PNPM on Windows./
-          );
-          return true;
-        }
+        runWithRetry(
```
sobolk, line 162: I don't recall this test being unstable. let's remove retry from here.

packages/integration-tests/src/package_manager_sanity_checks.test.ts
```diff
@@ -81,76 +82,109 @@ void describe('getting started happy path', async () => {
...
+              env: { npm_config_user_agent: 'yarn/1.22.21' },
+            }
+          );
+        } else {
+          await runPackageManager(
+            packageManager,
+            ['create', amplifyAtTag, '--yes'],
+            tempDir
+          ).run();
+        }
```
sobolk, line 103: I would wrap just this part of the test with retry. this is what installs packages and causes intermitent problems.

packages/integration-tests/src/amplify_app_pool.ts
```diff
@@ -173,6 +178,17 @@ class DefaultAmplifyAppPool implements AmplifyAppPool {
...
   };
+
+  private retryableOperation = <T>(operation: () => Promise<T>) => {
+    return runWithRetry(
+      operation,
+      (error) => {
+        // Add specific error conditions here that warrant a retry
+        return error.message.includes('Unexpected token');
+      },
+      3 // maxAttempts
```
rtpascual, line 189: `maxAttempts` is already defaulting to 3 so this isn't needed. There are two other places as well with the same thing.

rtpascual (commented):
This is looking good.

rtpascual (dismissed):

packages/integration-tests/src/package_manager_sanity_checks.test.ts
```diff
@@ -138,20 +155,29 @@ void describe('getting started happy path', async () => {
...
           );
-          return true;
         }
-      );
-    }
+      },
+      (error) => {
+        return error.message.includes(
+          'Unexpected token \'<\', "<!DOCTYPE "... is not valid JSON'
+        );
```
sobolk, line 179: Does PNPM throw this ? This looks related to app pool, not to PNPM.

I'd remove retry from this particular test for now. I don't recall problems with this particular test (as mentioned here https://github.com/aws-amplify/amplify-backend/pull/2209#discussion_r1835039392) .

sobolk (dismissed):

packages/integration-tests/src/amplify_app_pool.ts
```diff
@@ -173,6 +178,13 @@ class DefaultAmplifyAppPool implements AmplifyAppPool {
       }
     }
   };
+
+  private retryableOperation = <T>(operation: () => Promise<T>) => {
+    return runWithRetry(operation, (error) => {
+      // Add specific error conditions here that warrant a retry
+      return error.message.includes('Unexpected token');
+    });
```
sobolk, line 186:
```suggestion
    return runWithRetry(operation, (error) => {
      // Add specific error conditions here that warrant a retry
      return error.message.includes('Unexpected token') || error.message.includes('Bad control character');
    });
```

From recent run 
image

packages/integration-tests/src/retry.ts
```diff
@@ -0,0 +1,35 @@
...
+  for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+    try {
+      const result = await callable();
+      return result;
+    } catch (error) {
+      if (error instanceof Error) {
+        collectedErrors.push(error);
+
+        if (!retryPredicate(error) || attempt === maxAttempts) {
+          break;
```
sobolk, line 23:
```suggestion
          throw error;
```

It turns out that we shouldn't be throwing aggregate error if predicate doesn't match. The output is confusing.

image

sobolk (commented):
There are two things that showed up in this PR run that can be improved.

sobolk (approved):

rtpascual (approved):

PR #2224, opened 2024-11-12
feat: Add lamda data client support

stocaaro:
## Problem

Users would like to use the Gen2 data client interface to access their APIs from application functions (lambda's).

Our [docs offer instructions on this experience](https://docs.amplify.aws/react/build-a-backend/data/customize-authz/grant-lambda-function-access-to-api/#access-the-api-using-aws-amplify), which stop short of indicating how the customer should provide the model introspection schema needed to make the whole experience work.

## Changes
1. Build the model introspection schema
    1. Build the Model Introspection Schema (MIS) in the data factory
    2. Construct an S3 bucket with the MIS as an object
    3. Grant get permission to the MIS to any function that is granted `resource` permissions to the data schema
    4. Add the bucket name and object key for the MIS to SSM (where it will be read
1. Add config function for user ease
    1. Add a configuration function we can internally import from `import { getAmplifyClientsConfiguration } from "@aws-amplify/backend/function/runtime";`

#### Example: `amplify/functions/todo-count/handler.ts`
```ts
import { Amplify } from "aws-amplify";
import { generateClient } from "aws-amplify/data";
import type { Schema } from "../../data/resource";

import { getAmplifyDataClientConfig } from "@aws-amplify/backend/function/runtime";
import { env } from "$amplify/env/todo-count";

const { resourceConfig, libraryOptions } = await getAmplifyDataClientConfig(env)
Amplify.configure(resourceConfig, libraryOptions);

const client = generateClient<Schema>();

  const todos = (await client.models.Todo.list()).data;
  return todos.length;
};
```

https://github.com/aws-amplify/docs/pull/8096

## Validation
- Unit tests added
- Manually tested in an example application
- E2E tests added for data access and unrelated user use of `S3Client`

## Checklist


packages/backend-function/runtime/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "@aws-amplify/backend-function/runtime",
+  "main": "../dist/cjs/runtime/index.js",
+  "react-native": "../src/runtime/index.ts",
+  "browser": "../dist/esm/runtime/index.mjs",
+  "module": "../dist/esm/runtime/index.mjs",
+  "typings": "../dist/esm/runtime/index.d.ts"
+}
```
sobolk, line 8: Why do we need this file ?

packages/backend-function/src/function_data_config_generator.ts
```diff
@@ -0,0 +1,45 @@
...
+  private indentation: string = '  ';
+
+  /**
+   * Initialize data configuration file name and location
+   */
+  constructor(private readonly functionName: string) {
+    this.typeDefFilePath = `${process.cwd()}/.amplify/generated/data-config/${
+      this.functionName
+    }.ts`;
+  }
```
sobolk, line 27:
```suggestion
/**
 * Generates the data configuration imports
 */
export class FunctionAmplifyClientConfigGenerator {
  private typeDefFilePath: string;

  private indentation: string = '  ';

  /**
   * Initialize data configuration file name and location
   */
  constructor(private readonly functionName: string) {
    this.typeDefFilePath = `${process.cwd()}/.amplify/generated/amplify-client-config/${
      this.functionName
    }.ts`;
  }
```

It seems we're doing all of this to call `Amplify.configure` . The runtime objects this generates could be later extended to support more properties (just generating full amplify outputs would be most extreme expansion).
Therefore I think we should use a bit more generic naming here.

packages/backend-function/src/function_env_type_generator.test.ts
```diff
@@ -57,7 +56,6 @@ void describe('FunctionEnvironmentTypeGenerator', () => {
   });
 
   void it('generated type definition file has valid syntax', async () => {
-    const targetDirectory = await fsp.mkdtemp('func_env_type_gen_test');
```
sobolk, line 60: why do we change this test?

packages/backend-function/src/runtime/get_amplify_clients_configuration.ts
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * The implementation of `getAmplifyClientsConfiguration` that allows you to pass in an S3 client for testing purposes
+ * @param env - The environment variables for the data client
+ * @param s3Client - The S3 client to use for retrieving the schema
+ * @returns An object containing the `resourceConfig` and `libraryOptions`
+ */
+export const getAmplifyClientsConfigurationRetriever = async <T>(
+  env: T,
+  s3Client: S3Client
+): Promise<DataClientReturn<T>> => {
```
sobolk, line 121: This can be single function, you can

```
export const getAmplifyClientsConfigurationRetriever = async <T>(
  env: T,
  s3Client = new S3Client()
): Promise<DataClientReturn<T>> => {
```

OR do what ssm is doing here https://github.com/aws-amplify/amplify-backend/blob/9fd064261b1a8ee3c9f544885f68b128a82564db/packages/backend-function/src/lambda-shims/resolve_ssm_params.ts#L12

packages/backend-function/src/function_env_type_generator.test.ts
rtpascual: I discussed with @stocaaro about this and found this is not being used (maybe was used in a previous iteration of this test before we moved to `.amplify/generated/env`)

packages/backend-function/runtime/package.json
stocaaro: I'm not sure. I thought I had errors without it, but can't replicate them this morning. This is the pattern that is followed in js and elsewhere, but I'll remove it until/unless I encounter the issue it resolves.

packages/backend-function/src/function_env_type_generator.test.ts
stocaaro: Appreciate the help debugging @rtpascual

packages/backend-function/src/function_data_config_generator.ts
stocaaro: I had that same thought, however the DX I've been working towards with @josefaidt does all the work and top level `await` to prepare a data centric config (where the config offers specific errors if the MIS isn't available). My thinking is to namespace this tuned experience and if we offer a config function that behaves more generally, that we would offer that on a different import path (maybe the proposed path) and deprecate the data specific experience.

Working ahead of this would assume design work that hasn't happened yet and introduce an extra DX step to get the config from a function that Josef would prefer we avoid.

sobolk: Ok. If we thought about deprecation strategy then it's fine.

packages/backend-function/src/runtime/get_amplify_clients_configuration.ts
stocaaro: Got it. Making the change. Should the export function name be prefixed with internal. eg `internalGetAmplifyClientConfiguration` or similar?

sobolk: yes if customer is not supposed to call it themselves

packages/backend-function/src/runtime/get_amplify_clients_configuration.ts
```diff
@@ -0,0 +1,147 @@
...
+      throw new Error(
+        'Error retrieving the schema from S3. Please confirm that your project has a `defineData` included in the `defineBackend` definition.'
+      );
+    } else if (caught instanceof S3ServiceException) {
+      throw new Error(
+        `Error retrieving the schema from S3. You may need to grant this function authorization on the schema. ${caught.name}: ${caught.message}.`
+      );
+    } else {
+      throw caught;
+    }
```
Amplifiyer, line 139: any throw should be an instance of `AmplifyError`

packages/backend-function/src/function_data_config_generator.ts
stocaaro: In the conversation this morning, we decided to move to a `getAmplifyDataClientConfig` function exposed directly from `backend/function/runtime`. Resolving this with that update.

packages/backend-function/API.md
```diff
@@ -6,14 +6,35 @@
...
+        DataClientConfig,
+        DataClientReturn,
+        getAmplifyDataClientConfig
+    }
+}
+export { __export__get_amplify_clients_configuration }
+
+declare namespace __export__runtime {
+    export {
+        getAmplifyDataClientConfig
```
stocaaro, line 25: It's probably redundant to export this from the index path where it is exported from. Should I remove this, or is it good to keep it since this is where it is consumed?

packages/backend-function/src/runtime/get_amplify_clients_configuration.ts
```diff
@@ -127,13 +139,14 @@ export const getAmplifyDataClientConfig = async <T>(
...
+        message:
+          'Error retrieving the schema from S3. Please confirm that your project has a `defineData` included in the `defineBackend` definition.',
+      });
     } else if (caught instanceof S3ServiceException) {
-      throw new Error(
-        `Error retrieving the schema from S3. You may need to grant this function authorization on the schema. ${caught.name}: ${caught.message}.`
-      );
+      throw new AmplifyFault('S3ConfigFault', {
+        message: `Error retrieving the schema from S3. You may need to grant this function authorization on the schema. ${caught.name}: ${caught.message}.`,
+      });
```
Amplifiyer: Sorry I realized later that this code runs in lambda and not in the ampx binary. These should be fine with `Error`.

stocaaro: I'll revert this back to using errors. Thanks for the update.

packages/backend-function/API.md
sobolk: We should only have `index`.

I.e. match this entry point from package.json.

```
    "./runtime": {
      "types": "./lib/runtime/index.d.ts",
      "import": "./lib/runtime/index.js",
      "require": "./lib/runtime/index.js"
```

sobolk: the goal is to have api extractor "see" that entry point and report types accessible from there.

packages/backend-function/src/index.internal.ts
```diff
@@ -0,0 +1,11 @@
...
+import * as __export__runtime from './runtime/index.js';
+// eslint-disable-next-line @typescript-eslint/naming-convention
+import * as __export__get_amplify_clients_configuration from './runtime/get_amplify_clients_configuration.js';
+
+/*
+ Api-extractor does not ([yet](https://github.com/microsoft/rushstack/issues/1596)) support multiple package entry points
+ Because this package has a submodule export, we are working around this issue by including that export here and directing api-extract to this entry point instead
+ This allows api-extractor to pick up the submodule exports in its analysis
+ */
+export { __export__runtime, __export__get_amplify_clients_configuration };
```
sobolk, line 11:
```suggestion
// eslint-disable-next-line @typescript-eslint/naming-convention
import * as __export__runtime from './runtime/index.js';

/*
 Api-extractor does not ([yet](https://github.com/microsoft/rushstack/issues/1596)) support multiple package entry points
 Because this package has a submodule export, we are working around this issue by including that export here and directing api-extract to this entry point instead
 This allows api-extractor to pick up the submodule exports in its analysis
 */
export { __export__runtime };
```

packages/backend-function/API.md
stocaaro: Updated.

packages/backend/src/index.internal.ts
```diff
@@ -1,4 +1,6 @@
 export * from './index.js';
+// eslint-disable-next-line @typescript-eslint/naming-convention
+import * as __export__function_runtime from './function/runtime/index.js';
```
sobolk, line 3:
```suggestion
import * as __export__function__runtime from './function/runtime/index.js';
```

`__` is converted to `/` in api checks.

packages/backend-function/src/runtime/get_amplify_clients_configuration.ts
```diff
@@ -0,0 +1,177 @@
...
+
+/* eslint-disable @typescript-eslint/naming-convention */
+export type ResourceConfig = {
+  API: {
+    GraphQL: {
+      endpoint: string;
+      region: string;
+      defaultAuthMode: string;
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      modelIntrospection: any;
```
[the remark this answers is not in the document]
stocaaro: It would not. The [ResourceConfig](https://github.com/aws-amplify/amplify-js/blob/main/packages/core/src/singleton/types.ts#L48-L57) type is full of optional inputs, but the [API type for graphql](https://github.com/aws-amplify/amplify-js/blob/main/packages/core/src/singleton/API/types.ts#L21-L48) starts to get pickier and the [MIS typing](https://github.com/aws-amplify/amplify-js/blob/main/packages/core/src/singleton/API/types.ts#L143-L153) is narrower still. I started reproducing enough typing to make this happy and ended up >100 loc of type before I decided it was better to advocate for a more permissive type in this case.

The experience with `Record<string, unknown>` is:
image

The experience with ` unknown` is:
image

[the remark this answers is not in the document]
stocaaro: My minimal type for MIS added 122 loc, and 22 types. Recommend keeping `any` and I have added a comment for context.

.changeset/brave-cheetahs-repeat.md
```diff
@@ -0,0 +1,8 @@
+---
+'@aws-amplify/backend': patch
+'@aws-amplify/backend-function': patch
+'@aws-amplify/backend-data': patch
+'@aws-amplify/integration-tests': patch
+---
+
+feat: Add lambda data client
```
sobolk, line 8:
```suggestion
---
'@aws-amplify/backend': minor
'@aws-amplify/backend-function': minor
'@aws-amplify/backend-data': patch
---

Add lambda data client
```

Explanation:
1. We don't need integration tests here. "Private" packages are excluded from changeset checks.
2. `backend-function` is adding new API. This warrants minor bump.
3. `backend` inherits max of function/data bumps.
4. We're not using conventional commits in our changelogs. See example changelog https://github.com/aws-amplify/amplify-backend/blob/main/packages/backend-function/CHANGELOG.md .

sobolk (commented):
LGTM

packages/backend-function/src/runtime/get_amplify_clients_configuration.ts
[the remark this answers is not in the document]
stocaaro: To resolve this discussion. Offline we decided to keep this `any`.

sobolk (dismissed):

packages/integration-tests/src/test-project-setup/data_and_function_project.ts
```diff
@@ -0,0 +1,192 @@
...
+      }
+    );
+    return project;
+  };
+}
+
+/**
+ * The data and function test project.
+ */
+class DataAndFunctionTestProject extends TestProjectBase {
```
Amplifiyer, line 90: Can we rename this `DataAccessFromFunctionTestProject` to be more specific.

packages/integration-tests/src/test-projects/data-and-function/amplify/functions/noop-import/handler.ts
```diff
@@ -0,0 +1,22 @@
...
+const { resourceConfig, libraryOptions } = await getAmplifyDataClientConfig(
+  env
+);
+
+Amplify.configure(resourceConfig, libraryOptions);
+
+const client = generateClient<Schema>();
+
+export const handler: Handler = async () => {
+  const _s3Client = new S3Client();
```
Amplifiyer, line 19: Just for my understanding, what's the significance of this `noop-import` function?

stocaaro: I added this test as a followup to conversation with @sobolk about having a function that uses `S3Client` separate from the usage in `getAmplifyDataClientConfig` that gets the MIS. 

My understanding is that with the SSM integration there was an issue that occurred when the user used SSM in their handler, which is [tested for here](https://github.com/aws-amplify/amplify-backend/blob/main/packages/integration-tests/src/test-projects/advanced-auth-and-functions/amplify/func-src/handler_with_ssm.ts)

Amplifiyer: Got it, I'd think if the related SSM test works then it would work for other sdk clients as well, and we have this other test https://github.com/aws-amplify/amplify-backend/blob/main/packages/integration-tests/src/test-projects/advanced-auth-and-functions/amplify/func-src/handler_with_aws_sdk.ts

If there is still value for this test, I'd rename this as `customer-s3-import` or something similar to make it clear the purpose of this test.

sobolk: > I'd think if the related SSM test works then it would work for other sdk clients as well

Unfortunately, asserting that "SSM imports in banner" are in such a way that don't affect customer's SSM imports does not imply that S3 scenario works. These are two instances of similar pattern, but their "content" does not overlap, hence should be asserted separately.

sobolk: Regarding https://github.com/aws-amplify/amplify-backend/blob/main/packages/integration-tests/src/test-projects/advanced-auth-and-functions/amplify/func-src/handler_with_aws_sdk.ts .

That project was created to assert that we're bundling SDKs. 
It does not have `data` however, so can't be used to assert that "S3 client works AND data client works at the same time". Then, adding `data` there would make it overloaded.

sobolk: Perhaps this should be `AdvancedDataScenariosTestProject` where primary purpose is to dump scenarios originating from data (whether they involve other resources or not).

Amplifiyer: My general recommendation is to start specific and go wide if/when there is need be. The next time someone adds something to this test, it would be purposely added instead of "anything data related that seems advanced to me".

sobolk: Sounds good. Let's do `DataAccessFromFunctionTestProject` then.

packages/integration-tests/src/test-project-setup/data_access_from_function_project.ts
```diff
@@ -0,0 +1,195 @@
...
+        query noopImport {
+          noopImport
+        }
+      `,
+      variables: {},
+    });
+
+    assert.deepEqual(response.data, { noopImport: 'STATIC TEST RESPONSE' });
+  };
+}
```
sobolk, line 195: I don't seem to see any code that would call the lambda.
If the feature is about lambda accessing appsync, we should be invoking lambda and asseserting that lambda intracted with appsync.

The test as is doesn't prove that lambda is working.

Some ideas how to test lambda can be found [here](https://github.com/aws-amplify/amplify-backend/blob/d227f96793fd6a3e181af96bbb1ef2f0d3ab7ae4/packages/integration-tests/src/test-project-setup/data_storage_auth_with_triggers.ts#L217-L241) - we assert that lambda interacts with s3 bucket there, similar strategy can be applied here for appsync.

stocaaro: Both of these tests call a query, which each [call a lambda](https://github.com/aws-amplify/amplify-backend/pull/2224/files#diff-6ab6d1ba968c8f4bd0c0f18c3c6ab9f5c80cc036055ed09b00bd3a1a4bb21f32R13-R25) and the response assertions are specific to the lambda being called ([todoCount](https://github.com/aws-amplify/amplify-backend/pull/2224/files#diff-2807ad44b84ffd9d18ec3423889231abb0a54bf424b0be74c3295685625c94d3R19) and [noopImport](https://github.com/aws-amplify/amplify-backend/pull/2224/files#diff-623f11d82c5ce3d595e11768cfbcaf371e3405fb30668a217bd76eea61ade840R21)).

I considered testing the lambda's in isolation (without wiring these calls through AppSync query calls), but that use-case is a simpler subset of the use-cases covered by these tests. 

Can you help me understand the coverage gap that needs to be filled?

sobolk: I see. There's no coverage gap then.

Can you add JSDoc comment to both `private assert...` methods explaining that they indeed call the lambda and how ? For the reader it's not obvious that GQL calls translate to lambda invocations here.

stocaaro: Understood. JSDoc added

packages/integration-tests/src/test-project-setup/data_access_from_function_project.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    const clientConfig = await generateClientConfig(backendId, '1.1');
+    if (!clientConfig.data?.url) {
+      throw new Error('Data and function project must include data');
+    }
+    if (!clientConfig.data.api_key) {
+      throw new Error('Data and function project must include api_key');
+    }
+
+    // const dataUrl = clientConfig.data?.url;
```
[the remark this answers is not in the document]
stocaaro: You mean the commented code (removed), or is there something about the error thrown validating the api key that isn't good?

packages/integration-tests/src/test-project-setup/data_access_from_function_project.ts
```diff
@@ -0,0 +1,195 @@
...
+  constructor(
+    name: string,
+    projectDirPath: string,
+    projectAmplifyDirPath: string,
+    cfnClient: CloudFormationClient,
+    amplifyClient: AmplifyClient,
+    private readonly lambdaClient: LambdaClient = new LambdaClient(
+      e2eToolingClientConfig
+    ),
+    private readonly cognitoIdentityProviderClient: CognitoIdentityProviderClient = new CognitoIdentityProviderClient(
```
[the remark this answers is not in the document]
stocaaro: Removed

packages/integration-tests/src/test-project-setup/data_access_from_function_project.ts
```diff
@@ -0,0 +1,195 @@
...
+// TODO: this is a work around
+// it seems like as of amplify v6 , some of the code only runs in the browser ...
+// see https://github.com/aws-amplify/amplify-js/issues/12751
+if (process.versions.node) {
+  // node >= 20 now exposes crypto by default. This workaround is not needed: https://github.com/nodejs/node/pull/42083
+  if (new SemVer(process.versions.node).major < 20) {
+    // @ts-expect-error altering typing for global to make compiler happy is not worth the effort assuming this is temporary workaround
+    globalThis.crypto = crypto;
+  }
+}
```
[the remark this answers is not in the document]
stocaaro: Removed

sobolk (approved):

packages/integration-tests/src/test-project-setup/data_access_from_function_project.ts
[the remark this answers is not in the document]
sobolk: > You mean the commented code (removed)

Yes.

PR #2228, opened 2024-11-13
test: Add e2e test for lambda data client

stocaaro:
## Problem
The lambda data client needs to be covered by an e2e test.

## Changes
- Adds an e2e test for the lambda data client

## Validation
Tested using
`npm run test:dir packages/integration-tests/lib/test-e2e/sandbox/data_and_function_project.sandbox.test.js`


packages/integration-tests/src/test-projects/data-and-function/amplify/functions/todo-count/handler.ts
```diff
@@ -0,0 +1,18 @@
+import type { Handler } from 'aws-lambda';
+import { Amplify } from 'aws-amplify';
+import { generateClient } from 'aws-amplify/data';
+import type { Schema } from '../../data/resource.js';
+// @ts-ignore
```
stocaaro, line 5: Esbuild during synth is fine with all of this, but the ts check before synth is unhappy. It looks like a workaround was used for `data-storage-auth-with-triggers-ts` of setting up fake files to make the pre-synth check happy. Should I replicate that here? It would avoid the exclude change in the tsconfig below too.

Seems fine, but also very fake since this file isn't being built in the project.

stocaaro:
This change has been merged into https://github.com/aws-amplify/amplify-backend/pull/2224

PR #2233, opened 2024-11-14
add resourceGroupName prop to function

rtpascual:
## Problem

Certain usage patterns with functions end up with `Caused By: ❌ Deployment failed: Error [ValidationError]: Circular dependency between resources: ...`

This happens in scenarios where it could be avoided if functions created with `defineFunction` were not in the same stack.

## Changes

Add `resourceGroupName` prop to `defineFunction` in order to provide the flexibility to designate a stack for a function resource.

Also introduces `AmplifyResourceGroupName` type which has default Amplify resource group names for each resource (example: `defineData`'s resourceGroupName = 'data') that will show up in IntelliSense:
Screenshot 2024-11-14 at 4 36 49 PM

So with `defineFunction` it will be possible to:
- Group with existing Amplify resources (example: group auth triggers with auth resources by setting `resourceGroupName = 'auth'`)
- Be in its own group with a custom string

## Validation

Two new E2E test projects for common use cases that run into circular dependency issues: `circular-dep-auth-data-func` and `circular-dep-data-func`. I have them in two separate projects instead of one to isolate use cases. I also considered using `TestProjectBase.getUpdates` to swap out all files from one use case to another to keep things in one project but this will take longer to run.

## Checklist

### DIFF
--- a/packages/backend-function/src/factory.ts
+++ b/packages/backend-function/src/factory.ts
@@ -155,6 +155,8 @@ export type FunctionProps = {
   /**
    * Group the function with existing Amplify resources or separate the function into its own group.
    * @default 'function' // grouping with other Amplify functions
+   * @example
+   * resourceGroupName: 'auth' // to group an auth trigger with an auth resource
    */
   resourceGroupName?: AmplifyResourceGroupName;
 };
--- a/packages/backend-function/src/factory.ts
+++ b/packages/backend-function/src/factory.ts
@@ -9,6 +9,7 @@ import {
   TagName,
 } from '@aws-amplify/platform-core';
 import {
+  AmplifyResourceGroupName,
   BackendOutputStorageStrategy,
   BackendSecret,
   BackendSecretResolver,
@@ -150,6 +151,12 @@ export type FunctionProps = {
    * Options for bundling the function code.
    */
   bundling?: FunctionBundlingOptions;
+
+  /**
+   * Group the function with existing Amplify resources or separate the function into its own group.
+   * @default 'function' // grouping with other Amplify functions
+   */
+  resourceGroupName?: AmplifyResourceGroupName;
 };
 
 export type FunctionBundlingOptions = {
@@ -208,6 +215,7 @@ class FunctionFactory implements ConstructFactory<AmplifyFunction> {
       schedule: this.resolveSchedule(),
       bundling: this.resolveBundling(),
       layers,
+      resourceGroupName: this.props.resourceGroupName ?? 'function',
     };
   };
 
@@ -339,12 +347,14 @@ class FunctionFactory implements ConstructFactory<AmplifyFunction> {
 type HydratedFunctionProps = Required<FunctionProps>;
 
 class FunctionGenerator implements ConstructContainerEntryGenerator {
-  readonly resourceGroupName = 'function';
+  readonly resourceGroupName: AmplifyResourceGroupName;
 
   constructor(
     private readonly props: HydratedFunctionProps,
     private readonly outputStorageStrategy: BackendOutputStorageStrategy<FunctionOutput>
-  ) {}
+  ) {
+    this.resourceGroupName = props.resourceGroupName;
+  }
 
   generateContainerEntry = ({
     scope,
--- a/packages/backend-storage/src/construct.ts
+++ b/packages/backend-storage/src/construct.ts
@@ -8,6 +8,7 @@ import {
   IBucket,
 } from 'aws-cdk-lib/aws-s3';
 import {
+  AmplifyResourceGroupName,
   BackendOutputStorageStrategy,
   ConstructFactory,
   FunctionResources,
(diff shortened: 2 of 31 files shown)


packages/backend-storage/src/construct.ts
```diff
@@ -63,6 +64,13 @@ export type AmplifyStorageProps = {
       ConstructFactory<ResourceProvider<FunctionResources>>
     >
   >;
+  /**
+   * Group the storage resource with existing Amplify resources or separate the storage resource into its own group.
+   *
+   * Warning: Changing this is a destructive change for the storage resource.
```
rtpascual, line 70: I'm not sure if there is a way to prompt if `resourceGroupName` value changes for storage since this package isn't CLI-like.

I was also thinking if we should add another prop like `allowDestructiveChanges = true` and error if `resourceGroupName` has a value and this is false.

packages/plugin-types/src/amplify_resource_group_name.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Represents the types of resource group name
+ */
+export type AmplifyResourceGroupName =
+  | 'auth'
+  | 'conversationHandlerFunction'
```
awsluja, line 6: should this be here?

packages/plugin-types/src/amplify_resource_group_name.ts
```diff
@@ -0,0 +1,14 @@
...
+  | 'data'
+  | 'function'
```
awsluja, line 8: same with function i think only core resource groups should be here, auth data and storage

packages/backend-function/src/factory.ts
```diff
@@ -150,6 +151,12 @@ export type FunctionProps = {
    * Options for bundling the function code.
    */
   bundling?: FunctionBundlingOptions;
+
+  /**
+   * Group the function with existing Amplify resources or separate the function into its own group.
+   * @default 'function' // grouping with other Amplify functions
+   */
```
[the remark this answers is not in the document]
awsluja: @josefaidt , my 2 cents, start with "Optional ... " and include recommendations that cover base use cases, ie "if you want to add an auth trigger... set this to... " etc

packages/backend-storage/src/construct.ts
awsluja: We're supporting resourceGroupName for storage too?

awsluja: I don't think we should support this for storage just yet , but if we need it to cover an existing use case then we'll need a test around it

packages/plugin-types/src/amplify_resource_group_name.ts
rtpascual: I was thinking to add all the resource groups we define to be consistent across the board. Also in the case where customers change the resource group of a function and want to change it back the resource group for functions is discoverable.

sobolk: I think narrowing down this makes sense given that arbitrary string can also be provided. This can be expanded later.

The consistency of our internal stuff is secondary concern here.

packages/backend-storage/src/construct.ts
rtpascual: Currently there's no use case reported, I'll remove it for now and we can add this back when we need to expand to storage.

packages/backend-function/src/factory.ts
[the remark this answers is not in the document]
rtpascual: Good point, added an example.

Amplifiyer (approved):

sobolk (dismissed):

sobolk (approved):

packages/backend-function/src/factory.ts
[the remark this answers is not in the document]
josefaidt: looking at the default, is "function" a valid value?

Amplifiyer (approved):

josefaidt:
cross-posting to main thread, for the jsdoc `@default`, is `"function"` a valid value? we can show the default is undefined if not https://github.com/aws-amplify/amplify-backend/pull/2233/files#diff-9e302baf9085cc0328df9e0e7779db3be80f85e6ec8157baffae911472642f51R30

packages/backend-function/src/factory.ts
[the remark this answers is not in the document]
rtpascual: "function" is what we currently have as the resourceGroupName in our construct factory, see https://github.com/aws-amplify/amplify-backend/blob/main/packages/backend-function/src/factory.ts#L342. This change will update the value if customers pass in a value for `resourceGroupName`, otherwise it falls back to default behavior.

sobolk:
> cross-posting to main thread, for the jsdoc `@default`, is `"function"` a valid value? we can show the default is undefined if not https://github.com/aws-amplify/amplify-backend/pull/2233/files#diff-9e302baf9085cc0328df9e0e7779db3be80f85e6ec8157baffae911472642f51R30

@josefaidt  https://github.com/aws-amplify/amplify-backend/blob/f83d6d58cf9b167c6b4aa59e6be92ee4e9051ab5/packages/backend-function/src/factory.ts#L342

PR #2260, opened 2024-11-21
fix: truncate large error messages before printing to customer

Amplifiyer:
## Problem

Sometimes CDK CLI prints a massive uglified code on the stderr when an error happens. This may mask the actual error message to be shown to customers.

## Changes

Truncate and keep the later part of the error message.

## Validation

Tested manually. I couldn't figure out a clean way to mock execa child process to be able to test this.

## Checklist


packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -199,14 +199,23 @@ export class CDKDeployer implements BackendDeployer {
       // However if the cdk process didn't run or produced no output, then we have nothing to go on with. So we throw
       // this error to aid in some debugging.
       if (aggregatedStderr.trim()) {
+        // If the string is more than 65KB, truncate and keep the last portion.
         // eslint-disable-next-line amplify-backend-rules/prefer-amplify-errors
-        throw new Error(aggregatedStderr);
+        throw new Error(this.truncateString(aggregatedStderr, 65000));
```
sobolk, line 204: Would it make sense to do something like:

```
new Error(`${this.head(aggregatedStderr, 55000)}... truncated ... ${this.tail(aggregatedStderr)}`);
```
so that there's a chance to see fragment of root cause if it's at the end?

If we don't then w should at least indicate that message is truncated.
```
new Error(`${this.truncateString(aggregatedStderr, 65000)}... truncated`);
```

packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -199,14 +199,23 @@ export class CDKDeployer implements BackendDeployer {
...
       }
     }
   };
 
+  private truncateString = (str: string, size: number) => {
+    const encoder = new TextEncoder();
+    const decoder = new TextDecoder();
+    const truncated = encoder.encode(str).slice(-size); // truncate the beginning
+
+    return decoder.decode(truncated);
```
sobolk, line 216: I'd suggest to simplify this and just take 65k characters instead of bytes.

The reason is that:
1. it seems node is using UTF-8 encoding for strings by default.
2. this transformation is transforming it back and forth to `Uint8Array`.

That seems that effectively "number of bytes" is "number of characters". (or perhaps relationship is at least constant in case default encoding is different than what I mentioned above).

Amplifiyer: That's what I had it in the beginning but learned that UTF-8 is [actually 1-4 bytes](https://www.ibm.com/docs/en/db2-for-zos/12?topic=unicode-utfs), and the VARCHAR used in DB counts bytes and not characters.

packages/backend-deployer/src/cdk_deployer.ts
Amplifiyer: Updated

packages/backend-deployer/src/cdk_deployer.ts
sobolk: I see.
Could you check if `Buffer.byteLength(string, encoding)` could be used here?
It may be a bit faster and avoid some data copies.

packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -199,14 +199,24 @@ export class CDKDeployer implements BackendDeployer {
...
       }
     }
   };
 
+  private truncateString = (str: string, size: number) => {
+    const encoder = new TextEncoder();
+    const decoder = new TextDecoder();
+    const encoded = encoder.encode(str);
+    return encoded.byteLength > size
+      ? '...truncated...' + decoder.decode(encoded.slice(-size))
```
sobolk, line 216: nit: I'd expect the `...truncated...` to be at the end.

sobolk (approved):

packages/backend-deployer/src/cdk_deployer.ts
Amplifiyer: We are truncated from the beginning, in the hopes of the real error message to be there. Hence "slice(**-size**)"

ShadowCat567 (approved):

PR #2263, opened 2024-11-21
add more forms of transform errors to cdk error mapping

rtpascual:
## Problem

There are more forms of transform errors:

```
Error [TransformError]: 
You installed esbuild for another platform than the one you're currently using.
This won't work because esbuild is written with native code and needs to
install a platform-specific binary executable.

Specifically the @esbuild/linux-arm64 package is present but this platform
needs the @esbuild/darwin-arm64 package instead. People often get into this
situation by installing esbuild on Windows or macOS and copying node_modules
into a Docker image that runs Linux, or by copying node_modules between
Windows and WSL environments.

If you are installing with npm, you can try not copying the node_modules
directory when you copy the files over, and running npm ci or npm install
on the destination platform after the copy. Or you could consider using yarn
instead of npm which has built-in support for installing a package on multiple
platforms simultaneously.

If you are installing with yarn, you can try listing both this platform and the
other platform in your .yarnrc.yml file using the supportedArchitectures
feature: https://yarnpkg.com/configuration/yarnrc/#supportedArchitectures
Keep in mind that this means multiple copies of esbuild will be present.

Another alternative is to use the esbuild-wasm package instead, which works
the same way on all platforms. But it comes with a heavy performance cost and
can sometimes be 10x slower than the esbuild package, so you may also not
want to do that.

    at ...
```

```
Error: Transform failed with 1 error:
/some-path/amplify/storage/resource.ts:7:0: ERROR: Expected identifier but found }
    at ...
```

```
Error [TransformError]: The package esbuild-package could not be found, and is needed by esbuild.

If you are installing esbuild with npm, make sure that you don't specify the
--no-optional or --omit=optional flags. The optionalDependencies feature
of package.json is used by esbuild to install the correct binary executable
for your current platform.
    at ...
```

## Changes

CDK error mapping update and addition.

## Validation

Unit tests.

## Checklist

### DIFF
--- a/packages/backend-deployer/src/cdk_error_mapper.ts
+++ b/packages/backend-deployer/src/cdk_error_mapper.ts
@@ -248,11 +248,22 @@ export class CdkErrorMapper {
     {
       // If there are multiple errors, capture all lines containing the errors
       errorRegex: new RegExp(
-        `\\[TransformError\\]: Transform failed with .* error(s?):${this.multiLineEolRegex}(?<esBuildErrorMessage>(.*ERROR:.*${this.multiLineEolRegex})+)`
+        `(\\[TransformError\\]|Error): Transform failed with .* error(s?):${this.multiLineEolRegex}(?<esBuildErrorMessage>(.*ERROR:.*${this.multiLineEolRegex})+)`
       ),
       humanReadableErrorMessage: '{esBuildErrorMessage}',
       resolutionMessage:
         'Fix the above mentioned type or syntax error in your backend definition.',
+      errorName: 'ESBuildError',
+      classification: 'ERROR',
+    },
+    {
+      // Captures other forms of transform error
+      errorRegex: new RegExp(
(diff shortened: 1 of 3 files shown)


packages/backend-deployer/src/cdk_error_mapper.ts
```diff
@@ -248,14 +248,25 @@ export class CdkErrorMapper {
...
       classification: 'ERROR',
     },
+    {
+      // Captures other forms of transform error
+      errorRegex: new RegExp(
+        `Error \\[TransformError\\]:(${this.multiLineEolRegex}|\\s)?(?<esBuildErrorMessage>(.*(${this.multiLineEolRegex})?)+)`
+      ),
+      humanReadableErrorMessage: '{esBuildErrorMessage}',
+      resolutionMessage:
+        'Make sure esbuild is installed and is for the platform you are currently using.',
```
sobolk, line 266:
```suggestion
        'Make sure esbuild is installed and is compatible with the platform you are currently using.',
```

sobolk (dismissed):

awsluja (approved):

sobolk (approved):

PR #2042, opened 2024-09-23
`@aws-amplify/backend` should inherit version bump of same kind from dependencies

ShadowCat567:
## Problem

`@aws-amplify/backend` sometimes does not properly inherit the right kind of version bump from its dependencies.

**Issue number, if available:** https://github.com/aws-amplify/amplify-backend/issues/1693

## Changes

- refactored check_changeset_completeness.ts to be split into functions
- added a function to check_changeset_completeness.ts that checks whether backend should be inheriting a version bump and whether it is currently inheriting the right kind of version bump
**Corresponding docs PR, if applicable:**

## Validation

- ran manual tests using different version bumps for `backend` and its dependencies

## Checklist

- [ ] If this PR includes a functional change to the runtime behavior of the code, I have added or updated automated test coverage for this change.
- [ ] If this PR requires a change to the [Project Architecture README](../PROJECT_ARCHITECTURE.md), I have included that update in this PR.
- [ ] If this PR requires a docs update, I have linked to that docs PR above.
- [ ] If this PR modifies E2E tests, makes changes to resource provisioning, or makes SDK calls, I have run the PR checks with the `run-e2e` label set.

_By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license._

PR #2273, opened 2024-11-26
feat: add api id and amplify environment name to stash

dpilch:
## Problem

Customers are not able to use DDB batch operations in custom JS resolvers. The customer needs the DDB table name to perform batch operations. The customer can construct the table name from the GraphQL API ID and the Amplify environment name (`<model-name>-<graphql-api-id>-<environment-name>`). 

**Issue number, if available:** https://github.com/aws-amplify/amplify-category-api/issues/408#issuecomment-2307154077

## Changes

Add GraphQL API ID and the Amplify environment name to resolver context stash.

**Corresponding docs PR, if applicable:** https://github.com/aws-amplify/docs/pull/8145

## Validation

* Add unit test
* Test in sample app

## Checklist

### DIFF
--- a/.changeset/old-hornets-type.md
+++ b/.changeset/old-hornets-type.md
@@ -1,0 +1,5 @@
+---
+'@aws-amplify/backend-cli': patch
+---
+
+add validation for sandbox set command secret-name
--- a/.changeset/swift-mirrors-enjoy.md
+++ b/.changeset/swift-mirrors-enjoy.md
@@ -1,0 +1,5 @@
+---
+'@aws-amplify/model-generator': patch
+---
+
+Updated error handling with S3 Client
--- a/packages/cli/src/commands/sandbox/sandbox-secret/sandbox_secret_set_command.test.ts
+++ b/packages/cli/src/commands/sandbox/sandbox-secret/sandbox_secret_set_command.test.ts
@@ -1,7 +1,10 @@
 import { beforeEach, describe, it, mock } from 'node:test';
 import { AmplifyPrompter, printer } from '@aws-amplify/cli-core';
 import yargs, { CommandModule } from 'yargs';
-import { TestCommandRunner } from '../../../test-utils/command_runner.js';
+import {
+  TestCommandError,
+  TestCommandRunner,
+} from '../../../test-utils/command_runner.js';
 import assert from 'node:assert';
 import { SandboxBackendIdResolver } from '../sandbox_id_resolver.js';
 import {
@@ -11,6 +14,7 @@ import {
 import { SandboxSecretSetCommand } from './sandbox_secret_set_command.js';
 import { ReadStream } from 'node:tty';
 import { PassThrough } from 'node:stream';
+import { AmplifyError } from '@aws-amplify/platform-core';
 
 const testSecretName = 'testSecretName';
 const testSecretValue = 'testSecretValue';
@@ -143,6 +147,23 @@ void describe('sandbox secret set command', () => {
       testSecretName,
       testSecretValue,
     ]);
+  });
+
+  void it('throws AmplifyUserError if invalid secret name is provided', async () => {
+    const invalidSecretName = 'invalid@';
+    await assert.rejects(
+      () => commandRunner.runCommand(`set ${invalidSecretName}`),
+      (err: TestCommandError) => {
+        assert.ok(AmplifyError.isAmplifyError(err.error));
+        assert.strictEqual(
+          err.error.message,
+          'Invalid secret name provided: invalid@'
+        );
+        assert.strictEqual(err.error.name, 'InvalidCommandInputError');
+        return true;
+      }
+    );
+    assert.equal(secretSetMock.mock.callCount(), 0);
   });
 
   void it('show --help', async () => {
--- a/packages/cli/src/commands/sandbox/sandbox-secret/sandbox_secret_set_command.ts
+++ b/packages/cli/src/commands/sandbox/sandbox-secret/sandbox_secret_set_command.ts
@@ -6,6 +6,7 @@ import { ArgumentsKebabCase } from '../../../kebab_case.js';
 import { SandboxCommandGlobalOptions } from '../option_types.js';
 import { once } from 'events';
 import { ReadStream } from 'node:tty';
+import { AmplifyUserError } from '@aws-amplify/platform-core';
 
 /**
  * Command to set sandbox secret.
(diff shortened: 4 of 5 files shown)


packages/backend-data/src/convert_js_resolvers.ts
```diff
@@ -71,12 +47,38 @@ export const convertJsResolverDefinition = (
...
+       * Customer-provided handlers are added as a Functions list in `pipelineConfig.functions`
+       *
+       * Use inline code to remove circular dependency when adding the API ID.
+       */
+      code: `
+        /**
+         * Pipeline resolver request handler
+         */
+        export const request = (ctx) => {
+          ctx.stash.apiId = '${amplifyApi.apiId}';
```
palpatim, line 71: I'd normally recommend we namespace all amplify variables, as in:
- `amplifyAppsyncApiId`
- `amplifyEnvironmentName`
- `amplifyFooBar...`

However! Since we already have an output name for API ID, I wonder if we should use that:

- `awsAppsyncApiId`

For parallelism, that makes the other variable something like:

- `awsAmplifyEnvironmentName`

sobolk (changes requested):

packages/backend-data/src/convert_js_resolvers.ts
```diff
@@ -71,12 +47,38 @@ export const convertJsResolverDefinition = (
...
+          ctx.stash.awsAmplifyEnvironmentName = '${amplifyEnvironmentName}';
+          return {};
+        };
+        /**
+         * Pipeline resolver response handler
+         */
+        export const response = (ctx) => {
+          return ctx.prev.result;
+        };
+      `,
```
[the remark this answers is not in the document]
dpilch: I've changed to use the compiled file. We still have the risk of malformed JS because we need to add in the GraphQL API ID and Amplify environment name after compilation. Let me know if we think we still need a E2E test to cover this.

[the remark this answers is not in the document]
sobolk: thanks.

> Let me know if we think we still need a E2E test to cover this.

if app id and env injection can be run as part of unit/integration test then e2e is not required. 

i.e. in the unit/integration test we could:
1. run injection logic and dump resulting file on disk.
2. dynamically import it
3. run test scenarios on dynamic import.

packages/backend-data/src/convert_js_resolvers.test.ts
```diff
@@ -28,6 +33,29 @@
...
 
+void describe('defaultJsResolverCode', () => {
+  void it('returns the default JS resolver code with api id and env name in valid JS', async () => {
+    const code = defaultJsResolverCode('testApiId', 'testEnvName');
+    assert(code.includes("ctx.stash.awsAppsyncApiId = 'testApiId';"));
+    assert(code.includes("ctx.stash.amplifyBranchName = 'testEnvName';"));
+
+    const tempDir = tmpdir();
+    const filename = join(tempDir, 'js_resolver_handler.js');
+    writeFileSync(filename, code);
```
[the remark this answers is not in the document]
dpilch: Shouldn't be a concern here because this is in a unit test.

sobolk (dismissed):

sobolk (dismissed):

packages/backend-data/src/convert_js_resolvers.ts
```diff
@@ -18,17 +19,22 @@ const JS_PIPELINE_RESOLVER_HANDLER = './assets/js_resolver_handler.js';
...
     '../../lib',
     JS_PIPELINE_RESOLVER_HANDLER
   );
 
-  return new Asset(scope, 'default_js_resolver_handler_asset', {
-    path: resolveEntryPath(resolvedTemplatePath),
-  });
+  return readFileSync(resolvedTemplatePath, 'utf-8')
+    .replace('${amplifyApiId}', amplifyApiId)
+    .replace('${amplifyEnvironmentName}', amplifyEnvironmentName);
```
palpatim, line 37: For future proofing: If templates reference the match patterns more than once, this will fail. Consider a global regexp:
```suggestion
    .replace(new RegExp('\$\{amplifyApiId\}', 'g'), amplifyApiId)
    .replace(new RegExp('\$\{amplifyEnvironmentName\}', 'g'), amplifyEnvironmentName);
```

If you do adopt this change, please add tests to verify.

sobolk (dismissed):

.changeset/strong-toes-sniff.md
```diff
@@ -0,0 +1,6 @@
+---
+'@aws-amplify/backend-data': minor
+'@aws-amplify/backend': minor
+---
```
Amplifiyer, line 4:
```suggestion
---
'@aws-amplify/backend-data': minor
---
```

packages/backend-data/src/convert_js_resolvers.ts
```diff
@@ -71,12 +78,15 @@ export const convertJsResolverDefinition = (
 
     const resolverName = `Resolver_${resolver.typeName}_${resolver.fieldName}`;
 
+    const amplifyEnvironmentName =
+      scope.node.tryGetContext('amplifyEnvironmentName') ?? 'NONE';
```
Amplifiyer, line 82: This context var `amplifyEnvironmentName` looks like is set by data construct exclusively? Is this the `branchName` for CI/CD deployments and `NONE` for sandbox?

.changeset/strong-toes-sniff.md
sobolk: that won't pass checks, see https://github.com/aws-amplify/amplify-backend/pull/2042 .

packages/backend-data/src/convert_js_resolvers.ts
dpilch: Yes, this is this the branchName for CI/CD deployments and NONE for sandbox.

.changeset/strong-toes-sniff.md
Amplifiyer: So this is still an issue?

packages/backend-data/src/convert_js_resolvers.ts
Amplifiyer: than we are using the wrong naming here. `amplifyEnvironmentName` suggests amplify wide name for the environment which is not the case here. May I suggest `amplifyApiEnvironmentName`?

.changeset/strong-toes-sniff.md
sobolk: Not an issue.
We enforce that `backend` has to inherit version bump from `backend-X` packages. We may revisit if that check makes sense, but that's why we have both in this PR.

packages/backend-data/src/convert_js_resolvers.ts
dpilch: The value is already used in the data construct. To use the new name without a breaking change we would need to set both variables, then it might be confusing why there are two variables with the same value. I think it is a better option to continue using the existing one.

Amplifiyer: I'm not asking to change in the data construct but rather the new variables created in the `backend-data` to which this value is assigned. As well as the new variable set in the `convert_js_resolvers.ts`. It's not a branch name since it's applicable for sandboxes as well.

dpilch: Oh sorry, you mean `const amplifyEnvironmentName` not what is context?

Amplifiyer: yes, I understand context vars are harder to change, but we should not create more vars with this naming (such as `amplifyBranchName`)

Amplifiyer (approved):

sobolk (approved):

packages/backend-data/src/convert_js_resolvers.ts
```diff
@@ -71,12 +78,15 @@ export const convertJsResolverDefinition = (
 
     const resolverName = `Resolver_${resolver.typeName}_${resolver.fieldName}`;
 
+    const amplifyApiEnvironmentName =
+      scope.node.tryGetContext('amplifyApiEnvironmentName') ?? 'NONE';
```
Amplifiyer, line 82: I thought we didn't want to change the context parameter, is this intended?

dpilch: Oops, hasty find and replace.

sobolk (approved):

Amplifiyer (approved):

Issue #2282, opened 2024-11-29
Add ephemeralStorageSize option to defineFunction


fossamagna:
### Environment information

```plain text
System:
  OS: macOS 14.7
  CPU: (10) arm64 Apple M1 Max
  Memory: 210.70 MB / 32.00 GB
  Shell: /bin/zsh
Binaries:
  Node: 20.11.1 - ~/.nvm/versions/node/v20.11.1/bin/node
  Yarn: 1.22.19 - ~/.nvm/versions/node/v20.11.1/bin/yarn
  npm: 10.2.4 - ~/.nvm/versions/node/v20.11.1/bin/npm
  pnpm: 8.15.6 - ~/.nvm/versions/node/v20.11.1/bin/pnpm
NPM Packages:
  @aws-amplify/auth-construct: 1.5.0
  @aws-amplify/backend: 1.7.0
  @aws-amplify/backend-auth: 1.4.0
  @aws-amplify/backend-cli: 1.4.2
  @aws-amplify/backend-data: 1.2.0
  @aws-amplify/backend-deployer: 1.1.9
  @aws-amplify/backend-function: 1.7.5
  @aws-amplify/backend-output-schemas: 1.4.0
  @aws-amplify/backend-output-storage: 1.1.3
  @aws-amplify/backend-secret: 1.1.4
  @aws-amplify/backend-storage: 1.2.2
  @aws-amplify/cli-core: 1.2.0
  @aws-amplify/client-config: 1.5.2
  @aws-amplify/deployed-backend-client: 1.4.1
  @aws-amplify/form-generator: 1.0.3
  @aws-amplify/model-generator: 1.0.9
  @aws-amplify/platform-core: 1.2.0
  @aws-amplify/plugin-types: 1.4.0
  @aws-amplify/sandbox: 1.2.5
  @aws-amplify/schema-generator: 1.2.5
  aws-amplify: 6.8.2
  aws-cdk: 2.167.1
  aws-cdk-lib: 2.167.1
  typescript: 5.6.3
No AWS environment variables
No CDK environment variables
```

### Describe the feature

I would like to add an option to defineFunction to specify the size of ephemeral storage.

```ts
defineFunction({
  entry: './test-assets/default-lambda/handler.ts',
  ephemeralStorageSize: 1024, // <- like this
})
```

See: 
- https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda.Function.html#ephemeralstoragesize
[...]

PR #2283, opened 2024-11-29
Add ephemeralStorageSizeMB option to defineFunction

fossamagna:
## Problem

#2282

## Changes

https://github.com/aws-amplify/docs/pull/8142

## Validation

Added test.

## Checklist


packages/backend-function/API.md
```diff
@@ -52,6 +52,7 @@ export type FunctionProps = {
     entry?: string;
     timeoutSeconds?: number;
     memoryMB?: number;
+    ephemeralStorageSize?: number;
```
josefaidt, line 55: can we add an `MB` suffix similar to `memoryMB`?

packages/backend-function/src/factory.ts
```diff
@@ -115,6 +115,13 @@ export type FunctionProps = {
    */
   memoryMB?: number;
 
+  /**
+   * The size of the function's /tmp directory in MiB.
+   * Must be a whole number.
+   * Default is 512MiB.
```
josefaidt, line 121:
```suggestion
   * @default 512
```

fossamagna: @josefaidt 
Thank you for your review and suggestion.
I have modified it as you suggested.

josefaidt (dismissed):
Thanks for taking the time to file this @fossamagna ! Change looks good to me from a dx perspective

packages/backend-function/API.md
```diff
@@ -93,6 +93,7 @@ export type FunctionProps = {
...
```
sobolk, line 96: Can we suffix this with `MB` ? we do that in property above.

packages/client-config/API.md
```diff
@@ -627,7 +627,7 @@ export type CustomClientConfig = {
 export const DEFAULT_CLIENT_CONFIG_VERSION: ClientConfigVersion;
 
 // @public
-export const generateClientConfig: <T extends "1" | "1.1" | "1.2" | "1.3" | "0">(backendIdentifier: DeployedBackendIdentifier, version: T, awsClientProvider?: AWSClientProvider<{
+export const generateClientConfig: <T extends "1.3" | "0" | "1" | "1.1" | "1.2">(backendIdentifier: DeployedBackendIdentifier, version: T, awsClientProvider?: AWSClientProvider<{
```
sobolk, line 630: the changes in `packages/client-config/API.md` and `packages/platform-core/API.md` are unexpected.

You can either revert these two files.
Or run `npm run clean && npm install && npm run build && npm update:api` . I.e. re-generate api reports from clean workspace.

packages/backend-function/API.md
sobolk: Some context:
It seems that CDK is just passing through this value:
see https://github.com/aws/aws-cdk/blob/b5c21890cf67ddda78c18eb0425f11a59fcadb34/packages/aws-cdk-lib/aws-lambda/lib/function.ts#L1057
and perhaps multiplies it by 1024 if necessary.
see https://github.com/aws/aws-cdk/blob/62a0638e325024fc0469ed50f9accd448d87bac7/packages/aws-cdk-lib/core/lib/size.ts#L133

But the lambda claims it's `MB` not `MiB` here https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-ephemeralstorage.html .

To confirm this. It might be good to deploy lambda with `ephemeralStorageSize` given to CDK construct and checking what value is printed in the AWS Console.

fossamagna: @sobolk
In CloudFormation and AWS Console, the unit of ehpemeralStorageSize for Lambda was `MB`.
I understood that only CDK uses `MiB` as a unit and it is appropriate to give MB to the suffix in Amplify. 
I fixed the part of JSDoc for ephemeralStorageSizeMB that used `MiB` has been corrected to `MB`. 
Thank you very much.

sobolk (approved):

packages/backend-function/src/factory.ts
```diff
@@ -324,6 +332,27 @@ class FunctionFactory implements ConstructFactory<AmplifyFunction> {
...
+      return ephemeralStorageSizeDefault;
+    }
+    if (
+      !isWholeNumberBetweenInclusive(
+        this.props.ephemeralStorageSizeMB,
+        ephemeralStorageSizeMin,
+        ephemeralStorageSizeMax
+      )
+    ) {
+      throw new Error(
```
Amplifiyer, line 349: Please throw an `AmplifyUserError` instead of a regular error. See line 327 for an example
https://github.com/aws-amplify/amplify-backend/pull/2283/files#diff-e354c6c4617ce06e0e16948f4521eca5f9269445265b80ed16720372d336c886R327

Amplifiyer (commented):
Changes look good, just need one update.

Amplifiyer (approved):

sobolk (approved):

PR #2292, opened 2024-12-03
fix: add a required input prompt for use in region input

Amplifiyer:
## Problem

Currently we accept "Enter" as acceptable input for region while configuring the profile. This results in `""` value to be stored as region which later fails validation

## Changes

fix: add a required input prompt for use in region input

## Checklist

### DIFF
--- a/packages/cli-core/API.md
+++ b/packages/cli-core/API.md
@@ -15,6 +15,9 @@ export class AmplifyPrompter {
         message: string;
         defaultValue?: string;
     }) => Promise<string>;
+    static requiredInput: (options: {
+        message: string;
+    }) => Promise<string>;
     static secretValue: (promptMessage?: string) => Promise<string>;
     static yesOrNo: (options: {
         message: string;
--- a/.changeset/shaggy-pens-sort.md
+++ b/.changeset/shaggy-pens-sort.md
@@ -1,0 +1,6 @@
+---
+'@aws-amplify/cli-core': patch
+'@aws-amplify/backend-cli': patch
+---
+
+add a required input prompt for use in region input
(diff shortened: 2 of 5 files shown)


packages/cli-core/API.md
```diff
@@ -15,6 +15,9 @@ export class AmplifyPrompter {
         message: string;
         defaultValue?: string;
     }) => Promise<string>;
+    static requiredInput: (options: {
+        message: string;
+    }) => Promise<string>;
```
sobolk, line 20: how about
```
    static input: (options: {
        message: string;
        defaultValue?: string;
        required?: boolean
    }) => Promise<string>;
```

above?

sobolk: (or some type union for `options` to make default and required mutually exclusive)

Amplifiyer: combining it was making it slightly cryptic without making a break in the API.

e.g.
```ts
{
    message: string;
    required: false;
    defaultValue?: string;
  } | {
    message: string;
    required: true;
  }
```
is a breaking change and
```ts
{
    message: string;
    defaultValue?: string;
  } | {
    message: string;
    required: true;
  }
```
allows both `defaultValue` and `required` shows up in the intellisense which is confusing API
```ts
await AmplifyPrompter.input({
    message: `Hit [enter] when complete`,
    required: true,
    defaultValue: ''
});
```
I figured having an explicit API is probably better to not touch the existing uses of `input`

sobolk (approved):

packages/cli-core/API.md
sobolk: I did see this pattern in bedrock SDK. might be worth trying.

```
{
    message: string;
    required?: never;
    defaultValue?: string;
  } | {
    message: string;
    required: true;
    defaultValue?: never;
  }
```

Amplifiyer: ohh that's neat. Updated.

sobolk (approved):

ShadowCat567 (approved):

PR #2301, opened 2024-12-06
validate branch and app id inputs for pipeline-deploy command

rtpascual:
## Problem

This PR addresses 2 problems:

1. There is no validation to confirm `--branch` and `--app-id` have inputs for the `pipeline-deploy` command. So something like this can be passed `npx ampx pipeline-deploy --branch --app-id` which results in the following error when we start tagging the root stack:
```
ValidationError: 2 validation errors detected: Value '' at 'tags.1.member.value' failed to satisfy constraint: Member must have length greater than or equal to 1; Value '' at 'tags.2.member.value' failed to satisfy constraint: Member must have length greater than or equal to 1
```

2. For `generate` commands, if `--branch`, `--app-id`, or `--stack` have no inputs (eg. `generate outputs --stack`) we incorrectly fallback to sandbox resolver.

## Changes

- Validate that `--branch` and `--app-id` options have an input with a length of at least 1 for pipeline-deploy command.
- Make fallback logic more strict for when we fallback to sandbox resolver for backend id

## Validation

Unit tests.

## Checklist

### DIFF
--- a/packages/cli/src/commands/pipeline-deploy/pipeline_deploy_command.ts
+++ b/packages/cli/src/commands/pipeline-deploy/pipeline_deploy_command.ts
@@ -122,9 +122,9 @@ export class PipelineDeployCommand
       })
       .check(async (argv) => {
         if (argv['branch'].length === 0 || argv['app-id'].length === 0) {
-          throw new AmplifyUserError('MissingCommandInputError', {
-            message: 'Missing --branch and/or --app-id',
-            resolution: 'Branch/app id must be at least 1 character',
+          throw new AmplifyUserError('InvalidCommandInputError', {
+            message: 'Invalid --branch or --app-id',
+            resolution: '--branch and --app-id must be at least 1 character',
           });
         }
       });
--- a/packages/cli/src/commands/pipeline-deploy/pipeline_deploy_command.ts
+++ b/packages/cli/src/commands/pipeline-deploy/pipeline_deploy_command.ts
@@ -119,6 +119,14 @@ export class PipelineDeployCommand
         type: 'string',
         array: false,
         choices: Object.values(ClientConfigFormat),
+      })
+      .check(async (argv) => {
+        if (argv['branch'].length === 0 || argv['app-id'].length === 0) {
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/pipeline-deploy/pipeline_deploy_command.ts
```diff
@@ -119,6 +119,14 @@ export class PipelineDeployCommand
         type: 'string',
         array: false,
         choices: Object.values(ClientConfigFormat),
+      })
+      .check(async (argv) => {
+        if (argv['branch'].length === 0 || argv['app-id'].length === 0) {
+          throw new AmplifyUserError('MissingCommandInputError', {
+            message: 'Missing --branch and/or --app-id',
+            resolution: 'Branch/app id must be at least 1 character',
```
sobolk, line 127:
```suggestion
            message: 'Invalid --branch or --app-id',
            resolution: '--branch and --app-id must be at least 1 character',
```

Amplifiyer:
Are all the error instances from `pipeline` command? Most `generate` commands also accept `branch` and `app-id` input.

rtpascual:
> Are all the error instances from `pipeline` command? Most `generate` commands also accept `branch` and `app-id` input.

Yes from what I've seen in telemetry, I'm suspecting customers may be using env vars or something similar to populate these inputs but they are not set when running the command. Interestingly enough, `generate output` doesn't fail when passing empty `branch` and `app-id` inputs and still generate correct when I tested just now, I'll have to look more into this.

rtpascual:
Ah I see, it falls back to sandbox resolver which may not match customer's intentions, I'll add the same validation to `generate` commands as well.

Amplifiyer:
> Ah I see, it falls back to sandbox resolver which may not match customer's intentions, I'll add the same validation to `generate` commands as well.

We should prevent the fallback. if user provided `appId` and `branch`, falling back to `sandbox` backend identifier is a bug.

sobolk (approved):

Amplifiyer (approved):

PR #2313, opened 2024-12-09
Add error mapping for Amplify app not found in specified region

vigy02:
## Problem
This error specifically occurs when using the following generate commands with only the --branch option:

1. `npx ampx generate graphql-client-code --branch <branch-name>`
2. `npx ampx generate forms --branch <branch-name>`
3. `npx ampx generate outputs --branch <branch-name>`

When users run these commands with only the --branch parameter and the Amplify app doesn't exist in the current AWS region, they will receive the error message:
"No Amplify app found with name '{appName}' in region '{region}'"

To properly use these commands, users should either:
- Include both --app-id and --branch parameters
- Ensure they are working in the correct AWS region where their Amplify app exists

For example, the correct usage would be:
```bash
npx ampx generate graphql-client-code --branch <branch-name> --app-id <app-id>
```

## Changes

- Added a new error mapping for the "Amplify app not found" scenario
- Updated backend identifier resolver

## Validation & Reproduction Steps

To reproduce the error scenario:
1. Create an Amplify app in region A (e.g., us-east-1)
2. Switch to a different region B (e.g., us-west-2)
3. Run the generate command with just the --branch option in the CLI
4. The system will now show a clear error message indicating the app doesn't exist in region B

Screenshot 2024-12-09 at 5 19 51 PM

Screenshot 2024-12-10 at 3 05 50 PM
Screenshot 2024-12-11 at 3 42 44 PM

### DIFF
--- a/packages/cli/src/commands/generate/graphql-client-code/generate_graphql_client_code_command.ts
+++ b/packages/cli/src/commands/generate/graphql-client-code/generate_graphql_client_code_command.ts
@@ -111,7 +111,7 @@ export class GenerateGraphqlClientCodeCommand
       await result.writeToDirectory(out);
     } catch (error) {
       if (
-        error instanceof BackendOutputClientError &&
+        BackendOutputClientError.isBackendOutputClientError(error) &&
         error.code === BackendOutputClientErrorType.NO_APP_FOUND_ERROR
       ) {
         throw new AmplifyUserError(
--- a/packages/deployed-backend-client/src/stack-name-resolvers/app_name_and_branch_main_stack_name_resolver.ts
+++ b/packages/deployed-backend-client/src/stack-name-resolvers/app_name_and_branch_main_stack_name_resolver.ts
@@ -1,10 +1,6 @@
 import { MainStackNameResolver } from '@aws-amplify/plugin-types';
 import { AmplifyClient, ListAppsCommand } from '@aws-sdk/client-amplify';
 import { BackendIdentifierConversions } from '@aws-amplify/platform-core';
-import {
-  BackendOutputClientError,
-  BackendOutputClientErrorType,
-} from '../backend_output_client_factory.js';
 
 /**
  * Tuple of Amplify App name and branch
@@ -42,8 +38,7 @@ export class AppNameAndBranchMainStackNameResolver
     );
     const region = await this.amplifyClient.config.region();
     if (appMatches.length === 0) {
-      throw new BackendOutputClientError(
-        BackendOutputClientErrorType.NO_APP_FOUND_ERROR,
+      throw new Error(
         `No apps found with name ${this.appNameAndBranch.appName} in region ${region}`
       );
     } else if (appMatches.length >= 2) {
--- a/.changeset/ninety-coins-jog.md
+++ b/.changeset/ninety-coins-jog.md
@@ -1,5 +1,7 @@
 ---
+'@aws-amplify/deployed-backend-client': patch
+'@aws-amplify/client-config': patch
 '@aws-amplify/backend-cli': patch
 ---
 
-Added error mapping for app name not available in the region
+Add error mapping for app name not available in the region
--- a/packages/cli/src/commands/generate/outputs/generate_outputs_command.ts
+++ b/packages/cli/src/commands/generate/outputs/generate_outputs_command.ts
@@ -81,7 +81,7 @@ export class GenerateOutputsCommand
       );
     } catch (error) {
       if (
-        error instanceof BackendOutputClientError &&
+        BackendOutputClientError.isBackendOutputClientError(error) &&
         error.code === BackendOutputClientErrorType.NO_APP_FOUND_ERROR
       ) {
         throw new AmplifyUserError(
--- a/packages/client-config/src/unified_client_config_generator.ts
+++ b/packages/client-config/src/unified_client_config_generator.ts
@@ -103,6 +103,15 @@ export class UnifiedClientConfigGenerator implements ClientConfigGenerator {
               },
               error
             );
+          case BackendOutputClientErrorType.NO_APP_FOUND_ERROR:
+            throw new AmplifyUserError(
+              'AmplifyAppNotFoundError',
+              {
+                message: error.message,
+                resolution: `Ensure that an Amplify app exists in the region.`,
+              },
+              error
+            );
         }
       }
       throw error;


packages/backend-deployer/src/cdk_error_mapper.ts
```diff
@@ -497,6 +497,20 @@ export class CdkErrorMapper {
...
+    {
+      errorRegex:
+        /No apps found with name (?<appName>.*) in region (?<region>.*)/,
+      humanReadableErrorMessage:
+        'No Amplify app found with name "{appName}" in region "{region}".',
+      resolutionMessage:
+        'Ensure that an Amplify app named "{appName}" exists in the "{region}" region.',
+      errorName: 'AmplifyAppNotFoundError',
+      classification: 'ERROR',
+    },
```
Amplifiyer, line 513: Are you able to reproduce this error?

Amplifiyer: fyi, the error originates from https://github.com/aws-amplify/amplify-backend/blob/383884985f0da2e849ca71769ee6ba59041d33aa/packages/deployed-backend-client/src/stack-name-resolvers/app_name_and_branch_main_stack_name_resolver.ts#L42 and is not emitted from the CDK execution

packages/cli/src/commands/generate/graphql-client-code/generate_graphql_client_code_command.ts
```diff
@@ -89,20 +90,36 @@ export class GenerateGraphqlClientCodeCommand
...
+      );
 
-    await result.writeToDirectory(out);
+      if (appNotFoundMatch?.groups) {
+        const { appName, region } = appNotFoundMatch.groups;
+        throw new AmplifyUserError('AmplifyAppNotFoundError', {
+          message: `No Amplify app found with name "${appName}" in region "${region}".`,
+          resolution: `Ensure that an Amplify app named "${appName}" exists in the "${region}" region.`,
+        });
+      }
```
Amplifiyer, line 119: My question is still not answered. How are you able to reproduce this and validate it? I don't see `appName` as a user input anywhere so not sure how this can be easily categorized as a User Error. 
What situation(s) can lead to this error?

vigy02: I have moved the try-catch wrapper to a common resolver where all the generate command finally lands.I have updated the repro steps in the description. Seems to be happening only if the --branch option is set without the AppID

Amplifiyer: This is mostly happening in the console (based on the isCI flag in the telemetry) where customers might not be explicitly providing these values and the region should match if customers are using the Amplify Console to download the output file.

>Seems to be happening only if the --branch option is set without the AppID

Yes, when only `--branch` is set, the appName is loaded from the Amplify Service and filtered on the appName which is coming from the packageJson#Name (I think).

packages/cli/src/backend-identifier/backend_identifier_resolver.ts
```diff
@@ -30,21 +33,39 @@ export class AppBackendIdentifierResolver implements BackendIdentifierResolver {
...
+          branchName: args.branch,
+        };
+      }
+      return undefined;
+    } catch (error) {
+      const appNotFoundMatch = (error as Error).message.match(
+        /No apps found with name (?<appName>.*) in region (?<region>.*)/
+      );
+
+      if (appNotFoundMatch?.groups) {
```
Amplifiyer, line 58: which line here is throwing the exception that you are catching? Afaik `this.namespaceResolver.resolve();` doesn't throw this exception, but this does https://github.com/aws-amplify/amplify-backend/blob/383884985f0da2e849ca71769ee6ba59041d33aa/packages/deployed-backend-client/src/stack-name-resolvers/app_name_and_branch_main_stack_name_resolver.ts#L42 as I mentioned in my previous comment https://github.com/aws-amplify/amplify-backend/pull/2313#discussion_r1876837745

packages/cli/src/commands/generate/graphql-client-code/generate_graphql_client_code_command.ts
vigy02: > This is mostly happening in the console (based on the isCI flag in the telemetry) where customers might not be explicitly providing these values and the region should match if customers are using the Amplify Console to download the output file.

| Input Command | CI Percentage | Non-CI Percentage |
|--------------|--------------|-----------------|
| UnknownCommand | 31.03 | 68.97 |
| generate graphql-client-code | 0 | 100 |
| generate outputs | 56.33 | 43.67 |

packages/cli/src/backend-identifier/backend_identifier_resolver.ts
vigy02: The error originates in `app_name_and_branch_main_stack_name_resolver.ts`. Since we are unable to place AmplifyUserError in the deployed-backend-client, I've moved the error handling to 
the CLI layer where the error bubbles up. I've successfully implemented error catching at this level. You can see in the screenshot I've added to the description.

Screenshot 2024-12-10 at 2 39 21 PM (2)

packages/cli/src/commands/generate/graphql-client-code/generate_graphql_client_code_command.ts
vigy02: Its a valid concern, since appName is internally derived. When I tested out in the right region it works every time. When I'm working in the wrong region it fails. So we can safely call it as UserError.
This condition is triggered only when --branch is used with wrong region

Amplifiyer: Have you tested this in the Console? I understand you can reproduce this in the CLI, but the fact that this issue happens in CI points to the issue could be something else.

vigy02: Yes I was able to repro this issue on the console side. I edited the `amplify.yml` and added wrong build command which caused this issue. The changes which I have implemented in the cli package will be able to handle this gracefully.
Screenshot 2024-12-11 at 3 42 44 PM

Amplifiyer (commented):
this is looking good. Need to add tests as well.

packages/cli/src/commands/generate/graphql-client-code/generate_graphql_client_code_command.ts
```diff
@@ -89,20 +90,37 @@ export class GenerateGraphqlClientCodeCommand
...
+
+      const result = await this.generateApiCodeAdapter.invokeGenerateApiCode({
+        ...backendIdentifier,
+        ...formatParams,
+      } as unknown as InvokeGenerateApiCodeProps);
+
+      await result.writeToDirectory(out);
+    } catch (error) {
+      const appNotFoundMatch = (error as Error).message.match(
+        /No apps found with name (?<appName>.*) in region (?<region>.*)/
```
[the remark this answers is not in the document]
vigy02: Updated

packages/cli/src/commands/generate/outputs/generate_outputs_command.test.ts
```diff
@@ -249,3 +254,101 @@ void describe('generate outputs command', () => {
...
+  let clientConfigGenerator: ClientConfigGeneratorAdapter;
+  let backendIdentifierResolver: AppBackendIdentifierResolver;
+  let generateOutputsCommand: GenerateOutputsCommand;
+
+  beforeEach(() => {
+    // Mock the dependencies
+    clientConfigGenerator = {
+      generateClientConfigToFile: mock.fn(),
+    } as unknown as ClientConfigGeneratorAdapter;
+
```
Amplifiyer, line 268: Why not use the existing `describe` that is already doing all the mock setup

packages/cli/src/commands/generate/outputs/generate_outputs_command.test.ts
```diff
@@ -249,3 +254,101 @@ void describe('generate outputs command', () => {
...
+        branch: 'main',
+        format: undefined,
+        outDir: undefined,
+        'out-dir': undefined,
+        outputsVersion: '1.3',
+        'outputs-version': '1.3',
+        _: [],
+        $0: 'command-name',
+      });
+      assert.fail('Expected error was not thrown');
```
Amplifiyer, line 308: use `commandRunner.runCommand` as used in other tests.

packages/cli/src/commands/generate/graphql-client-code/generate_graphql_client_code_command.ts
```diff
@@ -89,20 +94,38 @@ export class GenerateGraphqlClientCodeCommand
...
+      const result = await this.generateApiCodeAdapter.invokeGenerateApiCode({
+        ...backendIdentifier,
+        ...formatParams,
+      } as unknown as InvokeGenerateApiCodeProps);
 
-    await result.writeToDirectory(out);
+      await result.writeToDirectory(out);
+    } catch (error) {
+      if (
+        error instanceof BackendOutputClientError &&
```
Amplifiyer, line 114: As per the lint, don't use instanceof as it might not work if there are multiple versions of DeployedBackendClient package in the node_modules.

packages/deployed-backend-client/src/stack-name-resolvers/app_name_and_branch_main_stack_name_resolver.ts
```diff
@@ -1,6 +1,10 @@
 import { MainStackNameResolver } from '@aws-amplify/plugin-types';
 import { AmplifyClient, ListAppsCommand } from '@aws-sdk/client-amplify';
 import { BackendIdentifierConversions } from '@aws-amplify/platform-core';
+import {
+  BackendOutputClientError,
+  BackendOutputClientErrorType,
+} from '../backend_output_client_factory.js';
```
Amplifiyer, line 7: `BackendOutputClientError` is probably not the right error type to use here since based on it's name, it should be used with `BackendOutputClient` only in the `DeployedBackendClient`.

packages/cli/src/commands/generate/outputs/generate_outputs_command.test.ts
vigy02: Updated it, keeping it consistent
I was planning to create Test suites separate for handling all negative test cases

packages/cli/src/commands/generate/outputs/generate_outputs_command.test.ts
vigy02: updated it

PR #2315, opened 2024-12-10
do not stream function logs if stack does not exist

rtpascual:
## Problem

When running `ampx sandbox --stream-function-logs`, if there are errors during deployment that are due to allowed destructive changes then we continue and start streaming function logs.

Streaming function logs will not work in the case when stack does not exist because it failed to deploy.

## Changes

- Only start streaming logs if stack exists
- Since this introduces another package which checks instanceof `BackendOutputClientError`, update detection of `BackendOutputClientError` similar to https://github.com/aws-amplify/amplify-backend/pull/2200

## Validation

Unit tests.

## Checklist

### DIFF
--- a/packages/sandbox/src/file_watching_sandbox.ts
+++ b/packages/sandbox/src/file_watching_sandbox.ts
@@ -173,7 +173,7 @@ export class FileWatchingSandbox extends EventEmitter implements Sandbox {
       // Stop streaming the logs so that deployment logs don't get mixed up
       this.functionsLogStreamer.stopStreamingLogs();
 
-      await this.deploy(options);
+      let deployResult = await this.deploy(options);
 
       // If latch is still 'deploying' after the 'await', that's fine,
       // but if it's 'queued', that means we need to deploy again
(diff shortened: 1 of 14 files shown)


sobolk (dismissed):

packages/sandbox/src/file_watching_sandbox.ts
```diff
@@ -184,16 +184,19 @@ export class FileWatchingSandbox extends EventEmitter implements Sandbox {
...
-        await this.backendIdSandboxResolver(options.identifier),
-        options.functionStreamingOptions
-      );
+      // if deployment was successful, start streaming function logs
+      if (deployResult) {
+        await this.functionsLogStreamer.startStreamingLogs(
+          await this.backendIdSandboxResolver(options.identifier),
+          options.functionStreamingOptions
+        );
+      }
```
Amplifiyer, line 199: This will prevent the log streaming to work after every failed deployment which we might not want to do. After a successful deployment, it should be always ok to let the log streamer working as many of the failures are simply syntax errors where customers are making live updates to their app.

Amplifiyer (approved):

sobolk (approved):

PR #2319, opened 2024-12-11
fix: add cdk error mapping for error 'no valid tables found'

awsluja:
## Problem
At this time, 100% of errors with 'no valid tables found' error message are coming from the 'generate schema-from-database' command. This reclassifies that error as it is a user error, and resolution is to make sure a primary key is defined on at least one table in the database that is being imported/generated from.

see: https://docs.amplify.aws/javascript/build-a-backend/data/connect-to-existing-data-sources/connect-postgres-mysql-database/#step-2---generate-typescript-representation-of-your-database-schema

## Changes

## Validation
Validated that this is not reproducible with standard deployment commands, and found that all errors were coming from the generate schema-from-database command, which requires primary keys on at least one table.

## Checklist


vigy02 (approved):

packages/backend-deployer/src/cdk_error_mapper.ts
```diff
@@ -486,6 +486,16 @@ export class CdkErrorMapper {
...
+    {
+      errorRegex:
+        /No valid tables found\. Make sure at least one table has a primary key\./,
+      humanReadableErrorMessage:
+        'No valid tables found. Make sure at least one table has a primary key.',
+      resolutionMessage:
+        "If you are using the 'generate schema-from-database' command, make sure the database has at least one table with a primary key.",
+      errorName: 'GenerateSchemaFromDatabaseError',
+      classification: 'ERROR',
+    },
```
sobolk, line 498: is `backend-deployer` involved in `generate schema-from-database` command ?

packages/backend-deployer/src/cdk_error_mapper.test.ts
```diff
@@ -641,6 +641,13 @@ npm error enoent`,
     errorName: 'CloudFormationDeletionError',
     expectedDownstreamErrorMessage: undefined,
   },
+  {
+    errorMessage: `No valid tables found. Make sure at least one table has a primary key.`,
```
ShadowCat567, line 645: I have 2 questions:
1. took a brief look into this error, it comes from the data team's repo, from here: https://github.com/aws-amplify/amplify-category-api/blob/e4d16911236c5751e02609c8c3b069543535883a/packages/amplify-graphql-schema-generator/src/ts-schema-generator/helpers.ts#L176 
I think we have a few other non-cdk errors that are getting mapped in this file as well, just want to confirm that this is ok.
2. You are committing some changes to package.lock. Is that intended?

awsluja: looking into this more (thanks to Kamil's comment), it does seem that we don't have tooling in place for classifying anything other than what is coming from backend-deployer

also yes it seems package.lock is being updated on clean build but i can try again

packages/backend-deployer/src/cdk_error_mapper.ts
sobolk: https://github.com/aws-amplify/amplify-backend/blob/560878f3af595d4a661f396545fe970fac39504d/packages/schema-generator/src/generate_schema.ts#L45

sobolk: But we should ask data team to throw typed error for this instead of attempting to match strings.

rtpascual: In cases like this I think we would want to catch and throw user error in `SchemaGenerator` or wherever makes sense. Unless we want to introduce error mapping for errors that can happen for CLI commands.

sobolk: > In cases like this I think we would want to catch and throw user error in SchemaGenerator

Correct. But it would be best if data team gave us better error to map from.

awsluja:
closing for now, we need an alternative way to map these kind of errors

Issue #2320, opened 2024-12-12
getAmplifyDataClientConfig() fails to check Amplify data variables when API name is defined


lengieng:
### Environment information

```plain text
System:
  OS: Windows 11 10.0.26100
  CPU: (12) x64 13th Gen Intel(R) Core(TM) i5-1335U
  Memory: 2.08 GB / 15.72 GB
Binaries:
  Node: 22.2.0 - C:\Program Files\nodejs\node.EXE
  Yarn: 1.22.22 - ~\AppData\Roaming\npm\yarn.CMD
  npm: 10.9.1 - C:\Program Files\nodejs\npm.CMD
  pnpm: undefined - undefined
NPM Packages:
  @aws-amplify/auth-construct: 1.5.1
  @aws-amplify/backend: 1.9.0
  @aws-amplify/backend-auth: 1.4.2
  @aws-amplify/backend-cli: 1.4.3
  @aws-amplify/backend-data: 1.2.2
  @aws-amplify/backend-deployer: 1.1.11
  @aws-amplify/backend-function: 1.9.0
  @aws-amplify/backend-output-schemas: 1.4.0
  @aws-amplify/backend-output-storage: 1.1.4
  @aws-amplify/backend-secret: 1.1.5
  @aws-amplify/backend-storage: 1.2.4
  @aws-amplify/cli-core: 1.2.1
  @aws-amplify/client-config: 1.5.3
  @aws-amplify/deployed-backend-client: 1.4.2
  @aws-amplify/form-generator: 1.0.3
  @aws-amplify/model-generator: 1.0.9
  @aws-amplify/platform-core: 1.3.0
  @aws-amplify/plugin-types: 1.6.0
  @aws-amplify/sandbox: 1.2.7
  @aws-amplify/schema-generator: 1.2.6
  aws-amplify: 6.10.2
  aws-cdk: 2.172.0
  aws-cdk-lib: 2.172.0
  typescript: 5.6.2
No AWS environment variables
No CDK environment variables
```

### Describe the bug

The doc shows an example of how to use `defineAuth` and `defineFunction` to create [Cognito post confirmation Lambda trigger](https://docs.amplify.aws/react/build-a-backend/functions/examples/create-user-profile-record/). Another related section is [grant Lambda function access to API](https://docs.amplify.aws/react/build-a-backend/data/customize-authz/grant-lambda-function-access-to-api/#access-the-api-using-aws-amplify).

[...]

PR #2324, opened 2024-12-12
fix: Lambda client env var name issue

stocaaro:
## Problem
When using the `getAmplifyDataClientConfig` function with a named `dataBackend`, the SSM variable names don't line up.

Example:
```
  name: 'CoolName',
  schema,
  ...
})
```
with 
```
// amplify/auth/post-confirmation/handler.ts
...
import { getAmplifyDataClientConfig } from '@aws-amplify/backend/function/runtime';
...
```

Results in a typescript error caused by mismatched env var names.

https://github.com/aws-amplify/amplify-backend/issues/2320

## Changes
- Use `AmplifyData` as a prefix (which is included even if a user defined name is provided
- Always pass the name and prefix through the `AMPLIFY_DATA_DEFAULT_NAME` SSM variable
- `getAmplifyDataClientConfig` now uses the default name variable to use the correct env variables
- Move `toScreamingSnakeCase` into the export surface of `'@aws-amplify/platform-core'` as SSM variables are normalized by this method and this normalization needs to be applied to consume these variables in `getAmplifyDataClientConfig`

## Validation
- Modified integ test to use `defineData({name: '', ...})`
- Add unit tests to cover the named case
- Manual validation

## Checklist


.changeset/warm-sloths-tickle.md
```diff
@@ -0,0 +1,7 @@
+---
+'@aws-amplify/backend-function': patch
+'@aws-amplify/backend-data': patch
+'@aws-amplify/platform-core': patch
```
sobolk, line 4:
```suggestion
'@aws-amplify/backend-function': minor
'@aws-amplify/backend-data': patch
'@aws-amplify/platform-core': minor
```

.changeset/warm-sloths-tickle.md
```diff
@@ -0,0 +1,7 @@
...
+---
+
+Lambda client env var name issue
```
sobolk, line 7: This is going to changelogs, can this be better?

stocaaro: You're right. This can be improved. Made a revision.

packages/backend-data/src/factory.ts
```diff
@@ -307,14 +308,20 @@ class DataGenerator implements ConstructContainerEntryGenerator {
...
           amplifyApi.resources.cfnResources.cfnGraphqlApi.attrGraphQlUrl,
-        [`${this.name}_MODEL_INTROSPECTION_SCHEMA_BUCKET_NAME`]:
+        [`${namePrefix}${this.name}_MODEL_INTROSPECTION_SCHEMA_BUCKET_NAME`]:
           modelIntrospectionSchemaBucket.bucketName,
-        [`${this.name}_MODEL_INTROSPECTION_SCHEMA_KEY`]:
+        [`${namePrefix}${this.name}_MODEL_INTROSPECTION_SCHEMA_KEY`]:
           modelIntrospectionSchemaKey,
+        ['AMPLIFY_DATA_DEFAULT_NAME']: `${namePrefix}${this.name}`,
+        // @deprecated
+        [`${this.name}_GRAPHQL_ENDPOINT`]:
```
Amplifiyer, line 323: will this not create a duplicate when `namePrefix` is empty?

packages/backend-function/src/runtime/get_amplify_clients_configuration.ts
```diff
@@ -125,6 +128,38 @@ export type DataClientReturn<T> = T extends DataClientEnv
...
+    !(
+      bucketName in env &&
+      keyName in env &&
+      endpointName in env &&
+      typeof env[bucketName] === 'string' &&
+      typeof env[keyName] === 'string' &&
+      typeof env[endpointName] === 'string'
+    )
+  ) {
+    throw new Error('The data environment variables are malformed');
```
Amplifiyer, line 148: Can you also add serialized `env` and `dataName` in the error message.

packages/backend-data/src/factory.ts
stocaaro: I was worried about the same thing. Up on [line 311](https://github.com/aws-amplify/amplify-backend/pull/2324/files#diff-ec4062ca129b817f0d020474942b0b354abaebe0d9fe5fa88a370c89797369d8R311), name prefix is left blank if the name is defaulted.

Amplifiyer: If it's not expected, we should avoid adding it. Also wanted to ask what's special about `GRAPHQL_ENDPOINT` that this behavior is not replicated for the other `_MODEL_INTROSPECTION_SCHEMA_BUCKET_NAME` and `_MODEL_INTROSPECTION_SCHEMA_KEY`

stocaaro: `GRAPHQL_ENDPOINT` predates the MIS fields and customers may be depending on it directly. Since the other two are newer, @sobolk advised that they can be modified without concern for breaking change.

Also `GRAPHQL_ENDPOINT` is included in the prefixing, however, we are leaving an unprefixed version in place with a @deprecation flag to enable removal with the next major version change.

packages/backend-function/src/runtime/get_amplify_clients_configuration.ts
stocaaro: Added

Amplifiyer (approved):

packages/backend-data/src/factory.ts
```diff
@@ -307,14 +308,32 @@ class DataGenerator implements ConstructContainerEntryGenerator {
 
     convertJsResolverDefinition(scope, amplifyApi, schemasJsFunctions);
 
+    const namePrefix = this.name === defaultName ? '' : defaultName;
+
+    const ssmEnvironmentScopeContext = {
```
sobolk, line 313:
```suggestion
    const ssmEnvironmentEntries = {
```

perhaps. this is a nit, feel free to ignore.

packages/backend-data/src/factory.ts
```diff
@@ -307,14 +308,32 @@ class DataGenerator implements ConstructContainerEntryGenerator {
...
+      [`${namePrefix}${this.name}_GRAPHQL_ENDPOINT`]:
+        amplifyApi.resources.cfnResources.cfnGraphqlApi.attrGraphQlUrl,
+      [`${namePrefix}${this.name}_MODEL_INTROSPECTION_SCHEMA_BUCKET_NAME`]:
+        modelIntrospectionSchemaBucket.bucketName,
+      [`${namePrefix}${this.name}_MODEL_INTROSPECTION_SCHEMA_KEY`]:
+        modelIntrospectionSchemaKey,
+      ['AMPLIFY_DATA_DEFAULT_NAME']: `${namePrefix}${this.name}`,
+    };
+
+    const backwardsCompatibleScopeContext =
```
sobolk, line 323: update this plz if you take above nit.

sobolk (approved):

packages/backend-data/src/factory.ts
stocaaro: This naming came from the function that receives these entries/vars `ssmEnvironmentEntriesGenerator.generateSsmEnvironmentEntries(scopeContext)`. I'm going to proceed with this naming and can revisit it in another change if we decide its clearer with different naming.

PR #2333, opened 2024-12-14
Test latest cdk deeply.

sobolk:
## Problem

We're not validating that latest CDK release is compatible with our scenarions.

## Changes

This PR:
1. Adds new workflow that invokes `health_checks` with a  version of CDK latest by default .
    1. Run sandbox and deployment tests (package manager and create amplify overlap with existing canaries and are using unreleased packages - not useful).
    2. Narrow down runtimes to Node 18 and Ubuntu. (Our logic is tested in full e2e before release, therefore minimizing run time and flakes here).
    3. Run this workflow daily until we have better way to detect CDK releases.
4. Parametrizes `health_checks`:
    1. Adds options to include/exclude - MacOS, Windows, Package Manager tests, Create Amplify tests.
    3. Adds validation for node version, cdk version (while working on this I discovered that we have a mess, i.e. jobs that claim node 20 ran with 18 or were using wrong cache).
    
    

Health checks inputs when running manually:
image

## Out of scope

It was itching me to extract e2e tests into separate sub-workflow and call it from both canaries and health checks.

This however will need more experiments and time. It is recommended to pursue this separately.

## Validation

This PR checks.

New workflow was validated in fork: https://github.com/sobolk/amplify-backend/actions/workflows/validate_cdk_release.yml 

And builds:
cleanup workflow: https://github.com/aws-amplify/amplify-backend/actions/runs/12333975501
health_checks from dispatch: https://github.com/aws-amplify/amplify-backend/actions/runs/12332932430

health_checks with broken CDK and minimal coverage: https://github.com/aws-amplify/amplify-backend/actions/runs/12355545138 (failure expected)
image

## Checklist


.github/actions/build_with_cache/action.yml
```diff
@@ -2,17 +2,36 @@ name: build_with_cache
 description: builds the source code if cache miss and caches the result
 inputs:
   node-version:
-    default: 18
+    required: true
+  cdk-version:
+    required: true
```
sobolk, line 7: When I was debugging this. I discovered that we have a mess.
I.e. tests claiming that they run on Node 20 were either using wrong node (18) or wrong install/build cache.

Therefore I made these parameters mandatory in couple of places to assert correctness. And I added runtime validation (it turns out that `required` matters only for intellisense really...).

packages/integration-tests/src/process-controller/predicated_action_queue_builder.ts
```diff
@@ -92,6 +93,7 @@ export class PredicatedActionBuilder {
       action: (strWithDeploymentTime: string) => {
         // the time can be in fractional or whole seconds. 24.3, 24, 24.22 etc.
         const regex = /^✨ {2}Total time: (\d*\.*\d*)s.*$/;
+        strWithDeploymentTime = stripANSI(strWithDeploymentTime);
```
sobolk, line 96: Latest CDK changed output a bit (see https://github.com/aws-amplify/amplify-backend/actions/runs/12332987817/job/34421507582)

image

.github/workflows/health_checks.yml
```diff
@@ -75,27 +185,40 @@ jobs:
...
   test_scripts:
     needs:
       - build
+      - resolve_inputs
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # version 4.1.4
       - uses: ./.github/actions/setup_node
+        with:
+          node-version: ${{ fromJSON(needs.resolve_inputs.outputs.node)[0] }}
```
Amplifiyer, line 200: Should we be explicit here instead of just getting the first version?

sobolk: To do this I would need to convert `NodeJS array` input to `Include Node 20?` and have 18 always included by default.

Otherwise this would fail if workflow was ran with `["20"]` exclusively because install and build cache would be not present.

That's another aspect I was a bit unsure what to do so I left it flexible.

wdyt?

Amplifiyer: Got it, can we run the build/install for 18,20 regardless every single time?. I'm worries that [18,20] and [20,18] will behave differently and will be harder to debug.

sobolk: we can do that too.

sobolk: Done.

see https://github.com/aws-amplify/amplify-backend/actions/runs/12361549810

.github/workflows/snapshot_release.yml
```diff
@@ -35,7 +50,12 @@ jobs:
     steps:
       - uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # version 4.1.4
       - uses: ./.github/actions/setup_node
+        with:
+          node-version: 18
       - uses: ./.github/actions/restore_build_cache
+        with:
+          node-version: 18
+          cdk-version: FROM_PACKAGE_LOCK
```
awsluja, line 58: where node-version is hard coded, we should extract to single place if possible with relevant name, maybe 'current supported node version'

sobolk: Can you see any alternative?

Amplifiyer (approved):

.github/workflows/snapshot_release.yml
Amplifiyer: If this introduces more complexity (more than what's already here), I don't see the value add for doing this, i.e. trying to extract default node version in a common place. Given that we would change this once every couple years?

awsluja: I was thinking environment variable set in workflow, but thats fine if not possible

awsluja: this was my thought if we were to set at the project level https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#defining-configuration-variables-for-multiple-workflows

awsluja (approved):

.github/workflows/snapshot_release.yml
sobolk: Yaml directives are processed before job is ran. Env vars are not available at this stage - they can be used inside jobs but not in their declarations.

sobolk: > https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#defining-configuration-variables-for-multiple-workflows

the `context vars` (not env vars) could work. but that's another trade-off - they would be outside of yml files in repo settings - not discoverable.

PR #2338, opened 2024-12-16
Add additional info to `The specified bucket does not exist`

ShadowCat567:
## Problem

Errors with the name `NoSuchBucket: The specified bucket does not exist` are classified as Unknown faults and have very vague messaging which does not tell the user what bucket is missing or how to restore that bucket. 

## Changes

`NoSuchBucket` errors (from `model-generator`) are now caught and classified as `AmpliftFaults` with clearer messaging about how to resolve them. 

## Validation

`NoSuchBucket` appears for 3 different commands, `generate outputs`, `generate graphql-client-code`, and `pipeline-deploy`. Was able to get a reproduction and confirmed that the new messaging appears for `generate outputs` and `pipeline-deploy`.
## Checklist


packages/model-generator/src/s3_string_object_fetcher.ts
```diff
@@ -14,15 +15,27 @@ export class S3StringObjectFetcher {
...
+      }
+      return schema;
+    } catch (caught) {
+      if (caught instanceof NoSuchBucket) {
+        throw new AmplifyUserError('BucketDoesNotExistError', {
+          message: `${caught.name} the bucket, ${bucket} does not exist.`,
+          resolution: `Ensure that ${bucket} exists and contains its expected contents before trying this command again.\n
+            To find the expected contents, create a new sandbox and investigate the contents of buckets with a similar name to this one.`,
+        });
+      } else {
```
Amplifiyer, line 35: The `bucketName` or the `s3Uri` is not a user input but managed by amplify. It's saved in the stack outputs during the CFN deployment and then retrieved here https://github.com/aws-amplify/amplify-backend/blob/1400e6f65cf8e156b8aaeddc4cb37311a37a3fc3/packages/model-generator/src/create_graphql_models_generator.ts#L122 before passing to this method.

Error here would typically mean
1. Bug in the Amplify Backend
2. Some validations missing internally allowing a bad s3 URI to be passed here
3. Customer manually deleted this bucket.

the third is indeed a user error but how can we ascertain that the first two aren't an issue?

Amplifiyer: That's the challenge, it's super hard with the current information available. My suggestion is to keep this as a `fault` for now and add more data in the error message that would help us and customers figure out what's the cause of this.

ShadowCat567: I was able to replicate this by manually deleting buckets, but there could very well be a bug with these commands. Especially because the buckets you have to delete to trigger this error all have a `model-schema.graphql` file in them so users might be less likely to get rid of them because of the 2-step process of emptying then deleting the bucket.

packages/model-generator/src/s3_string_object_fetcher.ts
```diff
@@ -14,15 +15,28 @@ export class S3StringObjectFetcher {
...
+      const schema = await getSchemaCommandResult.Body?.transformToString();
+      if (!schema) {
+        // eslint-disable-next-line amplify-backend-rules/prefer-amplify-errors
+        throw new Error('Error on parsing output schema');
+      }
+      return schema;
+    } catch (caught) {
+      if (caught instanceof NoSuchBucket) {
+        //eslint-disable-next-line amplify-backend-rules/prefer-amplify-errors
+        throw new AmplifyFault('NoSuchBucketFault', {
```
ShadowCat567, line 31: should I throw this as an Error or as an AmplifyFault?

Amplifiyer: `AmplifyFault` is correct here.

packages/model-generator/src/s3_string_object_fetcher.ts
```diff
@@ -14,15 +15,28 @@ export class S3StringObjectFetcher {
...
+          message: `${caught.name} the bucket, ${bucket} does not exist. \n
+            Ensure that ${bucket} exists and contains its expected contents before trying this command again.\n
+            To find the expected contents, create a new sandbox and investigate the contents of buckets with a similar name to this one.`,
+        });
```
Amplifiyer, line 35: add the `caught` as a cause to `AmplifyFault` as well.

packages/model-generator/src/s3_string_object_fetcher.ts
```diff
@@ -14,15 +15,28 @@ export class S3StringObjectFetcher {
...
-    if (!schema) {
-      // eslint-disable-next-line amplify-backend-rules/prefer-amplify-errors
-      throw new Error('Error on parsing output schema');
+    try {
+      const getSchemaCommandResult = await this.s3Client.send(
+        new GetObjectCommand({ Bucket: bucket, Key: key })
+      );
+      const schema = await getSchemaCommandResult.Body?.transformToString();
+      if (!schema) {
+        // eslint-disable-next-line amplify-backend-rules/prefer-amplify-errors
```
Amplifiyer, line 24: If the lint rule specifically asks for `AmplifyUserError` then it's an issue. Any `Amplify` error should be fine. Not for this PR, but if that's the case we should create a backlog tech debt item to fix it.

ShadowCat567: Lint might be fine with `AmplfiyFault` (will check), this is a relic of making a version with `Error`

Amplifiyer (approved):

packages/model-generator/src/s3_string_object_fetcher.ts
```diff
@@ -14,15 +15,31 @@ export class S3StringObjectFetcher {
...
+        throw new Error('Error on parsing output schema');
+      }
+      return schema;
+    } catch (caught) {
+      if (caught instanceof NoSuchBucket) {
+        throw new AmplifyFault(
+          'NoSuchBucketFault',
+          {
+            message: `${caught.name} the bucket, ${bucket} does not exist. \n
```
awsluja, line 33: what does this print out? ${caught.name} specifically

packages/model-generator/src/s3_string_object_fetcher.ts
```diff
@@ -14,15 +15,31 @@ export class S3StringObjectFetcher {
...
+            Ensure that ${bucket} exists and contains its expected contents before trying this command again.\n
+            To find the expected contents, create a new sandbox and investigate the contents of buckets with a similar name to this one.`,
```
awsluja, line 35: Are we suggesting customers recreate a bucket and its contents if the bucket doesn't exist anymore? Do we know why the bucket is going missing?

packages/model-generator/src/s3_string_object_fetcher.ts
ShadowCat567: `${caught.name}` prints out "`NoSuchBucket`"

packages/model-generator/src/s3_string_object_fetcher.ts
ShadowCat567: Yes, we are suggesting the customer recreate the bucket if it no longer exists. We don't know exactly why is goes missing to begin with, could be they accidentally deleted it, could be we have a bug that deletes it by accident in our code, so we can't reclassify this as a UserError. I have been able to get a repro by deleting buckets, but if it is a bug, finding it might require us to have a customer come to us with complaints about this or this needs to be tabled until we have the updated telemetry so we have more info about what is going on in the customer's environment to give us clues.

Amplifiyer: We should hold off on this resolution until we can confirm it's a customer issue. Because if it's not, customers would have no idea how to "recreate" this bucket.

A better resolution would be "Try deploying your changes again and if the error still persists, create a bug report here <url>". This is in a hope that the deployment would "recreate" the bucket and the problem might go away.

ShadowCat567: Then I'll update the message

Amplifiyer (approved):

awsluja (approved):

PR #2341, opened 2024-12-16
import named export from parcel/watcher

josefaidt:
## Problem

follow-up from #2332 

in a limited env like WebContainers we're seeing issues related to the import of the "default" export from the commonjs distribution of parcel/watcher. "default" exports are a concept in ESM, where commonjs will export on `module.exports` or `exports`. Depending on the build process these are sometimes rebound to a named `default` -> `module.exports.default = "something"` or merged onto `module.exports`

## Changes

This change removes the import of parcel/watchers default export in favor of the named `subscribe` export. `subscribe` was already in use for typing, so this change is small

## Validation

validated by re-running the sandbox command in the limited WebContainers env
CleanShot 2024-12-13 at 16 30 07

## Checklist


packages/sandbox/src/file_watching_sandbox.test.ts
```diff
@@ -39,7 +39,8 @@ import {
 
 // Watcher mocks
 const unsubscribeMockFn = mock.fn();
-const subscribeMock = mock.method(watcher, 'subscribe', async () => {
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const subscribeMock = mock.fn(async (_dir, _effect, _options) => {
```
sobolk, line 43:
```suggestion
const subscribeMock = mock.fn(async () => {
```

that should be fine, if not then something like this: https://github.com/aws-amplify/amplify-backend/blob/1400e6f65cf8e156b8aaeddc4cb37311a37a3fc3/packages/client-config/src/client-config-writer/client_config_writer.test.ts#L27

packages/sandbox/src/file_watching_sandbox.test.ts
```diff
@@ -1163,7 +1165,8 @@ const setupAndStartSandbox = async (
     testData.ssmClient,
     testData.functionsLogStreamer,
     printer as unknown as Printer,
-    testData.open ?? _open
+    testData.open ?? _open,
+    testData.subscribe ?? subscribe
```
sobolk, line 1169:
```suggestion
    subscribeMock as never
```

packages/sandbox/src/file_watching_sandbox.test.ts
```diff
@@ -1216,4 +1219,5 @@ type SandboxTestData = {
   ssmClient: SSMClient;
   functionsLogStreamer: LambdaFunctionLogStreamer;
   open?: typeof _open;
+  subscribe?: typeof subscribe;
```
sobolk, line 1222: this is not needed.

sobolk: (I think)

josefaidt: it's not given the previous suggestion, will remove

packages/sandbox/src/file_watching_sandbox.test.ts
josefaidt: this was to appease the checks like this one where we're checking that subscribe was called with a specific argument, https://github.com/aws-amplify/amplify-backend/pull/2341/files#diff-a3a9ef2ca41c1c9b876d1523afccb0fa5478ffd7ef185395e103467e10bd6484R967

I can pull in the parameters type to give the arguments a type. The underscore satisfies tsc but I also had to add the eslint-disable comment to appease no-unused-vars

josefaidt: perhaps
```ts
const subscribeMock = mock.fn(async (..._args: Parameters<typeof subscribe>) => {
  return { unsubscribe: unsubscribeMockFn };
});
```

sobolk: the `mock.fn<>` satisfies both tsc and linter. Let's define parameterless stub AND provides typing for argument assertion.

The `_dir` or `..._args` are not relevant for `mock` . this is for TSC - but that can be handled with `<>`

packages/sandbox/src/file_watching_sandbox.test.ts
```diff
@@ -40,7 +40,12 @@ import { EOL } from 'os';
 
 // Watcher mocks
 const unsubscribeMockFn = mock.fn();
-const subscribeMock = mock.method(watcher, 'subscribe', async () => {
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
sobolk, line 43: is this still needed ?

sobolk (commented):
LGTM

sobolk (approved):

rtpascual (approved):

PR #2347, opened 2024-12-17
Test on Node 22.

sobolk:
## Problem

Node 22 is active stable. Customers are using that. We're not testing it.

## Changes

Move upper bound Node to 22 in test matrix.

Change unit test resolution script due to https://github.com/nodejs/node/issues/50219 . 

Note: This PR makes an assumption that testing on lower and upper boundaries (18 and 22) is sufficient.

## Validation

PR checks.

Canaries: https://github.com/aws-amplify/amplify-backend/actions/runs/12417044898
CDK validation: https://github.com/aws-amplify/amplify-backend/actions/runs/12417254468

## Checklist


.github/workflows/health_checks.yml
```diff
@@ -47,7 +47,7 @@ on:
         description: 'Node versions list (as JSON array).'
         required: false
         type: string
-        default: '["18", "20"]'
+        default: '["18", "22"]'
```
awsluja, line 50: whats support guidance for 20? since we don't pull 18 up to 20

.github/workflows/health_checks.yml
```diff
@@ -512,6 +512,10 @@ jobs:
             node-version: 20
```
awsluja, line 512: should we remove old reference to 20 here?

sobolk: I saw this and left it alone. One can still trigger workflow with "20" included in parameter list, if this is removed the exclusion won't work.

.github/workflows/health_checks.yml
sobolk: I left a note in PR description about this.

The 18 is minimum we support. The 22 then becomes the max we know we support. I think this is enough. The presence of 18 should assert that we can handle "old node".

We definitely don't want to run all e2e tests on 18,20 and 22.

A lightweight solution that we could do is to run install-build-unit-tests on 18, 20 and 22, but I'm still not convinced we need this.

@Amplifiyer @rtpascual wdyt?

Amplifiyer: If most of the customers are on node 20, then I see some value of running at least some sanity tests with it. Though it's true that if the library functions well on `22` and on `18`, every version in between is covered, the real life situations situations such as special logic or considerations for `18` might break that.

My thought is to have some coverage for the majority use case. As soon as the `20` is no longer the majority shareholder, we can drop it.

rtpascual: I agree. While testing on `22` should cover `20`, there may be edge cases we might not catch if we're not explicitly testing on `20` and this goes for whenever future node versions are released. Where we set the line for dropping testing a node version we need to decide at some point.

sobolk: Ok.
Would `install-build-unit-tests` be good enough or do we want to bring some e2e into the mix ?

Amplifiyer: ok, look like majority is node `18` so we definitely don't need extensive coverage for `20`. Just the build and unit tests sounds good to me.

sobolk: Alright, we discussed this offline a bit.
Decided to try just adding Node 22 for now. We'll re-evaluate approach later. We still have room for ~164 more jobs.

rtpascual (approved):

awsluja (approved):

PR #2353, opened 2024-12-18
add dependabot.yml

rtpascual:
## Problem

We can be more proactive in keeping our dependencies up to date so we can avoid problem statements like in https://github.com/aws-amplify/amplify-backend/pull/2303.

## Changes

Adds `dependabot.yml` which will trigger Dependabot to check for version updates of our dependencies.

## Validation

Forked the repo and used `dependabot.yml` to create PRs to update dependencies, see https://github.com/rtpascual/amplify-backend-fork/pulls.

## Checklist


.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
+# Dependabot version updates raises a maximum of five pull requests each time it checks dependencies.
+# Note that there is some overlap with Dependabot security updates so some options can effect security updates as well,
+# see https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file.
+
+version: 2
+updates:
```
sobolk, line 6: I think we should exclude couple of deps from this:
1. CDK - we bump this one when we need new APIs. otherwise we leave range wide. (this might change in the future but this is what we do today).
2. Data packages. (same reason as CDK).

Or perhaps we should consider `increase-if-necessary` strategy from https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy and try that first.

I think we should try the strategy first. This will be least noisy option.

rtpascual: Makes sense. The `increase-if-necessary` strategy would still try to bump CDK and Data packages if there is a new major version so maybe we still want to exclude them?

sobolk: I'd keep them included if we adopt that strategy. We'll get notified about the fact that MV was shipped.

My main concern was to be less aggressive with updates for these (in package.jsons, we have `test_with_baseline_deps` in healthchecks to cover bottom of the range). The `increase-if-necessary` mitigates that concern.

rtpascual: Now my only concern is how `increase-if-necessary` plays with security updates as docs don't clearly state what happens (just that security updates are affected).

For example with https://github.com/aws-amplify/amplify-backend/pull/2253, since these were all patch versions would Dependabot even make the PR?

rtpascual: Looks like it will still work, I reverted the change in #2253 in my fork and added the versioning strategy. https://github.com/rtpascual/amplify-backend-fork/pull/12

sobolk: This https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/customizing-dependency-updates#impact-of-configuration-changes-on-security-updates 
says `These pull requests are always triggered by a security advisory for a dependency, rather than by the Dependabot schedule. However, they inherit relevant configuration settings from the dependabot.yml file unless you specify a different target branch for version updates.`

So we'll keep getting these PRs as soon as security alert is present.

Previous PRs usually changed package-lock. only which would correspond to `lockfile-only` or `auto` going into `lockfile-only` mode. If I understand this correctly the `increase-if-necessary` is `lockfile-only` for minor-patch bumps, package.json + lock file for mv bumps.

.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
...
+  # Maintain dependencies for npm ecosystem
+  - package-ecosystem: 'npm'
+    # Checks all directories from the current layer and below recursively for package.json files
+    directories:
+      - '**/*'
+    schedule:
+      # Runs every Monday
+      interval: 'weekly'
+    # Update if new version is outside of version range specified in package.json
```
sobolk, line 15:
```suggestion
    # Update package.json files if new version is outside of version range specified there. Otherwise lock file only.
```

sobolk (dismissed):

.github/dependabot.yml
rtpascual: Ah I missed that particular section. I was reading this which led me to security update related doc pages where I didn't find anything helpful:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
`In general, security updates use any configuration options that affect pull requests, for example, adding metadata or changing their behavior. For more information about security updates, see [Configuring Dependabot security updates](https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates).`

sobolk (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
...
+version: 2
+updates:
+  # Maintain dependencies for npm ecosystem
+  - package-ecosystem: 'npm'
+    # Checks all directories from the current layer and below recursively for package.json files
+    directories:
+      - '**/*'
+    schedule:
+      # Runs every Monday
+      interval: 'weekly'
```
Amplifiyer, line 14: why weekly and not say daily?

.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
...
+    # Update package.json files if new version is outside of version range specified there. Otherwise lock file only.
+    versioning-strategy: increase-if-necessary
```
Amplifiyer, line 16: If this strategy can update package.json files, how will we accommodate `changesets` in those PRs?

sobolk: That's a good point. Best if this could be automated.

If it can't:
PR won't pass checks until changeset is added (if our check works)
Maintainer adds changeset (this is doable using UI here https://github.com/aws-amplify/amplify-backend/pull/2353#issuecomment-2552462457 +/- quotes style but we could exclude changeset files from prettier.)

.github/dependabot.yml
rtpascual: I was thinking weekly to start so we aren't constantly bombarded with version update PRs, especially since we are using `increase-if-necessary` the PRs will most likely be for new MVs so we need to be thorough on checking for breaking changes.

Or we make sure E2E runs on these PRs and we assume if E2E passes then there are no breaks, in that case I'm fine with changing it to daily.

.github/dependabot.yml
rtpascual: Good point on `changesets`, I'll see if I can automate it.

rtpascual: With change in [d826612](https://github.com/aws-amplify/amplify-backend/pull/2353/commits/d826612363cdea40813b88cb42908c1a957e58e8), I used the new script locally for this PR https://github.com/rtpascual/amplify-backend-fork/pull/20 and this is the resulting changeset file that will be force pushed to the PR:
Screenshot 2024-12-19 at 5 18 28 PM

.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
...
+    labels:
+      - 'dependency'
```
sobolk, line 18: What's the purpose of `dependency` label?

scripts/dependabot_generate_changesets.ts
```diff
@@ -0,0 +1,93 @@
...
+      if (!match) {
+        continue;
+      }
+      // Add dependency name (match[1]) and new version (match[2]) to Map
+      updates.set(match[1].replace(/"/g, ''), match[2].replace(/"/g, ''));
+    }
+  }
+
+  return updates;
+};
```
sobolk, line 44: One other source to consider is parse PR title or description. I'm not sure if this will be any simple - please evaluate and adopt or reject this suggestion.

scripts/dependabot_generate_changesets.ts
```diff
@@ -0,0 +1,93 @@
...
+
+// Get modified package.json files that we care about for changesets
+const packageJsonFiles = changedFiles.filter(
+  (changedFile) =>
+    changedFile.startsWith('packages/') &&
+    !['packages/integration-tests', 'packages/eslint-rules'].some(
+      (packageName) => changedFile.startsWith(packageName)
+    ) &&
+    changedFile.endsWith('package.json')
+);
```
sobolk, line 75: You can use `private` instead of names, see https://github.com/aws-amplify/amplify-backend/blob/b574bce26ec2d55a4391aeceecc1d4957e79fba3/scripts/components/api-changes-validator/api_changes_validator.ts#L49

scripts/dependabot_generate_changesets.ts
```diff
@@ -0,0 +1,93 @@
...
+    modifiedPackageDir
+  );
+  packageNames.push(modifiedPackageName);
+}
+
+const fileName = `.changeset/dependabot-${headRef}.md`;
+const versionUpdates = await getVersionUpdates(packageJsonFiles);
+await createChangesetFile(versionUpdates, packageNames);
+await gitClient.commitAllChanges('add changeset');
+await gitClient.push({ force: true });
```
sobolk, line 93: How does the stop condition work?
I.e. this will presumably push a change to PR and trigger new workflow. Would that trigger this script again or not ? If this is handled already - please drop a comment somewhere - it's not easy to catch this detail.

Other thing is e2e tests.
In case script adds extra commit then we'll be kicking off 2 e2e runs next to each other. If possible we should kick off one.
For example this script could add run-e2e label before pushing. And always commit something (even if it's only pacakge-lock update).

scripts/dependabot_generate_changesets.ts
```diff
@@ -0,0 +1,93 @@
+import fsp from 'fs/promises';
```
sobolk, line 1: I'd rename this script to `dependabot_handle_version_update` or something like this.
It's doing a bit more than generating changeset (i.e. it's also pushing it to repo).


Also - consider writing a happy case test case for this. 
Something like https://github.com/aws-amplify/amplify-backend/blob/main/scripts/components/release_lifecycle.test.ts .

sobolk (commented):
The idea for automation looks good.

.github/dependabot.yml
rtpascual: Just keeping that label from past dependabot PRs

scripts/dependabot_generate_changesets.ts
rtpascual: Sometimes dependabot will group version updates together so the title can be something like `Bump the npm_and_yarn group across 1 directory with 3 updates`.

I think description will be best, since it gives us the version dependabot updated from since there will be a line like `Updates/Bumps <dependency> from <old-version> to <new-version>`.

scripts/dependabot_generate_changesets.ts
rtpascual: Good catch I don't have anything in place to stop pushing the changeset from triggering the script again, I'll see where I can handle that (possibly using the fact that the script may be adding the `run-e2e` label?).

That's a good idea on the script adding the `run-e2e` label to prevent the initial e2e run, then we can push an empty commit or empty changeset (in case of only package-lock update).

sobolk (commented):
LGTM,
Just couple of comments.

scripts/components/dependabot_version_update_handler.ts
```diff
@@ -0,0 +1,135 @@
...
+   *
+   * Running this method when either of the following are true results in a no-op:
+   * 1. GitHub context event is not a pull request
+   * 2. Branch PR does not start with 'dependabot/', meaning the branch isn't for a Dependabot PR
+   * 3. PR already has a changeset in list of files changed
+   */
+  handleVersionUpdate = async () => {
+    if (!this._ghContext.payload.pull_request) {
+      // event is not a pull request, return early
+      process.exit();
```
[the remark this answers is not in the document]
rtpascual: Ah that was leftover before moving this to components, updated.

sobolk (approved):

sobolk (approved):

Amplifiyer (approved):

.github/workflows/health_checks.yml
```diff
@@ -269,10 +269,31 @@ jobs:
...
+      - name: Generate changesets for packages with version updates
+        run: npx tsx scripts/dependabot_handle_version_update.ts "$BASE_SHA" "$HEAD_SHA"
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
   do_include_e2e:
     needs:
       - install
       - resolve_inputs
+      - generate_changeset
```
sobolk, line 296: @rtpascual 

I think we can just remove this dependency here to fix the workflow on main.

Because:
First iteration of dependabot PR will add label AND push change to PR branch.

PR branch will trigger second independent run of workflow which will have label already added.

PR #2359, opened 2024-12-19
feat: add data logging api in `defineData`

bobbyu99:
## Changes

This PR added data logging options in `defineData`. Summary of changes:
- bumped `data-construct` version to consume the new logging api
- added `DataLogLevel` type
- added enum converter `DataLogLevelConverter` and logging options parser `convertLoggingOptionsToCDK`
- modified data factory to parse logging options

## Validation

Unit tests added for:
- enum converter `DataLogLevelConverter`
- logging options parser `convertLoggingOptionsToCDK`
- `defineData` with logging options

## Checklist


packages/plugin-types/API.md
```diff
@@ -142,6 +142,9 @@ export type ConstructFactoryGetInstanceProps = {
     resourceNameValidator?: ResourceNameValidator;
 };
 
+// @public (undocumented)
+export type DataLogLevel = 'none' | 'error' | 'info' | 'debug' | 'all';
```
sobolk, line 146: This type should be moved to `backend-data` .
Similarly to https://github.com/aws-amplify/amplify-backend/blob/6dd93d17fd04df3bcc2984682a2aa6b383c9a445/packages/backend-function/src/factory.ts#L70 . The function just aliases it because it uses all values.

The `LogLevel` defined in platform types is superset of known log levels.

We can then define `DataLogLevel` either like you have it here. 
OR we can do something like this `export type DataLogLevel = Extract<LogLevel, 'info' | 'debug'>;` to assure that `export type DataLogLevel = Extract<LogLevel, 'info' | 'debug'>;` is indeed subset of `LogLevel` .

packages/platform-core/src/cdk/enum_converters.ts
```diff
@@ -95,3 +100,27 @@ export class LogLevelConverter {
     }
   };
 }
+
+/**
+ * Converts DataLogLevel to CDK types.
+ */
+export class DataLogLevelConverter {
```
sobolk, line 107: This converter can (and should be) moved to `backend-data`. There's no need for it to be visible (other converters must be accessible from 2 packages).

It can be then package-private (not exposed).

packages/client-config/API.md
```diff
@@ -627,7 +627,7 @@ export type CustomClientConfig = {
 export const DEFAULT_CLIENT_CONFIG_VERSION: ClientConfigVersion;
 
 // @public
-export const generateClientConfig: <T extends "1" | "1.1" | "1.2" | "1.3" | "0">(backendIdentifier: DeployedBackendIdentifier, version: T, awsClientProvider?: AWSClientProvider<{
+export const generateClientConfig: <T extends "0" | "1" | "1.1" | "1.2" | "1.3">(backendIdentifier: DeployedBackendIdentifier, version: T, awsClientProvider?: AWSClientProvider<{
```
sobolk, line 630: This drift should be reverted (we have an edge case with extractor in our repo).
Just keep it out of PR, or run clean build. `npm run clean && npm install && npm run build && npm run update:api`

Sorry for inconvenience here.

sobolk (commented):
LGTM.

We should move couple of things around to reduce public api additions.

packages/platform-core/src/cdk/enum_converters.ts
sobolk: There's one more alternative to try here.

1. The `DataLogLevelConverter` is removed.
2. The existing `LogLevelConverter` gets `LogLevelConverter.toCDKFieldLogLevel` method - handle subset, throw otherwise.
3. The `DataLogLevel` is moved to `backend-data`
4. `backend-data` calls `LogLevelConverter.toCDKFieldLogLevel` with `DataLogLevel` which works because `DataLogLevel` it's a subset.

The caveat is that we expose public api just for consistency sake. 

@Amplifiyer wdyt?

Amplifiyer: We are already doing that with `toCDKLambdaApplicationLogLevel`, so this suggestion is definitely in line with that.

sobolk: `LogLevelConverter.toCDKAppsyncFieldLogLevel(logLevel: LogLevel)` (we prefix  `ApplicationLogLevel` with `Lambda`).

packages/backend-data/src/types.ts
```diff
@@ -147,3 +157,78 @@ export type AmplifyDataError =
...
+   * - **'debug'**: Logs the following information only for the fields that are in the debug, info, and error categories:
+   *   - Debug-level messages.
+   *   - The user messages sent through `$util.log.info`, `$util.log.debug`, `console.log`, and `console.debug`.
+   *   - Field-level tracing and mapping logs are not shown.
+   * - **'all'**: The following information is logged for all fields in the query:
+   *   - Field-level tracing information.
+   *   - The generated request/response functions that were resolved for each field.
+   * @default 'none'
+   */
+  level?: DataLogLevel;
```
Amplifiyer, line 233: Should this be name `fieldLogLevel` just as you described in the default example in the docs above?

bobbyu99: Thanks for catching the discrepancy in the doc. `fieldLogLevel` is used in the cdk. But I think `level` as the API is more straightforward and aligns with what we already have in `defineFunction`. What do you think?

packages/platform-core/src/cdk/enum_converters.ts
bobbyu99: For this to work, we will need to expand `LogLevel`
https://github.com/aws-amplify/amplify-backend/blob/6dd93d17fd04df3bcc2984682a2aa6b383c9a445/packages/plugin-types/src/log_level.ts#L1
to
```
export type LogLevel = 'info' | 'debug' | 'warn' | 'error' | 'fatal' | 'trace' | 'none' | 'all';
```
and change `DataLogLevel` and `FunctionLogLevel`:
```
export type DataLogLevel = Extract<LogLevel, 'none' | 'all' | 'info' | 'debug' | 'error'>;
```
```
export type FunctionLogLevel = Extract<LogLevel, 'info' | 'debug' | 'warn' | 'error' | 'fatal' | 'trace'>;
```

Is that what you are suggesting?

sobolk: Yes. I did not notice that appsync has one extra value.

We should also cover this line in this case https://github.com/aws-amplify/amplify-backend/blob/e7f415cee5d84f4494538c2583af62a058707cdb/packages/backend-ai/src/conversation/factory.ts#L162 . (same transformation like in functions case).

bobbyu99: updated `ConversationHandlerFunctionLogLevel`

packages/backend-data/src/types.ts
sobolk: Does AppSync have other log levels ? for example `modelLogLevel` ?
If it does have it or have potential to have it in the future then - would we add second "level property" here or use this `level` to feed both?

bobbyu99: I don't find any other log level besides `fieldLogLevel`. I'm confirm if AppSync has plan to add other log levels.

bobbyu99: But thanks for the catch. I think it's better to name it `fieldLogLevel` for clarity and to differentiate in case AppSync has plan later to add other log level

.changeset/quick-dingos-whisper.md
```diff
@@ -0,0 +1,9 @@
+---
+'@aws-amplify/backend-function': minor
+'@aws-amplify/platform-core': minor
+'@aws-amplify/backend-data': minor
+'@aws-amplify/plugin-types': minor
+'@aws-amplify/backend-ai': minor
+---
```
sobolk, line 7:
```suggestion
---
'@aws-amplify/backend-function': minor
'@aws-amplify/platform-core': minor
'@aws-amplify/backend-data': minor
'@aws-amplify/plugin-types': minor
'@aws-amplify/backend-ai': minor
'@aws-amplify/backend': minor
---
```

sobolk (approved):

Amplifiyer (approved):

PR #2361, opened 2024-12-20
POC - Seed Secrets

ShadowCat567:
## Problem

## Changes

## Checklist


packages/backend-seed/src/namespace_resolver.ts
```diff
@@ -0,0 +1,35 @@
+import {
+  AmplifyUserError,
+  PackageJsonReader,
+} from '@aws-amplify/platform-core';
+import { EOL } from 'os';
+
+//not using this anymore
+//stole this from ../packages/cli/src/backend-identifier/local_namespace_resolver.ts for the POC
```
sobolk, line 8: these will probably go to `cli-core` .

packages/backend-seed/API.md
```diff
@@ -0,0 +1,12 @@
+## API Report File for "@aws-amplify/backend-seed"
+
+> Do not edit this file. It is a report generated by [API Extractor](https://api-extractor.com/).
+
+```ts
+
+// @public (undocumented)
+export const GetSeedSecret: (secretName: string) => Promise<string>;
```
sobolk, line 8: I'd just call it `getSecret` .

Other option we could consider is 

```
type SeedSecretClient {
  getSecret: () => Promise<string>;
}

// factory returns implementation
getSeedSecretClient = () => SeedSecretClient
```

We should think about packaging of this because APIs to deal with secrets will grow over time.
We'll also have `AuthSeedClient` or so or set of auth related APIs. it would be good to group secrets and auth.

An alternative would be to develop namespaces:
`@aws-amplify/backend-seed/secret` and `@aws-amplify/backend-seed/auth` .

We should capture these alternatives in API review.

ShadowCat567: I initially called it `getSecret`, but I got some errors since it conflicted with `getSecret` from the `backend-secret` package -- I'll probably experiment with it again

sobolk: This might be helpful syntax to resolve that problem https://github.com/aws-amplify/amplify-backend/blob/b574bce26ec2d55a4391aeceecc1d4957e79fba3/packages/backend-output-schemas/src/auth/index.ts#L2 .

or `import * as backendSecret from 'backend-secret'` and then `backendSecret.getSecret` .

PR #2380, opened 2024-12-30
add e2e test project reset and add reset to hotswap retries

rtpascual:
## Problem

When performing retries of hotswapping resources in sandbox E2E tests, we are not bringing the test project back to pre-update state (without the updated files) for each retry.

## Changes

Add `reset` method to `TestProjectBase` which does the following:
1. Cleans out project's `amplify` directory
2. Copies the contents of the source project's `amplify` directory to the emptied project `amplify` directory
3. Sets up `package.json` and `tsconfig.json` in the project's `amplify` directory again (since they were removed from step 1).

Use `testProject.reset()` for retries of hotswapping resources.

## Validation

E2E tests

## Checklist


packages/integration-tests/src/test-e2e/sandbox/sandbox.test.template.ts
```diff
@@ -86,8 +86,12 @@ export const defineSandboxTest = (testProjectCreator: TestProjectCreator) => {
           if (updates.length > 0) {
             // retry hotswapping resources if deployment time is higher than the threshold
             await runWithRetry(
-              async () => {
-                // keeping initial deployment in retry loop to reset app state for each hotswap to be a non no-op
+              async (attempt) => {
+                if (attempt > 1) {
```
sobolk, line 90: nit: we could skip this if. reset of fresh project is no-op.
this is more explicit though so feel free to disregard.

sobolk (approved):

packages/integration-tests/src/test-e2e/sandbox/sandbox.test.template.ts
rtpascual: That's true but I preferred being more explicit here so the project doesn't reset when it doesn't need to on first attempt of hotswap.

Amplifiyer (approved):

PR #2409, opened 2025-01-06
fix: Update getAmplifyDataClient to have strict env type and remove narrowing logic

stocaaro:
## Problem
Developers encountering function build issues are running into the `getAmplifyDataClient` invalidType error, which is a distraction from the sandbox errors with clear actionable resolution steps.

Since the `env` type in these cases is `any`, using plain typescript to validate the `env` argument for `getAmplifyDataClient` would not result in a type error and the function is already annotate with instructions on how to resolve the issue when the `env` type is missing the data variables caused by not having been granted resource authorization.

- https://github.com/aws-amplify/amplify-backend/issues/2320

## Changes
Remove the `env` type narrowing and use a clear `env` type.

## Validation
Validated manually in a test project.

## Checklist


packages/backend-function/src/runtime/get_amplify_clients_configuration.ts
```diff
@@ -111,7 +111,7 @@ const getLibraryOptions = (env: DataClientEnv): LibraryOptions => {
 };
 
 export type InvalidConfig = unknown & {
-  invalidType: 'Some of the AWS environment variables needed to configure Amplify are missing.';
+  invalidType: 'Some of the AWS environment variables needed to configure Amplify are missing. Check the sandbox output for an error with resolution guidance.';
```
Amplifiyer, line 114: why is this an `invalidType` error and why is it assigned to `resourceConfig` and `libraryOptions` in the return object instead of an error object with perhaps a resolution field.

stocaaro: Interesting. I was going to answer "yes", but in running a quick test, the env variable is `any` when the function build fails (link in the customer cases this is in response to). As a result, testing the `env` input type shows no errors when `env` is `any`. Maybe this is the better result for these cases.

The original requirement we where considering with this is that when a customer uses `getAmplifyDataClientConfig` in a function, but forgets to grant resource authorization to that function, the customer should get clear direction. A type error due to missing the `env` variables doesn't clearly tell the customer to add `resource` permissions. The `any` case wasn't carefully considered when putting this together.

Amplifiyer: >but in running a quick test, the env variable is any when the function build fails (like in the customer cases this is in response to). As a result, testing the env input type shows no errors when env is any. Maybe this is the better result for these cases.

Yes, when the build fails, it's better to not tag on another TS error and let the customer focus on **why** the build failed. When the build is successful and we have the `env` type narrowed down(`LambdaProvidedEnvVars`?), we then do type checking on the input to this function. If we can't do that, I'd much rather prefer a run time error than doing this redirection.

stocaaro: Testing around this, I agree with this enough that I've altered this change to adopt your suggestion. I have re-tested the experience with this surface and this is clearly what we should have been doing from the beginning.

Amplifiyer: Any chance this could cause breaking experience for customers who already started using the current types?

stocaaro: Technically, since we're changing the argument type, this is a breaking change. 

In practice, customers who have a working `env` value will have no change in experience and customers with incorrect `env` types will no longer receive a type error and their errors in the console will persist.

I can't think of a case where a working app would be broken by this change.

The exported types we're removing are all internal. Hypothetically, a customer could be extracting a type, in which case the extracted type is changing and I'm not sure what would happen in this case.
Ex:
```ts
type EnvType = Parameters<typeof getAmplifyDataClientConfig>[0];
```

sobolk: > Technically, since we're changing the argument type, this is a breaking change.

> In practice, customers who have a working env value will have no change in experience and customers with non-functional env types will no longer receive a type error for any and when they're missing the data types in env the type issue will be more about specific env vars.

And the existing error is not something customers would catch and handle programmatically. So I think we're safe here.

Amplifiyer (dismissed):

Amplifiyer (approved):

PR #2414, opened 2025-01-08
report cdk versions

rtpascual:
## Problem

We are blind to versions of CDK used in customer projects

## Changes

Report CDK version.

How we get the version is through customer project's lock file (which is different based on package manager) so included in this PR is:
- Abstractions for logic that will parse lock file based on package manager
- A generic way to pick the dependency version we want to report so we can add more if we want to

## Validation

Unit tests

## Checklist


packages/platform-core/src/lock-file-reader/lock_file_reader_factory.ts
```diff
@@ -0,0 +1,63 @@
+import { NpmLockFileReader } from './npm_lock_file_reader';
+import { PnpmLockFileReader } from './pnpm_lock_file_reader';
+import { YarnClassicLockFileReader } from './yarn_classic_lock_file_reader';
+import { YarnModernLockFileReader } from './yarn_modern_lock_file_reader';
+
+/**
+ * LockFileReaderFactory is a factory for an abstraction around reading lock files from different package managers
+ */
+export class LockFileReaderFactory {
```
sobolk, line 9: This looks somewhat similar to https://github.com/aws-amplify/amplify-backend/blob/main/packages/cli-core/src/package-manager-controller/package_manager_controller_factory.ts .

Reading lock files is a good idea - it's fastest way to get data and does not rely on child process.

However, I suggest to package this functionality a bit differently. I.e.:
1. Add `PackageManagerController.getDependencies(): Array<Dependency>` to [existing type](https://github.com/aws-amplify/amplify-backend/blob/22a286384e27fae69e3249c49fd46e7961080e7a/packages/plugin-types/src/package_manager_controller.ts#L26).
2. Move all these "lock file readers" to `cli-core` (not platform-core) and use them to implement `getDependencies` in each package manager controller subtype.
3. Get rid of "lock file reader factory" - it won't be needed if we move this functionality under package manager controller.

packages/platform-core/src/lock-file-reader/lock_file_reader_factory.ts
```diff
@@ -0,0 +1,63 @@
...
+        return new NpmLockFileReader();
+      case 'pnpm':
+        return new PnpmLockFileReader();
+      case 'yarn-classic':
+        return new YarnClassicLockFileReader();
+      case 'yarn-modern':
+        return new YarnModernLockFileReader();
+      default:
+        // defaults to npm lock file reader as it is the most used package manager by customers
+        return new NpmLockFileReader();
```
sobolk, line 31: if we can't detect package manager we can either throw or have NoOpLockFileReader.
(this comment might be irrelevant if we move this under package manager controller)

packages/platform-core/src/lock-file-reader/npm_lock_file_reader.ts
```diff
@@ -0,0 +1,59 @@
...
+      process.cwd(),
+      'package-lock.json'
+    );
+
+    let jsonLockParsedValue: Record<string, unknown>;
+    try {
+      const jsonLockContents = await fsp.readFile(packageLockJsonPath, 'utf-8');
+      jsonLockParsedValue = JSON.parse(jsonLockContents);
+    } catch (error) {
+      // We failed to get lock file contents either because file doesn't exist or it is not parse-able
```
sobolk, line 26: if we move this to cli-core then we should `printer."log debug"` that we filed to get lock file content.

similarly in other versions of this impl.

packages/platform-core/src/lock-file-reader/pnpm_lock_file_reader.test.ts
```diff
@@ -0,0 +1,98 @@
...
+import fsp from 'fs/promises';
+import { afterEach, describe, it, mock } from 'node:test';
+import path from 'path';
+import { PnpmLockFileReader } from './pnpm_lock_file_reader';
+
+void describe('PnpmLockFileReader', () => {
+  const fspReadFileMock = mock.method(
+    fsp,
+    'readFile',
+    // eslint-disable-next-line spellcheck/spell-checker
```
sobolk, line 11: `lockfile` can go to dictionary

packages/platform-core/src/lock-file-reader/lock_file_reader_factory.ts
rtpascual: Yeah I was referring to package manager controller to abstract the logic for different lock files. I didn't want to use a child process in order to get the data we need as well and over complicate our usage data retrieval.

I see, I was avoiding circular dependencies between `cli-core` and `platform-core` so I'll try this out, saves some duplicate code as well.

packages/platform-core/src/lock-file-reader/lock_file_reader_factory.ts
rtpascual: I was avoiding throwing to make sure we don't fail to send other usage data in case retrieving dependency versions fails (this would ultimately return an empty dependency array a little further down the line after failing to find the package-lock.json file).

packages/plugin-types/API.md
```diff
@@ -196,6 +202,16 @@ export type ImportPathVerifier = {
...
+// @public (undocumented)
+export type LockFileContents = {
+    dependencies: Array<Dependency>;
+};
+
+// @public (undocumented)
+export type LockFileReader = {
+    getLockFileContentsFromCwd: () => Promise<LockFileContents>;
+};
+
```
sobolk, line 214: it seems that these two types can be internalized in cli-core and not public.

packages/plugin-types/API.md
```diff
@@ -220,6 +236,7 @@ export type PackageManagerController = {
     runWithPackageManager: (args: string[] | undefined, dir: string, options?: ExecaOptions) => ExecaChildProcess;
     getCommand: (args: string[]) => string;
     allowsSignalPropagation: () => boolean;
+    getDependencies: () => Promise<Array<Dependency>>;
```
sobolk, line 239: I wonder if we should have
`tryGetDependencies: () => Promise<Array<Dependency> | undefined>;`

given opportunistic nature of this API.

return `undefined` if attempt failed. (which is different from `[]` which would mean project without dependencies - in theory - this won't happen in real projects for now)

sobolk (commented):
LGTM.

packages/plugin-types/API.md
rtpascual: That makes sense and distinguishes the two scenarios, I'll have to make a few tweaks to the readers since I don't think they handle lock files without dependencies gracefully.

sobolk (dismissed):

sobolk (approved):

packages/cli-core/src/package-manager-controller/lock-file-reader/yarn_classic_lock_file_reader.ts
```diff
@@ -0,0 +1,52 @@
...
+    const eolRegex = '[\r\n]';
+    const dependencies: Array<Dependency> = [];
+    const yarnLockPath = path.resolve(process.cwd(), 'yarn.lock');
+
+    try {
+      const yarnLockContents = await fsp.readFile(yarnLockPath, 'utf-8');
+      const yarnLockContentsArray = yarnLockContents
+        .trim()
+        .split(new RegExp(`${eolRegex}${eolRegex}`));
+
```
ShadowCat567, line 24: In general, looks good! I am excited to have this implemented. I just have a question about `YarnClassicLockFileReader`, the other 3 LockFileReaders have an early return if there is a LockFile but no dependencies, why doesn't yarn classic?

ShadowCat567 (approved):

packages/cli-core/src/package-manager-controller/lock-file-reader/yarn_classic_lock_file_reader.ts
rtpascual: Good question and I probably should have left a comment or be more explicit with this to be consistent with the others.

In Yarn classic, the lock file only contains the comment block and a few empty lines https://github.com/aws-amplify/amplify-backend/pull/2414/files#diff-43f1c04808a37bbc8864616b3975a823536369171b5e0d4d7ecacef025fea884R72-R76. This makes the for loop a no-op due to the slice that happens which results in:
```
for (const yarnDependencyBlock of []) {
...
}
```

rtpascual: It just happened that while iterating through these LockFileReaders to handle a project with no dependencies I didn't have to make any changes to `YarnClassicLockFileReader`

sobolk: If we have consistent test cases for each subtype then we should be covered here.

Issue #2257, opened 2024-11-20
support `sandbox secret remove --all`


josefaidt:
### Environment information

```plain text
n/a
```

### Describe the feature

It can be tedious to remove all secrets related to a sandbox environment without removing each individual secret, or by navigating to the AWS console to remove secrets manually. Support the `--all` flag on `sandbox secret remove` to remove all secrets in a single command	

### Use case

see above

ykethan:
Marking as feature request

ykethan:
Closing as this is now released on `@aws-amplify/backend-cli@1.5.0`

PR #2423, opened 2025-01-14
feat: add --all flag to remove all secrets in sandbox

fossamagna:
## Problem

Support the --all flag on sandbox secret remove to remove all secrets in a single command

#2257

## Changes

I change to use `yargs.command` instead of `yargs.positional` to handling the `secret-name` argument. 
We can not set `secret-name` as optional argument when use `yargs.positional` without default value.
Instead of we can set `secret-name` as optional argument that using `[secret-name]` syntax in `yargs.command`

## Validation

- Unit Tests
- Manual Tests

## Checklist


fossamagna:
@josefaidt
Hi, I would be grateful if you could review this PR.

packages/cli/src/commands/sandbox/sandbox-secret/sandbox_secret_remove_command.ts
```diff
@@ -41,28 +42,53 @@ export class SandboxSecretRemoveCommand
...
+      printer.print(`Successfully removed secret ${args.secretName}`);
+    } else if (args.all) {
+      const secrets = await this.secretClient.listSecrets(
+        sandboxBackendIdentifier
+      );
+      await Promise.all(
+        secrets.map((secret) =>
+          this.secretClient.removeSecret(sandboxBackendIdentifier, secret.name)
+        )
+      );
```
sobolk, line 59: The delete api has low TPS quota https://docs.aws.amazon.com/general/latest/gr/ssm.html#parameter-store .

I think we should add `secretClients.removeSecrets(backendIndentifer, secretNames: Array<string>)`

And use `DeleteParameters` batch api instead of this.

packages/cli/src/commands/sandbox/sandbox-secret/sandbox_secret_remove_command.ts
```diff
@@ -41,28 +42,53 @@ export class SandboxSecretRemoveCommand
...
+      })
+      .check(async (argv) => {
+        if (!argv.all && !argv['secret-name']) {
+          throw new AmplifyUserError('InvalidCommandInputError', {
+            message: 'Either secret-name or all flag must be provided',
+            resolution: 'Provide either secret-name or all flag',
+          });
+        }
+        return true;
+      });
```
sobolk, line 82: how does the `npx ampx sandbox secret remove --help` output look like before and after change?
could you please check and post the outputs before and after ?

sobolk: It seems we're loosing secret name description from help. (btw I logged https://github.com/aws-amplify/amplify-backend/issues/2579 to fix unrelated problem of excessive options presented in help output). 

> I change to use yargs.command instead of yargs.positional to handling the secret-name argument.
We can not set secret-name as optional argument when use yargs.positional without default value.
Instead of we can set secret-name as optional argument that using [secret-name] syntax in yargs.command

@fossamagna are you sure this is valid? I played with this a bit and I managed to retain `secret-name` in help output.

Changes I made locally:
1. `this.command = 'remove <secret-name>';` -> `this.command = 'remove [secret-name]';`
2. `    secretName: string;` -> `    secretName: string | undefined;` (in `SecretRemoveCommandOptionsKebabCase` definition)
3. `demandOption: true,` -> `demandOption: false,` (in positional call)

Output I got:

```
$ npx ampx sandbox secret remove --help
ampx sandbox secret remove [secret-name]

Remove a sandbox secret

Positionals:
  secret-name  Name of the secret to remove a                           [string]

Logs streaming
      --logs-filter    Regex pattern to filter logs from only matched functions.
                        E.g. to stream logs for a function, specify it's name, a
                       nd to stream logs from all functions starting with auth s
                       pecify 'auth' Default: Stream all logs            [array]
      --logs-out-file  File to append the streaming logs. The file is created if
                        it does not exist. Default: stdout              [string]

Options:
      --debug       Print debug logs to the console   [boolean] [default: false]
      --identifier  An optional identifier to distinguish between different sand
                    boxes. Default is the name of the system user executing the
                    process                                             [string]
      --profile     An AWS profile name.                                [string]
      --once        Execute a single sandbox deployment without watching for fut
                    ure file changes                                   [boolean]
  -h, --help        Show help                                          [boolean]
```

fossamagna: @sobolk 
I was also able to retain the description for the `secret-name` parameter in help output by calling positional as you suggested. Thank you very much.

sobolk (approved):

Amplifiyer (approved):

PR #2426, opened 2025-01-15
Lint rule for cause propagation when wrapping errors

ShadowCat567:
## Problem

No lint rule in place to ensure that we are propagating the underlying cause for errors we wrap

## Changes

Added lint rule that mandates cause is added to all errors we wrap from catch blocks.

## Validation

Added new tests, lint fails if an `AmplifyFault`, `AmplifyUserError`, or `Error` is declared without a cause property.

## Checklist


rtpascual (dismissed):

sobolk (commented):
We discussed this offline.

the rule should flag `catch` blocks where `error` is not passed to ctor.

Example valid blocks:
```
} catch (e) {
  throw new Error('foo', { cause: e});
}

} catch (e) {
  throw new AmplifyError('foo', { ..some props...}, e);
}
```

Sample invalid blocks:
```
} catch (e) {
  throw new Error('foo');
}

} catch (e) {
  throw new AmplifyError('foo', { ..some props...});
}

} catch {
  throw new Error('foo');
}
```

In other words the rule should:
1. Process `CatchClause`.
2. Detect if there's any `NewExpression` inside with Error/AmplifyError/Fault
3. Assert that "cause" (`e`) is passed in new expression.

ShadowCat567:
Still a Work in Progress. Currently this will work for most `AmplifyUserErrors` and `AmplifyFaults`, aside from ones that are hidden inside if and/or switch statements like this: 
image
Currently working on getting this to apply to `Error` as well

ShadowCat567:
> We discussed this offline.
> 
> the rule should flag `catch` blocks where `error` is not passed to ctor.
> 
> Example valid blocks:
> 
> ```
> } catch (e) {
>   throw new Error('foo', { cause: e});
> }
> 
> } catch (e) {
>   throw new AmplifyError('foo', { ..some props...}, e);
> }
> ```
> 
> Sample invalid blocks:
> 
> ```
> } catch (e) {
>   throw new Error('foo');
> }
> 
> } catch (e) {
>   throw new AmplifyError('foo', { ..some props...});
> }
> 
> } catch {
>   throw new Error('foo');
> }
> ```
> 
> In other words the rule should:
> 
> 1. Process `CatchClause`.
> 2. Detect if there's any `NewExpression` inside with Error/AmplifyError/Fault
> 3. Assert that "cause" (`e`) is passed in new expression.
This has been completed, now to see if our checks will be happy

packages/cli-core/src/package-manager-controller/execute_with_debugger_logger.ts
```diff
@@ -28,6 +28,7 @@ export const executeWithDebugLogger = (
     );
 
     return childProcess;
+    // eslint-disable-next-line amplify-backend-rules/propagate-error-cause
```
sobolk, line 31: I'd address this one.

I patched it in one of snapshot releases I shipped to troubleshoot one of faults before. Didn't get it to main, but it would be good to do so.

packages/eslint-rules/src/rules/propagate_error_cause.ts
```diff
@@ -0,0 +1,191 @@
...
+
+            if (
+              body.type !== 'ThrowStatement' &&
+              body.type !== 'IfStatement' &&
+              body.type !== 'SwitchStatement' &&
+              body.type !== 'BlockStatement'
+            ) {
+              for (const newBody of node.body.body) {
+                if (newBody.type === 'VariableDeclaration') {
+                  //@ts-expect-error assumes incorrect type for body.declarations[0].id
```
sobolk, line 43: instead of `@ts-expect-error` you can `(body.declarations[0].id as <desiredType>).name`
this would be preferable way of doing this. See example here https://github.com/aws-amplify/amplify-backend/blob/040272d7d05427d2de82658e8de2ec0d60981b46/scripts/components/api-changes-validator/api_usage_statements_generators.ts#L225

The reason is that casting with `as` dismisses less potential problems than `@ts-expect-error`. We should use `@ts-expect-error` only if we exhausted other options.

Please apply this to other places.

packages/eslint-rules/src/rules/propagate_error_cause.ts
```diff
@@ -0,0 +1,191 @@
...
+            const causeVarNames = [];
+            causeVarNames.push(node.param.name);
+            const queue = [];
+            let body = node.body.body[0];
+
+            if (
+              body.type !== 'ThrowStatement' &&
+              body.type !== 'IfStatement' &&
+              body.type !== 'SwitchStatement' &&
+              body.type !== 'BlockStatement'
```
sobolk, line 39: We should add some "valid" samples for each type of nested blocks in the test.

packages/eslint-rules/src/rules/propagate_error_cause.ts
```diff
@@ -0,0 +1,191 @@
...
+                          prop.key.name === 'cause' &&
+                          causeVarNames.includes(prop.value.name)
+                        ) {
+                          return; // the error was thrown with a cause and the cause was valid
+                        }
+                      }
+                      context.report({
+                        messageId: 'noCausePropagation',
+                        node,
+                      });
```
sobolk, line 86: I would suggest to split the algorithm into two pieces.

1. traverse catch block AST and just find all throw statements with new expressions. Accumulate them in separate array. (the reason is that we could then extract this algorithm to separate class/method if we were to build similar rule that needs to find nested elements.
2. After traversing AST, loop through accumulated "blocks of interest" and assert/report violations.

scripts/update_tsconfig_refs.ts
```diff
@@ -52,6 +52,7 @@ for (const packagePath of packagePaths) {
   let tsconfigObject: Record<string, unknown>;
   try {
     tsconfigObject = JSON.parse(fs.readFileSync(tsconfigPath, 'utf-8'));
+    // eslint-disable-next-line amplify-backend-rules/propagate-error-cause
```
sobolk, line 55: let's address this one too.

sobolk (commented):
Looks good overall.

packages/eslint-rules/src/rules/propagate_error_cause.ts
sobolk: For the 1. I'd extract traversal to something like this:

```
findNestedNodes(root, predicate: (node) => boolean);

// Usage:
findNestedNodes(catchBlock, (node) => {               return  body.argument &&
                  body.argument?.type === 'NewExpression' &&
                  body.argument.callee.name === errorType});
```

sobolk: Or perhaps `handleNestedBlocks(root, onNode: (node) => void);` because we may also track the assignments at some point (i.e. not only predicate).

ShadowCat567: in the rule itself, I should have something like this (I have return type of `findNestedNodes` as an array): 
image
and for the nested nodes function itself, I would have something like this -- use the predicate to figure out whether I want the throwStatement: 
image

packages/create-amplify/src/get_project_root.ts
```diff
@@ -39,6 +39,7 @@ export const getProjectRoot = async () => {
     printer.log(`Creating directory ${projectRoot}`, LogLevel.DEBUG);
     try {
       await fsp.mkdir(projectRoot, { recursive: true });
+      // eslint-disable-next-line amplify-backend-rules/propagate-error-cause
```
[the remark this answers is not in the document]
ShadowCat567: this one is a little annoying because there is a test that matches against the output of this error so that should get changed as well

packages/eslint-rules/src/rules/find_nested_nodes.ts
```diff
@@ -0,0 +1,50 @@
...
+  root: TSESTree.CatchClause,
+  addToCauseNames: (name: string) => void
+) {
+  const queue = [];
+  let body = root.body.body[0];
+
+  if (body.type === 'VariableDeclaration') {
+    for (const newBody of root.body.body) {
+      if (newBody.type === 'VariableDeclaration') {
+        addToCauseNames((body.declarations[0].id as TSESTree.Identifier).name);
```
sobolk, line 17: This is a very use case specific logic.

I'd bubble this up to caller and drop `addToCauseNames` parameter.

Btw. I think we can handle this with single for loop at the caller. Because presumably variable declarations are always before their usage, so populating `causeVarNames` and processing other conditions at the same time should work fine.

sobolk (approved):

rtpascual (approved):

PR #2452, opened 2025-01-27
adds check command

ykethan:
## Problem

folks will need to run sandbox command to compile/synthesize their backend for any error.
The `npx ampx check` will dry run with just tsc and synth without deploying any resources.

https://github.com/user-attachments/assets/5033c257-a22f-443f-954d-7be10cfb2d11

image

## Changes

## Checklist

### DIFF
--- a/packages/cli/src/commands/check/check_command.ts
+++ b/packages/cli/src/commands/check/check_command.ts
@@ -1,0 +1,48 @@
+import { BackendDeployer } from '@aws-amplify/backend-deployer';
+import { format, printer } from '@aws-amplify/cli-core';
+import { BackendIdentifier } from '@aws-amplify/plugin-types';
+import { Argv, CommandModule } from 'yargs';
+
+/**
+ * Command that runs type checking and CDK synthesis without deployment
+ */
+export class CheckCommand implements CommandModule {
+  command = 'check';
+  describe = 'Run type checking and CDK synthesis without deployment';
+
+  /**
+   * Creates a new instance of CheckCommand
+   * @param backendDeployer - The deployer used to run CDK synthesis
+   */
+  constructor(private readonly backendDeployer: BackendDeployer) {}
+
+  handler = async () => {
+    const backendId: BackendIdentifier = {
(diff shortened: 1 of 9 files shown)


packages/cli/src/commands/check/check_command.ts
```diff
@@ -0,0 +1,48 @@
+import { BackendDeployer } from '@aws-amplify/backend-deployer';
+import { format, printer } from '@aws-amplify/cli-core';
+import { BackendIdentifier } from '@aws-amplify/plugin-types';
+import { Argv, CommandModule } from 'yargs';
+
+/**
+ * Command that runs type checking and CDK synthesis without deployment
+ */
+export class CheckCommand implements CommandModule {
+  command = 'check';
```
Amplifiyer, line 10: Not a big fan of the command name `check`. It's more closely related to `build` and perhaps `--dryRun` isn't bad either for the existing `sandbox` command. This way customers can utilize the `watch` nature of sandbox for builds as well.

josefaidt: I agree that it's more of a "build", though the `check` name was inspired by other projects that provide similar functionality with the intent of a sort of "preflight" task before merging changes. There is another request to provide a "diff" function and integrate cdk-nag to perform diagnostics and/or highlight best practices that I think would fit nicely with the "check" theme

https://docs.astro.build/en/reference/cli-reference/#astro-check
https://svelte.dev/docs/cli/sv-check

Amplifiyer: There is definitely overlap between `check` and `build`.  We might be slightly different in the sense that we "synth" as well. We discussed offline to do an API review of this change will all the proposals.

dwil618: Have you considered the option of `synth` too? 

I was looking for both `diff` and `synth` commands as a long time user of CDK alongside Amplify gen1. Given that CDK does type checking with tsc during synth too it seems functionally similar, and perhaps nice to have consistency in naming across the two services.

PR #2454, opened 2025-01-28
Seed POC: passwordless and MFA

ShadowCat567:
Used admin credentials to avoid role assumption complexities in seed command. If there are any thoughts about sign in and passkeys, I would like to hear them.

### Passkeys: what have I attempted
1. see `passwordless_authentication.ts` for attempt with `associateWebAuthnCredential()` from the JS library -- doesn't work because Web credentials stuff from the JS library does not want to be used in a node environment, seems to only work if it is in a browser
2. registering and using a passkey with SDK -- [Cognito Documentation on Passkeys](https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-authentication-flow-methods.html#amazon-cognito-user-pools-authentication-flow-methods-passkey) -- see `Register a passkey (SDK)`, this got confusing when it came to actually being able to get access the passkey to be able to use it, could not figure out how to do it (see `Sign in with a passkey`)
3. registering a passkey with managed login, did some research into this, crafting and using the links are the main problem that would make this difficult (see `Register a passkey (managed login)` in passkey documentation)


Amplifiyer:
can this be closed?

ShadowCat567:
> can this be closed?

yep, we probably don't need this for now

PR #2468, opened 2025-01-30
EACCES: Permission Denied Errors

ShadowCat567:
## Problem

Currently `EACCES: permission denied` errors appear with two commands, `generate outputs` and `configure profile`. I was able to replicate the error for `generate outputs` by removing permissions from `amplify_outputs.json`. And I was able to replicate both instances of the error for `configure profile` by removing write access and removing read access from `~/.aws/config`. These errors can only be replicated on Mac or Linux, if you attempt to replicate on Windows, you get `EPERM` errors.

Note about `fs.appendfile()` and `fs.writefile()`, mode only has an effect if this is the first time the file has been created: https://nodejs.org/api/fs.html#fsappendfilepath-data-options-callback -- I did attempt to do `fs.writefile(... {mode: 600})` for when the error appears with `generate outputs`, it did not fix the error. 

## Changes
Updated error messaging and error handling for `EACCES` errors that appear with `generate outputs` and `configure profile`

## Validation

Got a repro for both `generate outputs` and `configure profile` and was able to verify the new messaging shows up. 
Tests were added to `profile_controller.test.ts` and `client_config_writer.test.ts`.
## Checklist


packages/cli/src/commands/configure/profile_controller.ts
```diff
@@ -77,7 +78,23 @@ export class ProfileController {
...
+          'PermissionsError',
+          {
+            message: `You do not have the permissions to write to this file: ${filePath}`,
+            resolution: `Ensure that you have the right permissions to write to ${filePath}.`,
+          },
+          error
+        );
+      } else {
+        throw error;
+      }
```
rtpascual, line 96: Should we add tests for this and `client_config_writer.ts`?

ShadowCat567: that would probably be a good idea

ShadowCat567: the way test for `client_config_writer.ts` are set up are not really conducive to being able to test file permissions, but I was able to add tests for `profile_controller.ts`

rtpascual: Oh nice you are able to actually test file permissions. I was thinking to mock file system to throw `EACCES` error and some other error, here's an example of mocking `fsp.readFile` https://github.com/aws-amplify/amplify-backend/blob/56ec47ec5225dbf44af419036fa2d64c39bacf20/packages/cli-core/src/package-manager-controller/lock-file-reader/npm_lock_file_reader.test.ts#L56.

packages/cli/src/commands/configure/profile_controller.test.ts
```diff
@@ -218,6 +220,70 @@ void describe('profile controller', () => {
...
+      }
+
+      const expectedErr = new AmplifyUserError('PermissionsError', {
+        message: `You do not have the permissions to read this file: ${configFilePath}.`,
+        resolution: `Ensure that you have the right permissions to read from ${configFilePath}.`,
+      });
+      chmod(configFilePath, 0o000, () => {
+        //do nothing
+        return;
+      });
```
sobolk, line 236: we should either use `chmodSync` from `fs` or `import { chmod } from fs/promises` and await it. 
Otherwise this code has a potential race condition (rare but still).

packages/cli/src/commands/configure/profile_controller.ts
sobolk: I would mock as well, but for a simple test like this using `chmod` is acceptable imo.

sobolk (approved):

rtpascual (approved):

Issue #2413, opened 2025-01-07
functions with more than 10 secrets (including generated vars) fail to resolve at runtime


josefaidt:
### Environment information

```plain text
n/a
```

### Describe the bug

when using defineFunction with more than 10 secrets, the SDK fails to resolve due to a validation on the `Names` arg
```
Message: Value '[...]' at 'names' failed to satisfy constraint: 
Member must have length less than or equal to 10
```

```
ERROR   Uncaught Exception      {"errorType":"ValidationException","errorMessage":"1 validation error detected: Value '[...]' at 'names' failed to satisfy constraint: Member must have length less than or equal to 10","name":"ValidationException","$fault":"client","$metadata":{"httpStatusCode":400,"requestId":"05425c47-78fd-4b06-9757-3b01f61b984b","attempts":1,"totalRetryDelay":0},"__type":"ValidationException","message":"1 validation error detected: Value '[...]' at 'names' failed to satisfy constraint: Member must have length less than or equal to 10","stack":["ValidationException: 1 validation error detected: Value '[...]' at 'names' failed to satisfy constraint: Member must have length less than or equal to 10","    at throwDefaultError (/var/runtime/node_modules/@aws-sdk/node_modules/@smithy/smithy-client/dist-cjs/index.js:840:20)","    at /var/runtime/node_modules/@aws-sdk/node_modules/@smithy/smithy-client/dist-cjs/index.js:849:5","    at de_CommandError (/var/runtime/node_modules/@aws-sdk/client-ssm/dist-cjs/index.js:7021:14)","    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)","    at async /var/runtime/node_modules/@aws-sdk/node_modules/@smithy/middleware-serde/dist-cjs/index.js:35:20","    at async /var/runtime/node_modules/@aws-sdk/node_modules/@smithy/core/dist-cjs/index.js:165:18","    at async /var/runtime/node_modules/@aws-sdk/node_modules/@smithy/middleware-retry/dist-cjs/index.js:320:38","    at async /var/runtime/node_modules/@aws-sdk/middleware-logger/dist-cjs/index.js:34:22","    at async resolveSecrets (file:///var/task/index.mjs:1:690)","    at async internalAmplifyFunctionResolveSsmParams (file:///var/task/index.mjs:1:1390)"]}
```

### Reproduction steps

```ts
defineFunction({
  environment: {
    one: secret("one"),
    two: secret("two"),
[...]

PR #2480, opened 2025-02-04
paginate getParameter call when there are more than 10 parameters

Amplifiyer:
## Problem

when using defineFunction with more than 10 secrets, the SDK call fails to resolve due to a validation on the `Names` arg
```
Message: Value '[...]' at 'names' failed to satisfy constraint: 
Member must have length less than or equal to 10
```

```
ERROR   Uncaught Exception      {"errorType":"ValidationException","errorMessage":"1 validation error detected: Value '[...]' at 'names' failed to satisfy constraint: Member must have length less than or equal to 10","name":"ValidationException","$fault":"client",...
```

**Issue number, if available:** https://github.com/aws-amplify/amplify-backend/issues/2413

## Changes

Paginate the request to call `getParameters` with 10 parameters at a time.

## Validation

Unit tests and manual validation

## Checklist


packages/backend-function/src/lambda-shims/chunk_arrays.ts
rtpascual, line 1: We don't need this file since the code is included and tested for in the shims.

rtpascual (commented):
I'm curious how easily we can hit the limit of the Lambda's `Init` phase of 10 seconds. Would you need to max out the function's env vars to 4KB and have the max amount of extensions added as layers?

rtpascual (approved):

packages/backend-function/src/lambda-shims/chunk_arrays.ts
Amplifiyer: yeah I forgot to remove it earlier. Removed it in new revision

Amplifiyer:
> I'm curious how easily we can hit the limit of the Lambda's `Init` phase of 10 seconds. Would you need to max out the function's env vars to 4KB and have the max amount of extensions added as layers?

I haven't tested what would break the limit of the `init` phase. If that ever happens, we may need to revisit this feature (adding extensions would likely be extra cost)

PR #2503, opened 2025-02-12
Pretty sandbox: integrate with aws cdk toolkit

Amplifiyer:
## Problem

Sandbox output is too noisy as it streams cdk cli output as-is. 

## Changes

Integrate with [CDK Toolkit](https://docs.aws.amazon.com/cdk/api/toolkit-lib/) to use CDK's programmatic API and structured API responses instead of calling CDK cli through a child process. Consume events through CDK's Toolkit API for showing sandbox output to customers. Here are the top level changes

- [packages/backend-deployer/src/cdk_deployer.ts](https://github.com/aws-amplify/amplify-backend/pull/2503/files#diff-d3e53744e47d7ba2e6cf7029d26ce38ceefb14fa0a26471f00a39021c73d1e82) -> Integrate/Replace cdk-cli with Toolkit API
- [packages/cli-core/src/loggers/amplify_io_events_bridge.ts](https://github.com/aws-amplify/amplify-backend/pull/2503/files#diff-be6a683e46b23231d358bd7877f4bd338c9b7461c49879c9c1033021cbd5223f) -> Implements CDK IOHost while extending it for amplify events. This creates a bridge that allows multiple consumers to "subscribe" to the events and do whatever they want with it.
- [packages/cli-core/src/loggers/amplify_event_loggers.ts](https://github.com/aws-amplify/amplify-backend/pull/2503/files#diff-2e734deff77e428261d8bff3e026a3efe0ed2a112bc68c76c074247aadf94136) -> Loggers used for displaying information to customers. These loggers subscribe to the bridge and manage how/what to display to customers.
- [packages/cli-core/src/loggers/cfn-deployment-progress/cfn_deployment_progress_logger.ts](https://github.com/aws-amplify/amplify-backend/pull/2503/files#diff-ce3ca5b6c40d810db81f3a41bcfe1dece530b10397f5617a57b60c7fe774949c) -> Given a stream of CDK/CFN deployment events, it collates, formats the display of deployment progress information in a [self-refreshing "window"](https://github.com/aws-amplify/amplify-backend/pull/2503/files#diff-32c97208d0aadb05ce4de8bc1ee0d8f76b7c889544b10243243f8d342d79fdf4) in terminal.
- Printer and format changes for supporting the new display format

### Things still left
- Completely remove aws-cdk dependency from the library (we shouldn't need it)
- Tests for new loggers

## Checklist


packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -18,72 +20,118 @@ import {
...
+          on the CDK side resulting in the app not getting synthesized properly. So we send a signal/message
+          to the same process and catch it in backend package where App is initialized to explicitly perform synth
+     */
+    const cx = await this.cdkToolkit.fromAssemblyBuilder(
+      async () => {
+        await tsImport(
+          pathToFileURL(this.backendLocator.locate()).toString(),
+          import.meta.url
+        );
+        process.emit('message', 'amplifySynth', undefined);
```
sobolk, line 83: I have some more thoughts about this solution after seeing it again. Mostly related to synchronization.

I'm assuming the event handler is called synchronously on the amplify app side by Node right?

If yes, then - does Node document/guarantee such behavior, i.e. is there any risk that this gets de-synchronized unintentionally?

Some mitigations that I have in mind that might be worth trying:
1. It seems that `emit` can accept extra parameters that could be received on the other side by listener. We could pass an unresolved Promise to amplifySynth and use it to signal back completion, on the emitter side await that promise after dispatching signal - if the event handler is synchronous it's no-op, otherwise it would prevent race condition.
2. An alternative to event could be that amplify app registers itself somewhere under `global` and can be discovered here. This strongly assumes that there's only one amplify app, which I think is safe to do. (I don't think this is elegant solution, but removes reliance on events).

packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -128,16 +198,50 @@ export class CDKDeployer implements BackendDeployer {
...
+      contextParams[CDKContextKey.BACKEND_NAMESPACE] = backendId.namespace;
+      contextParams[CDKContextKey.BACKEND_NAME] = backendId.name;
+      contextParams[CDKContextKey.DEPLOYMENT_TYPE] = backendId.type;
+      assembly = await this.cdkToolkit.fromAssemblyBuilder(
+        async () => {
+          await tsImport(
+            pathToFileURL(this.backendLocator.locate()).toString(),
+            import.meta.url
+          );
+          return new CloudAssembly(this.relativeCloudAssemblyLocation);
```
sobolk, line 228: afaik currently this does run synth (with ipc implementation). should we synth here?

packages/backend-deployer/src/cdk_deployer_singleton_factory.ts
```diff
@@ -55,18 +57,37 @@ export class BackendDeployerFactory {
...
   getInstance(): BackendDeployer {
     if (!BackendDeployerFactory.instance) {
+      let profile = undefined;
+      if (process && process.argv) {
+        for (let i = 2; i < process.argv.length; i++) {
+          if (process.argv[i] == '--profile') {
+            profile = process.argv[i + 1];
+          }
+        }
+      }
```
sobolk, line 76: this is temporary ?

We should `getInstance(props? : {profile?:string})`

packages/backend/src/project_environment_main_stack_creator.ts
```diff
@@ -14,7 +14,13 @@ export class ProjectEnvironmentMainStackCreator implements MainStackCreator {
   constructor(
     private readonly scope: Construct,
     private readonly backendId: BackendIdentifier
-  ) {}
+  ) {
+    process.on('message', (message) => {
+      if (message === 'amplifySynth') {
+        (scope as App).synth({ errorOnDuplicateSynth: false });
```
sobolk, line 20: Should this be in `createDefaultStack` ?

packages/cli-core/API.md
```diff
@@ -479,17 +522,24 @@ export class PackageManagerControllerFactory {
 // @public
 export class Printer {
     constructor(minimumLogLevel: LogLevel, stdout?: WriteStream | NodeJS.WritableStream, stderr?: WriteStream | NodeJS.WritableStream, refreshRate?: number, enableTTY?: boolean);
+    clearConsole: () => void;
+    // (undocumented)
+    readonly enableTTY: boolean;
```
sobolk, line 527:
```suggestion
    readonly ttyEnabled: boolean;

```

packages/cli-core/src/loggers/amplify_io_events_bridge.ts
```diff
@@ -0,0 +1,52 @@
...
+   * Receive amplify events and fan out notifying anyone interested.
+   */
+  notify<T>(msg: AmplifyIoHostEventMessage<T>): Promise<void> {
+    if (!this.eventHandlers.notify) {
+      return Promise.resolve();
+    }
+    const promises: Promise<void>[] = this.eventHandlers.notify?.flatMap(
+      (handler) => handler(msg)
+    );
+    return Promise.any(promises);
```
sobolk, line 33: should this be `all` ? 

i.e. why would some handler be special?

packages/cli-core/src/loggers/amplify_io_events_bridge.ts
```diff
@@ -0,0 +1,52 @@
...
+  async requestResponse<T, U>(
+    msg: AmplifyIoHostEventRequestMessageIoRequest<T, U>
+  ): Promise<U> {
+    if (!this.eventHandlers.requestResponse) {
+      return Promise.resolve(msg.defaultResponse);
+    }
+    const promises: Promise<U>[] = this.eventHandlers.requestResponse.flatMap(
+      (handler) => handler(msg)
+    );
+    // return the first undefined response if available, else the default response
```
sobolk, line 48:
```suggestion
    // return the first defined response if available, else the default response
```

packages/cli-core/src/loggers/amplify_io_events_bridge.ts
```diff
@@ -0,0 +1,52 @@
...
+    const response = (await Promise.all(promises)).find((response) => response);
+    return response ?? msg.defaultResponse;
```
sobolk, line 50: This would be kind of random if there are many handlers and I don't see any example call sites.

packages/cli-core/API.md
```diff
@@ -6,10 +6,48 @@
...
+}
+
+// @public (undocumented)
+export const amplifyIOEventsBridgeFactory: AmplifyIOEventsBridgeSingletonFactory;
+
+// @public
+export class AmplifyIOEventsBridgeSingletonFactory {
+    constructor();
+    getInstance: () => AmplifyIOEventsBridge;
+}
```
sobolk, line 49: Does `AmplifyIOEventsBridge` have to be public?
I.e. could factory return `AmplifyIOHost` ?

packages/cli-core/API.md
```diff
@@ -6,10 +6,48 @@
...
 
+import { AmplifyIOHost } from '@aws-amplify/plugin-types';
+import { AmplifyIoHostEventMessage } from '@aws-amplify/plugin-types';
+import { AmplifyIoHostEventRequestMessageIoRequest } from '@aws-amplify/plugin-types';
 import { PackageManagerController } from '@aws-amplify/plugin-types';
 import { WriteStream } from 'node:tty';
 import z from 'zod';
 
+// @public
+export class AmplifyEventLogger {
```
sobolk, line 17: can this be package internal ?

sobolk (commented):
Overall looks good.

packages/backend-deployer/package.json
```diff
@@ -23,10 +23,12 @@
     "@aws-amplify/plugin-types": "^1.8.0",
     "execa": "^9.5.1",
     "tsx": "^4.6.1",
-    "strip-ansi": "^6.0.1"
+    "strip-ansi": "^7.1.0",
+    "@aws-cdk/toolkit-lib": "0.1.4"
   },
   "peerDependencies": {
-    "aws-cdk": "^2.168.0",
+    "aws-cdk": "^2.1000.2",
```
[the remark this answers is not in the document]
Amplifiyer: It's one of the things left. It touches a lot of places and so I'm thinking to remove it as a followup.

packages/backend-deployer/package.json
```diff
@@ -23,10 +23,12 @@
     "@aws-amplify/plugin-types": "^1.8.0",
     "execa": "^9.5.1",
     "tsx": "^4.6.1",
-    "strip-ansi": "^6.0.1"
+    "strip-ansi": "^7.1.0",
+    "@aws-cdk/toolkit-lib": "0.1.4"
```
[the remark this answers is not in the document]
Amplifiyer: >If there's relationship between toolkit version and cdk-lib version

There isn't. And they have actually disconnected those two versions as well. Moreover, I don't think customers would be interfacing with the toolkit, if they need they can execute their app with cdk cli.

packages/backend-deployer/src/cdk_deployer.ts
Amplifiyer: we don't have access to the `app` here and `synth` only exists on the `app` interface.

sobolk: I mean `amplifySynth` via events.

To test if we need this we can:
1. `npx ampx sandbox --once`
2. Nuke `.amplify` folder (or have fresh copy of the project by any other mean)
3. `npx ampx sandbox delete` <-- see if this works or not.

packages/backend-deployer/src/cdk_deployer_singleton_factory.ts
Amplifiyer: I'm having difficulty patching this correctly. `profile` is an input to the `CdkToolkit` constructor. However `profile` is only available at command runtime while we are injecting/creating `CdkToolkit` in the factories instantiation.

sobolk: We can have logic like this, we do this here for example https://github.com/aws-amplify/amplify-backend/blob/32bcc0bde671937151b078f3e6cb58d608642a26/packages/cli-core/src/printer.ts#L3-L5 .

But this should move to cli package I would say, the argv access.

packages/backend-deployer/src/cdk_deployer.ts
Amplifiyer: ohh I misunderstood, I thought this comment was in `deploy` method. Yeah I'll test and add if synth is needed here.

packages/backend-deployer/src/cdk_deployer_singleton_factory.ts
Amplifiyer: yeah that was the inspiration. I'll try to bring this value from cli.

packages/cli-core/src/loggers/amplify_io_events_bridge.ts
Amplifiyer: yup, should be all. Updated

packages/cli-core/API.md
Amplifiyer: Yup, I missed that. Updated

packages/cli-core/API.md
Amplifiyer: Yes, updated.

packages/cli-core/src/loggers/amplify_io_events_bridge.ts
Amplifiyer: I was also not sure how to model that. Currently we don't even have a use case for this, i.e. there are no request response loggers. If we were to do that, we would need a runtime validation that only one subscriber is registered else fail.

sobolk: If we don't have a use case then we could just descope this api for now perhaps.

packages/backend-deployer/src/cdk_deployer.ts
Amplifiyer: From node documentation
>Synchronously calls each of the listeners registered for the event named eventName, in the order they were registered, passing the supplied arguments to each. Also, the `emit` is happening in a closure that is not executed until the `toolkit.synth` is called. 


@mrgrain are there any concerns with race conditions here with `App` being synthesized from the event emitted through this closure?

packages/backend-deployer/src/cdk_deployer_singleton_factory.ts
Amplifiyer: I moved this to be passed in the factory constructor as it's not really an instance variable, i.e. we are not going to have multiple instances per profile.

packages/backend/src/project_environment_main_stack_creator.ts
Amplifiyer: Updated

packages/backend-deployer/src/cdk_deployer.ts
sobolk: that should do then, until synth ever becomes async. thanks for checking.

packages/backend-deployer/src/cdk_deployer_singleton_factory.ts
Amplifiyer: A side effect of this change is an API change to `cdk_deployer` that it can't accept `profile` anymore as a runtime argument. Now the question is if it's considered a public facing breaking change for us to do major version release or move instantiation of `CDKToolkit` inside `deploy/destroy` methods.

sobolk: But... It might be impossible to honor method level override with this new design. So let's do MV.

packages/cli-core/src/loggers/amplify_io_events_bridge.ts
Amplifiyer: Updated, the impl will not accept any `requestResponse` handlers

packages/backend-deployer/src/cdk_deployer_singleton_factory.ts
Amplifiyer: Updated.

packages/backend-deployer/API.md
```diff
@@ -15,7 +17,7 @@ export type BackendDeployer = {
 
 // @public
 export class BackendDeployerFactory {
-    constructor(packageManagerController: PackageManagerController, formatter: BackendDeployerOutputFormatter);
+    constructor(packageManagerController: PackageManagerController, formatter: BackendDeployerOutputFormatter, backendDeployerIOHost: AmplifyIOHost, sdkProfileResolver: SDKProfileResolver);
```
sobolk, line 20: This is going to break [these places](https://github.com/search?q=repo%3Aaws-amplify%2Famplify-backend+%22new+BackendDeployerFactory%22&type=code) when newer backend-deployer is installed but other packages remain old.

Some options to solve this:
1. Ignore the problem, this might be a niche scenario.
2. Make new parameters optional and fall back to something that's safe - DX might be crappy but at least it shouldn't crash.
3. Ship MV of `backend-deployer`

sobolk (commented):
LGTM

packages/backend-deployer/API.md
Amplifiyer: This would be a same problem in `printer` as well where some of the APIs have changed (spinner Ids removed)

1. I don't see why customers would be pinning some dependent versions of cli.
2. Most of the wiring is done through `cli-core` and `cli` packages. Also if they have old packages of `platform-core` or `plugin-types` the project won't compile as the types would be missing. So defaults also won't help.
3. This could be a safer option than # 1, however I'm unsure of any downsides of some packages on different MV than the others. If we do go this route, we would be forced to increase MV for other packages as well, such as plugin-types or platform-core

My thinking is we go with # 1 and if we there are customers pinning these versions for whom the project won't compile, we ask them to upgrade/update.

packages/backend-deployer/src/cdk_deployer_singleton_factory.ts
```diff
@@ -1,19 +1,30 @@
...
 import { BackendDeployerOutputFormatter } from './types.js';
+import { Toolkit } from '@aws-cdk/toolkit-lib';
 
 export type DeployProps = {
   secretLastUpdated?: Date;
   validateAppSources?: boolean;
+  /**
+   * @deprecated CDK toolkit now accepts profile only in the constructor instead of at runtime. This param will be removed in next MV
+   * Use `sdkProfileResolver` in `BackendDeployerFactory` instead to set the profile
+   */
```
sobolk, line 19: if we do MV then we can just remove. (and below as well).

sobolk (dismissed):

packages/backend-deployer/src/cdk_deployer_singleton_factory.ts
Amplifiyer: These props are coming from `sandbox` and `cli` package, I don't think we want to do MV for them just yet.

Amplifiyer: https://github.com/aws-amplify/amplify-backend/pull/2503/files#diff-19a3328a512ce2869b263ce83233f5c1c00f3c4506ed966a8fab5a4f204ab3dfR41-R44

rtpascual (dismissed):
LGTM.

Not blocking but noticed `dependency-review` is failing for SPDX expressions and licenses not on our allow list https://github.com/aws-amplify/amplify-backend/actions/runs/13930686107/job/38991411600?pr=2503#step:3:13. Wondering if we should evaluate and add these to our config.

packages/backend-deployer/src/cdk_deployer_singleton_factory.ts
Amplifiyer: Updated the sandbox to have MV and removed profile options.

sobolk (approved):

sobolk (approved):

PR #2535, opened 2025-02-21
pin cdk versions for create amplify

rtpascual:
## Problem

We are not pinning to versions of CDK being used in our repo for create amplify.
We also don't have an automated way to update these pinned versions.

## Changes

- Pin versions of CDK for create amplify taken from `package-lock.json`
- Moved create amplify dependencies to a separate file to contain changes
- Add git hook and command to run update create amplify dependencies script
- Add new job that runs a check which attempts to update create amplify dependencies to match what is in `package-lock.json`, then if there is any git diff it fails.

## Validation

Unit tests and https://github.com/aws-amplify/amplify-backend/pull/2534, which is a PR built on top of this one that has changes to CDK versions in `package-lock.json`.

## Checklist


packages/create-amplify/src/default_packages.ts
```diff
@@ -0,0 +1,12 @@
...
+  '@aws-amplify/backend-cli',
+  'aws-cdk@2.1000.2',
+  'aws-cdk-lib@2.180.0',
+  'constructs@^10.0.0',
+  'typescript@^5.0.0',
+  'tsx',
+  'esbuild',
+];
+
+export const defaultProdPackages = ['aws-amplify'];
```
sobolk, line 12: could this be plain JSON file ?

scripts/components/create_amplify_dep_updater.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  dependenciesToUpdate.forEach((dep) => {
+    // taken from lodash escapeRegExp and used to sanitize dep.name before using it in RegExp
+    // can be replaced by RegExp.escape once it is available
+    const safeDepName = dep.name.replace(/[\\^$.*+?()[\]{}|]/g, '\\$&');
+    const depRegex = new RegExp(`'${safeDepName}.*'`);
+    defaultPackagesContent = defaultPackagesContent.replace(
+      depRegex,
+      `'${dep.name}@${dep.version}'`
+    );
```
sobolk, line 38: if we make `default_packages.json` we could get rid of regexes

.github/workflows/health_checks.yml
```diff
@@ -318,6 +318,30 @@ jobs:
           BASE_SHA: ${{ github.event.pull_request.base.sha }}
           # The dependabot_handler_version_update script needs to add the 'run-e2e' pull request label
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  update_create_amplify_deps:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
```
sobolk, line 323: The automation for this is great idea.

In addition to this we should have:
1. A check that asserts that some of `default_packages.ts` are matching package.lock content. The check should block PR merges. This would be a gate to stop inconsistency, therefore this is a mandatory addition to have I would think.
2. Git hook to short circuit feedback loop to local machine that either does validation or upgrade or both. This is nice to have.

sobolk (commented):
overall looking good, please see comments.

packages/create-amplify/src/default_packages.ts
rtpascual: You're right this can be a plain JSON and will remove the need for regex

.github/workflows/health_checks.yml
rtpascual: 1. I agree there may be inconsistencies in the beginning, but won't the automation and this new check be redundant? Since each time package.lock is updated automation will update cdk versions in `default_packages.ts` accordingly
2. That's a great idea for some nice QOL, I'll look into adding this

scripts/update_create_amplify_deps.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+const baseRef = process.argv[2];
+if (baseRef === undefined) {
+  throw new Error('No base ref specified for update create amplify deps');
+}
+
+if (!ghContext.payload.pull_request) {
+  // event is not a pull request, return early
+  process.exit();
```
sobolk, line 23: can we use `throw` and `return` instead of process exists ?

.github/workflows/health_checks.yml
sobolk: 1. If we were to craft extremely minimalistic automation, then the check would be automated and updates would be left manual. The goal of check is to prevent inconsistency, the goal of update is to reduce operational load on operators, those are different goals. But to your point about redundancy - if this job can be made is such a way that it `update_and_checks_create_amplify_desp` i.e. it either updates and early exists or asserts that versions are in sync as last step then it would fulfill the need for "check".

rtpascual: Added the automated check and the not-so-manual update that runs on git hook and can be run manually for tighter feedback loop.

sobolk (approved):

packages/create-amplify/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": { "rootDir": "src", "outDir": "lib" },
+  "include": ["**/*", "src/**/*.json"],
```
rtpascual, line 4: Leaving a comment here since checks would fail if I left it in this file.

The default include value of "**/*" does not include `.json` extension by default so we have to explicitly add it here for the new `default_packages.json`. See https://www.typescriptlang.org/tsconfig/#include:

> If the last path segment in a pattern does not contain a file extension or wildcard character, then it is treated as a directory, and files with supported extensions inside that directory are included (e.g. .ts, .tsx, and .d.ts by default, with .js and .jsx if [allowJs](https://www.typescriptlang.org/tsconfig/#allowJs) is set to true).

sobolk: Did you try `resolveJsonModule` and loading json via usual `import` ?

rtpascual: Yeah `resolveJsonModule` is true based on `tsconfig.base.json`, issue was the json file wasn't included in the resulting `lib` outDir without this change. But now that I think of it, I should be able to import the json instead of using file system now.

sobolk (approved):

scripts/components/create_amplify_dep_updater.ts
```diff
@@ -0,0 +1,60 @@
...
+    return dependenciesToUpdate.has(depName) ? expectedDepString : depString;
+  });
+
+  if (update) {
+    await fsp.writeFile(
+      defaultPackagesPath,
+      JSON.stringify(
+        {
+          defaultDevPackages: newDevPackages,
+          defaultProdPackages,
```
Amplifiyer, line 53: For the sake of completeness we should run this in both dev and prod packages, or change this method name to `createAmplifyDevDepUpdater`

scripts/components/get_dependencies_from_package_lock.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    // Version may not exist if package is a symbolic link
+    if (dependencyVersion) {
+      // Remove "node_modules/" prefix
+      const dependencyName = key.replace(/^node_modules\//, '');
+      dependencies.push({
+        name: dependencyName,
+        version: dependencyVersion,
+      });
+    }
```
Amplifiyer, line 29: what happens if a dep is hoisted under nested package e.g.
`node_modules/backend-cli/node_modules/aws-cdk` and `node_modules/aws-cdk` but with different versions.

scripts/components/create_amplify_dep_updater.ts
rtpascual: I kept the name as is to open it up to when/if we start pinning prod packages and would need to do the same thing, but we can do the same logic now to prod packages.

scripts/components/get_dependencies_from_package_lock.ts
rtpascual: This replaces only the first `node_modules`, so with your example the result would be
```ts
[
  {
    name: 'backend-cli/node_modules/aws-cdk',
    version: '2.0.0',
  },
  {
    name: 'aws-cdk',
    version: '2.1.0',
  }
]
```
and I believe we only care about the version of `node_modules/aws-cdk` which would be used to update pinned create amplify dependencies.

scripts/components/create_amplify_dep_updater.ts
rtpascual: Done in https://github.com/aws-amplify/amplify-backend/pull/2535/commits/d0aefc00fb2477b5b58512f3263ab1e0096c9cb9

Amplifiyer (approved):

sobolk (approved):

PR #2546, opened 2025-02-27
Seed

ShadowCat567:
## Problem

## Changes

## Checklist


packages/seed/src/auth-seed/mfa_flow.ts
```diff
@@ -0,0 +1,225 @@
...
+    } else if (
+      passwordSignIn.nextStep.signInStep === 'CONFIRM_SIGN_IN_WITH_EMAIL_CODE'
+    ) {
+      let challengeResponse: string;
+      if (!user.signUpChallenge) {
+        challengeResponse = await this.prompter.input({
+          message: `Please input one-time password from EMAIL for ${user.username}:`,
+        });
+      } else {
+        assert.strictEqual(user.mfaPreference, 'EMAIL');
```
ShadowCat567, line 130: EMAIL and SMS have an assert here to ensure they are not using the same `signUpChallenge` as TOTP. I would rather not do this because `mfaPreference` is an optional prop and I want to keep it that way, so I would like suggestions for alternatives to this.

packages/seed/src/auth-seed/mfa_flow.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    assert.strictEqual(
+      signInResult.nextStep.signInStep,
+      'CONFIRM_SIGN_IN_WITH_NEW_PASSWORD_REQUIRED'
+    );
+
+    let passwordSignIn = await this.authApi.confirmSignIn({
+      challengeResponse: user.password,
+      options: {
+        userAttributes: {
```
ShadowCat567, line 40: there must be a better way to assign `userAttributes,` I cannot do `userAttributes = user.userAttributes` because of the camelcase vs snakecase

packages/cli/src/seed-policy-generation/generate_seed_policy_template.ts
```diff
@@ -0,0 +1,66 @@
...
+    });
+  }
+  const userPoolId = clientConfig.auth?.user_pool_id;
+
+  const userpoolOutput = await cognitoIdProvider.send(
+    new DescribeUserPoolCommand({ UserPoolId: userPoolId })
+  );
+  const userpoolArn = userpoolOutput.UserPool?.Arn;
+  // so long as there is an auth resource we should always be able to get an Arn for the userpool
+  assert.ok(userpoolArn);
```
sobolk, line 41: This will translate to `UnknownFault` if happens. Should we throw some `<Known>Fault` instead?

packages/seed/src/auth-seed/auth_api_factory.ts
```diff
@@ -0,0 +1,43 @@
+import { AuthClient } from './auth_client.js';
```
sobolk, line 1: I'd call this file just `auth_api`. I don't see a factory here.

sobolk (commented):
Overall this is looking very good.

Please see comments.

packages/cli/src/seed-policy-generation/generate_seed_policy_template.ts
```diff
@@ -0,0 +1,66 @@
+import { Effect, PolicyDocument, PolicyStatement } from 'aws-cdk-lib/aws-iam';
```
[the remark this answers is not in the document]
ShadowCat567: `generate_seed_policy_template.ts` could also move to the seed folder which already has a dependency on `aws-cdk-lib,` though it fractures the generate-policy command implementation across 2 packages

packages/seed/src/auth-seed/mfa_flow.test.ts
```diff
@@ -0,0 +1,346 @@
...
+
+      await mfaFlow.mfaSignUp(
+        {
+          username: testUsername,
+          password: testPassword,
+          signInAfterCreation: true,
+          signInFlow: 'MFA',
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
+          signUpChallenge: async (totpSetupDetails) =>
+            Promise.resolve({ challengeResponse: challengeResponse }),
```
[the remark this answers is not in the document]
ShadowCat567: signUpChallenge has 2 alternatives for its type, which is why I was considering splitting off the challenges based on MFA flow in one of my initial comments

packages/seed/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "@aws-amplify/auth-construct": "^1.6.0",
+    "@aws-amplify/backend-secret": "^1.1.5",
+    "@aws-amplify/cli-core": "^1.2.3",
+    "@aws-amplify/client-config": "^1.5.5",
+    "@aws-amplify/platform-core": "^1.6.0",
+    "@aws-amplify/plugin-types": "^1.8.0",
+    "@aws-sdk/client-cognito-identity-provider": "^3.624.0",
+    "aws-amplify": "^6.0.16",
+    "aws-cdk-lib": "^2.168.0"
+  }
```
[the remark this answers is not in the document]
ShadowCat567: I think I was doing something in tests with them, but looking at the tests, it does not make sense to keep those dependencies

packages/cli/src/seed-policy-generation/generate_seed_policy_template.ts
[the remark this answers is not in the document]
sobolk: Yep. we should not do that.

However, I looked again at the deps and the `cli` already depends on `aws-cdk-lib` via `plugin-types`. So that voids some of my concerns above.

Some more thoughts:
1. Perhaps we should make `aws-cdk-lib` `devDependency` of cli and use `import type { Effect, PolicyDocument, PolicyStatement } from 'aws-cdk-lib/aws-iam';` . To assure that cdk-lib logic is not used by cli at runtime (types would be fine).
2. If we can't do above we should make sure `aws-cdk-lib` is peerDep to cli.

packages/seed/package.json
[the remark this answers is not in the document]
sobolk: sg. please trim this list to viable minimum.

packages/cli/src/seed-policy-generation/generate_seed_policy_template.ts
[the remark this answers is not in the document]
ShadowCat567: 1. works for some parts but not for others, `PolicyStatement` can use `import type`, but this doesn't work for `Effect` and `PolicyDocument`

packages/seed/src/auth-seed/auth_api_factory.ts
```diff
@@ -0,0 +1,43 @@
...
+} from '../types.js';
+import { ConfigReader } from './config_reader.js';
+
+const authClient = new AuthClient(new ConfigReader());
+
+/**
+ * Creates and signs up a new user
+ * @param newUser - contains properties required to create new user
+ * @returns - Username and Sign up flow used by the new user
+ */
```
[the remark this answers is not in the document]
ShadowCat567: does this only apply to the APIs that utilize APIs from AmplifyJS?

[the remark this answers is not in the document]
sobolk: > does this only apply to the APIs that utilize APIs from AmplifyJS?

correct.

packages/seed/src/auth-seed/mfa_flow.ts
ShadowCat567: how exactly would I be able to use an implementation of `toSnakeCase` (similar to `toScreamingSnakeCase`) do this? 
The problem is that `userAttributes` looks like this: 
```
{
  family_name,
  given_name,
  middle_name
  nickname,
  preferred_username
}
```
and we enforce that our variable names are in camelcase so the equivalent section looks like this in `AuthSignUp.userAttributes`:
```
{
  familyName,
  givenName,
  middleName
  nickname,
  preferredUsername
}
```
and because of the difference in naming convention, I have to destructure `AuthSignUp.userAttributes` when doing assignment to `userAttributes` -- for the short period of time before I ran lint, the assignment that I want to do was possible without destructuring (because `AuthSignUp.userAttributes` also had snakecase properties)

sobolk: Ah I see. This is a compile time problem right?
The `NamingConventions` won't be useful here then - it's runtime conversion.
Please disregard the suggestion. It seems we would need some type transformation like this https://stackoverflow.com/questions/60269936/typescript-convert-generic-object-from-snake-to-camel-case - to satisfy the compiler and the runtime conversion as well to map data. Which would make solution hard to reason about and debug. What you have right now is probably best.

packages/seed/src/auth-seed/mfa_flow.ts
```diff
@@ -0,0 +1,229 @@
...
+      passwordSignIn = await this.authApi.confirmSignIn({
+        challengeResponse: user.mfaPreference,
+      });
+    }
+
+    if (
+      passwordSignIn.nextStep.signInStep === 'CONTINUE_SIGN_IN_WITH_TOTP_SETUP'
+    ) {
+      if (!user.totpSignUpChallenge) {
+        throw new AmplifyFault('MissingTOTPChallengeFault', {
```
Amplifiyer, line 92: AmplifyUserError?

packages/seed/src/auth-seed/mfa_flow.ts
```diff
@@ -0,0 +1,229 @@
...
+   * @param user - properties to sign up user with MFA
+   * @param tempPassword - temporary password that was generated for sign up
+   * @returns - true if user is successfully signed up, false otherwise
+   */
+  mfaSignUp = async (user: AuthSignUp, tempPassword: string) => {
+    assert.strictEqual(user.signInFlow, 'MFA');
+    const signInResult = await this.authApi.signIn({
+      username: user.username,
+      password: tempPassword,
+    });
```
Amplifiyer, line 30: If `CONFIRM_SIGN_IN_WITH_NEW_PASSWORD_REQUIRED` is common among the both flows, can we move it to the `auth_client`?

packages/seed/src/auth-seed/mfa_flow.ts
```diff
@@ -0,0 +1,229 @@
...
+        challengeResponse = challengeOutput.challengeResponse;
+      }
+      const totpSignIn = await this.authApi.confirmSignIn({
+        challengeResponse: challengeResponse,
+      });
+      return totpSignIn.nextStep.signInStep === 'DONE';
+    } else if (
+      signInResult.nextStep.signInStep === 'CONFIRM_SIGN_IN_WITH_SMS_CODE'
+    ) {
+      let challengeResponse: string;
```
Amplifiyer, line 194: This whole thing looks very similar to `mfaSignUp` method. Can we combine?

packages/seed/src/types.ts
```diff
@@ -0,0 +1,76 @@
...
+  username: string;
+};
+
+export type ChallengeResponse = {
+  challengeResponse: string;
+};
+
+export type PasswordSignInFlow = {
+  signInFlow: 'Password';
+  password: string;
```
Amplifiyer, line 28: why don't we do this?

```suggestion
  password: string | BackendSecret;
```

@sobolk did we discuss allowing developers to just pass the secrets reference?

packages/seed/src/auth-seed/mfa_flow.ts
ShadowCat567: most of this is similar except for TOTP, and `signIn` and `signUp` use different types

packages/seed/src/auth-seed/auth_client.ts
```diff
@@ -0,0 +1,190 @@
...
+                resolution: `Create a user called ${user.username}`,
+              },
+              error
+            );
+          } else {
+            throw err;
+          }
+        }
+
+        return signInResult.nextStep.signInStep === 'DONE';
```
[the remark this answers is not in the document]
ShadowCat567: that would probably be better for consistency

ShadowCat567:
> are these types not available from aws-sdk? e.g.
> import { Policy} from '@aws-sdk/client-iam'
`Policy` is available, but none of the other types are accessible

packages/seed/src/auth-seed/mfa_flow.ts
ShadowCat567: The entire persistent password flow is a part of the MFA flow, in theory I could entirely get rid of sign up for Persistent Password and have MFA split off if 'DONE' is not reached by `confirmSignUp`
My main concern about doing this is that Passwordless with come eventually and have an entirely different set up than either MFA or Password and I already have a section of code that will be incompatible with Passwordless and need to be stuck in an if block, this will make the if block larger.

packages/seed/src/types.ts
sobolk: That would have to be different `BackendSecret`. 
The existing one is tightly coupled to `Construct` , see https://github.com/aws-amplify/amplify-backend/blob/aec5e08590195d9eb9a0d73be06fe36fb1a09856/packages/plugin-types/src/backend_secret_resolver.ts#L10 .

We could potentially invent `SeedSecret` as a type/class.
And then have `get/setSecret` absorbed into it.

But the thing is. We agreed to provide APIs to deal with passwords using secure storage, but we don't aim to secure in memory runtime access.

packages/cli/src/seed-policy-generation/generate_seed_policy_template.ts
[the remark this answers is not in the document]
ShadowCat567: I tried, `Policy` is available, but none of the types I am currently use are

packages/seed/src/auth-seed/mfa_flow.ts
ShadowCat567: figured out a way to set this up so I don't have to tear apart signUp for PersistentPassword, it will be coming in the next commit with e2e tests

packages/cli/src/commands/sandbox/sandbox-seed/sandbox_seed_command.ts
```diff
@@ -35,15 +36,24 @@ export class SandboxSeedCommand implements CommandModule<object> {
...
+        stdio: 'inherit',
+        env: {
+          AMPLIFY_BACKEND_IDENTIFIER: JSON.stringify(backendID),
+        },
+      });
+    } catch (err) {
+      printer.stopSpinner();
+      throw err;
+    }
+    printer.stopSpinner();
```
sobolk, line 54:
```suggestion
    } finally {
      printer.stopSpinner();
    }
```

packages/integration-tests/src/test-project-setup/seed_test_project.ts
```diff
@@ -0,0 +1,226 @@
...
+  override async deploy(
+    backendIdentifier: BackendIdentifier,
+    environment?: Record<string, string>,
+  ) {
+    await super.deploy(backendIdentifier, environment);
+
+    const command = execaSync('npx', ['which', 'ampx'], {
+      cwd: this.projectDirPath,
+    }).stdout.trim();
+    const seedPolicyProcess = await execa(
```
sobolk, line 123:
```suggestion
    const seedPolicyProcessResult = await execa(
```

if we await, process already exited.

packages/integration-tests/src/test-projects/seed-test-project/amplify/seed/seed.ts
```diff
@@ -0,0 +1,54 @@
+import { createAndSignUpUser, signInUser } from '@aws-amplify/seed';
+import * as auth from 'aws-amplify/auth';
+import type { Schema } from './../data/resource.js';
+import { Amplify } from 'aws-amplify';
+// @ts-ignore typescript does not like this import
+import outputs from '../../amplify_outputs.json';
```
sobolk, line 6: other option is perhaps `fs.readFile` before calling configure.

packages/integration-tests/src/test-projects/seed-test-project/amplify/seed/seed.ts
```diff
@@ -0,0 +1,54 @@
...
+    // @ts-expect-error altering typing for global to make compiler happy is not worth the effort assuming this is temporary workaround
+    globalThis.crypto = crypto;
+  }
+}
+
+Amplify.configure(outputs);
+
+const dataClient = generateClient<Schema>();
+
+const username1 = 'testUser@testing.com';
```
sobolk, line 26:
```suggestion
const username1 = 'testUser@amazon.com';
```

Just in case.

packages/integration-tests/src/test-projects/seed-test-project/amplify/seed/seed.ts
```diff
@@ -0,0 +1,54 @@
...
+const password1 = 'T3st_Passw0rd*';
```
sobolk, line 27: Would it be possible to generate some part of password? So that we don't hardcode credential in the repo?

For example `T3st_Passw0rd*${Math.random()}` or so.

packages/integration-tests/src/test-project-setup/seed_test_project.ts
```diff
@@ -0,0 +1,226 @@
...
+          }
+        }
+      `,
+    });
+
+    assert.strictEqual(
+      content.data.listTodos.items[0].content,
+      `Todo list item for ${testUsername}`,
+    );
+  }
```
sobolk, line 225: The Todos created with cognito auth should have `owner` field populated.

Could you add `owner` filed to the list call and assert that it's non-empty string?

sobolk (commented):
LGTM.

I left couple of comments.

packages/integration-tests/src/test-projects/seed-test-project/amplify/seed/seed.ts
ShadowCat567: so something like this: 
```
import outputs from '../../amplify_outputs.json';
import { readFile } from 'node:fs/promises';

const outputFile = await readFile('../../amplify_outputs.json', { encoding: 'utf8'});
Amplify.configure(outputFile);
```

sobolk:
```
const outputFile = JSON.pars(await readFile('../../amplify_outputs.json', { encoding: 'utf8'}));
Amplify.configure(outputFile);
```

sobolk (approved):

sobolk (dismissed):

packages/integration-tests/src/test-projects/seed-test-project/amplify/seed/seed.ts
ShadowCat567: the seed e2e tests are failing with `ENOENT: no such file or directory, open '../../amplify_outputs.json'`, switching back to using the import

ShadowCat567: switching back to using `@ts-ignore`. `@ts-expect-error` caused seed tests to hang indefinitely due to `Unused '@ts-expect-error' directive.` during type checks. Looks like for this test the only option is using `@ts-ignore`, however this is not the case for customer applications (in my test project I have successfully used `readFile` and `@ts-expect-error` is only needed when the outputs file does not exist yet)

sobolk: @ShadowCat567 I think the problem with `fs.read` was that CLI is executing with `cwd` in root of test project. So attempt to traverse dir up twice (`../..`) is going to fail.

To work around that we could do something like this https://github.com/aws-amplify/amplify-backend/blob/342397409333b2e4fb4e4388fde037730f8b57d3/packages/backend-auth/src/reference_construct.ts#L38-L40 .

I.e. resolve path relatively to TS file location.

(the version above is "module" style. in case seed is commonjs for some reason the alternative is to use implicit `__dirname`).

packages/integration-tests/src/test-projects/seed-test-project/amplify/seed/seed.ts
```diff
@@ -2,11 +2,14 @@ import { createAndSignUpUser, signInUser } from '@aws-amplify/seed';
 import * as auth from 'aws-amplify/auth';
 import type { Schema } from './../data/resource.js';
 import { Amplify } from 'aws-amplify';
-// @ts-ignore typescript does not like this import
-import outputs from '../../amplify_outputs.json';
 import { generateClient } from 'aws-amplify/api';
+// @ts-ignore this file will not exist until sandbox is created
+// import outputs from '../../amplify_outputs.json';
```
sobolk, line 7: can we remove this now ?

sobolk (approved):

packages/cli/src/commands/sandbox/sandbox-seed/sandbox_seed_command.ts
```diff
@@ -35,15 +36,22 @@ export class SandboxSeedCommand implements CommandModule<object> {
    * @inheritDoc
    */
   handler = async (): Promise<void> => {
+    printer.startSpinner('');
```
Amplifiyer, line 39: Can we do this to make it more descriptive?
```suggestion
    printer.startSpinner('Running seed...');
```

Amplifiyer (dismissed):

sobolk (approved):

Amplifiyer (approved):

PR #2547, opened 2025-02-27
switch to using default_packages.json and add automation

rtpascual:
## Changes

Follow up from https://github.com/aws-amplify/amplify-backend/pull/2544. Part 2 where we switch to using `default_packages.json` and add automation for keeping pinned cdk versions up to date for create amplify.

Original PR: https://github.com/aws-amplify/amplify-backend/pull/2535

## Validation

Unit tests, e2e, and https://github.com/aws-amplify/amplify-backend/pull/2534

## Checklist


Amplifiyer (dismissed):

packages/create-amplify/src/amplify_project_creator.ts
```diff
@@ -4,27 +4,24 @@ import { PackageManagerController } from '@aws-amplify/plugin-types';
...
+import { fileURLToPath } from 'url';
 
 const LEARN_MORE_USAGE_DATA_TRACKING_LINK =
   'https://docs.amplify.aws/react/reference/telemetry';
 
+const defaultPackagesContent = await fsp.readFile(
+  fileURLToPath(new URL('./default_packages.json', import.meta.url)),
+  'utf-8'
+);
+const parsedDefaultPackagesFile = JSON.parse(defaultPackagesContent);
```
rtpascual, line 17: So I switched back to using fsp instead of importing the json file because:

1. import fails with `ERR_IMPORT_ASSERTION_TYPE_MISSING` so add import assertion/attribute
2. now getting `Import attributes/attributes are only supported when the '--module' option is set to 'esnext', 'nodenext', or 'preserve'.` so add @ts-expect-error with explanation
3. code fails to build because compiler option for module is Node16, where JSON modules and import assertions are experimental so the type assertion/attribute are unexpected characters for the import

So we would either have to update `tsconfig.base.json` to use one of the suggested module settings from 2
OR
Use fsp

So I opted for the option which wouldn't affect other packages.

sobolk (approved):

Amplifiyer (approved):

PR #2552, opened 2025-02-28
update naming of storage access policies to also use stack

rtpascual:
## Problem

When we create policies for storage access, the id for each new policy is `storageAccess<stack node children length>`. This works to keep policies unique within a stack, but there is overlap when it comes to reference auth since each role will have a newly created policy overwriting the existing policy.

## Changes

Prepend storage access policy id with stack to stop ids from clashing when using reference auth. This way auth and unauth roles get a new storage access policy when being referenced in another app.

## Validation

Updated E2E test to ensure storage access policies start with storage stack.

## Checklist

### DIFF
--- a/packages/backend-storage/src/storage_access_policy_factory.ts
+++ b/packages/backend-storage/src/storage_access_policy_factory.ts
@@ -62,7 +62,7 @@ export class StorageAccessPolicyFactory {
 
     return new Policy(
       this.stack,
-      `storageAccess${this.stack.node.children.length}`,
+      `${this.stack.toString()}Access${this.stack.node.children.length}`,
       {
         statements,
       }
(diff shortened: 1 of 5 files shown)


packages/backend-storage/src/storage_access_policy_factory.ts
```diff
@@ -62,7 +62,7 @@ export class StorageAccessPolicyFactory {
 
     return new Policy(
       this.stack,
-      `storageAccess${this.stack.node.children.length}`,
+      `${this.stack.toString()}Access${this.stack.node.children.length}`,
```
sobolk, line 65: Should we `this.stack.node.path` to make it explicit? 
https://github.com/aws/constructs/blob/ccc691a5df523a596082715c47f9a6339bc114c5/src/construct.ts#L494

rtpascual: Yeah that works, didn't see `this.stack.node.path`

sobolk (approved):

ShadowCat567 (approved):

PR #2554, opened 2025-02-28
Handle signature mismatch error

sobolk:
## Problem

Signature mismatch errors are misclassified as fautls.

## Changes

Add logic in:
- amplify error (handles generate outputs command for example)
- cdk error mapper (handles deployments)

## Validation

added tests

## Checklist

### DIFF
--- a/.changeset/pink-grapes-flow.md
+++ b/.changeset/pink-grapes-flow.md
@@ -1,0 +1,6 @@
+---
+'@aws-amplify/backend-deployer': patch
+'@aws-amplify/platform-core': patch
+---
+
+Handle signature mismatch error
--- a/package-lock.json
+++ b/package-lock.json
@@ -35466,7 +35466,7 @@
     },
     "packages/ai-constructs": {
       "name": "@aws-amplify/ai-constructs",
-      "version": "1.2.3",
+      "version": "1.2.4",
       "license": "Apache-2.0",
       "dependencies": {
         "@aws-amplify/backend-output-schemas": "^1.4.0",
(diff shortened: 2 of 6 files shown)


Amplifiyer:
Typically `signature mismatch errors` are either caused by clockskew issues or problems with aws-sdk. Invalid/incorrect credentials don't result in signature mismatch error imo.

sobolk:
> Typically `signature mismatch errors` are either caused by clockskew issues or problems with aws-sdk. Invalid/incorrect credentials don't result in signature mismatch error imo.

@Amplifiyer 

This is what is coming from smithy:
```
SignatureDoesNotMatch	The request signature we calculated does not match the signature you provided. 
Check your AWS Secret Access Key and signing method. Consult the service documentation for details.	NULL	
throwDefaultError	node_modules/@smithy/smithy-client/dist-cjs/index.js	835	NULL	node_modules/@smithy/smithy-client/dist-cjs/index.js	
```

But they also have this https://github.com/smithy-lang/smithy-typescript/blob/5aeb781faf8535a352161fbfaaf5f8013fd6a35d/packages/service-error-classification/src/constants.ts#L8-L14 .

So the source of these kind of errors might be both wrong creds (or part of them) OR clock skew.

Imo we should keep it as is and add hint about clock skew to our creds resolution message.

WDYT?

Amplifiyer:
> Check your AWS Secret Access Key and signing method

Yes, this is typically for cases when you are calculating your own SIGV4, but we use aws-sdk for that. We can use some generic resolution messages, like "You can try retrying your request, validate your system time is correctly set and refresh your credentials"

rtpascual (approved):

Amplifiyer (approved):

PR #2560, opened 2025-03-05
Notices

sobolk:
## Changes

This implements notice rendering system:
1. Ability to render notices after deployment or command execution
2. Ability to list notices and acknowledge them (new command group `ampx notices`
3. New concept of "typed configuration file" to support notices tracking and caching.

## Out of scope

I have punted the implementation of `backendComponent` because:
1. It might be easier to do it when pretty sandbox work goes in (shared memory)
2. It will be rather large change, so I want to have it in separate PR.

For the time being we'll refrain from defining `backendComponent` predicates on hosting side.
The implementation is just ignoring them so adding them later should not be breaking.

## Validation

Added e2e test.
Unit tests

## Checklist


rtpascual (commented):
Looks really good

packages/cli/src/notices/notices_controller.ts
```diff
@@ -0,0 +1,111 @@
...
+    params: {
+      includeAcknowledged?: boolean;
+    } & NoticesRendererParams,
+  ): Promise<Array<Notice>> => {
+    const noticesManifest =
+      await this.noticesManifestFetcher.fetchNoticesManifest();
+    let notices = noticesManifest.notices;
+    if (!params.includeAcknowledged) {
+      notices = await this.filterAcknowledgedNotices(notices);
+    }
```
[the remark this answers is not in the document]
sobolk: yeah. I intentionally left `includeAcknowledge` somewhere in filtering logic to later introduce `all` or perhaps more filters some day.

> And on this note, would acknowledged notices be displayed differently from unacknowledged ones 

We could, @josefaidt wdyt?

packages/cli/src/notices/notice_predictes_evaluator.ts
rtpascual, line 1: typo, should be `notice_predicates_evaluator.ts`

packages/cli/src/notices/notice_predictes_evaluator.ts
```diff
@@ -0,0 +1,184 @@
...
+    expectedFrequency: 'command' | 'deployment' | 'once' | 'daily',
+    event: NoticesRendererParams['event'],
+    noticeId: string,
+  ): Promise<boolean> => {
+    const projectName = await this.namespaceResolver.resolve();
+    const tracker = await this.noticesPrintingTrackerFile.read();
+    if (expectedFrequency === 'command') {
+      return event === 'postCommand' || event === 'listing';
+    } else if (expectedFrequency === 'deployment') {
+      return event === 'postDeployment' || event === 'listing';
```
Amplifiyer, line 103: what's a `listing` event? Is it `notices --list`?

packages/cli/src/notices/notice_predictes_evaluator.ts
```diff
@@ -0,0 +1,184 @@
...
+    private readonly _process = process,
+  ) {}
+
+  evaluate = async (
+    notice: Notice,
+    opts: NoticesRendererParams,
+  ): Promise<boolean> => {
+    // If it's post deployment suppress notice by default. So that we show
+    // only notices that are allow listed for deployment.
+    let result = opts.event !== 'postDeployment';
```
Amplifiyer, line 29: I didn't understand this. Why is it not part of the `evaluateFrequency` predicate?

packages/cli/src/notices/notices_files.ts
```diff
@@ -0,0 +1,56 @@
...
+import { noticesManifestSchema } from '@aws-amplify/cli-core';
+
+const acknowledgementFileSchema = z.object({
+  projectAcknowledgements: z.array(
+    z.object({
+      projectName: z.string(),
+      noticeId: z.string(),
+      acknowledgedAt: z.number(),
+    }),
+  ),
```
Amplifiyer, line 12: do we need a separate file just for acknowledgements? Can't we use `notices_printing_tracker`?

packages/cli/src/notices/notices_controller.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  acknowledge = async (noticeId: string) => {
+    const acknowledgementFileContent = await this.acknowledgementFile.read();
+    acknowledgementFileContent.projectAcknowledgements.push({
+      projectName: await this.namespaceResolver.resolve(),
+      noticeId,
+      acknowledgedAt: Date.now(),
+    });
+    await this.acknowledgementFile.write(acknowledgementFileContent);
+  };
```
Amplifiyer, line 59: is this de-duped somehow? i.e. can a customer just acknowledge same notice multiple times?

Amplifiyer (commented):
Overall looks good

packages/cli/src/notices/notice_predictes_evaluator.ts
rtpascual, line 1: Looks like a typo here and with the test file, should be `notice_predicates_evaluator.ts`

packages/cli/src/notices/notices_controller.ts
rtpascual: Also what happens if a customer acknowledges a notice id that does not exist/another string that is not id-like? The former being a bigger issue, if a customer acknowledges a notice by id before it exists then they may never see the notice once it is published.

.github/workflows/health_checks.yml
```diff
@@ -532,6 +532,37 @@ jobs:
...
+      - uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # version 4.1.4
+      - uses: ./.github/actions/setup_node
+        with:
+          node-version: 18
+      - uses: ./.github/actions/restore_build_cache
+        with:
+          node-version: 18
+          cdk-lib-version: ${{ needs.resolve_inputs.outputs.cdk_lib_version }}
+          cdk-cli-version: ${{ needs.resolve_inputs.outputs.cdk_cli_version }}
+      - run: cd packages/cli && npm link
```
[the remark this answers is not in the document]
sobolk: Yes. It was failing without it.

Similarly to these:
https://github.com/aws-amplify/amplify-backend/blob/8f59d16ed66e8a3b4b1c871dfe0858008a7705fd/.github/workflows/health_checks.yml#L592
https://github.com/aws-amplify/amplify-backend/blob/8f59d16ed66e8a3b4b1c871dfe0858008a7705fd/.github/actions/run_with_e2e_account/action.yml#L53

packages/cli/src/notices/notice_predictes_evaluator.ts
sobolk: Correct. perhaps `event` is wrong name. `purpose` might be better.
Any ideas ?

packages/platform-core/src/config/local_configuration_controller_factory.ts
```diff
@@ -32,7 +32,6 @@ export class ConfigurationControllerFactory {
     }
 
     this.controllers[configFileName] = new LocalConfigurationController(
-      'amplify',
       configFileName,
```
rtpascual, line 35: This change will affect customers who opted out of telemetry (they will have to opt in again)

rtpascual (commented):
Looking good, left a few comments.

packages/cli/src/notices/notice_predictes_evaluator.ts
sobolk: I'll think about how to do this cleaner.

The problem I'm trying to solve here is:
1. for listing and postCommand we show a notice by default unless it's suppressed by some predicate. i.e. we start from `true` and check if any predicate yields `false` later. I.e. "we assume notice is supposed to be shown and check if there's reason to not show it"
2. for postDeployment, all notices are suppressed by default. Then either one of non-frequency predicates rules it out or frequency predicate allows it to render. I.e. "we assume notice is hidden unless predicates say otherwise".

The interesting edge case for this is when notice has zero predicates and we're in post-deployment. Such notice should be shown after command but not after deployment. In this case, the logic for frequency predicate is not evaluated at all.

There are couple of ways how to solve it. We could massage existing logic/comments to make it easier.
Or perhaps add frequency postCommand predicate by default if there's none in the list. (I think I like this while writing this comment).

packages/cli/src/notices/notices_files.ts
sobolk: I was going back and forth between 1 file and multiple files.

The reason I extracted ack file is that this is most impactful resource should corruption happen. If it's broken then notices will come back to customer.

I was debating whether to unify printing and cache into `notices_metadata.json` file. Loosing these two has minimal impact.

@Amplifiyer wdyt?

packages/cli/src/notices/notice_predictes_evaluator.ts
Amplifiyer: I think `event` is fine since it seems to answer `when` rather than `where` or `why`. Maybe the `listing` should be `listingNoticesCommand`

packages/cli/src/notices/notices_controller.ts
sobolk: it isn't. good call. I'll fix this.

packages/platform-core/src/config/local_configuration_controller_factory.ts
rtpascual: Ah I missed that

packages/cli/src/notices/notices_controller.ts
sobolk: > Also what happens if a customer acknowledges a notice id that does not exist/another string that is not id-like? The former being a bigger issue, if a customer acknowledges a notice by id before it exists then they may never see the notice once it is published.


that is a good call. I think we should validate that notice exists while acknowledging. but later we don't care (in case we unpublish notice).

packages/cli/src/notices/notices_files.ts
sobolk: one more thing I was looking at is write frequencly.

ack write is rare event. needs customer action

cache or tracker writes are performed more often.

therefore separating write patterns decreases probablity of ack file corruption. (for example if node is killed mid write).

Amplifiyer: sounds good! We can go with two files!

packages/cli/src/notices/notice_predictes_evaluator.ts
Amplifiyer: >Or perhaps add frequency postCommand predicate by default if there's none in the list. (I think I like this while writing this comment).

Yes if we can model that in the manifest/schema even better, or worst as a runtime validation while publishing notice.

sobolk: It makes me think that frequency should be a property on a `notice` not a predicate.

packages/cli/src/notices/notices_controller.ts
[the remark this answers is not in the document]
josefaidt: makes sense to me!

packages/cli/src/notices/notice_predictes_evaluator.ts
Amplifiyer: I can see it either way, similar to how validity period could be a property. These two are less "Project" specific and more metadata/notices related.

packages/cli/src/notices/notices_controller.ts
```diff
@@ -49,18 +52,38 @@ export class NoticesController {
...
-    });
-    await this.acknowledgementFile.write(acknowledgementFileContent);
+    const projectName = await this.namespaceResolver.resolve();
+    const noticesManifest =
+      await this.noticesManifestFetcher.fetchNoticesManifest();
+    const notice = noticesManifest.notices.find((item) => item.id === noticeId);
+    if (!notice) {
+      throw new AmplifyUserError('NoticeNotFoundError', {
+        message: `Notice with id=${noticeId} does not exist.`,
+        resolution: 'Ensure that notice being acknowledged exists.',
```
Amplifiyer, line 62:
```suggestion
        resolution: 'Ensure that notice being acknowledged exists. Run `notices list` to find all the available notices', // or something like that
```

sobolk: done

packages/cli/src/notices/notice_predictes_evaluator.ts
sobolk: Alright, I pulled both frequency and validity period up to notice type.

sobolk: I couldn't find a good logic or modeling that would keep frequency as predicate.

Amplifiyer (approved):
🚀

rtpascual (approved):

PR #2562, opened 2025-03-05
Notices manifest schema

sobolk:
## Problem

## Changes

## Checklist


packages/cli-core/src/notices/notices.ts
```diff
@@ -0,0 +1,41 @@
+import z from 'zod';
+
+const noticePredicateSchema = z.discriminatedUnion('type', [
+  z.object({
+    type: z.literal('packageVersion'),
```
Amplifiyer, line 5: I don't think the `node` version will be captured by this, if not, we should add that as a predicate. What about OS as well?

packages/cli-core/src/notices/notices.ts
```diff
@@ -0,0 +1,41 @@
...
+    command: z.enum(['sandbox', 'pipeline-deploy', 'generate', 'configure']),
+  }),
+  z.object({
+    type: z.literal('errorMessage'),
+    errorMessage: z.string(),
+  }),
+  z.object({
+    type: z.literal('frequency'),
+    frequency: z.enum(['command', 'deployment', 'once', 'daily']),
+  }),
```
Amplifiyer, line 24: Didn't we also want the expiration to be part of the notices?

packages/cli-core/src/notices/notices.ts
```diff
@@ -0,0 +1,41 @@
...
+  title: z.string(),
+  details: z.string(),
+  link: z.string().optional(),
+  predicates: z.array(noticePredicateSchema),
+});
+
+export type Notice = z.infer<typeof noticeSchema>;
+
+export const noticesManifestSchema = z.object({
+  currentNotices: z.array(noticeSchema),
```
Amplifiyer, line 38: why `current` not just `notices`?

packages/cli-core/src/notices/notices_manifest_validator.ts
```diff
@@ -0,0 +1,64 @@
...
+  ) {}
+
+  validate = async (noticesManifest: NoticesManifest): Promise<void> => {
+    const ids = new Set<string>();
+    for (const notice of noticesManifest.currentNotices) {
+      await this.validateNotice(notice);
+      assert.ok(
+        !ids.has(notice.id),
+        `Notice ids must be unique. ${notice.id} is duplicated.`
+      );
```
Amplifiyer, line 28: should we throw AmplifyFault instead?

packages/cli-core/src/notices/notices.ts
```diff
@@ -0,0 +1,41 @@
...
+
+const noticePredicateSchema = z.discriminatedUnion('type', [
+  z.object({
+    type: z.literal('packageVersion'),
+    packageName: z.string(),
+    versionRange: z.string(),
+  }),
+  z.object({
+    type: z.literal('backendComponent'),
+    backendComponent: z.enum(['data', 'auth', 'function', 'storage']),
```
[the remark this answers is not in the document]
sobolk: we can add it.

packages/cli-core/src/notices/notices.ts
sobolk: good ideas, will do.

packages/cli-core/src/notices/notices.ts
```diff
@@ -0,0 +1,41 @@
+import z from 'zod';
+
+const noticePredicateSchema = z.discriminatedUnion('type', [
+  z.object({
+    type: z.literal('packageVersion'),
+    packageName: z.string(),
+    versionRange: z.string(),
```
ShadowCat567, line 7: I have seen some cdk notices with an upper bound and lower bound for the range (EX: Affected versions: cli: >=2.0.0 <=2.1005.0) in these cases, would the version range look like this '>=2.0.0 <=2.1005.0'?

sobolk: Yes. Here's the spec https://docs.npmjs.com/cli/v6/using-npm/semver#ranges .

sobolk: or here as well https://www.npmjs.com/package/semver#ranges

Amplifiyer (approved):

ShadowCat567 (approved):

Issue #2474, opened 2025-01-31
add `addEnvironment` for custom function


ykethan:
### Environment information

```plain text
System:
  OS: macOS 15.2
  CPU: (8) arm64 Apple M1
  Memory: 172.80 MB / 16.00 GB
  Shell: /bin/zsh
Binaries:
  Node: 20.18.1 - ~/.nvm/versions/node/v20.18.1/bin/node
  Yarn: 1.22.22 - /opt/homebrew/bin/yarn
  npm: 10.8.2 - ~/.nvm/versions/node/v20.18.1/bin/npm
  pnpm: 9.15.2 - ~/Library/pnpm/pnpm
NPM Packages:
  @aws-amplify/auth-construct: 1.6.0
  @aws-amplify/backend: 1.14.0
  @aws-amplify/backend-auth: 1.5.0
  @aws-amplify/backend-cli: 1.4.8
  @aws-amplify/backend-data: 1.4.0
  @aws-amplify/backend-deployer: 1.1.15
  @aws-amplify/backend-function: 1.12.1
  @aws-amplify/backend-output-schemas: 1.4.0
  @aws-amplify/backend-output-storage: 1.1.4
  @aws-amplify/backend-secret: 1.1.5
  @aws-amplify/backend-storage: 1.2.4
  @aws-amplify/cli-core: 1.2.3
  @aws-amplify/client-config: 1.5.5
  @aws-amplify/deployed-backend-client: 1.5.0
  @aws-amplify/form-generator: 1.0.3
  @aws-amplify/model-generator: 1.0.12
  @aws-amplify/platform-core: 1.6.0
  @aws-amplify/plugin-types: 1.8.0
  @aws-amplify/sandbox: 1.2.10
  @aws-amplify/schema-generator: 1.2.7
  aws-amplify: 6.12.1
  aws-cdk: 2.176.0
  aws-cdk-lib: 2.176.0
  typescript: 5.7.3
No AWS environment variables
No CDK environment variables
```

### Describe the feature

currently to add environment variables such as GraphQL API endpoint you need to access the environment property using 
```
const env = {
  variables: {
    APPSYNC_API_ENDPOINT: backend.data.graphqlUrl,
  },
};
backend.customAPIFunction.resources.cfnResources.cfnFunction.environment = env;
```

but adding this would override any env variables directly set on the function, to add the GraphQL url or reference other resources you need to override the property

```
// backend.ts
[...]

PR #2598, opened 2025-03-20
add addEnvironment for provided functions

rtpascual:
## Problem

Currently `addEnvironment` does not exist for custom functions.

**Issue number, if available:** https://github.com/aws-amplify/amplify-backend/issues/2474

## Changes

- Adds `addEnvironment` for custom functions, for secrets the value will be replaced with placeholder text so the secret value isn't visible until runtime.
- Change shape of `AMPLIFY_SSM_ENV_CONFIG` to one that makes more sense as we will be documenting how to resolve secrets from SSM from the paths contained in it

Out of scope and would have to be done manually:
- Secret value fetching
- Add permission for the Lambda function to get parameters from SSM for secrets that don't use `secret` (e.g. secrets made through CDK)

The reason being to account for secrets with custom parameter paths that may not match the pattern we expect from our `BackendSecretResolver`.

For example:
```ts
const nodeFunction = defineFunction((scope) => {
  return new NodejsFunction(scope, 'nodeFunction', {
    ...
  });
});

const backend = defineBackend({
  nodeFunction
});

backend.nodeFunction.addEnvironment('TEST_ENV', 'test env value');
backend.nodeFunction.addEnvironment('TEST_SECRET', secret('secretValue'));
```
In Lambda console the environment variables and their values will be:
```
AMPLIFY_SSM_ENV_CONFIG: {
  TEST_SECRET: {
    path: <amplify branch/sandbox path>,
    sharedPath: <amplify shared path>,
  }
}
TEST_ENV: 'test env value'
TEST_SECRET: <placeholder text>
```

## Validation

Unit and E2E tests

## Checklist


sobolk (commented):
> The reason being to account for secrets with custom parameter paths that may not match the pattern we expect from our BackendSecretResolver.

Could you share example of what custom parameter path is and how this wouldn't play with expected patterns?

I would think that accepting `secret('foo')` in custom function could result in return of `BackendSecret.resolvePath` being populated into environment variable.

packages/backend-function/src/provided_function_factory.test.ts
```diff
@@ -0,0 +1,148 @@
...
+  void it('allows adding environment variables that are secrets after defining the function', () => {
+    const functionFactory = providedNodeLambda;
+    const lambda = functionFactory.getInstance(getInstanceProps);
+    lambda.addEnvironment('testSecret', new TestBackendSecret('secretValue'));
+    const template = Template.fromStack(lambda.stack);
+    template.resourceCountIs('AWS::Lambda::Function', 1);
+    template.hasResourceProperties('AWS::Lambda::Function', {
+      Environment: {
+        Variables: {
+          testSecret: '<value will be resolved during runtime>',
```
sobolk, line 119: I wonder if we should descope `secret` from custom function `addEnvironment` and just handle plain strings for now.

With this one can build custom logic to use SSM (or other secure storage) on both ends (IaC and runtime).

rtpascual: The more I think of it the more I'm inclined to go this route. Incorporating `secret` I believe should mean that we would also have to consider custom CDK secrets at some point (function is made with cdk why can't the secrets used in it kind of scenario)

sobolk:
> we add an env var that contains the paths similar to AMPLIFY_SSM_ENV_CONFIG (which will need more to handle custom CDK secret case mentioned above)

I would think we want something like that because we would make it work with `sandbox secret set` .
If customer wants to go beyond that then I guess they must glue some cdk primitives with addEnv on function L2 reference.

sobolk:
This is also possible.

```
import {defineFunction} from "@aws-amplify/backend";
import {NodejsFunction} from "aws-cdk-lib/aws-lambda-nodejs";
import path from "node:path";
import * as secretsmanager from 'aws-cdk-lib/aws-secretsmanager';

const filename = fileURLToPath(import.meta.url);
const dirname = path.dirname(filename);

export const func = defineFunction((scope) => {

    const templatedSecret = new secretsmanager.Secret(scope, 'TemplatedSecret', {
        generateSecretString: {
            secretStringTemplate: JSON.stringify({ username: 'postgres' }),
            generateStringKey: 'password',
            excludeCharacters: '/@"',
        },
    });

    const nodeJsFunction = new NodejsFunction(scope, 'myFunc', {
        entry: path.join(dirname, 'handler.ts')
    });

    nodeJsFunction.addEnvironment('test', templatedSecret.secretArn)
    return nodeJsFunction;
});


```

rtpascual:
> I would think we want something like that because we would make it work with sandbox secret set .
If customer wants to go beyond that then I guess they must glue some cdk primitives with addEnv on function L2 reference.

I see that makes sense. So instead of descoping all secrets we would include `secret` and either do something like the following for `backend.providedFunction.addEnvironment('foo', secret('bar'))`:
- Replace value of `foo` with result of `BackendSecret.resolvePath`
```
'foo': {
  path: '/amplify/path/foo',
  sharedPath: '/amplify/shared/path/foo'
}
```
- Add new env var (possibly `AMPLIFY_SSM_ENV_CONFIG`?) that contains all paths (if we copy shape of `AMPLIFY_SSM_ENV_CONFIG` this might be hard to parse)

> This is also possible.
> ...
Yeah we may advise something like this for CDK secrets and this would be handled as plain string.

packages/backend-function/src/function_env_translator.ts
```diff
@@ -88,31 +88,29 @@ export class FunctionEnvironmentTranslator {
...
-        this.lambda.addEnvironment(key, this.ssmValuePlaceholderText);
-        this.ssmEnvVars[branchSecretPath] = {
-          name: key,
-          sharedPath: sharedSecretPath,
-        };
-        this.ssmPaths.push(branchSecretPath, sharedSecretPath);
-      }
-      this.amplifyBackendEnvVarNames.push(key);
+    if (key === amplifySsmEnvConfigKey) {
+      throw new Error(
```
Amplifiyer, line 92: AmplifyUserError

packages/backend-function/src/function_env_translator.ts
```diff
@@ -88,31 +88,29 @@ export class FunctionEnvironmentTranslator {
...
+    } else if (typeof value === 'string') {
+      this.lambda.addEnvironment(key, value);
+    } else {
+      const { branchSecretPath, sharedSecretPath } =
+        this.backendSecretResolver.resolvePath(value);
+      this.lambda.addEnvironment(key, ssmValuePlaceholderText);
+      this.ssmEnvVars[key] = {
+        path: branchSecretPath,
+        sharedPath: sharedSecretPath,
+      };
```
Amplifiyer, line 110: Is there any possibility of breaking existing customers who might be relying on the shape of this env var?

Amplifiyer (approved):

packages/integration-tests/src/test-project-setup/circular_dep_data_func.ts
```diff
@@ -62,6 +78,8 @@ class CircularDepDataFuncTestProject extends TestProjectBase {
     import.meta.url,
   );
 
+  private readonly testSecretNames = ['amazonSecret'];
```
sobolk, line 81: I would think that `advanced_auth_and_functions` test project would be better place to host this new coverage.

sobolk (commented):
lgtm.

Amplifiyer (approved):

packages/integration-tests/src/test-project-setup/circular_dep_data_func.ts
rtpascual: This was just to ensure E2E that secrets work with `addEnvironment`. I believe this still belongs here as adding graphql endpoint as an environment variable to custom function can introduce circular dependency (encountered in a previous iteration before adding `resourceGroupName: 'data'` to the custom function).

sobolk: sg.

sobolk (approved):

rtpascual:
> To double check. it might be worth running a simple experiment where we:
...

Makese sense. I confirmed that redeploying to same function with new code works and the function is still able to resolve secrets. Good thing the change to `AMPLIFY_SSM_ENV_CONFIG` and SSM shim are in the same package so there's no risk of one of them being out of sync.

Issue #2559, opened 2025-03-03
Using the `name` property on `defineAuth` results in doubled name string


chrisl777:
### Environment information

```plain text
System:
  OS: macOS 15.3.1
  CPU: (14) arm64 Apple M3 Max
  Memory: 82.42 MB / 36.00 GB
  Shell: /bin/zsh
Binaries:
  Node: 20.18.3 - ~/.asdf/installs/nodejs/20.18.3/bin/node
  Yarn: 1.22.22 - ~/.asdf/shims/yarn
  npm: 10.8.2 - ~/.asdf/installs/nodejs/20.18.3/bin/npm
  pnpm: undefined - undefined
NPM Packages:
  @aws-amplify/ai-constructs: 1.2.4
  @aws-amplify/auth-construct: 1.6.0
  @aws-amplify/backend: 1.14.1
  @aws-amplify/backend-ai: Not Found
  @aws-amplify/backend-auth: 1.5.0
  @aws-amplify/backend-cli: 1.4.11
  @aws-amplify/backend-data: 1.4.0
  @aws-amplify/backend-deployer: 1.1.18
  @aws-amplify/backend-function: 1.12.2
  @aws-amplify/backend-output-schemas: 1.4.0
  @aws-amplify/backend-output-storage: 1.1.4
  @aws-amplify/backend-secret: 1.1.6
  @aws-amplify/backend-storage: 1.2.4
  @aws-amplify/cli-core: 1.3.0
  @aws-amplify/client-config: 1.5.7
  @aws-amplify/data-construct: 1.15.0
  @aws-amplify/data-schema: 1.18.0
  @aws-amplify/deployed-backend-client: 1.5.1
  @aws-amplify/form-generator: 1.0.4
  @aws-amplify/model-generator: 1.0.12
  @aws-amplify/platform-core: 1.6.3
  @aws-amplify/plugin-types: 1.8.0
  @aws-amplify/sandbox: 1.2.11
  @aws-amplify/schema-generator: 1.2.7
  aws-amplify: 6.13.2
  aws-cdk: 2.1001.0
  aws-cdk-lib: 2.181.1
  typescript: 5.8.2
No AWS environment variables
No CDK environment variables
```

### Describe the bug

We're looking to name the auth resource (among other resources), in order to distinguish between different sandboxes and our other environments as well. 

```
[...]

PR #2605, opened 2025-03-24
fix: auth name issue #2559

daikimatsura:
## Problem

The Auth Construct in Amplify Backend Gen 2 was not explicitly setting the Cognito UserPool name, causing inconsistency between the construct ID and the actual AWS resource name. This created issues particularly when deploying to sandbox environments where resource naming consistency is important.

**Issue number, if available:** #2559

## Changes

- Modified the `AmplifyAuth` class in `packages/auth-construct/src/construct.ts` to explicitly set the `userPoolName` property to match the construct's `name` property
- This ensures consistent naming between the construct ID and the actual AWS Cognito UserPool resource
- The change is minimal and focuses only on addressing the naming inconsistency

## Validation

- Tested the fix by deploying to a sandbox environment using the command `npx @aws-amplify/backend-cli sandbox --profile sandbox-matsura --once`
- Verified that the UserPool resource is created with the expected name
- Confirmed that the deployment completes successfully and all resources use consistent naming
image

## Checklist


packages/auth-construct/src/construct.ts
```diff
@@ -155,11 +155,10 @@ export class AmplifyAuth
...
     this.computedUserPoolProps = this.getUserPoolProps(props);
 
-    this.userPool = new cognito.UserPool(
-      this,
-      `${this.name}UserPool`,
-      this.computedUserPoolProps,
-    );
+    this.userPool = new cognito.UserPool(this, `${this.name}UserPool`, {
+      ...this.computedUserPoolProps,
+      ...(this.name ? { userPoolName: this.name } : {}),
```
sobolk, line 160: Can you move this logic to `getUserPoolProps` and use `props.name` value there as source?

daikimatsura: @sobolk 
Thank you for your review. 
I've made the requested changes by moving the userPoolName property to the getUserPoolProps function. 
I've also re-run the specific tests for auth-construct, and all 92 tests are passing successfully. 
This confirms that the fix is working as expected. Please let me know if you would like any further adjustments.

PR #2637, opened 2025-03-28
Seed uses tsImport instead of execa

ShadowCat567:
## Problem

When seed used `execa`, we would not see the actual error message that were thrown to the customer while seed was running. instead all we would see is something like: `"message": "ExecaError: Command failed with exit code 1: tsx amplify/seed/seed.ts"`.
This appears as an unknown fault in our telemetry and is unhelpful for debugging the actual issue the customer may have run into.

## Changes

Switches seed from using `execa` to run the seed script to using `tsImport`. `tsImport` allows us to run the seed script in the same process as the seed command so any error messages that appear will go through `stderr`.

## Validation

Tested manually to ensure error messages still appear when running seed. Ensured seed command tests still operate correctly.

## Checklist

### DIFF
--- a/packages/cli/src/commands/sandbox/sandbox-seed/sandbox_seed_command.ts
+++ b/packages/cli/src/commands/sandbox/sandbox-seed/sandbox_seed_command.ts
@@ -2,7 +2,7 @@ import { Argv, CommandModule } from 'yargs';
 import path from 'path';
 import { existsSync } from 'fs';
 import { SandboxBackendIdResolver } from '../sandbox_id_resolver.js';
-import { AmplifyUserError } from '@aws-amplify/platform-core';
+import { AmplifyError, AmplifyUserError } from '@aws-amplify/platform-core';
 import { SandboxCommandGlobalOptions } from '../option_types.js';
 import { format, printer } from '@aws-amplify/cli-core';
 import { tsImport } from 'tsx/esm/api';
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/sandbox/sandbox-seed/sandbox_seed_command.ts
```diff
@@ -39,13 +40,28 @@ export class SandboxSeedCommand implements CommandModule<object> {
...
+        throw new AmplifyUserError(
+          'SeedScriptNotFoundError',
+          {
+            message: `There is no file that corresponds to ${seedPath}`,
+            resolution: `Please make a file that corresponds to ${seedPath} and put your seed logic in it`,
+          },
+          error,
+        );
+      } else {
+        throw e;
```
Amplifiyer, line 61: We should still wrap in UserError and throw. This error is thrown from user land code which we don't control. However, if the caught error is already an AmplifyError (which could be fault coming from our libraries) then we don't wrap and throw as is

ShadowCat567: so I would have something like this: 
```
if(error.name === 'Error') {
  throw new AmplifyUserError('SeedError', {
    message: "...",
    resolution: "..."},
    error);
} else {
  // this is AmplifyFault or error we wrap somewhere else
  throw e
}
```

Amplifiyer: something like this

https://github.com/aws-amplify/amplify-backend/blob/ece77e73cf3e4e311cba6fbb315bdea1bc9cab1c/packages/backend-data/src/factory.ts#L239-L252

sobolk (approved):

rtpascual (approved):

Issue #2569, opened 2025-03-10
Type error when trying to bring in your own existing bucket using backend.addOutput


ashika112:
### Environment information

```plain text
System:
  OS: macOS 15.3.1
  CPU: (12) arm64 Apple M2 Pro
  Memory: 179.63 MB / 32.00 GB
  Shell: /bin/zsh
Binaries:
  Node: 22.12.0 - ~/.nvm/versions/node/v22.12.0/bin/node
  Yarn: 1.22.21 - /opt/homebrew/bin/yarn
  npm: 10.9.0 - ~/.nvm/versions/node/v22.12.0/bin/npm
  pnpm: undefined - undefined
NPM Packages:
  @aws-amplify/auth-construct: 1.6.0
  @aws-amplify/backend: 1.14.1
  @aws-amplify/backend-ai: Not Found
  @aws-amplify/backend-auth: 1.5.0
  @aws-amplify/backend-cli: 1.4.11
  @aws-amplify/backend-data: 1.4.0
  @aws-amplify/backend-deployer: 1.1.17
  @aws-amplify/backend-function: 1.12.2
  @aws-amplify/backend-output-schemas: 1.4.0
  @aws-amplify/backend-output-storage: 1.1.4
  @aws-amplify/backend-secret: 1.1.6
  @aws-amplify/backend-storage: 1.2.4
  @aws-amplify/cli-core: 1.2.4
  @aws-amplify/client-config: 1.5.7
  @aws-amplify/data-construct: 1.15.0
  @aws-amplify/data-schema: 1.18.0
  @aws-amplify/deployed-backend-client: 1.5.0
  @aws-amplify/form-generator: 1.0.4
  @aws-amplify/model-generator: 1.0.12
  @aws-amplify/platform-core: 1.6.3
  @aws-amplify/plugin-types: 1.8.0
  @aws-amplify/sandbox: 1.2.11
  @aws-amplify/schema-generator: 1.2.7
  aws-amplify: 6.13.1
  aws-cdk: 2.1001.0
  aws-cdk-lib: 2.181.0
  typescript: 5.7.3
No AWS environment variables
No CDK environment variables
```

### Describe the bug

When trying to add storage info through backend.addOutput a type error occurs when you bring in info of usergroups.

Screenshot of error site provided below

<img width="903" alt="Image" src="https://github.com/user-attachments/assets/c8437a8c-41a7-4c0d-bfa6-8a641aa4f9e8" />

Error message

```
[...]

PR #2696, opened 2025-04-07
Resolve storage output typing issues

ShadowCat567:
## Problem

Typing for storage client config is different from typing for storage outputs, which is unexpected behavior. See issue for additional context.

**Issue number, if available:** #2569

## Changes

Introduces a new version of the client config (v1.4) which has typing for storage that matches the typing generated by outputs for storage. 
Updates typing for zod schema of storage to be closer to storage outputs and storage client config.
Updates default client config to 1.4 and updates tests to match.
Adds new section to post deployment e2e test that asserts the outputs generated by an application match the expected typing.

## Validation

Tested manually to ensure `backend.addOutput()` does not show an error when given expected storage input. 
Expanded e2e tests to ensure typing to outputs matches client config, the new tests pass.

## Checklist


packages/integration-tests/src/test-project-setup/test_project_base.ts
```diff
@@ -213,6 +219,25 @@ export abstract class TestProjectBase {
...
+    const outputFile = await getClientConfigPath(
+      ClientConfigFileBaseName.DEFAULT,
+      dir ?? this.projectAmplifyDirPath,
+      format,
+    );
+    const outputs = JSON.parse(await fsp.readFile(outputFile, 'utf-8'));
+
+    assert.ok(
+      outputs as Partial<clientConfigTypesV1_4.AWSAmplifyBackendOutputs>,
+    );
```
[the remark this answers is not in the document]
Amplifiyer: That's correct, we need runtime validation against the json schema.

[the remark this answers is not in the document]
ShadowCat567: it seems that directly using the json schema for 1.4 does not work with the json schema package because of `tsTypes` -- the json schema package does not know what to do with it and gives errors, any alternate suggestions?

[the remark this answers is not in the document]
ShadowCat567: offline came to the conclusion that if we want to use the json Validator to validate the schema generated by outputs in our e2e tests, we need to do some pre-processing and flip the `additionalProperties` flag to true for `amplify_storage_access_rule`

packages/integration-tests/src/test-projects/custom-outputs/amplify/backend.ts
```diff
@@ -45,3 +45,27 @@ backend.addOutput({
     user_pool_id: fakeCognitoUserPoolId,
   },
 });
+
+const fakeBucketName = 'fakeStorageBucket';
+backend.addOutput({
+  version: '1.4',
+  storage: {
+    aws_region: sampleRegion,
+    bucket_name: fakeBucketName,
```
ShadowCat567, line 54: `bucket_name` and `aws_region` are not auto-generated when using storage with `addOutput`, unlike with Amplify-created storage outputs

packages/backend-output-schemas/src/storage/v1.ts
```diff
@@ -10,13 +10,14 @@ const storageAccessActionEnum = z.enum([
...
-    resource: z.array(storageAccessActionEnum).optional(),
-  }),
+  z.intersection(
+    z.object({
+      guest: z.array(storageAccessActionEnum),
+      authenticated: z.array(storageAccessActionEnum),
+      resource: z.array(storageAccessActionEnum),
+    }),
+    z.record(z.string(), z.array(storageAccessActionEnum)),
+  ),
```
Amplifiyer, line 20: This change might not be forward compatible. i.e. Apps created with new version, when referenced with an older version might fail the zod parsing. Or does the Zod schema not fail at runtime since it was already working with group names that were not part of thois schema.

packages/integration-tests/package.json
```diff
@@ -51,5 +51,8 @@
     "uuid": "^11.1.0",
     "zod": "^3.22.2"
   },
-  "license": "Apache-2.0"
+  "license": "Apache-2.0",
+  "dependencies": {
+    "jsonschema": "^1.5.0"
```
sobolk, line 56: this can be dev dependency

sobolk (approved):

packages/integration-tests/src/test-project-setup/test_project_base.ts
```diff
@@ -213,6 +219,63 @@ export abstract class TestProjectBase {
...
+
+    const validator = new Validator();
+
+    const validSchema = validator.validate(outputs, schema, {
+      preValidateProperty: (object, key) => {
+        const value = object[key];
+
+        // check if we have a storage path, if we do check that it's properties match what we expect to see
+        if (key.endsWith('/*')) {
```
Amplifiyer, line 250: Is there a more deterministic way to determine that this is a storage path then `key.endsWith('/*')` since in future we may add paths like that?

packages/integration-tests/src/test-project-setup/test_project_base.ts
```diff
@@ -213,6 +219,63 @@ export abstract class TestProjectBase {
...
+              guestCounter = 1;
+              // ensure that if we see groups and entity in a property, they appear at the front of the property name
+              // and ensure that the property name is longer than 'groups' or 'entity' (which is where the 6 comes from)
+            } else if (
+              (k.indexOf('groups') == 0 || k.indexOf('entity') == 0) &&
+              k.length > 6
+            ) {
+              continue;
+            } else {
+              assert.fail(`Unexpected key in ${key}`);
```
Amplifiyer, line 270:
```suggestion
              assert.fail(`Unexpected key ${k} in ${key}`);
```

packages/backend-output-schemas/src/storage/v1.ts
Amplifiyer: Updates to this test would be helpful https://github.com/aws-amplify/amplify-backend/blob/90cd755d7a3c169c26640b293dd2bc3b6e39972a/packages/integration-tests/src/test-e2e/amplify_outputs_backwards_compatibility.test.ts#L160

ShadowCat567: for the Zod schema change, I am assuming the latter (will not fail at runtime), since it previously worked with the more restrictive typing and this actually makes the typing looser than what we allow through client config -- since I was not able to find a better way to reference dynamically named keys with zod than through records. But I'll update the test and we'll see what happens

packages/integration-tests/src/test-project-setup/test_project_base.ts
ShadowCat567: a better way may be to use the buckets key and iterate through the buckets if it exists, since it is unlikely that another category will get a buckets key

Amplifiyer (dismissed):

sobolk (approved):

Amplifiyer (approved):

sobolk (approved):

PR #2698, opened 2025-04-09
Emit notices metrics

sobolk:
## Problem

**Specific problem**
We don't know if notice rendering system is heathy (the part that renders notices after command/deployment as best effort).

**Generic problem**
We don't have any way to collect data from pieces of logic that are not happening at the end of process lifespan or end of sandbox deployment.
This is especially affecting activities that are optional/best effort like notices. I.e. they don't disrupt main control flow even if they fail. 

## Changes

1. Added a concept of `UsageDataCollector` - a long living instance that can be injected into logical component and gather data. Some notes here:
   1. I was wondering if this new functionality should be in `UsageDataEmitter` type or not. But decided this should be a new type, so that we inject narrow interface into components. In other words components should not have access to `emitX` interfaces, just to data collection.
   2. I was wondering if `UsageDataCollector` instance should be injected into `UsageDataEmitter` and other components. But decided to make `UsageDataCollector` a property on `UsageDataEmitter` since they are supposed to share lifetime scope (process long). So effectively at implementation layer the "collector" is kind of a view of "emitter" but for the rest of the world they are separate entities and we leave room to refactor implementation if needed.
   3. The implementation assumes Gen1 telemetry event, so mapping collected data into payload is gross. This should be changed in telemetry V2, i.e. telemetry event should be expanded to cover this new functionality (not sure what should go first to main tbh).
2. Added data collection into notices renderer.

## Validation

Added unit tests.

## Checklist

### DIFF
--- a/.changeset/bitter-news-tickle.md
+++ b/.changeset/bitter-news-tickle.md
@@ -1,0 +1,6 @@
+---
+'@aws-amplify/platform-core': minor
+'@aws-amplify/backend-cli': patch
+---
+
+Add ability to collect notices metrics
--- a/packages/cli/src/ampx.ts
+++ b/packages/cli/src/ampx.ts
@@ -41,7 +41,10 @@ attachUnhandledExceptionListeners(usageDataEmitter);
 
 verifyCommandName();
 
-const noticesRenderer = new NoticesRenderer(packageManagerController);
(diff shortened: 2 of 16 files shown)


Amplifiyer:
can this be closed?

PR #2485, opened 2025-02-06
feat: telemetry v2

rtpascual:
## Changes

Introduces new version of telemetry.

### Not in scope
Removal and cleanup of old version of telemetry

### High level overview of changes (what's new in v2):
- Start reporting versions of public amplify backend packages
- Instead of reporting subsequent events in`sandbox` command with `Sandbox`, report as `SandboxEvent` to differentiate more from the `sandbox` command
- Reusing `telemetry.enabled` config so current users don't have to opt out of telemetry again
- localProjectId is a uuid generated with current timestamp and namespace and is saved in `usage_data_preferences.json`
  - Example:
  ```json
    {
      "telemetry": {
        "enabled": false,
      },
      "<path to project without homedir>": {
        "projectId": "<uuid>"
      }
    }
  ```
- Error nesting is reversed (e.g. top level error is most nested error)
- Start reporting on aborted commands (e.g. user force closing on a prompt with `ctrl+c`)
- Deployment times are more accurate by including type checking time

Overall I try to call `telemetryDataEmitter` wherever `usageDataEmitter` is called, only time I don't is for deployments because:
- CDK toolkit provides latency data on successful deployments so `telemetryDataEmitter.emitSuccess` is called in `AmplifyEventLogger`
- For failures, since they are thrown in `CDKDeployer` and latency data is available, `telemetryDataEmitter.emitFailure` is called there

Example emitSuccess payload:
```
{
  "identifiers": {
    "payloadVersion": "1.0.0",
    "sessionUuid": "93f649e1-15c3-43b3-88fd-2697406a3ad4",
    "eventId": "aaa05c30-0a07-4c65-bf45-242009108f22",
    "timestamp": "2025-02-18T00:16:29.000Z",
    "localProjectId": "6bd6a407-cc63-56e5-a701-517e6bd2f876",
    "accountId": "d7f969f3-d9d4-54a5-824d-f7c6c0991b9d",
    "awsRegion": "us-west-1"
  },
  "event": {
    "state": "SUCCEEDED",
    "command": {
      "path": [
        "configure",
        "telemetry",
        "disable"
      ],
      "parameters": [
        ""
      ]
    }
  },
  "environment": {
    "os": {
      "platform": "darwin",
      "release": "24.3.0"
    },
    "shell": "/bin/zsh",
    "npmUserAgent": "npm/7.0.0 node/v15.0.0 darwin x64",
    "ci": false,
    "memory": {
      "total": 34359738368,
      "free": 777535488
    }
  },
[...]

PR #2717, opened 2025-04-16
switch to new version of telemetry

rtpascual:
## Changes

Started from https://github.com/aws-amplify/amplify-backend/pull/2485#discussion_r2036206461.

Use OpenTelemetry API for new version of telemetry.

The payload shape and contents sent does not differ from https://github.com/aws-amplify/amplify-backend/pull/2485, just the way we are collecting data throughout the process using `span`. Config file/telemetry flag does not change so customers who have opted out with `ampx configure telemetry disable` are still opted out.

Two points where spans are created:
- `packages/cli/src/ampx.ts` ("command" span)
  - Captures lifecycle of command executions
- `packages/sandbox/src/file_watching_sandbox.ts` ("sandbox" span)
  - Child span of span created in `packages/cli/src/ampx.ts`
  - Captures lifecycle of sandbox deployments
  - Contains latency data specific to deployments (synthesis, hotSwap, deployment)

So if running `ampx sandbox` and there is an error during deployment, the command span will have `SUCCEEDED` state and the sandbox span will have `FAILED` state and are two separate telemetry payloads.

Tried to make setting span attributes (where all the telemetry data is stored) as type safe as possible as this can be very prone to typos. To solve this there is the new `setSpanAttributesFromObject` helper which, with the `TelemetryPayload` type, can be used like (with intellisense goodness):
```ts
const data: DeepPartial<TelemetryPayload> = {
  event: {
    state: 'SUCCEEDED',
    command: {
      path: ['command1'],
      parameters: ['option1', 'option2'],
    },
  },
  ...
};
setSpanAttributesFromObject(span, data);
```

The span attributes' keys are flattened dot notation of telemetry payload keys. Later these are unflattened, mapped to payload shape, parsed with the zod schema then sent to the endpoint.
```ts
span = {
  attributes: {
    'event.state': 'SUCCEEDED',
    'event.command.path': ['command1'],
    'event.command.parameters': ['option1', 'option2'],
    ...
  },
  ...
}
```

## Validation

Unit tests and local project console logging payload.

## Checklist


rtpascual:
> - packages/sandbox/src/file_watching_sandbox.ts ("sandbox" span)
>  ...
>    - Does NOT have dependency data

We can always match this with telemetry payload from parent span by matching `sessionId`s. Maybe for convenience I will try getting this from the parent span before sending payload for sandbox span.

sobolk (commented):
Overall looking good.

packages/cli/src/ampx.ts
```diff
@@ -8,62 +14,124 @@ import { extractSubCommands } from './extract_sub_commands.js';
...
+    new PackageManagerControllerFactory().getPackageManagerController();
+  const dependencies = await packageManagerController.tryGetDependencies();
+  setSpanAttributesFromObject(span, 'project', {
+    dependencies: JSON.stringify(dependencies),
+  });
 
-const noticesRenderer = new NoticesRenderer(packageManagerController);
-const parser = createMainParser(libraryVersion, noticesRenderer);
-const errorHandler = generateCommandFailureHandler(parser, usageDataEmitter);
+  attachUnhandledExceptionListeners();
```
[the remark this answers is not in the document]
rtpascual: We can probably do that, the recommendation from otel is to configure otel before anything else but that may only be the case when using SDK.

packages/platform-core/API.md
```diff
@@ -222,15 +224,32 @@ export class ParameterPathConversions {
     static toResourceReferenceFullPath(backendId: BackendIdentifier, referenceName: string): string;
 }
 
+// @public
+export const setSpanAttributesFromObject: (span: Span, prefix: TelemetryPayloadKeys, obj: Record<string, any>) => void;
```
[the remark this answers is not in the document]
rtpascual: Yeah and it removes the `any`. My initial thought was to make this generic enough for adding anything to span and later in exporter we pick the ones we care about from the span attributes for the payload.

packages/platform-core/API.md
```diff
@@ -483,20 +505,6 @@ export const typedConfigurationFileFactory: TypedConfigurationFileFactory;
...
-export type UsageDataEmitter = {
-    emitSuccess: (metrics?: Record<string, number>, dimensions?: Record<string, string>) => Promise<void>;
-    emitFailure: (error: AmplifyError, dimensions?: Record<string, string>) => Promise<void>;
-};
-
-// @public
-export class UsageDataEmitterFactory {
-    getInstance: (libraryVersion: string, dependencies?: Array<Dependency>) => Promise<UsageDataEmitter>;
-}
-
```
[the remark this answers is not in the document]
rtpascual: I'll update the changeset.

packages/cli/src/ampx.ts
```diff
@@ -8,62 +14,136 @@ import { extractSubCommands } from './extract_sub_commands.js';
...
+        init: initTime,
+      },
+      project: {
+        dependencies: filteredDependencies,
+      },
+    };
+    setSpanAttributesFromObject(span, data);
+    span.end();
+    // wait a little to try to have span be exported before ending the process
+    await new Promise((resolve) => setTimeout(resolve, 200));
```
rtpascual, line 91: For this and below, I was seeing instances where span ends but is not exported before process ends so we end up with telemetry not being sent. This is a small enough delay to ensure telemetry being sent most of the time without significant delays in the terminal.

packages/cli/src/ampx.ts
```diff
@@ -8,62 +14,136 @@ import { extractSubCommands } from './extract_sub_commands.js';
...
+    });
+  }
 
-verifyCommandName();
+  const packageManagerController =
+    new PackageManagerControllerFactory().getPackageManagerController();
+  const dependencies = await packageManagerController.tryGetDependencies();
+  const filteredDependencies = dependencies?.filter((dep) =>
+    dependenciesToReport.includes(dep.name),
+  );
```
rtpascual, line 64: Had to preemptively filter dependencies here instead of adding all dependencies to the span attributes and filtering in exporter since there is a limit of 128 attributes.

rtpascual: Fixed how this was implemented so now dependencies don't go to payload through span attributes.

packages/cli/src/ampx.ts
Amplifiyer: Does this help https://github.com/open-telemetry/opentelemetry-js-api/issues/164?
If not, can we emit events from the exporter and listen here before exiting? I'm worried that this timeout value would not be consistent across different caliber of machines

packages/cli/src/ampx.ts
sobolk: > Had to preemptively filter dependencies here instead of adding all dependencies to the span attributes

This is good but I think collecting process wide data like deps, node version, os, and so on should be happening outside of span. (In the emitter perhaps). Spans should only be used to collect data from deeply nested call sites.
It might be worth describing that in some comment in the code (if not already explained somewhere).

> there is a limit of 128 attributes.

This worries me a bit. What happens if we reach this limit?
In some dynamic scenarios (like 20 nested `error.cause`) this might become problematic.
It seems we could raise this limit per https://opentelemetry.io/docs/specs/otel/common/#attribute-limits . Does JS implementation support this configuration?

packages/cli/src/extract_command_info.ts
```diff
@@ -0,0 +1,27 @@
...
+    return undefined;
+  }
+  const argv = yargs.parsed.argv;
+  const path = argv._.map((arg) => arg.toString());
+  const parameters: string[] = [];
+  // keys that are in yargs.parsed.argv that we want to filter when getting parameters
+  const ignoredArgvKeys = ['_', '$0'];
+  Object.keys(argv)
+    .filter((key) => !ignoredArgvKeys.includes(key) && argv[key])
+    .sort()
```
Amplifiyer, line 20: why sort?

packages/cli/src/extract_command_info.ts
```diff
@@ -0,0 +1,27 @@
...
+    .forEach((option) => parameters.push(option));
```
Amplifiyer, line 21: I believe some of the parameters may include absoluteFilePaths, would we be removing these later during the exporting?

packages/cli/src/ampx.ts
```diff
@@ -8,62 +14,126 @@ import { extractSubCommands } from './extract_sub_commands.js';
...
+    process.exit(code);
+  };
+  process.on('beforeExit', (code) => void handleAbortion(code));
+  try {
+    await parser.parseAsync(hideBin(process.argv));
+    const metricDimension: Record<string, string> = {};
+    const subCommands = extractSubCommands(parser);
+
+    if (subCommands) {
+      metricDimension.command = subCommands;
```
Amplifiyer, line 97: should this be `metricDimension.subCommands`

packages/platform-core/src/index.ts
```diff
@@ -3,18 +3,21 @@ export * from './backend_entry_point_locator.js';
...
 export {
   ErrorDetails,
   TelemetryPayload,
+  TelemetryPayloadKeys,
   telemetryPayloadSchema,
-} from './telemetry-data/telemetry_data.js';
+} from './telemetry/telemetry_payload.js';
+export * from './telemetry/set_span_attributes_from_object.js';
+export * from './telemetry/translate_error_to_error_details.js';
+export { TELEMETRY_TRACKING_ENABLED } from './telemetry/constants.js';
```
Amplifiyer, line 23:
```suggestion
export { TELEMETRY_ENABLED } from './telemetry/constants.js';
```

packages/platform-core/src/telemetry/get_telemetry_url.ts
```diff
@@ -1,12 +1,11 @@
 import url, { UrlWithStringQuery } from 'node:url';
-import { latestApiVersion } from './constants.js';
 
 let cachedUrl: UrlWithStringQuery;
 
-const prodUrl = `https://api.cli.amplify.aws/${latestApiVersion}/metrics`;
+const prodUrl = `https://telemetry.cli.amplify.aws/metrics`;
```
Amplifiyer, line 5: I was imagining sending telemetry to both endpoints for a while until we have good adoption for new versions carrying new telemetry. If we do a hard switch, we will have a tough time combining the metrics/data.

packages/platform-core/src/telemetry/telemetry_payload_exporter.ts
```diff
@@ -0,0 +1,201 @@
...
+    private readonly sessionId = uuidV4(),
+    private readonly url = getUrl(),
+    private readonly accountIdFetcher = new AccountIdFetcher(),
+    private readonly regionFetcher = new RegionFetcher(),
+  ) {
+    const targetDependencies = [
+      '@aws-amplify/ai-constructs',
+      '@aws-amplify/auth-construct',
+      '@aws-amplify/backend',
+      '@aws-amplify/backend-ai',
```
Amplifiyer, line 40: we should get `aws-amplify` as well for customers building frontend and backend in the same app and it's been known the aws-amplify dependencies causing some issues

Amplifiyer: Also, can't we do `'@aws-amplify/*` match?

packages/sandbox/src/file_watching_sandbox.ts
```diff
@@ -287,31 +292,72 @@ export class FileWatchingSandbox extends EventEmitter implements Sandbox {
...
+              : 0,
+            synthesis: deployResult.deploymentTimes.synthesisTime
+              ? deployResult.deploymentTimes.synthesisTime * 1000
+              : 0,
+          },
+          event: {
+            state: 'SUCCEEDED',
+            command: {
+              path: ['SandboxDeployment'],
+              parameters: [],
```
Amplifiyer, line 318: we can pass the SandboxOptions (anonymized) here

packages/platform-core/src/telemetry/translate_error_to_error_details.ts
```diff
@@ -0,0 +1,40 @@
...
+      const errorDetail: TelemetryPayload['error'] = {
+        name: serializedError.name,
+        message: serializedError.message,
+        stack: serializedError.stack ?? '',
+      };
+
+      if (errorDetails) {
+        // this reverses the nesting so lowest level error is the top level error in our telemetry
+        errorDetail.cause = errorDetails;
+      }
```
Amplifiyer, line 28: If we do the reversal. the `cause` field doesn't make sense. May be
```suggestion
      if (errorDetails) {
        // this reverses the nesting so lowest level error is the top level error in our telemetry
        errorDetail.caused = errorDetails;
      }
```

Amplifiyer (commented):
Overall looking really good with healthy test coverage 💯

packages/cli/src/ampx.ts
sobolk: > Otel doesn't seem to error and just stops adding attributes after hitting the limit 

That might not be a bad strategy. Some data gets emitted without trying to push some insane payload over the network.

Two ways to move forward would be:
1.  bump the limit to high value. (number.max perhaps? or estimate how much bytes we'd like payload to be at most).
2. Have some way to detect that we're hitting the limit. Maybe emitter could report number of attributes when sending the event.

packages/cli/src/ampx.ts
rtpascual: I'll try looking into adding something to the shutdown logic to hopefully "flush" any spans that aren't sent.

packages/cli/src/extract_command_info.ts
rtpascual: This is more for infra side and querying logs to make order of parameters more predictable. Also if we want to add parameters to dashboard (not just commands) then we avoid having a bigger number of possible values based on how customers order their parameters.

packages/cli/src/extract_command_info.ts
rtpascual: So we only capture the different options that are passed but not the values of those options. For example, `ampx generate outputs --out-dir some/dir` we would only capture `out-dir` as a parameter here.

packages/platform-core/src/telemetry/telemetry_payload_exporter.ts
rtpascual: We can and I can adjust the filtering to account for the match, would we want to capture all packages under `@aws-amplify`? Just searching npm comes up with 1000+ results https://www.npmjs.com/search?q=%40aws-amplify.

svidgen (commented):
Haven't gone deep on this; just pointing out one thing that jumps out at me.

packages/platform-core/src/telemetry/telemetry_payload_exporter.ts
rtpascual: Currently I think we can be explicit and add to this list. The list can be moved to `telemetry/constants.ts` if it gets too large.

packages/cli/src/ampx.ts
rtpascual: this is for `DefaultUsageDataEmitter`, https://github.com/aws-amplify/amplify-backend/blob/main/packages/platform-core/src/usage-data/usage_data_emitter.ts#L152.

packages/cli/src/ampx.ts
rtpascual: So `SimpleSpanProcessor` has a `forceFlush` method which we can use to make sure our exporter sends all spans. Removed the arbitrary timeouts

packages/platform-core/API.md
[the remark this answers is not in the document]
rtpascual: Spoke offline but looks like we will be keeping telemetry v1 for now so no major version bump here.

packages/cli-core/src/loggers/amplify_event_loggers.ts
```diff
@@ -202,6 +206,21 @@ export class AmplifyEventLogger {
...
           );
         }
+        const span = trace.getSpan(context.active());
+        if (this.isHotSwap && span) {
+          setSpanAttributesFromObject(span, {
+            latency: {
+              hotSwap: msg.data.duration,
+            },
+          });
+          this.isHotSwap = false;
```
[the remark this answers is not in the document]
rtpascual: Good catch, so for this I'm guessing we notify that deployment fails and use it to reset the flag

packages/cli/src/ampx.ts
rtpascual: Now that I'm thinking of this maybe there's 3rd option that's slightly different from Option 2:

3. Check if span.attributes = limit, then send bare bones payload (so parsing/validations pass) with error that span has reached number of attributes limit

This way we don't risk sending partial data but we won't know how much over the span attribute limit we are unless we keep a counter somewhere on the span to check in exporter to attach to the error message.

sobolk: > This way we don't risk sending partial data.

I don't see a problem with that. Unless payload is rejected by service.

I'm fine with either option. But I would just bump attribute limit to some big number and not worry about it.

packages/platform-core/src/telemetry/region_fetcher.ts
```diff
@@ -0,0 +1,24 @@
...
+    if (this.region) {
+      return this.region;
+    }
+    try {
+      this.region = await this.stsClient.config.region();
+      return this.region;
+    } catch {
+      return;
+    }
+  };
```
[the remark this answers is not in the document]
rtpascual: TIL too. I updated both accountId and region fetchers but I agree the only real way this race condition can be hit is with very quick sandbox deployments (maybe multiple file changes that fail fast). Other times these functions are invoked are during individual command calls where they are only called once per "session".

packages/platform-core/src/telemetry/telemetry_payload_exporter.ts
```diff
@@ -0,0 +1,206 @@
...
+        error: unflattened.error,
+      });
+      return payload;
+    } catch {
+      // Don't propogate errors related to not being able to get telemetry payload
+      return;
+    }
+  };
+
+  // Helper to unflatten dot notation span attributes into telemetry payload
```
sobolk, line 152:
```suggestion
  // Unflattens dot notation span attributes into telemetry payload
```

!!

scripts/check_package_versions.ts
```diff
@@ -12,8 +12,8 @@ const getExpectedMajorVersion = (packageName: string) => {
   switch (packageName) {
     case 'ampx':
       return '0.';
-    case '@aws-amplify/cli-core':
     case '@aws-amplify/backend-deployer':
+    case '@aws-amplify/cli-core':
```
sobolk, line 16: if we keep old toys, we shouldn't do MV.

sobolk: sounds good. I didn't notice this change just became ordering change.

sobolk (approved):

packages/backend-deployer/src/cdk_deployer.ts
```diff
@@ -154,6 +154,14 @@ export class CDKDeployer implements BackendDeployer {
             : HotswapMode.FULL_DEPLOYMENT,
       });
     } catch (error) {
+      await this.ioHost.notify({
+        message: 'Deployment failed',
+        code: 'DEPLOY_FAILED',
+        action: 'amplify',
+        time: new Date(),
+        level: 'debug',
```
Amplifiyer, line 162: nit: this should be `error` in case there are consumers interested in this event in future.

Amplifiyer (approved):

sobolk (approved):

PR #2821, opened 2025-05-20
Temporary fix for canaries

ShadowCat567:
## Problem

Our canaries were running out of heap space.

Our canaries also started running into errors like this: 
```
Error: packages/platform-core/src/config/typed_configuration_file.test.ts(3,19): error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("zod")' call instead.
  To convert this file to an ECMAScript module, change its file extension to '.mts' or create a local package.json file with `{ "type": "module" }`.
```
this most likely comes from the most recent release of `zod`, where they changed their types to use `esm` instead of `commonjs`: https://github.com/colinhacks/zod/commit/13226e033722dcb41dc9dbbf5fd03ab7ecca5042

## Changes

Pinning `zod` fixed the heap space problem as well.

Pins `zod` in `backend-output-schemas` and `platform-core` to the version they are currently using (this is a temporary solution to get us out of alarm, we will need a more permanent solution). Only `backend-output-schemas` and `platform-core` because they are our only `commonjs` packages that use `zod`. 
Part of this change also involves an update to `zod` since the version I was initially trying to pin to has security vulnerabilities so I upgraded to a version that has fixed the vulnerability.

## Validation

Successful canary run: https://github.com/aws-amplify/amplify-backend/actions/runs/15144107604

## Checklist


.github/workflows/canary_checks.yml
```diff
@@ -19,6 +19,10 @@ jobs:
           rm package-lock.json
           ./scripts/retry.js npm install
           npm run build
+        env:
+          # increase v8's old memory section max size to 5 GB
+          # see https://nodejs.org/api/cli.html#--max-old-space-sizesize-in-megabytes
+          NODE_OPTIONS: '--max-old-space-size=5120'
```
Amplifiyer, line 25: Do we need to increase this after pinning the zod?

ShadowCat567: we might not, I'll try removing it and we'll see

Amplifiyer (approved):

sobolk (dismissed):

Amplifiyer (approved):

sobolk (approved):

PR #2868, opened 2025-06-25
[DevTools] Dependencies and Basic Structure

megha-narayanan:
Includes dependencies, very elementary facade of react app, and basic devtools command functionality which just prints a statement instead of opening the app.


.github/dependency_review_config.yml
```diff
@@ -28,3 +28,5 @@ allow-licenses:
   - X11
   - zlib-acknowledgement
   - Zlib
+  - LicenseRef-scancode-unknown-license-reference AND MIT
```
Amplifiyer, line 31: Create a github issue to later fix the issues before we can merge to main.

packages/cli/scripts/build-devtools.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
Amplifiyer: who's invoking this script?

packages/cli/src/commands/sandbox/port_checker.ts
```diff
@@ -0,0 +1,46 @@
...
+    });
+  }
+
+  /**
+   * Checks if DevTools is running by checking if the default DevTools port is in use
+   * @returns A promise that resolves to true if DevTools is running, false otherwise
+   */
+  async isDevToolsRunning(): Promise<boolean> {
+    return this.isPortInUse(PortChecker.defaultDevtoolsPort);
+  }
```
Amplifiyer: Ideally this shouldn't be in the `PortChecker` class. We can either create a wrapper over this for devtools or just use `isPortInUse` at places where we need to determine if devtools is running

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/README.md
```diff
@@ -0,0 +1,54 @@
+# React + TypeScript + Vite
+
+This template provides a minimal setup to get React working in Vite with HMR and some ESLint rules.
+
+Currently, two official plugins are available:
+
+- [@vitejs/plugin-react](https://github.com/vitejs/vite-plugin-react/blob/main/packages/plugin-react) uses [Babel](https://babeljs.io/) for Fast Refresh
+- [@vitejs/plugin-react-swc](https://github.com/vitejs/vite-plugin-react/blob/main/packages/plugin-react-swc) uses [SWC](https://swc.rs/) for Fast Refresh
+
+## Expanding the ESLint configuration
```
Amplifiyer: This file should be removed.

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/index.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
```
Amplifiyer: There seems to be two sets of index.html and index.css. Do we need both?

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/tsconfig.app.json
```diff
@@ -0,0 +1,27 @@
+{
+  "compilerOptions": {
+    "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.app.tsbuildinfo",
+    "target": "ES2020",
+    "useDefineForClassFields": true,
```
Amplifiyer, line 5: do we need all these `tsconfig` files?

packages/cli/src/commands/sandbox/sandbox_command.ts
```diff
@@ -91,12 +93,19 @@ export class SandboxCommand
...
   ): Promise<void> => {
+    printer.log('DEBUG: Starting sandbox handler execution', LogLevel.DEBUG);
     const sandbox = await this.sandboxFactory.getInstance();
+    printer.log('DEBUG: Sandbox instance created', LogLevel.DEBUG);
     this.sandboxIdentifier = args.identifier;
     this.profile = args.profile;
+    printer.log(
+      `DEBUG: Sandbox identifier: ${String(this.sandboxIdentifier)}, profile: ${String(this.profile)}`,
+      LogLevel.DEBUG,
+    );
```
Amplifiyer: Do we need all these logger statements? Also we don't need to prepend `DEBUG` in each statement, it should be done by the printer if required.

packages/cli/package.json
```diff
@@ -15,7 +15,13 @@
     "#package.json": "./package.json"
   },
   "scripts": {
-    "update:api": "api-extractor run --local"
+    "update:api": "api-extractor run --local",
+    "clean": "rm -rf lib && rm -rf src/commands/sandbox/sandbox-devtools/public",
+    "build:devtools": "cd src/commands/sandbox/sandbox-devtools/react-app && npm install && npm run build",
```
Amplifiyer, line 20: when you `cd`, don't you also need to come back to the `cwd`?

packages/cli/package.json
```diff
@@ -15,7 +15,13 @@
...
+    "build:ts": "tsc --build --force",
+    "copy:devtools": "mkdir -p lib/commands/sandbox/sandbox-devtools/react-app/dist && cp -r src/commands/sandbox/sandbox-devtools/react-app/dist/* lib/commands/sandbox/sandbox-devtools/react-app/dist/ || echo 'No React app build found'",
+    "build": "npm run clean && npm run build:devtools && npm run build:ts && npm run copy:devtools && chmod +x lib/ampx.js",
```
Amplifiyer, line 23: why clean every time we build?

packages/cli/package.json
```diff
@@ -15,7 +15,13 @@
...
+    "postinstall": "npm run build:devtools && npm run copy:devtools"
```
Amplifiyer: who's calling/running this command?

packages/cli/src/commands/sandbox/sandbox_event_handler_factory.ts
```diff
@@ -25,6 +25,23 @@ export class SandboxEventHandlerFactory {
...
+            );
+          } catch (error) {
+            printer.print(
+              `${format.error(
+                'Failed to handle resource configuration change.',
+              )} ${format.error(error)}`,
+            );
+          }
+        },
+      ],
```
Amplifiyer: who's emitting this event? Should this be in PR2?

packages/cli/scripts/build-devtools.sh
megha-narayanan: Accidental relic from old build process-- Deleting it.

.github/dependency_review_config.yml
megha-narayanan: Done. See: https://github.com/aws-amplify/amplify-backend/issues/2872#issue-3184105758

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/README.md
megha-narayanan: Done.

packages/cli/src/commands/sandbox/sandbox_event_handler_factory.ts
megha-narayanan: yep, removing this and the corresponding field in the type def

packages/cli/package.json
megha-narayanan: For build, mostly developers (me). The app gets rebuilt often to show changes.  Postinstall is no longer useful, removing it.

packages/cli/package.json
megha-narayanan: Earlier in development build was giving me lots of trouble with changing app assets and file structure, and cleaning every time was the best way to ensure old assets weren't affecting my app. Now, builds are much less frequent, so we can probably take out the clean, but maybe have another script for build:clean that does run clean and run build?

packages/cli/package.json
megha-narayanan: in npm scripts, each command runs in its own shell subprocess, so when the command finishes, that subprocess exits and youre back in the `cwd`

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/index.html
megha-narayanan: It looks like my placeholder files were referencing both, but they can easily be consolidated-- done.

packages/cli/src/commands/sandbox/port_checker.ts
megha-narayanan: got it. going for the second option.  My worry was that if we ever change the devtools port there isn't a clear place that constant should be, but at least for the forseeable future sandbox is the the only process using this, so I'm just defining it there.

packages/cli/src/commands/sandbox/sandbox_command.ts
megha-narayanan: They were useful in development, but I agree they can be deleted now. Cutting a bunch of the excess debug logging.

packages/cli/src/commands/sandbox/sandbox_command.test.ts
```diff
@@ -206,6 +213,9 @@ void describe('sandbox command', () => {
   });
 
   void it('Prints stopping sandbox and instructions to delete sandbox when users send ctrl+c', async (contextual) => {
+    // Ensure isPortInUse returns false for this test
+    //contextual.mock.method(PortChecker.prototype, 'isPortInUse', () => Promise.resolve(false));
```
Amplifiyer: We can remove if not needed.

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/tsconfig.app.json
```diff
@@ -0,0 +1,27 @@
+{
+  "compilerOptions": {
+    "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.app.tsbuildinfo",
```
Amplifiyer, line 3: Is this filepath correct? Also do we need the 4 tsconfig we have in the react-app folder?

Amplifiyer (approved):
Looks good. Just left two non blocking questions.

package.json
```diff
@@ -114,6 +119,6 @@
     "*.md": "prettier --write"
   },
   "overrides": {
-    "minimatch": "10.0.1"
+    "minimatch": "10.0.3"
```
ShadowCat567, line 122: Is there a reason why you need minimatch v10.0.3? 
This was initially pinned to 10.0.1 because of an issue we had with 10.0.2 that broke our canaries and prevented new customers from setting up apps. Using 10.0.3 should be fine because the problem with 10.0.2 was allegedly fixed (https://github.com/isaacs/minimatch/issues/257) but be careful with changing the version of pinned packages

.gitignore
```diff
@@ -22,3 +22,8 @@ concurrent_workspace_script_cache.json
 
 testDir
 /.amplify
+
+# React app build files
```
ShadowCat567, line 26: This may be a stupid question, but why are these folders in the gitignore?

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/index.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" type="image/svg+xml" href="/vite.svg" />
```
rtpascual, line 5: Before we release this feature, we should replace this with an Amplify favicon, maybe this one that I saw from the UI repo https://github.com/aws-amplify/amplify-ui/blob/main/docs/public/favicon.ico.

rtpascual (dismissed):
Looking good, just left a comment that can be addressed in a later PR.

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/index.html
megha-narayanan: will do!

.gitignore
megha-narayanan: I wanted to leave the built files out of the repo, DevTools builds on install and then serves the locally built files.

package.json
Amplifiyer: Some of our dependencies moved to the newer version and so the overrides was causing issues. Upgrading it was the fastest way to unblock for now.

ShadowCat567 (dismissed):

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/tsconfig.app.json
megha-narayanan: Yep, I believe this follows standard architecture. I think the only file which is redundant is src/tsconfig since the one in the outer directory covers it.

Amplifiyer (approved):

PR #2879, opened 2025-07-02
[DevTools] PR2: Backend Services, Console Viewer, and Resources

megha-narayanan:
## Changes

Implements a functional DevTools UI for the Amplify Sandbox, building upon the foundation established in PR1. Key features include:

- Web-based interface for managing sandbox environments
- Real-time monitoring of sandbox status and deployment progress
- Console log viewer with formatted logs
- Resource explorer to view and manage deployed AWS resources
- WebSocket-based communication between UI and backend
- Enhanced sandbox lifecycle management (start, stop, delete)
- Improved error handling and status reporting

The implementation uses Express with Socket.IO for the backend server and React with Cloudscape Design components for the frontend UI.

## Validation

Unit tests have been included, though some files will only be further tested in later PRs

## Checklist


package.json
```diff
@@ -59,13 +59,15 @@
...
     "@aws-amplify/eslint-plugin-amplify-backend-rules": "^0.0.2",
     "@aws-sdk/client-amplify": "^3.750.0",
-    "@aws-sdk/client-cloudformation": "^3.750.0",
-    "@aws-sdk/client-cloudwatch-logs": "^3.750.0",
+    "@aws-sdk/client-cloudformation": "^3.828.0",
+    "@aws-sdk/client-cloudwatch-logs": "^3.835.0",
     "@aws-sdk/client-cognito-identity-provider": "^3.750.0",
     "@aws-sdk/client-dynamodb": "^3.750.0",
-    "@aws-sdk/client-iam": "^3.750.0",
+    "@aws-sdk/client-iam": "^3.835.0",
```
Amplifiyer: Do we need aws-sdk upgrade as part of this? If not, we should decouple.

package.json
```diff
@@ -118,6 +121,16 @@
...
+  "dependencies": {
+    "@aws-sdk/client-lambda": "^3.835.0",
+    "cookie-signature": "1.0.6",
+    "debounce-promise": "^3.1.2",
+    "esbuild": "0.25.5",
+    "express": "^5.1.0",
+    "express-rate-limit": "^7.5.0",
+    "jszip": "^3.10.1",
+    "ws": "^8.18.2"
+  },
```
Amplifiyer: Do we need all these dependencies, and in the root package?

packages/cli/src/commands/sandbox/port_checker.ts
```diff
@@ -1,10 +1,60 @@
...
+   */
+  private static readonly defaultDevtoolsPort = 3333;
+  /**
+   * Attempts to start the server on the specified port
+   * @param server The HTTP server
+   * @param port The port to use
+   * @returns A promise that resolves with the port when the server starts
+   * @throws Error if the port is already in use
+   */
+  async findAvailablePort(
```
Amplifiyer: The method name is not matching the documentation and what it is doing. And why are we starting a server in `PortChecker` class?

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/ConsoleViewer.tsx
```diff
@@ -0,0 +1,208 @@
...
+}
+
+interface ConsoleViewerProps {
+  logs: LogEntry[];
+}
+
+const cleanAnsiCodes = (text: string): string => {
+  // This regex handles various ANSI escape sequences including colors, bold, dim, etc.
+  return text.replace(/\u001b\[\d+(;\d+)*m|\[2m|\[22m|\[1m|\[36m|\[39m/g, '');
+};
```
Amplifiyer: Look into `stripANSI` which is already used in this repo.

packages/cli/src/commands/sandbox/sandbox-devtools/logging/cloudformation_format.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  return name || logicalId;
+};
+
+/**
+ * Normalizes a CDK construct path to create a more readable friendly name
+ * @param constructPath The CDK construct path
+ * @returns A normalized construct path
+ */
+const normalizeCDKConstructPath = (constructPath: string): string => {
```
Amplifiyer: This seems duplicated from here https://github.com/aws-amplify/amplify-backend/blob/146f60396e2e686843d1f957f30b9ca1cea5fe30/packages/cli-core/src/loggers/cfn-deployment-progress/cfn_deployment_progress_logger.ts#L237

We should see if we can refactor `cfn_deployment_progress_logger.ts` such that it can be used for both CLI and DevConsole instead of duplicating the logic

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/hooks/useResourceManager.ts
```diff
@@ -0,0 +1,211 @@
...
+  const [customFriendlyNames, setCustomFriendlyNames] = useState<
+    Record<string, string>
+  >({});
+  const [region, setRegion] = useState<string | null>(null);
+  const [backendName, setBackendName] = useState<string>('');
+
+  useEffect(() => {
+    if (!socket) return;
+
+    socket.emit('getCustomFriendlyNames');
```
Amplifiyer: Sockets usage should be abstracted away from the frontend. This abstraction should expose APIs that frontend can call and/or emit events that frontend can subscribe to.

packages/cli/src/commands/sandbox/sandbox-devtools/resource_console_functions.ts
```diff
@@ -0,0 +1,263 @@
...
+export const canProvideConsoleLink = (
+  resourceType: string,
+  resourceStatus: string,
+): boolean => {
+  const supportedResourceTypes = [
+    'AWS::Lambda::Function',
+    'AWS::Lambda::LayerVersion',
+    // 'AWS::DynamoDB::Table', // COMMENT OUT (UNTESTED)
+    'AWS::S3::Bucket',
+    // 'AWS::ApiGateway::RestApi', // COMMENT OUT (UNTESTED)
```
Amplifiyer: what's pending for these?

packages/cli/src/commands/sandbox/sandbox-devtools/resource_console_functions.ts
```diff
@@ -0,0 +1,263 @@
...
+ */
+export const isGlobalService = (service?: string): boolean => {
+  if (!service) return false;
+  return ['iam', 'cloudfront', 'route53'].includes(service.toLowerCase());
+};
+
+/**
+ * Generates an AWS console URL for a given resource
+ */
+export const getAwsConsoleUrl = (
```
Amplifiyer, line 91: All the methods in this file need extensive unit test coverage.

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
```diff
@@ -16,25 +75,395 @@ export class SandboxDevToolsCommand implements CommandModule<object, object> {
...
+    // Apply rate limiting to all routes
+    const limiter = rateLimit({
+      windowMs: 15 * 60 * 1000, // 15 minutes
+      max: 100, // Limit each IP to 100 requests per windowMs
+      standardHeaders: true, // Return rate limit info in the `RateLimit-*` headers
+      legacyHeaders: false, // Disable the `X-RateLimit-*` headers
+    });
+
+    // Apply the rate limiting middleware to all requests
+    app.use(limiter);
```
Amplifiyer, line 138: Do we need rate limiting for local servers?

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
```diff
@@ -16,25 +75,395 @@ export class SandboxDevToolsCommand implements CommandModule<object, object> {
...
+      getCloudFormationClient: () => new CloudFormationClient(),
+    });
+
+    // Get the sandbox instance but don't start it automatically
+    const sandboxFactory = new SandboxSingletonFactory(
+      sandboxBackendIdResolver.resolve,
+      new SDKProfileResolverProvider().resolve,
+      printer,
+      format,
+    );
```
Amplifiyer: We should add a command factory for devtools that instantiates this command with required dependencies injected. See https://github.com/aws-amplify/amplify-backend/blob/07b7f74a63bc950cdbd9a3211ca2d7a06fbad282/packages/cli/src/commands/sandbox/sandbox_command_factory.ts#L81 as example

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
```diff
@@ -16,25 +75,395 @@ export class SandboxDevToolsCommand implements CommandModule<object, object> {
...
+
+    // Function to determine the sandbox state and update related flags
+    const getSandboxState = async () => {
+      const previousState = sandboxState; // Store the previous state
+
+      // If the state is unknown, check if the stack exists
+      if (sandboxState === 'unknown') {
+        try {
+          // Check AWS stack state
+          const cfnClient = new CloudFormationClient();
```
Amplifiyer: All of these clients should be injected at runtime. Will make the unit testing much easier and simpler.

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
```diff
@@ -16,25 +75,395 @@ export class SandboxDevToolsCommand implements CommandModule<object, object> {
...
+          const stackName = BackendIdentifierConversions.toStackName(backendId);
+
+          try {
+            await cfnClient.send(
+              new DescribeStacksCommand({ StackName: stackName }),
+            );
```
Amplifiyer: You are not using the response here?

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
```diff
@@ -16,25 +75,395 @@ export class SandboxDevToolsCommand implements CommandModule<object, object> {
...
+    };
+
+    // Override printer.print (the lower-level method)
+    printer.print = function (message: string) {
+      // Call the original print method
+      originalPrint.call(this, message);
+      // Avoid double processing if this is called from printer.log
+      if (processingMessage) {
+        return;
+      }
```
Amplifiyer: We should avoid doing this. This creates sideeffects in other parts of code that would be very hard to debug/troubleshoot

packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers.ts
```diff
@@ -0,0 +1,537 @@
...
+        }
+
+        // Final check to ensure region is a string
+        if (region && typeof region !== 'string') {
+          region = String(region);
+        }
+      } catch (error) {
+        printer.log('Error processing region: ' + error, LogLevel.ERROR);
+        region = null;
+      }
```
Amplifiyer: See if we can reuse this https://github.com/aws-amplify/amplify-backend/blob/146f60396e2e686843d1f957f30b9ca1cea5fe30/packages/platform-core/src/telemetry/region_fetcher.ts#L6 through a DI, so we can reuse the cached results instead of finding the region every time.

packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers.ts
```diff
@@ -0,0 +1,537 @@
...
+  name: string;
+  resources: Array<
+    ResourceWithFriendlyName & { metadata?: Record<string, unknown> }
+  >;
+};
+
+/**
+ * Service for handling socket events
+ */
+export class SocketHandlerService {
```
Amplifiyer, line 52: This class also needs required unit test coverage

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/ConsoleViewer.tsx
svidgen: Have we already discussed keeping the styles offline with something like [`ansi-to-html`](https://www.npmjs.com/package/ansi-to-html)?

package.json
megha-narayanan: reverting.

package.json
megha-narayanan: Nope. Removing

packages/cli/src/commands/sandbox/sandbox-devtools/logging/cloudformation_format.ts
```diff
@@ -0,0 +1,118 @@
...
+      '\\[22m',
+      '\\[1m',
+      '\\[36m',
+      '\\[39m',
+    ].join('|'),
+    'g',
+  );
+
+  return text.replace(ansiEscapeCodesPattern, '');
+};
```
svidgen: What's functionally different about this cleaner and the one below?

packages/cli/src/commands/sandbox/sandbox-devtools/logging/cloudformation_format.ts
```diff
@@ -0,0 +1,118 @@
...
+  const parts = stackName.split('-');
+
+  if (parts && parts.length === 7 && parts[3] === 'sandbox') {
+    return parts[5].slice(0, -10) + ' stack';
+  } else if (parts && parts.length === 5 && parts[3] === 'sandbox') {
+    return 'root stack';
+  }
+
+  return undefined;
+};
```
svidgen: A good habit when baking arcane splits and magic numbers into a function like this is to show an example or two of the before/after in the docstring. If there *is* a good name you can give to the magic constants here, give them names. Else, it's helpful to indicate where they come from in the docstring examples.

In this case, the naming patterns might be completely obvious to others in the codebase. But, if someone new (it's always Day 1, right?) jumps in *at this point*, examples will help.

Leaving this comment on this particularly arcane looking set of manipulations. But, please apply anywhere you feel arcane incantations are being performed.

packages/cli/src/commands/sandbox/sandbox-devtools/logging/cloudformation_format.ts
```diff
@@ -0,0 +1,118 @@
+/**
+ * Creates a friendly name for a resource, using CDK metadata when available.
+ * @param logicalId The logical ID of the resource
+ * @param metadata Optional CDK metadata that may contain construct path
+ * @param metadata.constructPath Optional construct path from CDK metadata
+ * @returns A user-friendly name for the resource
+ */
+export const createFriendlyName = (
```
svidgen, line 20: Borderline arcane incantations herein. Partially cp-ing a comment from below. Examples in docstrings will help future maintainers.

packages/cli/src/commands/sandbox/sandbox-devtools/logging/cloudformation_format.ts
```diff
@@ -0,0 +1,118 @@
...
+export const isDeploymentProgressMessage = (message: string): boolean => {
+  const cleanedMessage = cleanAnsiCodes(message);
+  return (
+    cleanedMessage.includes('_IN_PROGRESS') ||
+    cleanedMessage.includes('CREATE_') ||
+    cleanedMessage.includes('DELETE_') ||
+    cleanedMessage.includes('UPDATE_') ||
+    cleanedMessage.includes('Deployment in progress') ||
+    cleanedMessage.includes('COMPLETE') ||
+    cleanedMessage.includes('FAILED') ||
```
svidgen: Wouldn't mind seeing something like this if it speaks to you:

```ts
[
  '_IN_PROGRESS',
  'CREATE_',
  'DELETE_',
  'UPDATE_',
  'Deployment in progress',
  'COMPLETE',
  'FAILED'
].some(pattern => cleanedMessage.includes(pattern))
```

But, with room for opinion here, so adopt only if it really speaks to you.

packages/cli/src/commands/sandbox/sandbox-devtools/logging/cloudformation_format.ts
```diff
@@ -0,0 +1,118 @@
...
+    // Match CloudFormation resource status patterns
+    /\d+:\d+:\d+\s+[AP]M\s+\|\s+[A-Z_]+\s+\|\s+.+\s+\|\s+.+/.test(
```
svidgen: This one's a bit arcane for me. Examples please.

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/ConsoleViewer.tsx
```diff
@@ -0,0 +1,208 @@
...
+    return matchesLevel && matchesSearch;
+  });
+
+  // Auto-scroll to bottom when logs change
+  useEffect(() => {
+    if (autoScroll && scrollContainerRef.current) {
+      scrollContainerRef.current.scrollTop =
+        scrollContainerRef.current.scrollHeight;
+    }
+  }, [filteredLogs, autoScroll]);
```
svidgen, line 98: Is `autoScroll` is always `true`? Plans for it not to be?

If a customer scrolls up in the container while it's being written to, will there be a mechanism to disable autoscroll until the customer scrolls back to the button and/or re-enables autoscroll?

(Or, does this already happen somehow and I'm missing it?)

packages/cli/src/commands/sandbox/sandbox-devtools/services/shutdown_service.ts
```diff
@@ -0,0 +1,81 @@
...
+      message: `DevTools server is shutting down (${String(reason)})...`,
+    });
+
+    // Close socket and server connections
+    await this.io.close();
+    this.server.close();
+
+    // Exit the process if requested
+    if (exitProcess) {
+      // Short delay to allow messages to be sent
```
svidgen: What messages? Why is `500 ms` the right amount of time to wait for them to send?

packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers.ts
```diff
@@ -0,0 +1,537 @@
...
+  /**
+   * Creates a new SocketHandlerService
+   */
+  constructor(
+    private io: Server,
+    private sandbox: Sandbox,
+    private getSandboxState: () => Promise<string>,
+    private backendId: { name: string },
+    private shutdownService: import('./shutdown_service.js').ShutdownService,
```
svidgen: Why `import()` `ShutdownService` as a function here instead of letting it hang out with its other `import`-ant friends?

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/hooks/useResourceManager.ts
```diff
@@ -0,0 +1,211 @@
...
+      setIsLoading(false);
+      setError(data.message);
+    };
+
+    socket.on('savedResources', handleSavedResources);
+    socket.on('deployedBackendResources', handleDeployedBackendResources);
+    socket.on('customFriendlyNames', handleCustomFriendlyNames);
+    socket.on('customFriendlyNameUpdated', handleCustomFriendlyNameUpdated);
+    socket.on('customFriendlyNameRemoved', handleCustomFriendlyNameRemoved);
+    socket.on('error', handleError);
```
svidgen: Just picking a random spot to leave this comment: Would be really good to have constants  imported from a central spot (at least per group) to match on instead of duplicated strings all over. In theory, a typo in one of those strings could be caught by a good suite of tests as well, but caught much sooner if caught statically.

(Bonus points if you include docstrings to indicate what the constants mean.)

Most IDEs can also help track down usage of a variable/constant more definitively than searching for a string literal.

packages/cli/src/commands/sandbox/sandbox-devtools/services/resource_service.ts
```diff
@@ -0,0 +1,185 @@
...
+        let region = null;
+
+        try {
+          if (typeof regionValue === 'function') {
+            // If it's an async function, we need to await it
+            if (regionValue.constructor.name === 'AsyncFunction') {
+              region = await regionValue();
+            } else {
+              region = regionValue();
+            }
```
svidgen: Q for the team: Do we have lint rules in place that would prevent this from just being `await regionValue()` either way? If not, there's usually not much reason to check whether a thing is async. If it *can* be async, just `await` it. *(Right?)*

packages/cli/src/commands/sandbox/sandbox-devtools/resource_console_functions.ts
```diff
@@ -0,0 +1,263 @@
...
+        ? physicalId.split(':function:')[1]
+        : physicalId;
+      return `${baseUrl}/lambda/home?region=${region}#/functions/${lambdaFunctionName}`;
+
+    // case 'AWS::DynamoDB::Table': // COMMENT OUT (UNTESTED)
+    //   // Check if physicalId is an ARN and extract table name if needed
+    //   const dynamoTableName = physicalId.includes(':table/')
+    //     ? physicalId.split(':table/')[1]
+    //     : physicalId;
+    //   return `${baseUrl}/dynamodb/home?region=${region}#/tables/view/${encodeURIComponent(dynamoTableName)}`;
```
svidgen: You mean not even manually tested? The others aren't covered by unit or integ tests yet, are they?

svidgen (commented):
Can't say I did a *full* review here &mdash; there's a lot here. 🎉  But, I left a handful of comments for you for things that jumped out at me.

My other big concern across the board is testing. Given that this is destined for a feature branch, and given that it's already huge, I could be convinced to see a separate PR focused on e2e testing. Some lower level unit/interaction testing would be good too. But, long term, before I'd put this in front of customers, I'd want to know I "can't" break it. And, I think e2e's are how I gain that confidence.

packages/cli/src/commands/sandbox/port_checker.ts
megha-narayanan: moving the server starting to devtools command -- this is a relic from when we tried multiple ports.

packages/cli/src/commands/sandbox/sandbox-devtools/logging/cloudformation_format.ts
megha-narayanan: already resolved in commit 87fa132 (duplicate from trimming down PR)

packages/cli/src/commands/sandbox/sandbox-devtools/logging/cloudformation_format.ts
megha-narayanan: pulled that function out and exported it to get rid of duplication.

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/ConsoleViewer.tsx
megha-narayanan: switched to stripANSI for now-- is there a benefit to taking it offline if stripAnsi already exists in the repo?

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/hooks/useResourceManager.ts
megha-narayanan: got it. refactored.

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
megha-narayanan: The github workflows wouldn't pass without it. I'm not sure if its strictly needed though-- any guidance?

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
megha-narayanan: got it, the factory right now is trivial, updating it to inject dependencies.

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
megha-narayanan: done. this does make unit testing a lot easier! I'll go ahead and make unit tests more robust in a separate commit

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/hooks/useResourceManager.ts
megha-narayanan: done. put constants in a shared location that frontend and backend can both access.

packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers.ts
megha-narayanan: yep, I hadn't put it in this PR. updated and added.

packages/cli/src/commands/sandbox/sandbox-devtools/resource_console_functions.ts
megha-narayanan: yep, just hadn't checked them out to this PR. done and updated.

packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers.ts
megha-narayanan: yep, this also makes testing easy. Adding it.

packages/cli/src/commands/sandbox/sandbox-devtools/services/resource_service.ts
megha-narayanan: resolved by eliminating this code by using RegionFetcher

packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers.ts
megha-narayanan: Moving it to the top with the other imports!

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/ConsoleViewer.tsx
megha-narayanan: good catch, yes, autoScroll was always true, now I've made it so that it disables if you scroll up in logs then re-enables if you scroll back down  to the bottom.

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
megha-narayanan: we're just checking if the stack exists. but maybe using exceptions for control flow is not appropriate. Would it be better if I used ListStacksCommand or validateTemplate. Or waitUntilStackExists with a short timeout?

packages/cli/src/commands/sandbox/sandbox-devtools/logging/cloudformation_format.ts
megha-narayanan: got it. this actually belongs in a different PR, so I'm getting rid of it here.

packages/cli/src/commands/sandbox/sandbox-devtools/logging/cloudformation_format.ts
megha-narayanan: will do.

packages/cli/src/commands/sandbox/sandbox-devtools/logging/cloudformation_format.ts
megha-narayanan: got it. actually figured out how to actually get non-null metadata which eliminates the need for some of these functions. but examples have been added in the remaining functions where I felt it was needed.

packages/cli/src/commands/sandbox/sandbox-devtools/services/shutdown_service.ts
megha-narayanan: I was sending a shutdown log but I guess that isn't super helpful because the process is, well, shutting down. If we want to log, we can always log on socket disconnect. Getting rid of the log and delay. This delay was also for the server close, which is async but doesn't return a promise. resolving that properly with a promise wrapper to make sure server closing is finished before we end the process.

packages/cli/src/commands/sandbox/sandbox-devtools/logging/cloudformation_format.ts
svidgen: Thank-you! Probably goes without saying, but explicit test coverage would also be ideal. Unless I'm overlooking it, just a small set of examples like you have here in a test loop would be good.

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/ConsoleViewer.tsx
svidgen: Thank-you!

packages/cli/src/commands/sandbox/sandbox-devtools/shared/socket_events.ts
svidgen, line 1: Works for me. In my head these docstrings were a little less repetition of the variable names and a little more "instructive". But, having a local "registry" of all events is a distinct improvement nonetheless. Thank-you!

packages/sandbox/src/file_watching_sandbox.ts
```diff
@@ -75,6 +75,13 @@ export class FileWatchingSandbox extends EventEmitter implements Sandbox {
   private watcherSubscription: Awaited<ReturnType<typeof _subscribe>>;
   private outputFilesExcludedFromWatch = ['.amplify'];
   private filesChangesTracker: FilesChangesTracker;
+  private state:
+    | 'running'
+    | 'stopped'
+    | 'deploying'
+    | 'deleting'
+    | 'nonexistent'
+    | 'unknown' = 'unknown';
```
svidgen: This enumeration looks like it occurs a handful of places. Does a more centralized named type make sense?

FWIW, we could also make a case for these being constants, but for reasons I can't quite articulate, it feels less clear-cut here. I'm curious for opinions on guiding principles from the more tenured CLI team members.

packages/deployed-backend-client/src/deployed-backend-client/deployed_resources_enumerator.ts
```diff
@@ -102,10 +109,74 @@ export class DeployedResourcesEnumerator {
...
+
+          metadata[logicalId] = {
+            constructPath: resourceMetadata['aws:cdk:path'],
+          };
+        },
+      );
+
+      return metadata;
+    } catch {
+      // If we can't get template metadata, return empty object
```
svidgen: An ideal to shoot for both on this comment and in all code comments *everywhere* (now and forever):

In general, don't tell me *what* the code does. It's usually pretty obvious *what the code does*. If you felt the need to leave a comment, my guess is that something else wasn't obvious as you were writing it. **Tell us the *unobvious* part!**

Why did we choose the behavior? Do any external requirements dictate it? Are some consumers inside or outside the package depending on it? Are there language or platform quirks it mitigates? Etc..

If what the code is doing *becomes* obvious as you refine the code, remove the comment.

If the code *itself* is not obvious:

1. Try to make it more obvious.
2. Try again! Decompose into well-named functions and variable names.
3. Finally, if the problem itself is inherently complex, explain the code, ideally using examples. (E.g., what I asked for with the "arcane" bits elsewhere.)

Should this comment be something like this?

```
// exception occurs when ____________.
// consumers of this don't need to know about these errors and instead prioritize
// just printing what they *do* know on the screen.
```

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
megha-narayanan: refactored to pass in my own printer. required taking sandboxfactory out of the injected dependencies -- let me know what you think about the approach I just pushed.

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.test.ts
```diff
@@ -0,0 +1,130 @@
...
+      };
+
+      await command.handler();
+
+      assert.strictEqual(printMock.mock.callCount(), 1);
+      assert.match(
+        printMock.mock.calls[0].arguments[0],
+        /DevTools server started at/,
+      );
+    });
```
svidgen: I'm a little confused as to what this is testing. If `command` and/or `printer` were just empty objects, wouldn't this still pass?

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.test.ts
```diff
@@ -0,0 +1,130 @@
...
+      command.handler = async () => {
+        const isInUse = await mockPortChecker.isPortInUse(3333);
+        const port = isInUse ? 4444 : 3333;
+        printer.print(`DevTools server started at http://localhost:${port}`);
+      };
+
+      await command.handler();
+
+      assert.strictEqual(portCheckerMock.mock.callCount(), 1);
+      assert.match(printMock.mock.calls[0].arguments[0], /localhost:3333/);
```
svidgen: Similar question here. I may be missing something or having bran fog. But, it looks like everything that's needed for these tests to pass is *in the test*.

packages/cli/src/commands/sandbox/sandbox-devtools/resource_console_functions.ts
megha-narayanan: manual testing. just completed though, and added these back in.

packages/cli/src/commands/sandbox/sandbox-devtools/resource_console_functions.ts
megha-narayanan: yep, I did manual testing now  though so I've added these back in.

svidgen (commented):
Lots here! A couple themes throughout to look at:

Less important: **Comment quality.** I left more detail on one comment in particular. But in general, please use them to communicate what's *not* obvious from the code itself that your future self needs to know before *changing* the code.

More important: **Test quality.** I'm seeing a broad mix of test quality. On the low end, some of the tests, at least to my eyes, look tautological. (Like they'd probably pass regardless of the correctness of the implementation.) I could have misread. But, just make sure tests clearly fail if the specific implementation code under test changes.

packages/sandbox/src/file_watching_sandbox.ts
megha-narayanan: yep, I have it defined in with my devtools code in cli but I agree that a better method would be to put it in sandbox package and import it, so then it can be used in both. making this change!

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.test.ts
megha-narayanan: totally mixed up my mocking in this file and forgot what I was testing -- I had to refactor sandbox_devtools_command to bring some methods out to make it all mock-able and testable, but I've now updated the tests to call the actual methods with mocked dependencies/external methods. goes for other comment regarding this file too.

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.test.ts
megha-narayanan: fixed

packages/deployed-backend-client/src/deployed-backend-client/deployed_resources_enumerator.ts
megha-narayanan: got it! definitely left myself some extraneous comments which just explain code-- I've fixed this comment and made some other misc comment edits in my recent commit. will definitely keep this in mind going forward.

package.json
```diff
@@ -59,13 +59,15 @@
     "@actions/github": "^6.0.0",
     "@aws-amplify/eslint-plugin-amplify-backend-rules": "^0.0.2",
     "@aws-sdk/client-amplify": "^3.750.0",
-    "@aws-sdk/client-cloudformation": "^3.750.0",
-    "@aws-sdk/client-cloudwatch-logs": "^3.750.0",
+    "@aws-sdk/client-cloudformation": "^3.828.0",
+    "@aws-sdk/client-cloudwatch-logs": "^3.835.0",
     "@aws-sdk/client-cognito-identity-provider": "^3.750.0",
     "@aws-sdk/client-dynamodb": "^3.750.0",
```
Amplifiyer: I still see these upgrades for some aws-sdk clients. We should avoid this

packages/cli-core/src/loggers/cfn-deployment-progress/cfn_deployment_progress_logger.ts
```diff
@@ -59,6 +60,12 @@ export class CfnDeploymentProgressLogger {
   private resourceNameIndentation = 0;
   private readonly getBlockWidth: () => number;
 
+  /**
+   * Extract nested stack names
+   * @deprecated Use the shared normalizeCDKConstructPath function from formatters/cdk_path_formatter.js instead
+   */
+  private normalizeCDKConstructPath = normalizeCDKConstructPath;
+
```
Amplifiyer: Is this necessary?

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/services/socket_client_service.ts
```diff
@@ -0,0 +1,244 @@
...
+    }
+  }
+
+  /**
+   * Emits an event to the server with exponential backoff
+   * @param event The event name
+   * @param data The event data
+   */
+  protected emit<T>(event: string, data?: T): void {
+    if (!this.socket) return;
```
Amplifiyer: Should this be a throw instead? Or at least an error?

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/services/socket_client_service.ts
```diff
@@ -0,0 +1,244 @@
...
+  /**
+   * Registers a handler for an event
+   * @param event The event name
+   * @param handler The event handler
+   * @returns A function to unsubscribe
+   */
+  protected on<T>(event: string, handler: (data: T) => void): () => void {
+    if (!this.socket) return () => {};
+    this.socket.on(event, handler);
+    return () => this.socket?.off(event, handler);
```
Amplifiyer: should this be a named method, called `unsubscribe` or something?

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/services/sandbox_client_service.ts
```diff
@@ -0,0 +1,114 @@
...
+  message?: string;
+  timestamp?: string;
+}
+
+/**
+ * Interface for sandbox options
+ */
+export interface SandboxOptions {
+  identifier?: string;
+  once?: boolean;
```
Amplifiyer: this should ideally come from sandbox package to avoid duplication and sandbox being the authority of what options it accepts.

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/ConsoleViewer.tsx
```diff
@@ -0,0 +1,241 @@
...
+  // Extract unique log levels from logs
+  useEffect(() => {
+    const uniqueLevels = [...new Set(logs.map((log) => log.level))];
+    setAvailableLogLevels(
+      uniqueLevels.map((level) => ({
+        label: level,
+        value: level,
+      })),
+    );
+  }, [logs]);
```
rtpascual:
```suggestion
const uniqueLevels = [...new Set(logs.map((log) => log.level))];
const availableLogLevels = uniqueLevels.map((level) => ({
  label: level,
  value: level,
}));
```
This may not need to be in a useEffect, see https://react.dev/learn/you-might-not-need-an-effect.

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/Header.tsx
```diff
@@ -0,0 +1,231 @@
...
+  const handleStopSandbox = () => {
+    setIsLoading(true);
+    onStopSandbox();
+  };
+
+  const handleDeleteSandbox = () => {
+    if (
+      window.confirm(
+        'Are you sure you want to delete the sandbox? This will remove all resources and cannot be undone.',
+      )
```
rtpascual: Why window instead of `Modal` here? Same in `handleStopDevTools`.

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/Header.tsx
```diff
@@ -0,0 +1,231 @@
...
+    ) {
+      onStopDevTools?.();
+    }
+  };
+
+  const getSandboxStatusIndicator = () => {
+    const statusText = sandboxIdentifier
+      ? `Sandbox ${sandboxStatus === 'nonexistent' ? '' : `(${sandboxIdentifier}) `}`
+      : 'Sandbox ';
```
rtpascual: Nit: I think it would be easier to read by moving the space after the status text to the `StatusIndicator` text. This avoids accidentally duplicating the space between words:
```tsx
<StatusIndicator>{statusText} Deploying</StatusIndicator>
```
Resulting in `<statusText>(two spaces)Deploying`

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/App.tsx
```diff
@@ -1,16 +1,494 @@
...
+  useEffect(() => {
+    if (sandboxStatus === 'unknown' && connected) {
+      const forceStatusCheck = setTimeout(() => {
+        console.log('Forcing sandbox status check after timeout');
+        sandboxClientService.getSandboxStatus();
+      }, 2000); // Force a check after 2 seconds
+
+      return () => clearTimeout(forceStatusCheck);
+    }
+  }, [sandboxStatus, connected, sandboxClientService]);
```
rtpascual: Can these useEffects be combined into one since the effect dependencies are identical?

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
```diff
@@ -1,10 +1,55 @@
...
+import { CommandModule } from 'yargs';
+import express from 'express';
+import rateLimit from 'express-rate-limit';
+import { createServer } from 'node:http';
+import { fileURLToPath } from 'node:url';
+import { dirname, join } from 'node:path';
+import { Server } from 'socket.io';
+import {
+  LogLevel,
+  format as formatUtil,
```
Amplifiyer:
```suggestion
  format as defaultFormat,
```

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
```diff
@@ -1,10 +1,55 @@
...
+  minimumLogLevel,
+  printer as printerUtil,
```
Amplifiyer:
```suggestion
  printer as defaultPrinter,
```

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
```diff
@@ -16,25 +61,427 @@ export class SandboxDevToolsCommand implements CommandModule<object, object> {
...
-  handler = async (): Promise<void> => {
-    printer.log('Sandbox DevTools UI is coming soon...', LogLevel.INFO);
-    printer.log('This is a placeholder implementation for PR 1', LogLevel.INFO);
-  };
+  async checkPortAvailability(port: number): Promise<void> {
+    const isInUse = await this.portChecker.isPortInUse(port);
+
+    if (isInUse) {
+      this.printer.log(
+        `Port ${port} is already in use. Please close any applications using this port and try again.`,
```
Amplifiyer:
```suggestion
        `Port ${port} is already in use. Close any applications using this port and try again.`,
```

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
```diff
@@ -16,25 +61,427 @@ export class SandboxDevToolsCommand implements CommandModule<object, object> {
...
+        LogLevel.ERROR,
+      );
+      throw new Error(
+        `Port ${port} is required for DevTools. Please ensure it's available.`,
```
Amplifiyer:
```suggestion
        `Port ${port} is required for DevTools. Ensure it's available and no other instance of DevTools is already running`,
```

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
```diff
@@ -16,25 +61,427 @@ export class SandboxDevToolsCommand implements CommandModule<object, object> {
...
+
+    // Create a new SandboxSingletonFactory with our custom logger
+    // Our server, and thus logger, are only created at runtime, so we cannot use the
+    // SandboxSingletonFactory as an injected dependency.
+    const localSandboxFactory = new SandboxSingletonFactory(
+      this.sandboxBackendIdResolver.resolve,
+      new SDKProfileResolverProvider().resolve,
+      devToolsLogger,
+      this.format,
+    );
```
Amplifiyer: Can we inject these through factory as well?

Amplifiyer (commented):
Looking really good. Still haven't gone through the entire PR, but wanted to leave this feedback early.

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
megha-narayanan: See the comment. I couldn't figure out a neat way to do it. Any input?

package.json
megha-narayanan: oops. missed that. fixed now!

packages/cli-core/src/loggers/cfn-deployment-progress/cfn_deployment_progress_logger.ts
megha-narayanan: I just wanted to be super careful about backward compatibility. If you're ok with it, I'll go ahead and remove.

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/Header.tsx
megha-narayanan: Good question! I've made them a custom Modal now.

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/App.tsx
megha-narayanan: Yep, done

packages/cli-core/src/loggers/cfn-deployment-progress/cfn_deployment_progress_logger.ts
Amplifiyer: Yup, as long as these are local variables and not exporting something out of package, it's fine to get rid of them

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
Amplifiyer: You can create a `DevToolsLoggerFactory`, inject it through devtools command factory, and use it to instantiate an instance at runtime.

Similarly, `sandboxFactory.getInstance()` can take an optional instance of logger which is available at runtime

packages/cli/src/commands/sandbox/sandbox_event_handler_factory.ts
```diff
@@ -25,6 +25,23 @@ export class SandboxEventHandlerFactory {
     clientConfigLifecycleHandler,
   }) => {
     return {
+      resourceConfigChanged: [
+        async () => {
+          try {
```
Amplifiyer: I can't find where this event is being emitted by the sandbox

packages/sandbox/src/sandbox.ts
```diff
@@ -23,12 +33,24 @@ export type Sandbox = {
    * Deletes this environment
    */
   delete: (options: SandboxDeleteOptions) => Promise<void>;
+
+  /**
+   * Gets the current state of the sandbox
+   * @returns The current state: 'running', 'stopped', 'deploying', 'deleting', 'nonexistent', or 'unknown'
+   */
+  getState: () => SandboxStatus;
```
Amplifiyer, line 41: This state management needs more unit test coverage.

packages/deployed-backend-client/src/deployed-backend-client/deployed_resources_enumerator.ts
```diff
@@ -102,10 +109,77 @@ export class DeployedResourcesEnumerator {
           region,
           accountId,
         ),
+        metadata:
+          templateMetadata[stackResourceSummary.LogicalResourceId || ''],
```
Amplifiyer: Should this be 
```suggestion
        metadata:
          templateMetadata[stackResourceSummary.LogicalResourceId] || '',
```

packages/cli/src/commands/sandbox/sandbox_event_handler_factory.ts
megha-narayanan: Yep, its extraneous now that it been replaced with successfulDeployment/failedDeployment for better error handling. removed in PR3 but forgot to propogate back to PR2. Doing that now.

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/services/socket_client_service.ts
megha-narayanan: I thought throwing an error would be too disruptive since the socket.io client has built-in reconnection logic, and so the socket might temporarily be null during reconnection attempts. emit uses a queue with exponential backoff to handle temporary disconnections. But I agree silently failing isn't ideal. I've added a console.warn instead. Does this seem like a good solution?

```
if (!this.socket) {
  console.warn(`Cannot emit ${event}: Socket is not initialized`);
  return;
}
```

Amplifiyer: The socket is only set at the initialization, right? In which scenario would this variable be null?

megha-narayanan: I thought throwing an error would be too disruptive since the socket.io client has reconnection logic, and so the socket might temporarily be null during reconnection attempts. Plus then we would need try/catch blocks everywhere. But I agree silently failing isn't ideal. I've added a console.error instead. Does this seem like a good solution?

```
if (!this.socket) {
  console.error(`Cannot emit ${event}: Socket is not initialized`);
  return;
}
```

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/services/socket_client_service.ts
megha-narayanan: It definitely can be, but I don't see a clear benefit. Is it for visibility in stack traces? This is called by this class and its subclasses, and whenever called, the return value is made a named function, so that seems to mitigate the stack trace issue.  I can absolutely change it just for style, just want to make sure I'm understanding this comment properly before acting on it.

Amplifiyer: It's the DX issue. If I'm understanding it correctly, currently the DX is
```ts
const subscription = clientService.on();
// To unsubscribe
subscription();
```

What I'm proposing is
```ts
const subscription = clientService.on();
// To unsubscribe
subscription.unsubscribe();
```

packages/sandbox/src/sandbox.ts
megha-narayanan: Absolutely. In writing my tests I also discovered a tiny edge case error in my state management -- fixed that and added tests.

packages/deployed-backend-client/src/deployed-backend-client/deployed_resources_enumerator.ts
megha-narayanan: EDIT: okay, turns out I actually misunderstood this. The reason I had my original code was because in case LogicalResourceId is undefined (which it can be), then we just look for the the metadata of ''. Though I imagine this is not the best way to handle that error.  The metadata property should be of type` { constructPath?: string } | undefined `, not `string`. I have now re-updated this to the following line, which is more intentional about handling the undefined case - instead of falling back to an empty string key (which likely doesn't exist in the metadata, but maybe in an error case, COULD), it explicitly returns undefined.  It avoids a potential issue where templateMetadata[''] might actually contain something unexpected. `stackResourceSummary.LogicalResourceId ? templateMetadata[stackResourceSummary.LogicalResourceId] : undefined,`

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/Header.tsx
```diff
@@ -70,52 +74,58 @@ const Header = ({
...
+  const cancelStopDevTools = () => {
+    setShowStopDevToolsConfirmation(false);
   };
 
   const getSandboxStatusIndicator = () => {
     const statusText = sandboxIdentifier
-      ? `Sandbox ${sandboxStatus === 'nonexistent' ? '' : `(${sandboxIdentifier}) `}`
-      : 'Sandbox ';
+      ? `Sandbox${sandboxStatus === 'nonexistent' ? '' : ` (${sandboxIdentifier})`}`
+      : 'Sandbox';
```
rtpascual: Taking another look at this, we can make this easier to read if we break this out of nested ternaries
```ts
const statusText = sandboxIdentifier && sandboxStatus !== 'nonexistent' ?
  `Sandbox (${sandboxIdentifier})` : 'Sandbox';
```

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/services/socket_client_service.ts
megha-narayanan: got it. I had them like this: `const unsubscribeError = resourceClientService.onError(handleError);` and then just call `unsubscribeError()` which matches all the other handlers. I'll go ahead and change them all, but then you have a bunch of calls like this `unsubscribeLog.unsubscribe();`

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/services/sandbox_client_service.ts
megha-narayanan: fixed with the temp fix we discussed and added a note about the issue in the comment

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/services/socket_client_service.ts
megha-narayanan: ah, I misunderstood the socket logic. yep, you're right. making it a throw.

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
megha-narayanan: got it.

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
Amplifiyer: can't the service return undefined as well that we should check against? https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-cloudformation/Interface/DescribeStacksCommandOutput/

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/services/socket_client_service.ts
Amplifiyer: Typically when you attach a handler or create a subscription, you get a subscription object which exposes methods to manage the subscription. An example https://github.com/parcel-bundler/watcher?tab=readme-ov-file#example

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
Amplifiyer: what's the error you are getting?

megha-narayanan: There's a CodeQL warning: https://codeql.github.com/codeql-query-help/javascript/js-missing-rate-limiting/. Leaving it in!

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/ConsoleViewer.tsx
```diff
@@ -0,0 +1,234 @@
...
+      case 'warning':
+        return 'warning';
+      case 'success':
+        return 'success';
+      case 'info':
+        return 'info';
+      case 'debug':
+        return 'pending';
+      default:
+        return 'info';
```
Amplifiyer, line 129: why is debug level a pending status and default info?

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/ConsoleViewer.tsx
```diff
@@ -0,0 +1,234 @@
...
+      ),
+      width: 150,
+    },
+    {
+      id: 'message',
+      header: 'Message',
+      cell: (item: LogEntry) => {
+        const cleanedMessage = stripAnsi(item.message);
+        return cleanedMessage;
+      },
```
Amplifiyer: can be just
```suggestion
      cell: (item: LogEntry) => stripAnsi(item.message),
```

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/Header.tsx
```diff
@@ -0,0 +1,264 @@
...
+  const [showDeleteConfirmation, setShowDeleteConfirmation] = useState(false);
+  const [showStopDevToolsConfirmation, setShowStopDevToolsConfirmation] =
+    useState(false);
+
+  // Reset loading state when sandbox status changes
+  useEffect(() => {
+    console.log(
+      `[CLIENT] Header: sandboxStatus prop changed to ${sandboxStatus}`,
+    );
+
```
Amplifiyer: do we need these log statements?

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/Header.tsx
```diff
@@ -0,0 +1,264 @@
...
+        } else {
+          return (
+            <SpaceBetween direction="horizontal" size="xs">
+              <Spinner size="normal" />
+              <StatusIndicator type="pending">
+                Status check taking longer than expected
+              </StatusIndicator>
+            </SpaceBetween>
+          );
+        }
```
Amplifiyer: We don't need this level of messaging. It's either "Waiting for status..." or "${Status}" to keep it simple.

packages/cli/src/commands/sandbox/sandbox-devtools/logging/cloudformation_format.test.ts
```diff
@@ -0,0 +1,116 @@
+/* eslint-disable spellcheck/spell-checker */
```
Amplifiyer, line 1: In one of future PR, we should remove all file level eslint suppression.

Amplifiyer (dismissed):
Great job 🚀

rtpascual (dismissed):
Overall LGTM 🚀

packages/cli/src/commands/sandbox/sandbox-devtools/logging/cloudformation_format.test.ts
megha-narayanan: yep, will do.

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/Header.tsx
megha-narayanan: nope, removing.

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/ConsoleViewer.tsx
megha-narayanan: This was just for the UI styling of the little icon that goes next to the level. This is the type: 
`export declare namespace StatusIndicatorProps {
    type Type = 'error' | 'warning' | 'success' | 'info' | 'stopped' | 'pending' | 'in-progress' | 'loading';
    type Color = 'blue' | 'grey' | 'green' | 'red' | 'yellow';
}` I just thought that the icon for pending fit debug best, and info was most versatile. Is there a choice for either that you think would be better?

Amplifiyer: got it. Maybe the name is slightly confusing. Perhaps `getStatusIndicatorIconForLogLevel` (verbose but self documenting)

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/Header.tsx
```diff
@@ -34,34 +33,13 @@ const Header = ({
...
-    if (sandboxStatus === 'unknown') {
-      const timer = setTimeout(() => {
-        setStatusCheckTimeout((prev) => prev + 1);
-      }, 3000);
-      return () => clearTimeout(timer);
-    } else {
-      // Reset timeout counter when we get a definitive status
-      setStatusCheckTimeout(0);
-    }
   }, [sandboxStatus]);
```
rtpascual, line 43: This can be in a follow up PR but I believe this useEffect can be removed since this component rerenders when sandboxStatus changes (since it is a prop) and you already initialize the state as `false` in line 35.

rtpascual (approved):

Amplifiyer (approved):

Issue #2901, opened 2025-07-17
Sandbox doesn't seem to work. SyntaxError: Named export 'OperationTypeNode' not found.


ChrisDean:
### Environment information

```plain text
(.venv) # npx ampx info
file:///<my_path>/test/node_modules/@graphql-codegen/core/node_modules/@graphql-tools/merge/node_modules/@theguild/federation-composition/esm/subgraph/state.js:1
import { Kind, OperationTypeNode, specifiedScalarTypes, DirectiveLocation, } from "graphql";
               ^^^^^^^^^^^^^^^^^
SyntaxError: Named export 'OperationTypeNode' not found. The requested module 'graphql' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'graphql';
const { Kind, OperationTypeNode, specifiedScalarTypes, DirectiveLocation, } = pkg;

    at ModuleJob._instantiate (node:internal/modules/esm/module_job:220:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:321:5)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:644:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:117:5)

Node.js v22.17.0
```

### Describe the bug

I appreciate this isn't exactly the expected output for npx ampx info but I'm simply trying to get the most basic project setup and running. In a totally vanilla environment I install the latest lts version of node, awscli and the basics for amplify

I run  ```npm create amplify@latest```
Which seems to install a project fine then asks me to run ````npx ampx sandbox ``` which gives me the error above.

If I try some sample projects from the aws samples I get the same error.

I've tried adding type module to the package.json too as the docs suggest.

have also tried pinning the version of graphql to the latest version, this seems to allow sandbox to run in part but ultimately the CDK stuff fails if I do that...

Any ideas? different package versions or node version recommended? 

Something else?

### Reproduction steps

in a new foolder
 ```
npm create amplify@latest
npx ampx sandbox
```

**Comment by markhker** [comment]
Same issue

**Comment by filipesbragio** [comment]
Apparently a GraphQLTools update broke Amplify's npm and it's being worked on #2900 
[ardatan/graphqltools#7290](https://github.com/ardatan/graphql-tools/issues/7290)

**Comment by sobolk** [comment]
The workaround for Yarn 4.x is to add the following to `package.json` and re-install deps.

```
[...]

PR #2903, opened 2025-07-18
Fix yarn

sobolk:
## Problem

Related to https://github.com/aws-amplify/amplify-backend/issues/2901 .

This addresses remaining problem Yarn4.0

## Changes

Bundle `form-gernerator` and `@graphql-codegen/core` together.

Reason for this is that other solutions attempted didn't work:
1. Resolutions - doesn't work due to locality https://github.com/yarnpkg/rfcs/blob/master/implemented/0000-selective-versions-resolutions.md#locality-of-the-resolutions .
2. Bundled dependencies don't work , https://yarnpkg.com/migration/guide#dont-use-bundledependencies
3. Moving dep to dev deps and keeping bundled doesn't work either. Yarn installs it.
4. Moving dep to root package jsona and keeping bundled doesn't work either. Yarn installs it with `*` version :( 

## Validation

Reproduced locally. Played with verdaccio and sample app.

Plus PR checks.

## Checklist


packages/form-generator/bundle.js
```diff
@@ -0,0 +1,20 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import esbuild from 'esbuild';
```
sobolk, line 2: This is ok. We don't ship this file to NPM. ESBuild is coming from root package.json.

Amplifiyer (dismissed):

PR #2909, opened 2025-07-21
[DevTools] PR 3.2 General Structure and Error Handling

megha-narayanan:
changes favicon, added error handling, minor type changes, general style and type updates


packages/cli/src/commands/sandbox/sandbox-devtools/error-handling/error_extractor.ts
```diff
@@ -0,0 +1,40 @@
...
+  message: string;
+  resolution?: string;
+};
+
+/**
+ * Extracts basic error information from any error object
+ * @param error The error to extract information from
+ * @returns Simple error information with name, message, and optional resolution
+ */
+export const extractErrorInfo = (error: unknown): SimpleErrorInfo => {
```
Amplifiyer: You should be able to use `AmplifyError.fromError()` that will guarantee return an `AmplifyError` and convert non amplify errors like in this function.
https://github.com/aws-amplify/amplify-backend/blob/f27477398f839bbb3babc31f52590a890e5b86bd/packages/platform-core/src/errors/amplify_error.ts#L107

megha-narayanan: Awesome! Even though its super short I'm going to keep it in the separate file for testing ease.

Amplifiyer: We should just reuse `AmplifyError` directly instead of creating this wrapper unless you believe there is a value add in this wrapper.

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/Header.tsx
```diff
@@ -19,6 +19,8 @@ interface HeaderProps {
   onStopDevTools?: () => void;
   onOpenSettings?: () => void;
   isStartingLoading?: boolean;
+  isStoppingLoading?: boolean;
+  isDeletingLoading?: boolean;
```
Amplifiyer, line 23: is `Loading` here meaningful? Can we just say `isStarting`,`isStopping` etc.?

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/Header.tsx
```diff
@@ -31,8 +33,10 @@ const Header = ({
   onStopDevTools,
   onOpenSettings,
   isStartingLoading = false,
+  isStoppingLoading = false,
+  isDeletingLoading = false,
 }: HeaderProps) => {
-  const [isLoading, setIsLoading] = useState(false);
+  // No longer need local loading state, using props from parent component
```
Amplifiyer: We can probably remove this comment. People reading this code would not have context on what was there before.

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/services/socket_client_service.test.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  void describe('disconnect', () => {
+    void it('calls socket.disconnect', () => {
+      service.disconnect();
+      // Our mock has disconnect as a mock function
+      assert.strictEqual(
+        (mockSocket.socket.disconnect as any).mock.callCount(),
+        1,
+      );
+    });
```
Amplifiyer: Can we also add another assertion `assert.strictEqual(service.isConnected(), false);`

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/services/socket_client_service.test.ts
```diff
@@ -0,0 +1,160 @@
...
+  });
+
+  void describe('disconnect', () => {
+    void it('calls socket.disconnect', () => {
+      service.disconnect();
+      // Our mock has disconnect as a mock function
+      assert.strictEqual(
+        (mockSocket.socket.disconnect as any).mock.callCount(),
+        1,
+      );
```
Amplifiyer, line 31: Can we also add another assertion `assert.strictEqual(service.isConnected(), false);`

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/Header.tsx
megha-narayanan: These are basically only for the actual visual component being in a "Loading" state or not, which sometimes differs from the actual state. The main example is isDeleting is different from isDeletingLoading, as there can be moments when `isDeletingLoading` (prop) and `isDeleting` (local variable), later in the file, are out of sync: when the user first clicks "Delete Sandbox": `isDeletingLoading` (prop) is set to true immediately (showing button loading state) but `isDeleting` remains false until the sandbox status updates to 'deleting'.  The loading states basically give us instant UI feedback for events that may have a short delay.

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/services/socket_client_service.test.ts
megha-narayanan: Sure, I do need to then add that to the mock socket, which I've done.

Amplifiyer (approved):

rtpascual (approved):

PR #2911, opened 2025-07-21
[DevTools] PR3.3 Local Storage

megha-narayanan:
Implements local storage and related functionality, including log settings and backend resource management and socket handlers. Also includes the backend functionality for testing lambda functions, as this is included in backend resource management which is closely tied to local storage.


Amplifiyer:
> Implements local storage and related functionality.

Can you add more detailed description on all the changes in the PR as I see functions related changes as well.

packages/cli/src/commands/sandbox/sandbox-devtools/local_storage_manager.ts
```diff
@@ -0,0 +1,838 @@
...
+
+      // Delete oldest files until we're under the limit
+      for (const file of logFiles) {
+        if (this.getLogsSizeInMB() <= this._maxLogSizeMB * 0.8) {
+          // Remove until we're at 80% of limit
+          break;
+        }
+
+        fs.unlinkSync(file.path);
+      }
```
svidgen: If I'm reading this right, looks like some good opportunity to DRY out some of this pruning logic and get some efficiency gains. E.g., a single function to collect file stats (including file size) as you're already doing in a handful of places here. When we need to calculate/display size, just sum the sizes from that list. When we need to prune, sort the same list by timestamp and start summing until we hit the pruning limit, after which we just start deleting. (Shouldn't need to repeatedly ask the filesystem "are we there yet" after each deletion.)

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/App.tsx
```diff
@@ -81,21 +93,22 @@ function AppContent() {
             console.log(
               `[CLIENT] Updating sandboxIdentifier from ${prevId} to ${data.identifier}`,
             );
+            // Store the identifier in localStorage for other components to use
+            if (data.identifier) {
+              window.localStorage.setItem('sandboxIdentifier', data.identifier);
```
svidgen: Running against localhost, this will be shared with any other dev job that uses the same port. Main concern being a leak of a customer's data into other apps they run on their machine that might manage/read/broadcast `localStorage` data.

I don't imagine this identifier is sensitive. But, is there anything else we're writing to `localStorage` that might be more sensitive?

svidgen (approved):

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/App.tsx
megha-narayanan: The identifier is not sensitive -- the sandbox prints it out on start.  Good news: this is actually dead code. I recognized the same thing and had it pass the identifier to the header as a prop instead, but it seems I forgot to remove the code that saves it. I'll remove this.

packages/cli/src/commands/sandbox/sandbox-devtools/local_storage_manager.ts
megha-narayanan: Just pushed this optimization

svidgen (approved):

packages/cli/src/commands/sandbox/sandbox-devtools/local_storage_manager.ts
```diff
@@ -0,0 +1,886 @@
...
+  resourceStatus?: {
+    resourceType: string;
+    resourceName: string;
+    status: string;
+    timestamp: string;
+    key: string;
+    statusReason?: string;
+    eventId?: string;
+  };
+  isGeneric?: boolean;
```
Amplifiyer: what does generic mean in this context? Can we use a more specific name for this property?

packages/cli/src/commands/sandbox/sandbox-devtools/local_storage_manager.ts
```diff
@@ -0,0 +1,886 @@
...
+  /**
+   * Appends a log entry to a resource's CloudWatch logs
+   * @param resourceId The ID of the resource
+   * @param logEntry The log entry to append
+   */
+  appendCloudWatchLog(resourceId: string, logEntry: CloudWatchLogEntry): void {
+    try {
+      const logs = this.loadCloudWatchLogs(resourceId);
+      logs.push(logEntry);
+      this.saveCloudWatchLogs(resourceId, logs);
```
Amplifiyer: this is ok for small files, but for larger files we'd end up doing a lot of IO (eg. let's assume we have 100MB file, adding 1 event would require us to read and load 100MB in memory and then write 100MB to file). Ideally we should be writing in a non-JSON format, e.g. a CSV or TSV

packages/cli/src/commands/sandbox/sandbox-devtools/local_storage_manager.ts
```diff
@@ -0,0 +1,886 @@
...
+        JSON.stringify(friendlyNames, null, 2),
+        { mode: 0o600 },
+      );
+    } catch (error) {
+      printer.log(
+        `Error saving custom friendly names: ${String(error)}`,
+        LogLevel.ERROR,
+      );
+    }
+  }
```
Amplifiyer: This might need to be re-evaluated as it has potential to cause confusion from customers believing that changing the name here will change in the backend as well. If we do end up merging this change, can you leave a TODO commend in this function
// TODO: Revaluate this functionality. E.g. either add tooltips for clarification, different naming or remove this altogether.

packages/cli/src/commands/sandbox/sandbox-devtools/local_storage_manager.ts
```diff
@@ -0,0 +1,886 @@
...
+      printer.log(
+        `LocalStorageManager: Error ensuring directories: ${String(error)}`,
+        LogLevel.ERROR,
+      );
+      if (error instanceof Error) {
+        printer.log(
+          `LocalStorageManager: Error stack: ${error.stack}`,
+          LogLevel.DEBUG,
+        );
+      }
```
Amplifiyer: On errors in this method, there should be a toggle to "disable" the local storage. For example if the permissions are not there, customers will keep seeing errors.

packages/cli/src/commands/sandbox/sandbox-devtools/local_storage_manager.ts
megha-narayanan: It was a flag for whether the event was directly tied to a resource (generally all are, but its technically possible that one isn't, or it can't identify the resource). Since the frontend relies on the resource info to display, I have separate display for events where we don't have a resource (error messages, generally). I realized I can just use the fact that resource is optional in this type though to do the job of the generic flag -- and just check if the resource exists to determine display.

packages/cli/src/commands/sandbox/sandbox-devtools/local_storage_manager.ts
megha-narayanan: yep, making these csv

packages/cli/src/commands/sandbox/sandbox-devtools/local_storage_manager.ts
megha-narayanan: got it. adding.

packages/cli/src/commands/sandbox/sandbox-devtools/local_storage_manager.ts
megha-narayanan: added this!

PR #2914, opened 2025-07-22
[DevTools] PR3.4 Logs

megha-narayanan:
Implements various parts of log steaming functionality
Note frontend tests for components do exist but are not included in this PR due to dependencies which require PR3.1 being merged.


packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers_logging.ts
```diff
@@ -0,0 +1,677 @@
...
+  public setupAdaptiveLogPolling(
+    resourceId: string,
+    logGroupName: string,
+    logStreamName: string,
+    socket?: Socket,
+  ): void {
+    let nextToken: string | undefined = undefined;
+    let currentLogStreamName = logStreamName;
+    let lastStreamCheckTime = Date.now();
+    const STREAM_CHECK_INTERVAL = 30000; // Check for new streams every 30 seconds
```
svidgen: Why 30 seconds?

packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers_logging.ts
```diff
@@ -0,0 +1,677 @@
...
+    let consecutiveEmptyPolls = 0;
+    let currentInterval = 2000; // Start with 2 seconds
```
svidgen: Why 2 seconds?

packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers_logging.ts
```diff
@@ -0,0 +1,677 @@
...
+
+              // Emit logs to all clients
+              this.io.emit(SOCKET_EVENTS.RESOURCE_LOGS, {
+                resourceId,
+                logs,
+              });
+
+              // Reset consecutive empty polls and speed up polling if needed
+              consecutiveEmptyPolls = 0;
+              if (currentInterval > 2000) {
```
svidgen: Should use a constant?

packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers_logging.ts
```diff
@@ -0,0 +1,677 @@
...
+                // Speed up polling if we're getting logs
+                clearInterval(pollingInterval);
+                currentInterval = 2000;
```
svidgen: Same?

packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers_logging.ts
```diff
@@ -0,0 +1,677 @@
...
+
+    // Helper function to handle empty poll results
+    const handleEmptyPoll = () => {
+      consecutiveEmptyPolls++;
+      if (consecutiveEmptyPolls > 3 && currentInterval < 10000) {
+        // Slow down polling after 3 empty polls
+        clearInterval(pollingInterval);
+        currentInterval = Math.min(currentInterval * 1.5, 10000);
+        pollingInterval = setInterval(fetchLogs, currentInterval);
+        this.activeLogPollers.set(resourceId, pollingInterval);
```
svidgen: Looks like some values that should be constants. Ideally annotated with *why* those constants were chosen. Whether and/or how they should be tuned.

Also, looks like some repeated logic for killing and re-registering the interval at a new level. Probably better as shared logic to avoid any drift in how the timers are managed.

packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers_logging.ts
```diff
@@ -0,0 +1,677 @@
...
+  ): void {
+    let nextToken: string | undefined = undefined;
+    let currentLogStreamName = logStreamName;
+    let lastStreamCheckTime = Date.now();
+    const STREAM_CHECK_INTERVAL = 30000; // Check for new streams every 30 seconds
+    let consecutiveEmptyPolls = 0;
+    let currentInterval = 2000; // Start with 2 seconds
+
+    const fetchLogs = () => {
+      void (async () => {
```
svidgen: Why is this wrapped in `void (async () => {})` instead of `const fetchLogs = async () => {}`?

packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers_logging.ts
```diff
@@ -0,0 +1,677 @@
...
+    socket?: Socket,
+  ): void {
+    let nextToken: string | undefined = undefined;
+    let currentLogStreamName = logStreamName;
+    let lastStreamCheckTime = Date.now();
+    const STREAM_CHECK_INTERVAL = 30000; // Check for new streams every 30 seconds
+    let consecutiveEmptyPolls = 0;
+    let currentInterval = 2000; // Start with 2 seconds
+
+    const fetchLogs = () => {
```
svidgen: Maybe I'm missing it, but this seems to optimistically assume any previous `fetchLogs` request has finished. But, that won't always be the case.

What happens if `fetchLogs` requests start to overlap? (Due to network congestion, service degradations, etc.)

packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers_logging.ts
megha-narayanan: Adding constants and explanation/documentation for each of these.

packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers_logging.ts
megha-narayanan: See above

packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers_logging.ts
megha-narayanan: yep, doing that. and extracting that into a helper function for shared logic!

packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers_logging.ts
megha-narayanan: I think I did this for lint at some point? anyway fixing that.

packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers_logging.ts
megha-narayanan: Good catch. I think I neglected this because of the delay between requests, but I added a check for that now.

PR #2917, opened 2025-07-22
[DevTools] PR3.5 Deployment Tracking 

megha-narayanan:
End-to-end deployment tracking. 
Note there had been an intention earlier to have Deployment Progress share more utilities with cfn_progress_logger, though partially due to time constraints, I have opted to stick with my current implementation that also provides critical persistence which is not possible by simply using cfn_progress_logger. Part of the purpose of devtools is being able to debug, which means the ability to go and see past deployment events is important.


packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers_resources.ts
```diff
@@ -0,0 +1,175 @@
...
+        // Add new events to the map (will overwrite duplicates)
+        formattedEvents.forEach((event) => {
+          const key =
+            event.resourceStatus?.eventId ||
+            `${event.timestamp}-${event.message}`;
+          eventMap.set(key, event);
+        });
+
+        // Convert map back to array
+        const mergedEvents = Array.from(eventMap.values());
```
svidgen: Same as [`mergeEvents()` method here](https://github.com/aws-amplify/amplify-backend/pull/2917/files#diff-fdeb3e66981a9164fe834d9bba697250c364868bb70fcddfadb96c50aa96db53R103). Might be good as a shared util.

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/DeploymentProgress.tsx
```diff
@@ -0,0 +1,612 @@
...
+    // Set up polling for CloudFormation events during deployment or deletion
+    let cfnEventsInterval: NodeJS.Timeout | null = null;
+    if (status === 'deploying' || status === 'deleting') {
+      console.log(
+        `Setting up polling for CloudFormation events (${status} state)`,
+      );
+      cfnEventsInterval = setInterval(() => {
+        console.log(`Polling for CloudFormation events, status: ${status}`);
+        deploymentClientService.getCloudFormationEvents();
+      }, 5000);
```
svidgen, line 246: Maybe better as a constant with note indicating why `5000`.

packages/cli/src/commands/sandbox/sandbox-devtools/react-app/src/components/DeploymentProgress.tsx
```diff
@@ -0,0 +1,612 @@
...
+      deploymentClientService.getCloudFormationEvents();
+    }
+
+    // Set up polling for CloudFormation events during deployment or deletion
+    let cfnEventsInterval: NodeJS.Timeout | null = null;
+    if (status === 'deploying' || status === 'deleting') {
+      console.log(
+        `Setting up polling for CloudFormation events (${status} state)`,
+      );
+      cfnEventsInterval = setInterval(() => {
```
svidgen, line 243: Having a hard time tracing through some of this; but are we polling from the local service for things that could be better *pushed* to the front-end via the socket?

Not a deal-breaker if we are; but if my sense is right on that, I'd at least appreciate a comment indicating why.

packages/cli/src/commands/sandbox/sandbox-devtools/services/socket_handlers_resources.ts
```diff
@@ -0,0 +1,175 @@
...
+import {
+  CloudFormationEventDetails,
+  CloudFormationEventsService,
+} from '../logging/cloudformation_format.js';
+import { SandboxStatus } from '@aws-amplify/sandbox';
+
+/**
+ * Service for handling socket events related to resources
+ */
+export class SocketHandlerResources {
```
svidgen, line 25: Where's this being used?

svidgen (changes requested):
Didn't go super deep on this yet; but ran into a newly introduced class I didn't actually see being imported anywhere. Is it supposed to be part of a different PR?

Ideally, let's make sure this PR only includes its own relevant changes -- or, if it's even possible, a convincing note indicating why a change is included here but not yet consumed.

PR #2918, opened 2025-07-23
[DevTools] PR4: Testing

megha-narayanan:
Various integration tests and misc tests that require previous PR3 changes to be merged. note this PR includes ONLY the test files, none of the actual files, so the tests don't actually run yet. It is merely to be able to view the tests.


packages/cli/src/commands/sandbox/sandbox-devtools/integration-tests/error_scenarios.test.ts
```diff
@@ -0,0 +1,334 @@
...
+    const errorLogCalls = mockLogFn.mock.calls.filter((call) =>
+      String(call.arguments[0]).includes('ThrottlingException'),
+    );
+    assert.strictEqual(errorLogCalls.length >= 1, true);
+  });
+
+  void it('should recover from storage corruption', async () => {
+    // Simulate corrupted storage by mocking loadResources to throw an error
+    const originalLoadResources = storageManager.loadResources;
+    storageManager.loadResources = () => null;
```
svidgen, line 176: Comments says "throw an error". Replacement function (which seems like maybe it should be a mock?) simply returns `null`.

packages/cli/src/commands/sandbox/sandbox-devtools/integration-tests/error_scenarios.test.ts
```diff
@@ -0,0 +1,334 @@
...
+    });
+
+    // Request saved resources
+    clientSocket.emit(SOCKET_EVENTS.GET_SAVED_RESOURCES);
+
+    // Wait for the response
+    const resources = await resourcesReceived;
+
+    // Should recover by returning an empty array instead of crashing
+    assert.deepStrictEqual(resources, []);
```
svidgen, line 192: The test description is pretty lofty -- touting "recovery" from "corruption." (Sounds *heroic*, actually!) But, this test doesn't really demonstrate recovery. It seems to demonstrate that a `null` result from `loadResources` equates to an empty array "return" value.

packages/cli/src/commands/sandbox/sandbox-devtools/integration-tests/error_scenarios.test.ts
```diff
@@ -0,0 +1,334 @@
...
+
+    // Set up a promise that will resolve when we receive resources
+    const resourcesReceived = new Promise<unknown>((resolve) => {
+      clientSocket.on(SOCKET_EVENTS.SAVED_RESOURCES, (data) => {
+        resolve(data);
+      });
+    });
+
+    // Request saved resources
+    clientSocket.emit(SOCKET_EVENTS.GET_SAVED_RESOURCES);
```
svidgen, line 186: Seeing this socket request/response pattern makes me wish we'd invested a little time in creating a protocol to handle these details under the hood -- to make these requests look more like "normal" RPC calls.

In lieu of that, for now, can we at lest rename "response" type event names to clearly indicate that they are in fact *in response* to a "call". Would be interested in a naming convention proposal. E.g., a straw man:

1. `{METHOD_NAME}_REQUEST`
2. `{METHOD_NANE}_RESPONSE`
3. `STREAM_{NAME}_CHUNK`

Or something like that.

PR #2937, opened 2025-07-31
[Geo] Construct: Phase 2 - API Key Support

manavgurnani21:
## Problem

With the creation of three resources within the Amplify Geo vertical (`AmplifyMap`, `AmplifyPlace`, and `AmplifyCollection`), developers could provision instances of these resources. While `AmplifyCollection` provisions a Geofence Collection through CDK, the initial version of `AmplifyMap` and `AmplifyPlace` did not provision any resources. Their entire purpose was to perform access control over customer-defined access permissions over AWS-managed Location resources. However, customers can integrate these resources within their application front-end through API keys. This PR focuses primarily on the feature support and output integration of AWS Location API keys for these resources.

## Changes

As part of the API key integration, there are three important segements to the implementation:

#### 1. API Key Provisioning

- Setup customer-end API key properties within `defineX()` (only one API key generated for Map/Place resource)
- Added `generateApiKey()` function for conditional provisioning
- Unit testing for API key provisioning

#### 2. Access Orchestration for API Keys

- Add API key definition to `GeoAccessBuilder` (maintains current DX of access orchestration)
- Conditional API key provisioning logic and key restriction mapping
- Additional error handling and unit testing for access orchestration

#### 3. Updating Amplify Outputs Schema
- Updating output schema within `backend-output-schemas/geo` to accomodate API keys
- Creating new amplify output schema to accomodate for geo API keys (`v1.4` -> `v1.5`)
- Bumping geo client configuration functionality to function with new schema
    - additional unit testing for new schema
    - new legacy conversion to `v0` schema (support for map styles deprecated)
    - sandbox command update (to run on `v1.5`)
- Additional unit testing for client config generation pipeline

### New DX for Geo Resources
``` typescript
export const generalMap = defineMap({
    name: 'testGeneralMap',
    access: (allow) => [
        allow.authenticated.to(['get']),
        allow.guest.to(['get']),
        allow.apiKey.to(['get']) // add API key access restrictions
    ],
    apiKeyProps: {
        apiKeyName: 'testKey'
    }
});

export const generalPlace = definePlace({
    name: 'testGeneralIndex',
    access: (allow) => [
        allow.authenticated.to(['search', 'geocode', 'autocomplete']),
        allow.guest.to(['search']),
        allow.apiKey.to(['geocode', 'search'])
    ],
    apiKeyProps: {
        apiKeyName: 'testIndexKey',
        noExpiry: true, // add API key proprerties
    }
});
```

## Validation

- 126 unit tests with >95% code coverage within `backend-geo`
- Additional tests added into `client-config` package pass tests
- Successful ESLint verification
- Successful API generation
- Successful deployment & amplify output validation on Sandbox project testing with above DX

### Unit Test Coverage
- Additional customer-end `defineX()` API testing
- Additional access orchestration testing
- API key generation testing (empty vs. non-empty)
- Error and edge-case handling tests for API keys
- Amplify backend output schema testing
- Amplify client configuration versioned testing

## Checklist

### DIFF
--- a/packages/backend-geo/src/geo_outputs_aspect.test.ts
+++ b/packages/backend-geo/src/geo_outputs_aspect.test.ts
@@ -195,7 +195,7 @@ void describe('AmplifyGeoOutputsAspect', () => {
       assert.equal(
         JSON.parse(
           addBackendOutputEntryMock.mock.calls[0].arguments[1].payload.maps,
-        ).items[0].api_key_name,
+        ).items[0].apiKeyName,
         'myKey',
       );
 
--- a/packages/backend-geo/src/geo_outputs_aspect.ts
+++ b/packages/backend-geo/src/geo_outputs_aspect.ts
@@ -12,8 +12,7 @@ export type GeoResource = AmplifyMap | AmplifyPlace | AmplifyCollection;
 
 type ResourceOutputs = {
   name?: string;
-  // eslint-disable-next-line @typescript-eslint/naming-convention
-  api_key_name?: string;
+  apiKeyName?: string;
 };
 
 /**
@@ -153,7 +152,7 @@ export class AmplifyGeoOutputsAspect implements IAspect {
     const mapOutputs: ResourceOutputs[] = maps.map((map): ResourceOutputs => {
       return {
         name: map.name,
-        api_key_name: map.resources.cfnResources.cfnAPIKey?.keyName,
+        apiKeyName: map.resources.cfnResources.cfnAPIKey?.keyName,
       };
     });
 
@@ -161,7 +160,7 @@ export class AmplifyGeoOutputsAspect implements IAspect {
       (place): ResourceOutputs => {
         return {
           name: place.name,
-          api_key_name: place.resources.cfnResources.cfnAPIKey?.keyName,
+          apiKeyName: place.resources.cfnResources.cfnAPIKey?.keyName,
         };
       },
     );
--- a/packages/client-config/API.md
+++ b/packages/client-config/API.md
@@ -11,7 +11,7 @@ import { DeployedBackendIdentifier } from '@aws-amplify/deployed-backend-client'
 import { S3Client } from '@aws-sdk/client-s3';
 
 // @public
-type AmazonCognitoStandardAttributes = "address" | "birthdate" | "email" | "family_name" | "gender" | "given_name" | "locale" | "middle_name" | "name" | "nickname" | "phone_number" | "picture" | "preferred_username" | "profile" | "sub" | "updated_at" | "website" | "zoneinfo";
+type AmazonCognitoStandardAttributes = 'address' | 'birthdate' | 'email' | 'family_name' | 'gender' | 'given_name' | 'locale' | 'middle_name' | 'name' | 'nickname' | 'phone_number' | 'picture' | 'preferred_username' | 'profile' | 'sub' | 'updated_at' | 'website' | 'zoneinfo';
 
 // @public
 type AmazonCognitoStandardAttributes_2 = 'address' | 'birthdate' | 'email' | 'family_name' | 'gender' | 'given_name' | 'locale' | 'middle_name' | 'name' | 'nickname' | 'phone_number' | 'picture' | 'preferred_username' | 'profile' | 'sub' | 'updated_at' | 'website' | 'zoneinfo';
@@ -65,7 +65,7 @@ interface AmazonLocationServiceConfig_6 {
 }
 
 // @public
-type AmazonPinpointChannels = "IN_APP_MESSAGING" | "FCM" | "APNS" | "EMAIL" | "SMS";
+type AmazonPinpointChannels = 'IN_APP_MESSAGING' | 'FCM' | 'APNS' | 'EMAIL' | 'SMS';
 
 // @public
 type AmazonPinpointChannels_2 = 'IN_APP_MESSAGING' | 'FCM' | 'APNS' | 'EMAIL' | 'SMS';
@@ -83,7 +83,7 @@ type AmazonPinpointChannels_5 = 'IN_APP_MESSAGING' | 'FCM' | 'APNS' | 'EMAIL' | 
 type AmazonPinpointChannels_6 = 'IN_APP_MESSAGING' | 'FCM' | 'APNS' | 'EMAIL' | 'SMS';
 
 // @public
-type AmplifyStorageAccessActions = "read" | "get" | "list" | "write" | "delete";
+type AmplifyStorageAccessActions = 'read' | 'get' | 'list' | 'write' | 'delete';
 
 // @public
 type AmplifyStorageAccessActions_2 = 'read' | 'get' | 'list' | 'write' | 'delete';
@@ -286,19 +286,19 @@ interface AWSAmplifyBackendOutputs {
             require_symbols: boolean;
         };
         oauth?: {
-            identity_providers: ("GOOGLE" | "FACEBOOK" | "LOGIN_WITH_AMAZON" | "SIGN_IN_WITH_APPLE")[];
+            identity_providers: ('GOOGLE' | 'FACEBOOK' | 'LOGIN_WITH_AMAZON' | 'SIGN_IN_WITH_APPLE')[];
             domain: string;
             scopes: string[];
             redirect_sign_in_uri: string[];
             redirect_sign_out_uri: string[];
-            response_type: "code" | "token";
+            response_type: 'code' | 'token';
         };
         standard_required_attributes?: AmazonCognitoStandardAttributes[];
-        username_attributes?: ("email" | "phone_number" | "username")[];
-        user_verification_types?: ("email" | "phone_number")[];
+        username_attributes?: ('email' | 'phone_number' | 'username')[];
+        user_verification_types?: ('email' | 'phone_number')[];
         unauthenticated_identities_enabled?: boolean;
-        mfa_configuration?: "NONE" | "OPTIONAL" | "REQUIRED";
-        mfa_methods?: ("SMS" | "TOTP")[];
+        mfa_configuration?: 'NONE' | 'OPTIONAL' | 'REQUIRED';
+        mfa_methods?: ('SMS' | 'TOTP')[];
         groups?: {
             [k: string]: AmplifyUserGroupConfig;
         }[];
@@ -345,7 +345,7 @@ interface AWSAmplifyBackendOutputs {
         bucket_name: string;
         buckets?: AmplifyStorageBucket[];
     };
-    version: "1.5";
+    version: '1.5';
 }
 
 // @public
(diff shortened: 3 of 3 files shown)


.changeset/fluffy-icons-wave.md
```diff
@@ -0,0 +1,18 @@
...
+'@aws-amplify/ai-constructs': patch
+'@aws-amplify/platform-core': patch
+'@aws-amplify/backend-auth': patch
+'@aws-amplify/backend-data': patch
+'@aws-amplify/plugin-types': patch
+'@aws-amplify/backend-geo': patch
+'@aws-amplify/backend-ai': patch
+'@aws-amplify/backend': patch
+'@aws-amplify/backend-cli': patch
+---
```
rtpascual, line 18: There should be a comment for every changeset on what changed here, in this case upgrading `aws-cdk-lib` peer dependencies.

packages/backend-geo/API.md
```diff
@@ -37,7 +43,9 @@ export type AmplifyMapFactoryProps = Omit<AmplifyMapProps, 'outputStorageStrateg
 // @public (undocumented)
 export type AmplifyMapProps = {
     name: string;
+    isDefault?: boolean;
```
rtpascual: I thought we can only define one map and place? Is there still a need for `isDefault` in these props?

packages/backend-geo/src/geo_access_policy_factory.test.ts
```diff
@@ -495,4 +495,134 @@ void describe('GeoAccessPolicyFactory', () => {
...
+        'geo-places:SearchText',
+        'geo-places:Suggest',
+      ]);
+    });
+
+    void it('generateKeyActions returns empty array for empty input', () => {
+      const keyActions = geoAccessPolicyFactory.generateKeyActions([]);
+
+      assert.deepStrictEqual(keyActions, []);
+    });
```
rtpascual: I'm wondering if we should allow this or throw to force users to add an action?

packages/client-config/src/client-config-schema/client_config_v1.5.ts
```diff
@@ -0,0 +1,322 @@
...
+     */
+    maps?: {
+      /**
+       * @minItems 1
+       */
+      items?: {
+        name?: string;
+        key?: string;
+        [k: string]: unknown;
+      }[];
```
rtpascual: Looks like we dropped `[k: string]: AmazonLocationServiceConfig;` here, is there a reason we are making this typing more loose?

packages/client-config/src/client-config-writer/client_config_to_legacy_converter.ts
```diff
@@ -23,9 +23,9 @@ export class ClientConfigLegacyConverter {
    */
   convertToLegacyConfig = (clientConfig: ClientConfig): ClientConfigLegacy => {
     // We can only convert from V1.4 of ClientConfig. For everything else, throw
```
rtpascual:
```suggestion
    // We can only convert from V1.5 of ClientConfig. For everything else, throw
```

packages/client-config/src/client-config-schema/client_config_v1.5.ts
```diff
@@ -0,0 +1,322 @@
...
+    /**
+     * Maps from Amazon Location Service
+     */
+    maps?: {
+      /**
+       * @minItems 1
+       */
+      items?: {
+        name?: string;
+        key?: string;
```
rtpascual: I believe we should rename to `api_key` like in the data part of the schema.

packages/client-config/src/client-config-schema/client_config_v1.5.ts
manavgurnani21: The `AmazonLocationServiceConfig` involved indexed map styles for every legacy map that was provisioned. This was a type used in older schemas and was not needed for the new version of the items schema (which only required names and API keys). The extra `[k: string]: unknown` was likely a schema conversion error since I intended on dropping it. I could alternatively change the internal type of `AmazonLocationServiceConfig` for the new schema and use that internally, but that could conflict with legacy types with the same name.

packages/backend-geo/src/geo_access_policy_factory.test.ts
manavgurnani21: We could throw a warning/error for users that have empty actions since each of the `allow` statements are optional anyways, so the user should only be calling them when they have actions to assign to them. Would this be the ideal DX for these functions?

packages/backend-geo/API.md
```diff
@@ -92,9 +104,37 @@ export type GeoActionBuilder = {
...
+    cfnResources: {
+        cfnAPIKey?: CfnAPIKey;
+    };
+};
+
+// @public (undocumented)
+export type ResourceOutputs = {
+    name: string;
+    key?: string;
+};
```
sobolk: Does this have to be public?
I don't see this being used by anything public.

packages/backend-geo/src/geo_outputs_aspect.test.ts
```diff
@@ -179,34 +263,82 @@ void describe('AmplifyGeoOutputsAspect', () => {
...
+            maps: JSON.stringify({
+              default: "testMapResource",
+              items: [{
+                name: "testMapResource",
+                apiKey: "TOKEN_STRING"
+              }]
+            })
+          }
+        }
+       */
```
sobolk: commented out code should be removed.

packages/backend-geo/src/geo_access_policy_factory.test.ts
rtpascual: Yes I believe we should throw an error here, if users write `allow.x.to()` there is intent to grant x some permission.

sobolk (commented):
Overall looks. Good.
Left couple of comments in addition to what @rtpascual already posted.

packages/client-config/src/client-config-schema/client_config_v1.5.ts
rtpascual: I see, since this is not a major version bump of the schema I say we should add `key` to `AmazonLocationServiceConfig` as optional and keep `items` level as
https://github.com/aws-amplify/amplify-backend/blob/f8d2b02fd55705b1d81dcde528b685e163b5a031/packages/client-config/src/client-config-schema/client_config_v1.4.ts#L223-L228.

Since the goal is that customers can switch from CDK to Amplify Geo constructs and their frontend does not break.

manavgurnani21: This makes sense, the only other concern is that the actual hidden value of the key cannot be accessed through CDK, so the `ref` value from the API key construct resolves to the name of the API key itself. Users would then have to use the CLI or SDK to get this API key value and use it within their resources. The items object would contain a key pointing to the same value, which I now realize is redundant. I could change items to be of type `AmazonLocationServiceConfig` and add that directly (maintains item layer but different typing) or we could find another way to get the hidden value or another structure. Let me know your thoughts on the same.

.changeset/fluffy-icons-wave.md
manavgurnani21: Done. PTAL.
https://github.com/aws-amplify/amplify-backend/blob/9d562af95b47dea765a0a04dc47cb9826aca2659/.changeset/fluffy-icons-wave.md?plain=1#L1-L20

packages/backend-geo/API.md
manavgurnani21: Removed. PTAL.
https://github.com/aws-amplify/amplify-backend/blob/9d562af95b47dea765a0a04dc47cb9826aca2659/packages/backend-geo/API.md?plain=1#L43-L49

packages/backend-geo/src/geo_outputs_aspect.test.ts
manavgurnani21: This was more of a note on structure to clarify where the TOKEN assertion was coming from since the value was loosely parsed. Would this be helpful or redundant?
https://github.com/aws-amplify/amplify-backend/blob/7566d1be63aa7151a7a628277c9f58dfcbc8dfa1/packages/backend-geo/src/geo_outputs_aspect.test.ts#L307-L319

packages/backend-geo/src/geo_access_policy_factory.test.ts
manavgurnani21: Done. PTAL.
https://github.com/aws-amplify/amplify-backend/blob/ae1e1881c8b60e6353e444a8d3f18cba73be8986/packages/backend-geo/src/geo_access_orchestrator.ts#L68-L75

packages/client-config/src/client-config-writer/client_config_to_legacy_converter.ts
manavgurnani21: Done. PTAL.
https://github.com/aws-amplify/amplify-backend/blob/ae1e1881c8b60e6353e444a8d3f18cba73be8986/packages/client-config/src/client-config-writer/client_config_to_legacy_converter.ts#L24-L30

sobolk (approved):

packages/client-config/src/client-config-schema/client_config_v1.5.ts
rtpascual: By hidden value of the key, do you mean during synthesis this is a CDK token? I believe there's no concern if we treat this key the same as the AppSync API key in data. Since it is added to stack outputs and also in `amplify_outputs.json`.

manavgurnani21: Changes made, PTAL. The change in the schema is now primarily for search indices as they include API keys as well, they have been made similar to the maps schema.
https://github.com/aws-amplify/amplify-backend/blob/8e594a3b430e2ea1ef1881a7555eb0221c365ec9/packages/client-config/src/client-config-schema/client_config_v1.5.ts#L226-L242

packages/client-config/src/client-config-schema/client_config_v1.5.ts
manavgurnani21: Renamed to `api_key_name` since hidden values are not CDK-accessible.
https://github.com/aws-amplify/amplify-backend/blob/8e594a3b430e2ea1ef1881a7555eb0221c365ec9/packages/client-config/src/client-config-schema/client_config_v1.5.ts#L299-L309

packages/backend-geo/API.md
manavgurnani21: Removed.

packages/backend-geo/src/geo_outputs_aspect.ts
```diff
@@ -6,10 +6,16 @@ import { AmplifyPlace } from './place_resource.js';
...
 import { GeoOutput, geoOutputKey } from '@aws-amplify/backend-output-schemas';
-import { GeoResourceType, ResourceOutputs } from './types.js';
+import { GeoResourceType } from './types.js';
 
 export type GeoResource = AmplifyMap | AmplifyPlace | AmplifyCollection;
 
+type ResourceOutputs = {
+  name?: string;
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  api_key_name?: string;
```
rtpascual: Nit: we can keep naming conventions here, i.e. `apiKeyName`, and in the contributer when we write to `amplify_outputs.json` that's when we change the field name to `api_key_name`.

This way we can remove this eslint-disable and keep naming consistent for each "layer" from outputs -> client config

manavgurnani21: Done. PTAL.
https://github.com/aws-amplify/amplify-backend/blob/6a843f15c52d96b0571cf1b0d1b580092b2c09da/packages/backend-geo/src/geo_outputs_aspect.ts#L12-L17

PR #2939, opened 2025-08-01
[DevTools] PR3.4 Logs NEW

megha-narayanan:
See this pull request: https://github.com/aws-amplify/amplify-backend/pull/2914

Same as above PR, but moved changes to an unstream branch instead of a personal fork


packages/create-amplify/src/default_packages.json
```diff
@@ -2,7 +2,7 @@
   "defaultDevPackages": [
     "@aws-amplify/backend",
     "@aws-amplify/backend-cli",
-    "aws-cdk-lib@2.204.0",
+    "aws-cdk-lib@2.201.0",
```
ShadowCat567, line 5: Any particular reason why `aws-cdk-lib` is getting downgraded here?

packages/cli/src/commands/sandbox/sandbox-devtools/sandbox_devtools_command.ts
```diff
@@ -516,6 +516,8 @@ export class SandboxDevToolsCommand implements CommandModule<object> {
       shutdownService,
       resourceService,
       storageManager,
+      undefined,
+      undefined,
```
ShadowCat567, line 520: I am assuming these are `activeLogPollers` and `toggleStartTimes`, do they get defined later on?

PR #2951, opened 2025-08-11
feat(ai): add cross-region inference support for conversation routes

esslam-ashour:
## Problem

This PR enables cross-region inference support for conversation routes in AI Kit by addressing the following two main limitations:

1. **Model ARN Resolution:**  
   `ConversationHandlerFunction` previously relied on manually constructed Bedrock ARNs for IAM policies, with no centralized, dynamic way to determine the correct set of ARNs, which is needed for cross-region inference as it requires access to wider model ARNs across multiple regions.

2. **Model ID Resolution:**  
   Runtime components such as `BedrockConverseAdapter` used `modelId` values from event configs directly, without resolving whether they should resolve to a foundation model ID or an inference profile ID based on cross-region inference availability and configuration. This made it impossible to use cross-region inference, which requires using inference profile IDs in place of foundation model IDs.

At its core, this solution depends on a static mapping of model IDs `models.json` and their cross-region inference configurations.

### Centralized Model ARN & ID Resolution
- **New:** `AiModelArnGeneratorConstruct` (CDK construct) with a Lambda-backed custom resource to:
  - Take `{ modelId, region, crossRegionInference }`
  - Return the full set of required ARNs for IAM policies
  - Support both foundation models and inference profiles
- **New:** `AiModelPropsResolver` class to:
  - Determine CRI requirements/support for a given model and resolve the appropriate model ID
  - Map between foundation model IDs and inference profile IDs
  - Return supported source regions for a given inference profile

### Deployment-Time Integration
- `ConversationHandlerFunction` updated to:
  - Accept `crossRegionInference?: boolean` in model configs
  - Use `AiModelArnGeneratorConstruct` output instead of hardcoded ARNs
- Schema & types updated to carry `crossRegionInference` from user-defined schema through to deployment.

### Runtime Integration
- `BedrockConverseAdapter` updated to:
  - Accept `crossRegionInference` in `ConversationTurnEvent.modelConfiguration`
  - Build an `AiModelConfig` at runtime and use `AiModelPropsResolver` to resolve the appropriate `modelId` before making Bedrock API calls.

### Lambda Implementation & Tests
- Lambda handler for ARN generator:
  - Generates ARNs
- Comprehensive unit tests for:
  - Construct and Lambda handler
  - Resolver logic (all CRI scenarios, invalid inputs, edge cases)
  - JSON schema (`models.json`) structure & data validity
  - `BedrockConverseAdapter` runtime behavior with and without CRI

## Validation
- **Unit Tests:** Verified both deployment-time ARN and runtime model ID resolution logic.
- **Manual Testing:**
  - Deployed stacks with CRI enabled/disabled in multiple regions
  - Tested sample AI app with all possible configurations to verify user-facing functionality

## Related PRs
https://github.com/aws-amplify/amplify-data/pull/610
https://github.com/aws-amplify/amplify-data/pull/616
https://github.com/aws-amplify/amplify-category-api/pull/3323

## Checklist


package.json
```diff
@@ -78,6 +78,7 @@
     "@types/prettier": "^3.0.0",
     "@typescript-eslint/eslint-plugin": "^8.23.0",
     "@typescript-eslint/parser": "^8.23.0",
+    "at-least-node": "^1.0.0",
```
sobolk: what is this for?

packages/ai-constructs/package.json
```diff
@@ -17,13 +17,17 @@
...
+      "types": "./lib/ai-model/index.d.ts",
+      "default": "./lib/ai-model/index.js"
     }
   },
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
   "scripts": {
     "post:compile": "npm run bundle:default-handler",
-    "bundle:default-handler": "esbuild src/conversation/runtime/default_handler.ts --bundle --platform=node --target=node20 --outfile=lib/conversation/runtime/default_handler_bundled/index.js",
+    "bundle:default-handler": "esbuild src/conversation/runtime/default_handler.ts --bundle --platform=node --target=node20 --external:@aws-sdk/* --external:@smithy/* --external:aws-lambda --external:buffer --external:crypto --external:fs --external:http --external:https --external:path --external:stream --external:url --external:util --minify --tree-shaking=true --outfile=lib/conversation/runtime/default_handler_bundled/index.js",
```
sobolk: What is the reason to add these flags here?
For example, we intentionally bundle pinned bedrock sdk, we shouldn't be marking it as external.

packages/ai-constructs/src/ai-model/ai_model_arn_generator_construct.ts
```diff
@@ -0,0 +1,77 @@
...
+    this.resourceCounter += 1;
+    const resourceId = `AiModelArnsResource${this.resourceCounter}`;
+
+    const resource = new CustomResource(this, resourceId, {
+      serviceToken: this.provider.serviceToken,
+      resourceType: AI_MODEL_ARN_GENERATOR_RESOURCE_TYPE,
+      properties: {
+        modelConfig,
+        // Helps CloudFormation detect property updates consistently
+        timestamp: Date.now(),
```
sobolk: We should remove or replace it with something else. I.e. we don't want custom resource to trigger on every deployment because results are highly cachable.

Instead of this. I think we should use hash of CRI mapping file.
Or just nothing because presumably when we update lambda sources, CFN will detect it and re-deploy.

packages/ai-constructs/src/ai-model/ai_model_types.ts
```diff
@@ -0,0 +1,32 @@
...
+  requiresCriIn: string[];
+};
+
+/** Model entry containing geography-specific configurations. */
+export type AiModelsEntry = {
+  geographies: Record<string, AiModelMapping>;
+};
+
+/** Root structure of the models JSON data. */
+export type AiModelsJson = {
```
sobolk:
```suggestion
export type AiModelsManifest = {
```

packages/platform-core/API.md
```diff
@@ -256,16 +256,16 @@ export const telemetryPayloadSchema: z.ZodObject<{
     }, "strip", z.ZodTypeAny, {
         payloadVersion: string;
         sessionUuid: string;
-        timestamp: string;
         eventId: string;
+        timestamp: string;
```
sobolk: changes in this file seem unwanted.

there's known problem that api extractor behaves inconsistently in local workspace.

`npm run clean && npm install && npm run build && npm run update:api` should fix this. I.e. rebuilding and regenerating api extracts from fresh workspace.

sobolk (commented):
Overall looks good.

packages/ai-constructs/package.json
```diff
@@ -32,6 +36,7 @@
     "@aws-amplify/plugin-types": "^1.10.1",
     "@aws-sdk/client-bedrock-runtime": "3.622.0",
     "@smithy/types": "^4.1.0",
+    "aws-lambda": "^1.0.7",
```
sobolk: could we try `@types/aws-lambda` as dev dep.

OR

we could move this `aws-lambda` to dev deps.
and use `import type {} from 'aws-lambda'` instead of normal imports.

packages/ai-constructs/package.json
eashour9: Removed, attempted to reduce size of bundled default handler but turns out the culprit was `aws-lambda` package

packages/ai-constructs/package.json
eashour9: Using only `@types/aws-lambda` in `devDependencies` doesn't seem to work, `aws-lambda` needs to be there as well. We do have a similar example in the `backend` package:

https://github.com/aws-amplify/amplify-backend/blob/6e5be9bb14ff2aeaaff5826b5de0d74a82e288ea/packages/backend/package.json#L50-L54

Added both to `devDependencies` for `ai-constructs`.

sobolk (dismissed):

.changeset/yellow-corners-like.md
```diff
@@ -0,0 +1,13 @@
+---
+'@aws-amplify/ai-constructs': minor
+'@aws-amplify/backend': patch
+'@aws-amplify/backend-ai': minor
+'create-amplify': patch
```
rtpascual: Nit: looks like there are only changes to `ai-constructs` and `backend-ai`, were the changes to `backend` and `create-amplify` lost or dropped at some point?

eashour9: Yup changes were dropped, will update changeset

rtpascual (dismissed):
LGTM!

sobolk (dismissed):

rtpascual (dismissed):

sobolk (approved):

rtpascual (approved):

PR #3019, opened 2025-10-07
fix: check_api_changes times out when steps are retried

iankhou:
## Problem

Screenshot 2025-10-07 at 11 55 40

The **check_api_changes** job is sometimes cancelled when the job exceeds the timeout of 20 minutes. This can happen when steps are retried (introduced in https://github.com/aws-amplify/amplify-backend/pull/3010).

## Changes

Increased the timeout from 20 minutes to 60 minutes. Being generous here, as a successful run of step `Check API changes with retry` should take around *14 minutes*. If we try that step three times, a 60 min timeout should be able to cover the time needed for 3 tries and other steps in the run.

## Validation

Run workflows on this PR.

## Checklist

- [X] If this PR requires a change to the [Project Architecture README](../PROJECT_ARCHITECTURE.md), I have included that update in this PR.
- [X] If this PR requires a docs update, I have linked to that docs PR above.

### DIFF
--- a/.github/workflows/health_checks.yml
+++ b/.github/workflows/health_checks.yml
@@ -245,7 +245,7 @@ jobs:
       - build
       - resolve_inputs
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    timeout-minutes: 60
     steps:
       - name: Checkout pull request ref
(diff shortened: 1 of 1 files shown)


.github/workflows/health_checks.yml
iankhou, line 273: Since this can run 3 times, we need the timeout for the job to be ~3x this.

PR #3058, opened 2025-11-25
chore: remove dependency on node built-in

rix0rrr:
The health checks are now failing with

```
error TS2307: Cannot find module 'node:crypto' or its corresponding type declarations.
```

What the test does is `cdk init` a project, then remove the `node_modules` directory in order to source the node modules from elsewhere.

Rather than try to figure out where the package is installed and what changed where in order to get the right definition of `@types/node` anywhere, I figured it would be easier to just cut the single dependency on the node built-in, and make the test sliiiightly more robust against what types are installed.


packages/integration-tests/src/test-cdk-projects/auth-cdk/test-cdk-project-auth-cdk-stack.ts
```diff
@@ -32,3 +32,12 @@ export class TestCdkProjectAuthCdkStack extends cdk.Stack {
     new AmplifyAuth(this, 'test-auth', authProps);
   }
 }
+
+function randomHex(n: number) {
```
mrgrain, line 36: not really that random, but yay!

mrgrain (approved):

PR #3053, opened 2025-11-17
feat: add passwordless authentication support

ahmedhamouda78:
## Problem

Adds first-class support for passwordless authentication in Amplify Gen 2 backend, enabling developers to configure Email OTP, SMS OTP, and WebAuthn (passkeys) without requiring CDK overrides.

**Issue number, if available:**

#2276 

## Changes

**Public API Changes:**
- Added `otpLogin?: boolean` property to email and phone login configurations in `defineAuth`
- Added `webAuthn?: WebAuthnLogin` property to `loginWith` configuration
- Added `WebAuthnOptions` type with `relyingPartyId` and `userVerification` settings

**Implementation:**
- Extended `auth-construct` to translate passwordless configurations to Cognito User Pool settings
- Added validation logic to prevent invalid configurations (e.g., WebAuthn-only without sign-up method, passwordless + MFA REQUIRED)
- Implemented automatic relying party ID resolution ('AUTO' resolves to localhost in sandbox, Amplify domain in branch mode)
- Added runtime warnings for immutable configuration changes

**Testing:**
- Added tests to verify correct Cognito User Pool and User Pool Client configuration

**Corresponding docs PR, if applicable:** N/A

## Validation

- Unit tests added for validation logic and configuration translation
- Tests verify:
  - Correct `AllowedFirstAuthFactors` configuration in User Pool (EMAIL_OTP, SMS_OTP, WEB_AUTHN)
  - `ALLOW_USER_AUTH` flow enabled in User Pool Client
  - WebAuthn configuration (RelyingPartyId, UserVerification) when applicable
- Manual testing with sample app verified end-to-end passwordless flows

## Checklist

- [x] If this PR includes a functional change to the runtime behavior of the code, I have added or updated automated test coverage for this change.
- [ ] If this PR requires a change to the [Project Architecture README](../PROJECT_ARCHITECTURE.md), I have included that update in this PR.
- [ ] If this PR requires a docs update, I have linked to that docs PR above.
- [ ] If this PR modifies E2E tests, makes changes to resource provisioning, or makes SDK calls, I have run the PR checks with the `run-e2e` label set.

_By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license._

PR #3065, opened 2025-12-08
feat(auth): add passwordless authentication options to client config

ahmedhamouda78:
## Problem

Add passwordless authentication configuration to the generated `amplify_outputs.json` file to enable frontend libraries to configure passwordless authentication flows (Email OTP, SMS OTP, and WebAuthn) and specify the preferred authentication challenge.

Related to #3053

## Changes

This PR adds passwordless configuration and preferred challenge to the client config output:

- Added `passwordlessOptions` field to auth output schema v1
- Added `preferredChallenge` prop to `AuthProps` (accepts `'EMAIL_OTP' | 'SMS_OTP' | 'PASSWORD' | 'WEB_AUTHN'`, defaults to `undefined`)
- Outputs `emailOtpEnabled`, `smsOtpEnabled`, `webAuthn` configuration, and `preferredChallenge`

N/A

## Validation

- Added unit test

## Checklist

### DIFF
--- a/.github/actions/run_with_e2e_account/action.yml
+++ b/.github/actions/run_with_e2e_account/action.yml
@@ -4,6 +4,9 @@ inputs:
   run:
     description: Script to run
     required: true
+  on_retry_command:
+    description: Script to run before a retry (e.g. a cleanup script).
+    required: false
   shell:
     description: Shell
     required: false
@@ -177,6 +180,13 @@ runs:
           if [ $ATTEMPT -lt $RETRY_ATTEMPTS ]; then
             echo "Retrying in $RETRY_DELAY seconds..."
             sleep $RETRY_DELAY
+            # Run on retry command if provided
+            if [ -n "${{ inputs.on_retry_command }}" ]; then
+              echo "Running on retry command..."
+              if ! run_with_timeout '${{ inputs.on_retry_command }}' "$SHELL_CMD"; then
+                echo "On retry command failed, but continuing with retry attempt"
+              fi
+            fi
           else
             echo "Script failed after $RETRY_ATTEMPTS attempts"
             exit 1
--- a/.github/workflows/health_checks.yml
+++ b/.github/workflows/health_checks.yml
@@ -700,6 +700,7 @@ jobs:
           node_version: ${{ matrix.node-version }}
           cdk-lib-version: ${{ needs.resolve_inputs.outputs.cdk_lib_version }}
           shell: bash
+          on_retry_command: git clean -f -d
           attempt_timeout_minutes: ${{ matrix.os == 'windows-2025' && 40 || 25 }}
           run: PACKAGE_MANAGER=${{matrix.pkg-manager}} npm run test:dir packages/integration-tests/src/package_manager_sanity_checks.test.ts
   lint:


packages/auth-construct/src/construct.ts
```diff
@@ -142,6 +142,11 @@ export class AmplifyAuth
    */
   private customSenderKMSkey: IKey | undefined;
 
+  /**
+   * The preferred authentication challenge
+   */
+  private readonly preferredChallenge: string | undefined;
```
ShadowCat567, line 148: Can you add some kind of happy path test for this that ensures if we get a `preferredChalllenge` then it is properly set? (this would probably go in `construct.test.ts`)

ShadowCat567 (commented):
Overall this looks good, remember to update the PR description since the default for `preferredChallenge` is `undefined` not `EMAIL_ONLY`.
My only comment is to add some unit testing related to `preferredChallenge` to the auth construct tests

packages/auth-construct/src/construct.test.ts
```diff
@@ -3329,5 +3330,94 @@ void describe('Auth construct', () => {
...
       });
     });
+
+    void it('includes preferredChallenge in client config when specified', () => {
+      const app = new App();
+      const stack = new Stack(app);
+      new AmplifyAuth(stack, 'test', {
+        loginWith: { email: true },
+        passwordlessOptions: {
+          preferredChallenge: 'EMAIL_OTP',
```
ShadowCat567, line 3344: Can we add some kind of warning for users if they try to set their preferred challenge to something that they have not enabled? 
EX: in this scenario, they should get a warning when they try to deploy their app if their preferred challenge is `SMS_OTP` or `WEBAUTHN` since those two are not enabled

ShadowCat567 (commented):
Generally looks good, looking at the new unit tests made me realize we should probably have some kind of mechanism to warn users if they try to deploy their app with a preferred challenge that is not in their available challenges list

packages/auth-construct/src/construct.ts
ahmedhamouda78: Added unit tests for `preferredChalllenge`

packages/auth-construct/src/construct.test.ts
ahmedhamouda78: Implemented a warning in case of challenge mismatch

packages/auth-construct/src/construct.ts
```diff
@@ -154,6 +159,11 @@ export class AmplifyAuth
     super(scope, id);
     this.name = props.name ?? '';
     this.domainPrefix = props.loginWith.externalProviders?.domainPrefix;
+    this.preferredChallenge = props.passwordlessOptions?.preferredChallenge;
+
+    // Validate preferredChallenge against enabled authentication methods
+    this.validatePreferredChallenge(props);
```
ShadowCat567, line 163: can you have this validation happen before `this.preferredChallenge` is set?

packages/auth-construct/src/construct.ts
```diff
@@ -1257,6 +1267,53 @@ export class AmplifyAuth
...
+    if (!enabledChallenges.includes(this.preferredChallenge)) {
+      process.stderr.write('\nWARNING:\n');
+      process.stderr.write(
+        `  • Preferred challenge "${this.preferredChallenge}" is not enabled in your authentication configuration.\n`,
+      );
+      process.stderr.write(
+        `    Enabled challenges: ${enabledChallenges.join(', ')}\n`,
+      );
+      process.stderr.write(
+        `    This will result in a broken authentication flow in the frontend.\n`,
```
ShadowCat567: if this breaks authentication flow in the frontend, then I think this warrants an error instead of a warning. I like the text you have for the body, so it can remain, just in the form of an error instead of a warning (use `throw new Error`, not the `AmplifyUserErrors` in this packages)

packages/auth-construct/src/construct.ts
ahmedhamouda78: I moved it

packages/auth-construct/src/construct.ts
ahmedhamouda78: Switched to an error instead of a warning

ShadowCat567 (approved):

Issue #3068, opened 2025-12-18
Support Lambda Durable Functions


fossamagna:
### Environment information

```plain text
System:
  OS: Linux 6.6 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
  CPU: (8) arm64 unknown
  Memory: 18.94 GB / 23.43 GB
  Shell: /bin/bash
Binaries:
  Node: 22.16.0 - /usr/local/bin/node
  Yarn: 1.22.22 - /usr/local/bin/yarn
  npm: 10.9.2 - /usr/local/bin/npm
  pnpm: undefined - undefined
NPM Packages:
  @aws-amplify/auth-construct: 1.8.1
  @aws-amplify/backend: 1.16.1
  @aws-amplify/backend-ai: Not Found
  @aws-amplify/backend-auth: 1.7.1
  @aws-amplify/backend-cli: 1.8.0
  @aws-amplify/backend-data: 1.6.1
  @aws-amplify/backend-deployer: 2.1.3
  @aws-amplify/backend-function: 1.14.1
  @aws-amplify/backend-output-storage: 1.3.1
  @aws-amplify/backend-secret: 1.4.0
  @aws-amplify/backend-storage: 1.4.1
  @aws-amplify/cli-core: 2.2.1
  @aws-amplify/client-config: 1.8.0
  @aws-amplify/data-construct: 1.16.3
  @aws-amplify/data-schema: 1.21.0
  @aws-amplify/deployed-backend-client: 1.8.0
  @aws-amplify/form-generator: 1.2.1
  @aws-amplify/model-generator: 1.2.0
  @aws-amplify/platform-core: 1.10.0
  @aws-amplify/plugin-types: 1.11.0
  @aws-amplify/sandbox: 2.1.2
  @aws-amplify/schema-generator: 1.4.0
  @aws-cdk/toolkit-lib: 1.1.1
  aws-amplify: 6.15.3
  aws-cdk-lib: 2.189.1
  typescript: 5.8.3
No AWS environment variables
No CDK environment variables
```

### Describe the feature

Support configure for [Lambda Durable Functions](https://docs.aws.amazon.com/lambda/latest/dg/durable-functions.html) in defineFunction.

### Use case

long time tasks without custom CDK construct. (ex: invoke from a scheduler, etc)

Alternatives:
- Configure a Step Functions StateMachine using a CDK Construct
- Use the custom function feature. Note that some features like Amplify secrets and environment variables are unavailable in this case.

PR #2968, opened 2025-08-22
Update Lambda logging configuration to use non-deprecated AWS CDK properties

nsteffens:
## Problem

**Issue number, if available:** #2965

## Changes

- Refactored `logging_options_parser.ts` to use modern, non-deprecated AWS CDK properties for Lambda logging
- Replaced the deprecated `logRetention` property with a dedicated `LogGroup` resource
- Updated `applicationLogLevelV2` and `loggingFormat` usage to follow AWS CDK best practices
- Updated tests to verify the new implementation works correctly
- Maintained backward compatibility for end users (no API changes)

- Updated documentation to have logging field documented in general. See PR https://github.com/aws-amplify/docs/pull/8428

### Why

AWS CDK marked several properties in `NodejsFunction` as deprecated:
- [`logRetention`](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda_nodejs.NodejsFunction.html#logretentionspan-classapi-icon-api-icon-deprecated-titlethis-api-element-is-deprecated-its-use-is-not-recommended%EF%B8%8Fspan) - Recommended to use a dedicated `LogGroup` resource
- Creating a `LogGroup` automatically to be used if logRetention is set. Will fallback to default `NodeJsFunction` when `undefined`.

This change ensures that Lambda functions continue to work properly with future AWS CDK versions.

## Testing

- Updated and extended unit tests to verify the new implementation
- Manually verified that logs are correctly configured with the proper retention settings
- Ensured backward compatibility for user-facing APIs

**Corresponding docs PR, if applicable:** TBD

## Validation

Have ran tests locally. Should be tested against an AWS account before going live.

## Checklist

- [x] If this PR includes a functional change to the runtime behavior of the code, I have added or updated automated test coverage for this change.
- [x] If this PR requires a change to the [Project Architecture README](../PROJECT_ARCHITECTURE.md), I have included that update in this PR.
- [x] If this PR requires a docs update, I have linked to that docs PR above.
- [ ] If this PR modifies E2E tests, makes changes to resource provisioning, or makes SDK calls, I have run the PR checks with the `run-e2e` label set.

_By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license._

PR #3069, opened 2025-12-18
feat: add durable function configuration support

fossamagna:
## Problem

#3068 

## Changes

- introduce support for durable function configuration in the `defineFunction` props.
- upgrades the `aws-cdk-lib` dependency across all relevant packages to version ^2.232.0 (and in one place to 2.232.2) to ensure compatibility with new features.

## Validation

- Unit Tests
- Manual Tests

## Checklist

### DIFF
--- a/.changeset/every-knives-warn.md
+++ b/.changeset/every-knives-warn.md
@@ -1,0 +1,7 @@
+---
+'@aws-amplify/backend-output-schemas': minor
+'@aws-amplify/auth-construct': minor
+'@aws-amplify/client-config': minor
+---
+
+Add passwordless authentication options to client config
--- a/.github/actions/run_with_e2e_account/action.yml
+++ b/.github/actions/run_with_e2e_account/action.yml
@@ -4,6 +4,9 @@ inputs:
   run:
     description: Script to run
     required: true
+  on_retry_command:
+    description: Script to run before a retry (e.g. a cleanup script).
+    required: false
   shell:
     description: Shell
     required: false
@@ -177,6 +180,13 @@ runs:
           if [ $ATTEMPT -lt $RETRY_ATTEMPTS ]; then
             echo "Retrying in $RETRY_DELAY seconds..."
             sleep $RETRY_DELAY
+            # Run on retry command if provided
+            if [ -n "${{ inputs.on_retry_command }}" ]; then
+              echo "Running on retry command..."
+              if ! run_with_timeout '${{ inputs.on_retry_command }}' "$SHELL_CMD"; then
+                echo "On retry command failed, but continuing with retry attempt"
+              fi
+            fi
           else
             echo "Script failed after $RETRY_ATTEMPTS attempts"
             exit 1
--- a/.github/workflows/health_checks.yml
+++ b/.github/workflows/health_checks.yml
@@ -700,6 +700,7 @@ jobs:
           node_version: ${{ matrix.node-version }}
           cdk-lib-version: ${{ needs.resolve_inputs.outputs.cdk_lib_version }}
           shell: bash
+          on_retry_command: git clean -f -d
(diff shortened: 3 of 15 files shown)


packages/backend-function/src/factory.ts
```diff
@@ -500,9 +525,68 @@ class FunctionFactory implements ConstructFactory<AmplifyFunction> {
...
+      executionTimeoutSeconds: this.props.durableConfig.executionTimeoutSeconds,
+      retentionPeriodDays: this.props.durableConfig.retentionPeriodDays ?? 14,
+    };
+  };
 }
 
-type HydratedFunctionProps = Required<FunctionProps>;
+type HydratedFunctionProps = {
+  durableConfig?: Required<FunctionProps['durableConfig']>;
+} & Required<Omit<FunctionProps, 'durableConfig'>>;
```
ShadowCat567, line 592: Is the reason why you have this kind of odd configuration here due to `durableConfig` not being able to be assigned to `{}` like logging?

packages/backend-function/src/factory.test.ts
```diff
@@ -860,6 +860,275 @@ void describe('AmplifyFunctionFactory', () => {
...
+            },
+          }).getInstance(getInstanceProps),
+        new AmplifyUserError('UnsupportedDurableFunctionRuntimeError', {
+          message: `Durable functions require runtime 22 or higher. Current runtime is 20.`,
+          resolution: `Set the function runtime to 22 or higher to use durable functions.`,
+        }),
+      );
+    });
+
+    void it('throws when durableConfig is used with default runtime', () => {
```
ShadowCat567: Can you remove this test? Default runtime will soon be Node 22, since Lambda will be deprecating Node 20 towards the middle of 2026
I would generally be somewhat hesitant to have tests that declare certain things will/won't work with the default runtime since the default runtime can change

.changeset/full-ends-teach.md
ShadowCat567, line 1: `@aws-amplify/backend` will need to get a minor bump too, see my comment here for a bit more context: https://github.com/aws-amplify/amplify-backend/pull/2968#issuecomment-3228723772

ShadowCat567 (commented):
Hello @fossamagna! Overall this PR looks good! We are excited to expand Amplify Functions to support durable functions :D. I left some comments for you to take a look at. 
You have a failing test related to dependency licenses, if you want to tackle fixing it, I think the fix would be to explicitly add `Apache-2.0 AND ISC AND MIT` as an allow listed license here: https://github.com/aws-amplify/amplify-backend/blob/main/.github/dependency_review_config.yml

packages/backend-function/src/factory.ts
fossamagna: That's correct. I not be able to assign `undefined` to `durableConfig` instead of `{}`.

packages/backend-function/src/factory.test.ts
fossamagna: That makes sense. I understand the reasoning, and I will remove this test.

ShadowCat567 (commented):
Hello @fossamagna!
You still have one failing check, `check_package_versions`. I'm not entirely sure why it is failing, but my best guess is something happened to your `package-lock.json` file and caused it to get into a state that check does not like.
Can you try re-running `npm install` to update your `package-lock.json` file and committing anything that changed?

fossamagna:
@ShadowCat567 
Thanks again for your help.
I re-ran `npm install` to try to update the `package-lock.json` file, but the file was not changed.
It appears that the step `npx --verbose --package @changesets/cli -- changeset version` in the `check_package_versions` workflow is timing out.
Since another workflow (https://github.com/aws-amplify/amplify-backend/actions/runs/20757199926/job/59603023230) also seems to be timing out around the same time, the cause is unclear, but it may be an issue on the GitHub Actions side.
Could you please try re-running the workflow?

PR #3071, opened 2025-12-23
chore: add permissions to validate cdk release workflow

svidgen:
## Problem

Runs of `validate_cdk_release` have been failing since 2025-12-12. E.g., [this run](https://github.com/aws-amplify/amplify-backend/actions/runs/20468038258).

> The workflow is not valid. .github/workflows/validate_cdk_release.yml (Line: 16, Col: 3): Error calling workflow 'aws-amplify/amplify-backend/.github/workflows/health_checks.yml@ffdc9dc5deea61bef70848c5efdf67ef3ee13ad0'. The nested job 'handle_dependabot_version_update' is requesting 'contents: write, pull-requests: write', but is only allowed 'contents: read, pull-requests: none'. .github/workflows/validate_cdk_release.yml (Line: 16, Col: 3): Error calling workflow 'aws-amplify/amplify-backend/.github/workflows/health_checks.yml@ffdc9dc5deea61bef70848c5efdf67ef3ee13ad0'. The nested job 'do_include_e2e' is requesting 'pull-requests: read', but is only allowed 'pull-requests: none'.

https://github.com/actions/runner/issues/4151 suggests that this may have started failing spontaneously due to new, stricter validations.

## Changes

Adds permissions used by child workflow to the parent workflow.

**Corresponding docs PR, if applicable:**

## Validation

Kicked off a run of the validation flow in this branch. It did not immediately fail with the aforementioned error. ✅ 

## Checklist

- [ ] If this PR includes a functional change to the runtime behavior of the code, I have added or updated automated test coverage for this change.
- [ ] If this PR requires a change to the [Project Architecture README](../PROJECT_ARCHITECTURE.md), I have included that update in this PR.
- [ ] If this PR requires a docs update, I have linked to that docs PR above.
- [ ] If this PR modifies E2E tests, makes changes to resource provisioning, or makes SDK calls, I have run the PR checks with the `run-e2e` label set.

_By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license._

PR #3080, opened 2026-01-08
chore: restore missing github token permissions

ShadowCat567:
## Problem

Some of our checks are missing permissions that they need to run properly. 
I assume the missing permissions are related to why this PR was created: https://github.com/aws-amplify/amplify-backend/pull/3071

## Changes

Restored write access to `update_package_versions` and `docs_build_and_publish` since they were failing with permissions issue. And if `update_package_versions` is unable to write to PRs then we cannot release.

## Checklist

### DIFF
--- a/.github/workflows/health_checks.yml
+++ b/.github/workflows/health_checks.yml
@@ -800,7 +800,7 @@ jobs:
           node-version: 18
           cdk-lib-version: ${{ needs.resolve_inputs.outputs.cdk_lib_version }}
       - run: npm run docs
-        uses: peaceiris/actions-gh-pages@4f9cc6602d3f66b9c108549d475ec49e8ef4d45e # version 4.0.0
+      - uses: peaceiris/actions-gh-pages@4f9cc6602d3f66b9c108549d475ec49e8ef4d45e # version 4.0.0
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/health_checks.yml
```diff
@@ -878,6 +880,8 @@ jobs:
       - install
       - resolve_inputs
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
```
mrgrain: Could this be enough?
```suggestion
      pull-requests: write
```

ShadowCat567: hopefully yes, but this will be difficult to tell until this PR is merged. If it turns out we do need `contents: write` I'll make a follow up PR

ShadowCat567:
> For defense in depth, let's move the `if` condition of `docs_build_and_publish` from the step to the job. Similar to how we do this for `update_package_versions`.

Poked around in our docs and old PRs to see if I could find out why the `docs_build_and_publish` job was set up like this and could not find anything. It also seems like there shouldn't be a reason why `npm run docs` should be run with the rest of the health checks since there shouldn't really be a reason why it should fail unless `typedocs` goes down.
Will be making this change, wanted to leave behind this info for the next person that may have to make a similar one in this repo.

Issue #3117, opened 2026-02-16
[Auth Seed] Wrongly getting `Attributes did not conform to the schema`


ClaPalo:
### Environment information

```plain text
System:
  OS: macOS 26.2
  CPU: (8) arm64 Apple M1
  Memory: 233.63 MB / 16.00 GB
  Shell: /bin/zsh
Binaries:
  Node: 22.17.0 - /Users/claudio/.nvm/versions/node/v22.17.0/bin/node
  Yarn: 1.22.22 - /Users/claudio/.nvm/versions/node/v22.17.0/bin/yarn
  npm: 10.9.2 - /Users/claudio/.nvm/versions/node/v22.17.0/bin/npm
  pnpm: 10.29.3 - /Users/claudio/.nvm/versions/node/v22.17.0/bin/pnpm
NPM Packages:
  @aws-amplify/ai-constructs: 1.6.0
  @aws-amplify/auth-construct: 1.11.0
  @aws-amplify/backend: 1.20.0
  @aws-amplify/backend-ai: Not Found
  @aws-amplify/backend-auth: 1.9.1
  @aws-amplify/backend-cli: 1.8.2
  @aws-amplify/backend-data: 1.6.3
  @aws-amplify/backend-deployer: 2.1.5
  @aws-amplify/backend-function: 1.16.0
  @aws-amplify/backend-output-schemas: 1.8.0
  @aws-amplify/backend-output-storage: 1.3.3
  @aws-amplify/backend-secret: 1.4.2
  @aws-amplify/backend-storage: 1.4.3
  @aws-amplify/cli-core: 2.2.3
  @aws-amplify/client-config: 1.10.0
  @aws-amplify/data-construct: 1.16.3
  @aws-amplify/data-schema: 1.23.0
  @aws-amplify/deployed-backend-client: 1.8.1
  @aws-amplify/form-generator: 1.2.6
  @aws-amplify/model-generator: 1.2.2
  @aws-amplify/platform-core: 1.10.4
  @aws-amplify/plugin-types: 1.11.2
  @aws-amplify/sandbox: 2.1.4
  @aws-amplify/schema-generator: 1.4.1
  @aws-cdk/toolkit-lib: 1.6.1
  aws-amplify: 6.16.0
  aws-cdk-lib: 2.236.0
  typescript: 5.9.3
npm warn exec The following package was not found and will be installed: cdk@2.1106.0
No AWS environment variables
No CDK environment variables
```

## Problem

[...]

PR #3118, opened 2026-02-18
Passing userAttributes to `AdminCreateUserCommand` instead of `confirmSignIn`

ClaPalo:
## Problem

When creating a new user, we create an `AdminCreateUserCommand` instance without passing the userAttributes defined. This was instead done later in `confirmSignin`, when also replacing the temp password with the one specified in the seed script. However, this causes a blocking issue if the user defined any userAttribute as `required`. Cognito rejects the creation of the user because the required attributes are missing.

**Issue number, if available:** #3117 

## Changes

Instead of passing the `userAttributes` in `confirmSignIn`, this PR passes them directly to `AdminCreateUserCommand` so that the required attributes are immediately available and Cognito does not reject the request.

## Validation

I manually validated the change by creating a test project with some required attributes. Before the change, I wasn't able to seed any new user. After the change, I am able to do so.
I also added two unit tests to ensure the user attributes are correctly passed to `AdminCreateUserCommand` and that undefined attributes are ignored.

## Checklist

### DIFF
--- a/packages/seed/src/auth-seed/auth_client.test.ts
+++ b/packages/seed/src/auth-seed/auth_client.test.ts
@@ -440,7 +440,7 @@ void describe('seeding auth APIs', () => {
       assert.strictEqual(mockMfaFlow.mfaSignUp.mock.callCount(), 1);
       assert.strictEqual(mockCognitoIdProviderClient.send.mock.callCount(), 1);
       assert.deepStrictEqual(userAttributes, [
-        { Name: 'phoneNumber', Value: testNumber },
+        { Name: 'phone_number', Value: testNumber },
       ]);
       assert.deepStrictEqual(output, {
         signInFlow: 'MFA',


packages/seed/src/auth-seed/auth_client.test.ts
```diff
@@ -220,6 +220,43 @@ void describe('seeding auth APIs', () => {
...
+        password: testPassword,
+        signInAfterCreation: true,
+        signInFlow: 'Password',
+        userAttributes: {
+          givenName: 'Test',
+          familyName: 'User',
+        },
+      });
+
+      assert.strictEqual(mockCognitoIdProviderClient.send.mock.callCount(), 1);
```
ShadowCat567, line 235: Can you also update this test (linked below) to have similar validation that you have here so we can also validate that MFA uses are created properly when they get `userAttributes`: https://github.com/aws-amplify/amplify-backend/blob/11184215e2fe8db636dfb0cc55fefae5b43e3046/packages/seed/src/auth-seed/auth_client.test.ts#L389-L406

ClaPalo: Thanks for reviewing! Added the check to ensure `phoneNumber` was passed and verified the test succeeds. I cannot approve workflows to run (only maintainers can)

Issue #3175, opened 2026-04-20
SAML provider not wired up in defineAuth externalProviders


hfurkanbozkurt:
## Bug Description

`defineAuth`'s `externalProviders` JSDoc says "Configure OAuth, OIDC, and SAML login providers" but SAML is not actually supported at the `defineAuth` level.

## Evidence

1. **JSDoc claims SAML support**: [`packages/backend-auth/src/types.ts#L188`](https://github.com/aws-amplify/amplify-backend/blob/88c4759a3d4809e779bd9038b3a0e15a96507503/packages/backend-auth/src/types.ts#L188)

2. **`ExternalProviderSpecificFactoryProps` has no `saml` property** — only Apple, Amazon, Facebook, OIDC, Google: [`packages/backend-auth/src/types.ts#L156`](https://github.com/aws-amplify/amplify-backend/blob/88c4759a3d4809e779bd9038b3a0e15a96507503/packages/backend-auth/src/types.ts#L156)

3. **No `translateSamlProps` in translate_auth_props.ts** — every other provider has one: [`packages/backend-auth/src/translate_auth_props.ts#L61-L93`](https://github.com/aws-amplify/amplify-backend/blob/88c4759a3d4809e779bd9038b3a0e15a96507503/packages/backend-auth/src/translate_auth_props.ts#L61-L93)

4. **Lower-level `auth-construct` fully supports SAML** (types, construct logic, tests): [`packages/auth-construct/src/types.ts#L337-L339`](https://github.com/aws-amplify/amplify-backend/blob/88c4759a3d4809e779bd9038b3a0e15a96507503/packages/auth-construct/src/types.ts#L337-L339)

## Impact

Any user attempting SAML configuration through `defineAuth` gets a TypeScript compilation error. The only workaround is CDK escape hatches via `backend.auth.resources.cfnResources`.

## Expected Behavior

`defineAuth` should expose a `saml` property in `externalProviders` that wires through to the existing `auth-construct` SAML support.

hfurkanbozkurt:
Update amplify gen2 agent skill when this is fixed

Issue #3088, opened 2026-01-15
Add secret() support for SAML metadata in defineAuth()


shivennn:
### Environment information

```plain text
System:
  OS: macOS 15.7.3
  CPU: (10) arm64 Apple M4
  Memory: 193.72 MB / 16.00 GB
  Shell: /bin/bash
Binaries:
  Node: 22.21.1 - /usr/local/bin/node
  Yarn: 1.22.22 - /usr/local/bin/yarn
  npm: 10.9.4 - /usr/local/bin/npm
  pnpm: 10.17.1 - /usr/local/bin/pnpm
NPM Packages:
  @aws-amplify/auth-construct: 1.10.0
  @aws-amplify/backend: 1.19.0
  @aws-amplify/backend-ai: Not Found
  @aws-amplify/backend-auth: 1.9.0
  @aws-amplify/backend-cli: 1.8.1
  @aws-amplify/backend-data: 1.6.2
  @aws-amplify/backend-deployer: 2.1.4
  @aws-amplify/backend-function: 1.15.2
  @aws-amplify/backend-output-schemas: 1.7.1
  @aws-amplify/backend-output-storage: 1.3.2
  @aws-amplify/backend-secret: 1.4.2
  @aws-amplify/backend-storage: 1.4.2
  @aws-amplify/cli-core: 2.2.3
  @aws-amplify/client-config: 1.9.1
  @aws-amplify/data-construct: 1.16.3
  @aws-amplify/data-schema: 1.22.1
  @aws-amplify/deployed-backend-client: 1.8.1
  @aws-amplify/form-generator: 1.2.6
  @aws-amplify/model-generator: 1.2.2
  @aws-amplify/platform-core: 1.10.3
  @aws-amplify/plugin-types: 1.11.1
  @aws-amplify/sandbox: 2.1.4
  @aws-amplify/schema-generator: 1.4.1
  @aws-cdk/toolkit-lib: 1.6.1
  aws-amplify: 6.15.9
  aws-cdk-lib: 2.216.0
  typescript: 5.9.3
No AWS environment variables
No CDK environment variables
```

### Describe the feature

Currently, Amplify Gen 2's secret() function works with OAuth provider credentials (clientId, clientSecret) but not with SAML metadataContent. This creates an inconsistency and forces developers to use environment variables for SAML metadata URLs, which are less secure than secrets stored in AWS Systems Manager Parameter Store.

```
import { defineAuth, secret } from '@aws-amplify/backend';

[...]

PR #3187, opened 2026-05-09
fix(auth): wire up SAML provider in defineAuth externalProviders

inariku:
## Summary

- Add `SamlProviderFactoryProps` type with `BackendSecret` support for `metadataContent`, enabling `secret()` usage for SAML metadata URLs
- Add `saml` property to `ExternalProviderSpecificFactoryProps`
- Add `translateSamlProps` function to resolve secrets when SAML metadata is configured via `secret()`

The `auth-construct` layer already fully supports SAML but the `backend-auth` factory layer was missing the type definition and translation logic. Users attempting SAML configuration through `defineAuth` would get TypeScript compilation errors.

### Usage

```typescript
import { defineAuth, secret } from '@aws-amplify/backend';

export const auth = defineAuth({
  loginWith: {
    email: true,
    externalProviders: {
      saml: {
        name: 'MicrosoftEntraIDSAML',
        metadata: {
          metadataType: 'URL',
          metadataContent: secret('AZURE_AD_SAML_METADATA_URL'),
        },
      },
      callbackUrls: ['http://localhost:3000/'],
      logoutUrls: ['http://localhost:3000/'],
    },
  },
});
```

**Issue number:** https://github.com/aws-amplify/amplify-backend/issues/3175, https://github.com/aws-amplify/amplify-backend/issues/3088

## Checklist


packages/backend-auth/src/translate_auth_props.ts
```diff
@@ -252,3 +263,33 @@ const translateGoogleProps = (
...
+  value: string | BackendSecret,
+): value is BackendSecret => {
+  return typeof value !== 'string';
+};
+
+const translateSamlProps = (
+  backendSecretResolver: BackendSecretResolver,
+  samlProviderProps?: SamlProviderFactoryProps,
+): SamlProviderProps | undefined => {
+  if (!samlProviderProps) {
```
sharonyajain, line 277: [Not-blocking]**unsafeUnwrap() disables CDK secret protection and diverges from sibling providers with no rationale comment**

Every other provider (translateGoogleProps etc.) passes backendSecretResolver.resolveSecret(...) through as a SecretValue, preserving CDK's secret-tracking guard. translateSamlProps instead calls .unsafeUnwrap() to fit the downstream SamlProviderProps.metadata.metadataContent type (a plain string). It is safe today only because CfnTokenBackendSecret.resolve() returns SecretValue.unsafePlainText(<Fn::GetAtt token>) - so unwrap yields a deploy-time CloudFormation token, not literal plaintext. That invariant is load-bearing but undocumented here (backend_secret.ts documents the mirror-image unsafePlainText call with a comment; this call site has none).

```
const metadataContent = isBackendSecret(metadata.metadataContent)
  ? backendSecretResolver
      .resolveSecret(metadata.metadataContent)
      .unsafeUnwrap()
  : metadata.metadataContent;
```

**Suggestion:** Add a comment at the call site documenting why unwrap is safe (metadataContent must be a deploy-time CFN token, mirroring backend_secret.ts), so the invariant is explicit and copy-paste-safe. Optionally centralize the 'resolve secret to string token' step in a shared helper used by any provider whose construct field is a plain string.

packages/backend-auth/src/translate_auth_props.test.ts
```diff
@@ -138,6 +145,13 @@ void describe('translateToAuthConstructLoginWith', () => {
           keyId: appleKeyId,
           privateKey: applePrivateKey,
         },
+        saml: {
```
sharonyajain, line 148: **Only the BackendSecret branch of translateSamlProps is tested; the plain-string metadataContent path is uncovered**

The single added case supplies metadataContent as new TestBackendSecret('samlMetadataUrl') and asserts the unwrapped string output, exercising only the isBackendSecret===true branch. The else branch (metadataContent supplied as a literal string) and the metadataType:'FILE' variant have no test.

```
saml: {
  name: 'MySAMLProvider',
  metadata: {
    metadataContent: 'samlMetadataUrl',
    metadataType: 'URL',
  },
},
```

**Suggestion:** Add a case where metadataContent is a plain string (asserting passthrough), and ideally one with metadataType:'FILE', so both branches of the isBackendSecret ternary are covered.

PR #3197, opened 2026-05-19
feat(backend-data): expose stackMappings for distributing resolvers across nested stacks

Simone319:
## Description

Adds a `stackMappings` option to `defineData()` that allows distributing AppSync resolvers across multiple nested CloudFormation stacks. This is a pass-through to the existing [`stackMappings` property](https://github.com/aws-amplify/amplify-category-api/blob/main/packages/amplify-graphql-api-construct/src/types.ts) in `@aws-amplify/graphql-api-construct`.

### Problem

CloudFormation has a 500-resource limit per stack. Projects with complex data models (many types, relationships, auth rules, indexes) can exceed this limit because all resolvers for a model are placed in a single nested stack. This is a Gen 1 → Gen 2 migration blocker — Gen 1 supported `stackMappings` to work around this, but Gen 2 did not expose it.

Related: [amplify-category-api#2550](https://github.com/aws-amplify/amplify-category-api/issues/2550)

### Solution

Expose the existing construct-level `stackMappings` property through `DataProps` as an optional pass-through. When provided, the mapping tells the transformer which nested stack to place each resolver in (by logical ID).

```typescript
  schema,
  authorizationModes: { defaultAuthorizationMode: 'apiKey' },
  stackMappings: {
    CreateOrderResolver: 'OrderMutations',
    UpdateOrderResolver: 'OrderMutations',
    DeleteOrderResolver: 'OrderMutations',
  },
});
```
### Validation

Invalid mapping keys (typos or stale entries that don't match any generated resolver) are detected during synthesis and reported as a warning:

```
[WARNING] stackMappings contains keys that don't match any generated resource: [CreateOrdersResolver].
These keys are ignored. Valid patterns: Create{Model}Resolver, Get{Model}Resolver, etc.
```

This validation logic will be in `amplify-category-api` repo.

Companion PR: https://github.com/aws-amplify/amplify-category-api/pull/3486

### Changes

| File | Change |
|------|--------|
| `packages/backend-data/src/types.ts` | Added `stackMappings?: Record<string, string>` to `DataProps` |
| `packages/backend-data/src/factory.ts` | Pass-through to underlying construct |
| `packages/backend-data/src/factory.test.ts` | 2 new tests verifying resolver placement and backward compatibility |
| `packages/backend-data/API.md` | Updated public API surface |

### How it works

The mapping flows through: `defineData()` → `AmplifyData` construct → `GraphQLTransform` → `StackManager.getScopeFor()`. When a resolver is created, the `StackManager` checks the mapping and places it in the specified CDK `NestedStack` (created on demand). Each nested stack synthesizes as a separate CloudFormation template, staying under the 500-resource limit.

Mapping keys are resolver logical IDs (e.g., `Create{Model}Resolver`, `Get{Model}Resolver`). Users can discover them via `npx ampx sandbox` output or `cdk synth`.

### Testing

- Unit tests verify resolvers are placed in custom nested stacks via CDK `Template` assertions
- Unit tests verify backward compatibility when `stackMappings` is omitted
- Manually verified via `npx ampx sandbox --once` deployment (resolvers distributed correctly, API functional, `amplify_outputs.json` generated correctly)

### Checklist

- [x] If this PR requires a change to the Project-Level Configuration Schema, I have added the schema change.

### DIFF
--- a/packages/backend-data/src/factory.test.ts
+++ b/packages/backend-data/src/factory.test.ts
@@ -1199,6 +1199,305 @@ void describe('Table Import', () => {
   });
 });
 
+void describe('DataFactory stackMappings', () => {
+  beforeEach(() => {
+    resetFactoryCount();
+  });
+
+  void it('passes stackMappings to the underlying construct', () => {
+    const schema = /* GraphQL */ `
+      type Todo @model @auth(rules: [{ allow: public }]) {
+        id: ID!
+        content: String!
+        done: Boolean
+      }
+
+      type Order @model @auth(rules: [{ allow: public }]) {
+        id: ID!
+        item: String!
+        quantity: Int
+        status: String
+      }
+    `;
+
+    const dataFactory = defineData({
+      schema,
+      authorizationModes: {
+        defaultAuthorizationMode: 'apiKey',
+        apiKeyAuthorizationMode: { expiresInDays: 7 },
+      },
+      stackMappings: {
+        CreateOrderResolver: 'OrderMutations',
+        UpdateOrderResolver: 'OrderMutations',
+        DeleteOrderResolver: 'OrderMutations',
+      },
+    });
+
+    const getInstanceProps = createInstancePropsBySetupCDKApp({
+      isSandboxMode: false,
+    });
+    const dataConstruct = dataFactory.getInstance(getInstanceProps);
+
+    // Verify that the OrderMutations nested stack was created
+    assert(
+      'OrderMutations' in dataConstruct.resources.nestedStacks,
+      'Expected OrderMutations nested stack to be created via stackMappings',
+    );
+
+    // Verify resolvers were placed in the OrderMutations stack
+    const orderMutationsTemplate = Template.fromStack(
+      dataConstruct.resources.nestedStacks['OrderMutations'],
+    );
+    orderMutationsTemplate.hasResourceProperties('AWS::AppSync::Resolver', {
+      FieldName: 'createOrder',
+      TypeName: 'Mutation',
+    });
+    orderMutationsTemplate.hasResourceProperties('AWS::AppSync::Resolver', {
+      FieldName: 'updateOrder',
+      TypeName: 'Mutation',
+    });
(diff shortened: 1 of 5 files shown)


packages/backend-data/src/types.ts
```diff
@@ -155,6 +155,21 @@ export type DataProps = {
    * Each element in the array represents a mapping for a specific branch.
    */
   migratedAmplifyGen1DynamoDbTableMappings?: AmplifyGen1DynamoDbTableMapping[];
+
+  /**
+   * Override the assigned nested stack on a per-resource basis.
+   * Only applies to resolvers, and takes the form { <logicalId>: <stackName> }.
+   * Use this to distribute resolver resources across multiple stacks.
+   *
+   * @example
```
adrianjoshua-strutt, line 165: Should we add something like 

```Resolver logical IDs follow patterns like Create{Model}Resolver, Get{Model}Resolver, etc. You can discover them by running `npx ampx sandbox` and examining the CloudFormation output, or by running `cdk synth` to see the generated template.```

File: packages/backend-data/src/factory.test.ts [hunk 2]
```diff
@@ -1199,6 +1199,104 @@ void describe('Table Import', () => {
   });
 });
 
+void describe('DataFactory stackMappings', () => {
```
adrianjoshua-strutt, line 1202: Should we add tests for:

1.  Multiple custom stacks (e.g., distributing resolvers across 3+ stacks)
2.  Empty object stackMappings: {}
3. Mix of mapped and unmapped resolvers (some Todo resolvers mapped, some in default stack)

packages/backend-data/src/types.ts
Simone319: I was originally adding the pattern example but then removed, to avoid misleading information if the naming is updated while the doc is stale. But adding instruction for customer to look for the resolver name is a good idea.

packages/backend-data/src/factory.test.ts
Simone319: Test cases added.

adrianjoshua-strutt (approved):

sarayev (approved):

PR #3198, opened 2026-05-19
feat(hosting): add skew protection via CloudFront Functions

adrianjoshua-strutt:
## Summary

Adds cookie-based skew protection to the hosting construct, ensuring users mid-session continue receiving assets from their original build during deployments.

## How it works

1. **Viewer-request CloudFront Function** reads a `__dpl` cookie to determine which build the user is pinned to. Rewrites the URI to the correct `builds/{buildId}/` prefix in S3.
2. **Viewer-response CloudFront Function** sets the `__dpl` cookie on HTML responses only (not JS/CSS/images), pinning new sessions to the current build.
3. Old builds are retained in S3 (existing `prune: false` behavior) and cleaned up by lifecycle rules (`buildRetentionDays`).

## Behavior

- **Enabled by default** — no configuration needed
- Opt-out: `skewProtection: { enabled: false }`
- Configurable max age: `skewProtection: { enabled: true, maxAge: 3600 }` (default: 24h)

## Framework-agnostic

Works identically for Next.js (OpenNext), Nuxt/Nitro, SPA, and custom adapters. The CloudFront Functions operate on raw HTTP paths — zero framework knowledge.

## Files changed

- `packages/hosting/src/constructs/skew_protection.ts` (NEW)
- `packages/hosting/src/constructs/cdn_construct.ts` (MODIFIED)
- `packages/hosting/src/constructs/hosting_construct.ts` (MODIFIED)
- `packages/hosting/src/index.ts` (MODIFIED)
- `packages/hosting/src/constructs/skew_protection.test.ts` (NEW)

## Tests

314 pass, 0 failures.

### DIFF
--- a/packages/hosting/src/constructs/cdn_construct.ts
+++ b/packages/hosting/src/constructs/cdn_construct.ts
@@ -990,10 +990,13 @@ export class CdnConstruct extends Construct {
     if (skewEnabled) {
       return new CloudFrontFunction(this, 'SkewProtectionRequestFunction', {
         code: FunctionCode.fromInline(
-          generateSkewProtectionViewerRequestCode(buildId),
+          generateSkewProtectionViewerRequestCode(buildId, redirects),
         ),
         runtime: CLOUDFRONT_FUNCTION_RUNTIME,
-        comment: `Skew protection: routes requests to build from cookie or current build ${buildId}`,
+        comment:
+          redirects.length > 0
+            ? `Skew protection: cookie-based routing + ${redirects.length} redirect rule(s)`
+            : `Skew protection: routes requests to build from cookie or current build ${buildId}`,
       });
     }
     return new CloudFrontFunction(this, 'BuildIdRewriteFunction', {
--- a/packages/hosting/src/constructs/cdn_construct.ts
+++ b/packages/hosting/src/constructs/cdn_construct.ts
@@ -48,7 +48,7 @@ import {
 } from 'aws-cdk-lib/aws-apigateway';
 import { HostingError } from '../hosting_error.js';
 import { prependBasePath } from '../adapters/shared/basepath.js';
-import { DeployManifest } from '../manifest/types.js';
+import { DeployManifest, Redirect } from '../manifest/types.js';
 import {
   ERROR_PAGE_KEY,
   generateAssetPrefixStripFunctionCode,
@@ -56,8 +56,16 @@ import {
   generateForwardedHostAndRedirectFunctionCode,
 } from '../defaults.js';
 import { createCustomHeadersPolicy } from './security_headers.js';
+import {
+  SkewProtectionConfig,
+  generateSkewProtectionViewerRequestCode,
+  generateSkewProtectionViewerResponseCode,
+} from './skew_protection.js';
 
 // ---- Constants ----
+
+/** Runtime version used for all CloudFront Functions in this construct. */
+const CLOUDFRONT_FUNCTION_RUNTIME = FunctionRuntime.JS_2_0;
 
 /**
  * CloudFront allows a maximum of 25 cache behaviors per distribution
@@ -136,6 +144,8 @@ export type CdnConstructProps = {
    * (`runtime = 'edge'`), one entry per route.
    */
   routeEdgeFunctions?: Map<string, IVersion>;
+  /** Cookie-based skew protection configuration. */
+  skewProtection?: SkewProtectionConfig;
 };
 
 // ---- Construct ----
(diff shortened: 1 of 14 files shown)


packages/hosting/src/constructs/cdn_construct.ts
```diff
@@ -146,16 +153,38 @@ export class CdnConstruct extends Construct {
...
-      {
-        code: FunctionCode.fromInline(generateBuildIdFunctionCode(buildId)),
-        runtime: FunctionRuntime.JS_2_0,
-        comment: `Rewrites request URIs to include build ID prefix: builds/${buildId}/`,
-      },
-    );
+    // When skew protection is enabled, the viewer-request function reads the
+    // __dpl cookie to route users to their pinned build. Otherwise,
+    // a simple static rewrite is used.
+    const viewerRequestFunction = skewEnabled
```
osama-rizk: A couple of suggestions here 

1.  Extract a shared runtime constant
Since FunctionRuntime.JS_2_0 is used in every CloudFrontFunction instantiation, worth pulling it into a constant so a future runtime upgrade is a one-line change:

```
const CLOUDFRONT_FUNCTION_RUNTIME = FunctionRuntime.JS_2_0;
```

2. Extract factory functions The inline ternaries with new expressions are a bit heavy to scan. Extracting createViewerRequestFunction and createViewerResponseFunction would clean up the call site nicely:

Suggested code 

```
const viewerRequestFunction = this.createViewerRequestFunction(buildId, skewEnabled);
const viewerResponseFunction = skewEnabled
  ? this.createViewerResponseFunction(buildId, skewMaxAge)
  : undefined;


private createViewerRequestFunction(
  buildId: string,
  skewEnabled: boolean,
): CloudFrontFunction {
  return skewEnabled
    ? new CloudFrontFunction(this, 'SkewProtectionRequestFunction', {
        code: FunctionCode.fromInline(generateSkewProtectionViewerRequestCode(buildId)),
        runtime: CLOUDFRONT_FUNCTION_RUNTIME,
        comment: `Skew protection: routes requests to build ${buildId}`,
      })
    : new CloudFrontFunction(this, 'BuildIdRewriteFunction', {
        code: FunctionCode.fromInline(generateBuildIdFunctionCode(buildId)),
        runtime: CLOUDFRONT_FUNCTION_RUNTIME,
        comment: `Rewrites URIs to builds/${buildId}/`,
      });
}

private createViewerResponseFunction(
  buildId: string,
  skewMaxAge: number,
): CloudFrontFunction {
  return new CloudFrontFunction(this, 'SkewProtectionResponseFunction', {
    code: FunctionCode.fromInline(generateSkewProtectionViewerResponseCode(buildId, skewMaxAge)),
    runtime: CLOUDFRONT_FUNCTION_RUNTIME,
    comment: `Skew protection: sets __dpl cookie to ${buildId}`,
  });
}



```

osama-rizk:
Heads-up on a regression in `generateSkewProtectionViewerRequestCode` (`packages/hosting/src/constructs/skew_protection.ts`).

The existing `generateBuildIdFunctionCode` in `packages/hosting/src/defaults.ts` resolves **two** directory-style cases before prepending the build-id prefix:

```js
if (uri.endsWith('/')) {
  uri = uri + 'index.html';                    // /foo/  -> /foo/index.html
} else {
  var lastSegment = uri.substring(uri.lastIndexOf('/') + 1);
  if (lastSegment.indexOf('.') === -1) {
    uri = uri + '/index.html';                 // /foo   -> /foo/index.html  (bare path)
  }
}
```

The skew-on variant only handles the trailing-slash case. So when skew protection is enabled (default), bare paths emitted by SSG builds — Next.js `output: 'export'`, Nuxt prerender, Astro — will break: `/about` will hit S3 at `/builds/<id>/about` instead of `/builds/<id>/about/index.html` and 404. The unit tests don't catch it because they only assert `code.includes("uri = uri + 'index.html'")`, and the e2e tests exercise the root `/` (trailing-slash branch) only.

This is fixable in a small follow-up — happy to absorb it. Plan: rebase #3200 onto `snapshot/iac-hosting` once this lands, then in the same PR extract the URI-resolution logic into a shared helper in `defaults.ts` so both functions can't drift again, and add a unit test that asserts both functions resolve `/about` -> `/builds/<id>/about/index.html`. That keeps your skew-protection PR scope clean and the regression visible in the next PR's diff.

Tracking on our side as part of #3200 (https://github.com/aws-amplify/amplify-backend/pull/3200).

osama-rizk (approved):

osama-rizk:
Here is health check run https://github.com/aws-amplify/amplify-backend/actions/runs/26452446381

PR #3200, opened 2026-05-21
feat(hosting): Lambda@Edge per-route, origin Cache-Control on SSR, framework redirects/headers

osama-rizk:
## Problem

The L3 hosting construct on `snapshot/iac-hosting` had four classes of latent issues that only surfaced under real-world stress: user security headers were silently overridden, header-rule cache behaviors shadowed SSR routes (403), the SSR cache policy was effectively disabled (defeating ISR/SWR), and operational guardrails (ResponseHeadersPolicy quota, real 404 status, `assetPrefix`, Lambda@Edge per-route wiring) were missing.

## Changes

Closes nine issues:

- **User security headers silently overridden** — Parse user `next.config.js headers()` (`x-frame-options`, HSTS, referrer-policy, CSP) into CloudFront's typed `securityHeadersBehavior` slots with `Override: true`; filter security-named headers out of `customHeadersBehavior.customHeaders[]` (CloudFront rejects them there).
- **Header-rule behaviors shadow SSR routes (403)** — Synthesized header-rule behaviors inherit the catch-all's origin (compute when `hasCompute`) instead of unconditionally pointing at S3 — fixes 403 on `/headers-test`, `/secure-headers`, etc.
- **SSR cache policy ignores origin `Cache-Control` (ISR/SWR broken)** — Custom `SsrCachePolicy` that honors origin `Cache-Control`; keys on Next.js App Router headers (`rsc`, `next-router-prefetch`, `next-router-state-tree`, `next-router-segment-prefetch`); cookies excluded; Brotli/Gzip via the typed flags (no `Accept-Encoding` in the allowList — CloudFront rejects that combination).
- **Next.js edge runtime not wired** — Wire per-route Lambda@Edge associations on every edge-runtime cache behavior.
- **SSG with real `404.html` returns SPA-fallback 200** — New `manifest.errorPages`; SPA adapter auto-detects `404.html`/`500.html`; L3 maps `404 → /404.html` with status 404 instead of `200 /index.html`.
- **`assetPrefix` not propagated end-to-end** — Adapter lifts `assetPrefix` from `.next/required-server-files.json`; L3 emits 4 prefixed CloudFront behaviors with a strip Function before the build-id S3 lookup.
- **Re-deploys churn the ResponseHeadersPolicy quota** — Stable content-derived construct IDs (SHA-256 fingerprint, 8 hex) for `CustomHeadersPolicy*` so re-deploys with unchanged header content are no-ops.

Also includes Windows correctness fixes (`shell: true` on `execFileSync`, Nitro bundle path resolution from #3201) and OpenNext patches (streaming-wrapper applied to all `index.mjs` bundles, edge encoder backslash escape).

## Validation

Deployed all four stress apps to a clean AWS account (us-east-1) and probed end-to-end:

| App | CloudFront URL | Issues verified | How we verified |
|---|---|---|---|
| Nuxt / Nitro | https://d32g3fdqyfxcqw.cloudfront.net | Security-header override; behavior shadowing; SSR cache policy | Opened the page that should ship a stricter X-Frame-Options and saw the user's value in the response — not the L3 default. The page that previously returned a 403 now loads correctly. The SWR page is cached for 30 seconds at the edge, matching what the framework asked for. |
| Next App Router | https://d1t4ms9orer7e6.cloudfront.net | Edge runtime; security-header override; assetPrefix | The edge-runtime API route returns its JSON. The secure-headers page reflects the user's strict X-Frame-Options. Static asset URLs prefixed with `/cdn-static/...` load from S3 (previously 404'd because no behavior existed at that prefix). |
| Next Pages Router | https://daur5x2f9ktll.cloudfront.net | SSG real-404 status; assetPrefix; SSR cache policy (ISR) | A non-existent path returns a real 404 instead of the homepage with a 200. Prefixed static assets resolve. The ISR page is cached at the edge for the revalidate window the framework declared. |
| Next Edge Stress | https://d1qcog3d2ramve.cloudfront.net | Edge runtime per-route wiring | Every route the developer marked as edge has a Lambda@Edge attached on its CloudFront behavior; default-runtime routes don't. Edge routes return their content end-to-end. |

All probes pass. Tests: 319 passing in the hosting package.

## Checklist

- [x] Functional changes are covered by unit tests (security_headers, cdn_construct, defaults, nitro_internal) and live-verified against the four CloudFront URLs above
- [ ] Project Architecture README update — N/A
- [ ] Docs PR — N/A
- [x] Run e2e on PR — already covered by `ci: auto-run e2e for PRs targeting snapshot/iac-hosting (#3199)`


packages/hosting/src/adapters/nextjs.ts
adrianjoshua-strutt, line 1: **Double `next build`** (around the `runNextBuild` call): This runs `next build` before OpenNext (which then runs it again internally). For large apps without `.next/cache`, this doubles build time (30-60s extra). The comment acknowledges it — worth a plan before merging to main. Could OpenNext expose an option to skip its internal `next build` when `.next/` already exists?

adrianjoshua-strutt, line 1: **`toManifestSource` is a no-op** — both branches return `source` unchanged:\n```ts\nconst toManifestSource = (source: string): string => {\n  return source.endsWith('/*') ? source : source;\n};\n```\nLeftover scaffolding? Either remove it or add a TODO explaining the intent.

packages/hosting/src/constructs/cdn_construct.ts
adrianjoshua-strutt, line 1: **Edge route caching inconsistency**: Edge route behaviors use `CACHING_DISABLED` while standard SSR routes now use the new `SsrCachePolicy` (honoring origin `Cache-Control`). If an edge function sets `Cache-Control: s-maxage=3600`, CloudFront won't cache it.\n\nEither use `ssrCachePolicy` for edge behaviors too, or add a comment explaining why edge routes are excluded (e.g. 'edge functions are stateless/dynamic by design').\n\n---\n\n**Lambda@Edge limit validation**: Each edge route creates a separate Lambda@Edge function. CloudFront soft limit is 25 behaviors, Lambda@Edge has 25 replicated functions per account. Worth a synth-time warning when approaching these limits.

packages/hosting/src/constructs/security_headers.ts
adrianjoshua-strutt, line 1: **CSP override → true is a behavioral change**: Previously origin CSP headers from Lambda would win. Now the L3's static CSP always wins. SSR apps with dynamic CSP (e.g. nonces) will break silently.\n\nWorth calling out in the changeset so teams know to pass dynamic CSP through `cdn.contentSecurityPolicy` instead.\n\n---\n\n**x-xss-protection: 0 silently dropped**: If a user sets `x-xss-protection: 0` (OWASP-recommended to disable), the value gets filtered from customHeaders but isn't parsed into the typed slot — which still emits `1; mode=block`. Low priority (header is deprecated), but worth documenting.

packages/hosting/src/constructs/cdn_construct.ts
adrianjoshua-strutt, line 1: **Route specificity sort bug with multiple wildcards**:\n\nThe formula `pattern.length * 1000 - wildcards * 1_000_000` with descending sort means `/*` (score: -998,000) ranks HIGHER than `/api/*/data/*` (score: -1,986,000). A root catch-all should be LESS specific than a multi-segment pattern.\n\nPractical impact is low (Next.js patterns rarely have 2+ wildcards), but worth fixing for correctness.

packages/hosting/src/constructs/security_headers.ts
osama-rizk: Both real — fixed in the next push.

**CSP override → true regression** is a real silent break for Next.js Server Components / Remix with per-request nonces. Fixed: `override` is now `true` only when the user explicitly supplied a CSP (via `headers()` or the `contentSecurityPolicy` prop). Default CSP emits with `override: false` so the origin's per-request CSP wins. Test updated + new test asserting both branches.

**`x-xss-protection: 0` silently dropped** — agreed, low priority since the header is deprecated and Chrome/Firefox both removed XSS Auditor years ago. Tracking as a follow-up; for now it just emits the L3 default.

packages/hosting/src/adapters/nextjs.ts
```diff
@@ -167,6 +288,50 @@
...
+ *
+ * Per OpenNext docs, `placement: 'global'` deploys to Lambda@Edge in
+ * us-east-1 and replicates to CloudFront PoPs. The Lambda@Edge converter
+ * (`aws-cloudfront`) reads CloudFront origin-request events directly —
+ * no API Gateway in front, so the body-hash 403 issue doesn't apply
+ * here (Lambda@Edge invocation isn't OAC-signed by CloudFront).
+ */
+const renderEdgeFunctionsBlock = (edgeRoutes: EdgeRoute[]): string => {
+  if (edgeRoutes.length === 0) return '';
+  const quote = (s: string) => `'${s.replace(/'/g, "\\'")}'`;
```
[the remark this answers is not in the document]
osama-rizk: Already fixed in commit ffc4b24f33 — `escapeForRegex` now escapes backslashes via `replace(/\\\\/g, '\\\\\\\\')` before any other special-char escapes. Should drop off the CodeQL alert list on the next scan.

packages/hosting/src/adapters/nextjs.ts
osama-rizk: Fixed in commit 86aa5a17a6. Removed the helper entirely and inlined `r.source` / `r.destination` / `h.source` at the three call sites — both branches of the ternary returned the same value, leftover scaffolding from an earlier refactor. -12 / +3.

packages/hosting/src/constructs/cdn_construct.ts
osama-rizk: Both real — addressed across two commits.

**Edge cache inconsistency** (`1518f25e23`): switched edge route behaviors from `CACHING_DISABLED` to the same `SsrCachePolicy` used by regional SSR. The policy's `defaultTtl: 0` means routes that don't emit `Cache-Control` still skip the CDN cache (preserving the safe default), while routes that DO emit `Cache-Control: s-maxage=N` opt into edge caching with `stale-while-revalidate` semantics — matches Vercel's developer experience ([Vercel Edge Function caching](https://vercel.com/docs/caching/cdn-cache#using-vercel-functions)). Auth-bearing edge routes must emit `Cache-Control: private` to opt out (cookies are not in the cache key). Expanded the JSDoc on `makeEdgeRouteBehavior` to call this out. New test asserts the edge behavior's `CachePolicyId` references the synthesized `SsrCachePolicy` (Ref), not the AWS-managed `CACHING_DISABLED` literal.

**Lambda@Edge limit validation** (`bcf09bacbf`): added synth-time check — `HostingError('TooManyEdgeRoutesError')` when the distribution declares >25 edge routes (the AWS account-level cap), with stderr warning at >20 (5-route headroom for other distributions in the account). Two new tests cover both branches. Converts the unwinnable case from a confusing CloudFormation rollback into a useful error with a resolution path (consolidate routes, or request a quota increase).

osama-rizk: Fixed in commit 73e81dea2d. Replaced the formula `length × 1000 - wildcards × 1_000_000` with `literalSegments × 1000 + length`.

Walking your case:
- `/api/*/data/*` → 2 literal segments, length 13 → 2_013
- `/*` → 0 literal segments, length 2 → 2

`/api/*/data/*` now ranks above `/*` as expected. The literal-segment count is the right primary axis because a pattern with more literal path segments constrains more of the URL, regardless of how many wildcards it contains. Length stays as the tiebreaker for patterns within the same literal-segment count.

New regression test asserts `/api/*/data/*` (2 literal segments) is emitted before `/_next/*` (1 literal segment) in the synthesized `CacheBehaviors` array. JSDoc updated with the new formula and worked examples.

packages/hosting/src/adapters/nextjs.ts
osama-rizk: Fixed in commit `9602d9566d`. Background and approach below.

**Why we added it.** OpenNext refuses to bundle `runtime: 'edge'` routes into the default function — it requires a `functions: { edgeN: { runtime: 'edge', routes, patterns } }` block in `open-next.config.ts`. To generate that block, the adapter needs to know which routes export `runtime: 'edge'`, and the canonical source of truth is Next.js's compiled `.next/server/middleware-manifest.json`. We ran `next build` once to produce it.

**Verified upstream first.** The OpenNext maintainers haven't exposed a skip-build flag — `@opennextjs/aws/dist/build/buildNextApp.js` unconditionally calls `execSync(buildCommand)`. The `buildCommand` config is just *what* to invoke, not whether to. Setting `buildCommand: 'true'` would be brittle (relies on OpenNext's bundler not validating that a fresh build just ran); separate upstream issue.

**The fix.** Skip our `runNextBuild` when:
1. `.next/server/middleware-manifest.json` already exists (user/CI ran `next build` first, or a previous deploy cached `.next/`), OR
2. The project's source declares no `runtime: 'edge'` / `'experimental-edge'` routes — `detectEdgeRoutes` returns empty, the generated `open-next.config.ts` has no `functions{}` block, and OpenNext's own build pass produces all the manifests we need.

Both branches eliminate one full Next.js build cycle (~10s warm, 30-60s cold). When neither applies (edge routes declared, manifest absent), `runNextBuild` fires as before — preserving correctness.

Two new exported helpers (`hasExistingMiddlewareManifest`, `projectHasEdgeRuntimeRoutes`) with `@internal` annotations + 8 unit tests covering: missing/present manifest, no-edge-routes, `runtime: 'edge'`, `experimental-edge`, `src/app/`, `node_modules` exclusion, dot-prefix exclusion, and projects with no `app/`/`pages/` directories.

adrianjoshua-strutt (approved):

PR #3220, opened 2026-06-03
feat(hosting): production-readiness batch + Lambda/S3/observability improvements

osama-rizk:
## Summary

Two-part hosting batch atop `snapshot/iac-hosting`. **All knobs new in this PR default off** so existing deploys see no behavior change unless they opt in.

#### Security (P0)

- **P0.1 / P0.2 — IPX SSRF + SVG.** Image-opt Lambda runtime now enforces `IMAGE_REMOTE_PATTERNS` / `IMAGE_ALLOWED_HOSTNAMES` / `IMAGE_ALLOW_SVG` (default-deny). Closes the SSRF primitive and the SVG-XSS gap. The construct already stamped these env vars; the Lambda runtime now actually reads them.
- **P0.3 — REST API stage gate.** Implemented in [#3218](https://github.com/aws-amplify/amplify-backend/pull/3218) with a deterministic origin-verify Referer secret. My earlier `randomBytes()` version was dropped during rebase in favor of the deterministic variant — same threat model, no CFN churn on each `cdk synth`.

#### Customer-visible correctness (P1)

- **P1.3** — Skew `__dpl` cookie skipped on 4xx/5xx HTML responses. No more sticky failure pinning users to a broken build.
- **P1.5** — Nitro cache plugin re-binds via `nitro.hook('storage:mounts')` + per-request fallback so a later-named user plugin can't silently shadow it.
- **P1.6** — Astro hybrid mode emits both bare and subtree static behaviors regardless of `trailingSlash` mode. Prerendered routes never silently fall through to the SSR Lambda.
- **P1.7** — Font Content-Type via per-extension `BucketDeployment`, ordered after the asset deployments via `addDependency()` so the right `font/woff2` Content-Type isn't overwritten back to `binary/octet-stream` by the asset pass.
- **P1.8** — Next.js trailing-slash redirects now read from `prerender-manifest`, `routes-manifest`, AND `app-paths-manifest`. Closes the `generateStaticParams` known-vs-unknown inconsistency.

#### Cost / quota (P2)

- **P2.1** — Default S3 build retention 365 → 30 days (`storage.buildRetentionDays` to extend).
- **P2.5** — Per-pattern RHP fingerprint includes CSP when the user explicitly sets `cdn.contentSecurityPolicy`. Narrower scope after #3218: the default path now uses CloudFront's managed `SECURITY_HEADERS` policy (zero quota), so this fix only matters for user-supplied CSP.
- **P2.7** — `assetPrefix` collapsed from 4 → 1 cache behavior, freeing 3 slots toward CloudFront's 24-additional-behavior cap.

#### Ops (P3)

- **P3.1 / P3.2** — Optional `monitoring: { enabled: true }` (off by default) wires CloudFront 5xx, Lambda error/throttle, image-opt error, and revalidation-DLQ-depth alarms. SNS topic ARN exposed as a CFN output.
- **P3.3** — Lambda log retention default 14 → 30 days, propagated to image-opt and middleware.
- **P3.5** — Deploy lock TTL 1h → 4h so first-time deploys with ACM validation aren't treated as stale.
- **P3.7** — `resolveProjectDir` walks up from cwd to the nearest `package.json` (override via `AMPLIFY_HOSTING_PROJECT_DIR`). Subdirectory deploys error clearly instead of failing deep in the adapter.

#### Drift gate (X.1)

- Adapters export `VERIFIED_OPENNEXT_RANGE` / `VERIFIED_NITRO_RANGE` / `VERIFIED_ASTRO_RANGE`. New unit test asserts each is a parseable semver range with an explicit upper bound — no open-ended pins.

### Part 2 — Improvements (Lambda / S3 / observability)

| ID | Surface | Default |
|---|---|---|
| 2.1 | `compute.warmup: { rate: Duration }` — EventBridge synthetic invoke | off |
| 2.2 | `compute.snapStart: true` — SnapStart toggle (forward-compat for Node) | off |
| 2.3 | Nitro cache plugin externalizes `@aws-sdk/client-s3` via `createRequire` | always; saves ~16 MB |
| 2.4 | IPX bundle prunes tests/docs/`.d.ts`/`.map`; `--omit=dev` | always; saves ~5-10 MB |
| 3.2 | `manifest.lifecycle?: Array<{prefix, days}>` — adapter-supplied lifecycle | always; Next decl `_next/data/` |
| 3.3 | `storage.inventory: { enabled: true }` — daily S3 inventory of `builds/` | off |
| 4.1 | `compute.tracing: { otelEndpoint, otelHeaders?, serviceName? }` — OTEL_* env on every Lambda | off |
| 4.1 | Image-opt receives `IMAGE_FORMATS` + `IMAGE_DEVICE_SIZES` env | always |
| 4.2 | `manifest.basicAuth[]` honored end-to-end via CloudFront Function gate | always |

### Notes on what's NOT in this PR

- **3.1 Brotli pre-compression** — 1+ week (new CF behaviors + variant-picker function).
- **4.2 `routeRules.proxy`** — design-level (per-pattern `HttpOrigin` + CF origin-rewrite).
- **4.2 `scheduledTasks`** — needs EventBridge wiring; deferrable.
- **4.2 WebSocket** — not fixable on existing infra (needs API GW WebSocket).
- **P1.4 / P1.9 / P1.10 / P3.4 / P3.6** — small follow-ups; tracked separately.

## Test plan

- [x] **Unit tests:** 652/652 pass on the merged tree (15 new tests from #3218 included).
- [x] **TypeScript:** clean.
- [x] **ESLint:** 0 errors with `--max-warnings=0`.
- [x] **Live deploy + URL probes** on us-east-1 (Next App Router stress app):
  - All P0 / P1 / P2 / P3 fixes verified end-to-end against a real CloudFront distribution.
  - 12/13 matrix items passed (M3.8 font Content-Type now passing after the per-deployment ordering fix).
- [x] **Hosting CI checks (#3220 latest run):**
  - `e2e_hosting` ✅
  - `vanilla_cdk_hosting_ssr` + `vanilla_cdk_hosting_spa` (3 platforms) ✅
  - `standalone_hosting_ssr` + `standalone_hosting_ssr_monorepo` ✅
  - `standalone_hosting_astro` / `standalone_hosting_nuxt` / `standalone_hosting_spa` — pending at time of writing.

### Pre-existing CI failures (not caused by this PR)

- `test_scripts` — `release_lifecycle.test.ts` calls out to `npx --package '@changesets/cli' -- changeset init` which fails on an upstream `human-id`/`@changesets/write` ESM/CJS interop regression. **The same failure reproduces on the base branch's most recent CI run** (`snapshot/iac-hosting` run [26940721212](https://github.com/aws-amplify/amplify-backend/actions/runs/26940721212)) — it's a registry-published transitive dep that broke between yesterday's green run and today.
.


adrianjoshua-strutt:
`Backwards compatibility`

That sounds really fishy. This was never released to customers. We are allowed to break stuff. We should not account for that and add unnecessary guards to account for customers that do not exist.

packages/hosting/src/adapters/ipx_lambda_template.ts
```diff
@@ -46,6 +46,89 @@ if (!bucket || !region) {
...
+    if (!matchHostname(host, p.hostname)) continue;
+    if (p.protocol && p.protocol + ':' !== parsed.protocol) continue;
+    if (p.port && p.port !== parsed.port) continue;
+    if (p.pathname) {
+      // Next.js pathname patterns allow trailing /** wildcard. Translate
+      // to a startsWith check (after stripping /**) for the common case.
+      const ptn = p.pathname.endsWith('/**')
+        ? p.pathname.slice(0, -3)
+        : p.pathname;
+      if (!parsed.pathname.startsWith(ptn)) continue;
```
adrianjoshua-strutt: This also matches `/images-secret/anything` -- if `ptn` is `/images`, `startsWith('/images')` passes for any path sharing that prefix. Should check for a path segment boundary (trailing slash or exact match, e.g. `ptn === parsed.pathname || parsed.pathname.startsWith(ptn + '/')`).

Also, this whole allowlist block (~90 lines of URL parsing, hostname matching, wildcard logic) is security-critical and has no unit tests. Worth adding coverage for the edge cases: protocol-relative URLs, hostname spoofing via `@` in userinfo, patterns without trailing path separators, encoded slashes, etc.

packages/hosting/src/defaults.ts
```diff
@@ -135,6 +135,92 @@ export const generateRedirectCheckSnippet = (
...
+    }
+    return {
+      source: rule.source,
+      realm: rule.realm ?? 'Restricted',
+      users: encodedUsers,
+    };
+  });
+  const table = JSON.stringify(encodedRules);
+  return `
+  var __ba = ${table};
```
adrianjoshua-strutt: Heads up -- these credentials end up in plaintext in the CloudFront Function source, which means they're visible in CFN templates, CloudTrail logs, and anyone with `cloudfront:GetFunction` permission. For pre-launch this is probably fine, but before GA you'll want to pull these from a parameter store or at minimum document the visibility scope so users understand what they're opting into.

Also no tests for this generator -- ~70 lines of security-relevant string templating (path matching, base64 comparison, 401 response shaping) seems worth covering with at least a few cases.

packages/hosting/src/constructs/hosting_construct.ts
```diff
@@ -703,6 +886,49 @@ export class AmplifyHostingConstruct extends Construct {
       }
     }
 
+    // ---- 9c. Default CloudWatch alarms (P3.1 + P3.2) ----
+    // Off by default so the construct stays cheap-by-default; opt in
+    // via `monitoring: { enabled: true }`. When enabled, surfaces the
+    // SNS topic ARN as a CloudFormation output so operators can wire
+    // subscriptions externally without touching the construct again.
+    if (props.monitoring?.enabled) {
```
adrianjoshua-strutt: Since this is pre-launch with zero deployed stacks, defaulting monitoring OFF means the first real users deploy blind to 5xx spikes, Lambda errors, and DLQ poison pills. The cost at idle is pennies/month per the comment above. I'd flip this to default ON and let users opt out -- you want maximum observability during the validation phase.

(The comment on L890-891 narrates a production history that hasn't happened yet, which might confuse future contributors reading this as validated operational experience rather than a design rationale.)

packages/hosting/src/constructs/hosting_construct.ts
```diff
@@ -590,12 +667,118 @@ export class AmplifyHostingConstruct extends Construct {
...
+    // CFN escape hatch so we don't depend on a CDK release that
+    // exposes `snapStart` for Node.
+    //
+    // Skipped for:
+    //   - Lambda@Edge: SnapStart is incompatible with replicated
+    //     edge functions.
+    //   - http-server Lambdas: the Web Adapter init path doesn't
+    //     benefit from SnapStart and the boot-process model
+    //     conflicts with snapshot semantics.
+    if (props.compute?.snapStart) {
```
adrianjoshua-strutt: Node.js Lambda doesn't support SnapStart yet -- this escape hatch will cause a CloudFormation deploy-time error (`SnapStart is not supported for the nodejs* runtime`) rather than silently no-op. Might want to gate this behind a runtime check (skip if `runtime.startsWith('nodejs')`) or at minimum document clearly that it's forward-looking and currently non-functional for the only runtime this construct targets.

packages/hosting/src/constructs/hosting_construct.ts
```diff
@@ -703,6 +886,49 @@ export class AmplifyHostingConstruct extends Construct {
...
+        snsTopic: userTopic,
+        distribution: this.distribution,
+        // Lambda@Edge functions don't accept the CW metric helpers we
+        // use; only attach when the SSR compute is a regional Lambda.
+        ssrFunction: ssrFn instanceof LambdaFunction ? ssrFn : undefined,
+        imageFunction: imgFn instanceof LambdaFunction ? imgFn : undefined,
+        revalidationDlq: this.revalidationDlq,
+      });
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (this as any).monitoringTopic = monitoring.topic;
```
adrianjoshua-strutt: Nit: Mutating a readonly field via `(this as any)` is TypeScript unsoundness -- if the field is readonly for a reason, this bypasses it silently. Is there a setter or builder pattern available instead? Or if the field genuinely needs to be set post-construction, making it non-readonly with a `@internal` annotation might be more honest.

packages/hosting/src/constructs/monitoring_construct.ts
```diff
@@ -0,0 +1,177 @@
...
+  imageFunction?: LambdaFunction;
+  /** Revalidation worker DLQ — depth alarm. */
+  revalidationDlq?: Queue;
+};
+
+/**
+ * CloudWatch alarm wiring construct. See module-level doc-comment for
+ * the rationale and the alarm set.
+ */
+export class MonitoringConstruct extends Construct {
```
adrianjoshua-strutt, line 69: This is 177 lines of alarm wiring with thresholds, SNS actions, and conditional logic -- but no unit tests. Worth adding at least a snapshot test that verifies the expected alarms are created (and not created) under different prop combinations, with the right thresholds and dimension values.

packages/hosting/src/constructs/cdn_construct.ts
```diff
@@ -671,11 +692,22 @@ export class CdnConstruct extends Construct {
...
     // When the framework's build emits asset URLs under a prefix
     // (Next.js `assetPrefix: '/shop-static'`), the non-prefixed
-    // `/_next/static/*` behavior won't match. Add prefixed copies that
-    // strip the prefix before falling through to the build-id S3 lookup.
+    // `/_next/static/*` behavior won't match. The previous
+    // implementation emitted four separate prefixed behaviors
+    // (`/<prefix>/_next/static/*`, `/_next/image*`, `/_next/data/*`,
+    // `/_next/*`) — each one consuming a slot of the CloudFront
+    // 24-additional-behavior budget. Customers with even modest
+    // route counts hit the cap.
```
adrianjoshua-strutt: Nit: This reads like a post-mortem for an incident that hasn't happened yet (pre-launch, zero customers). Consider rewriting as design rationale: "This approach scales poorly under N routes because each prefixed behavior consumes a slot" rather than "Customers with even modest route counts hit the cap." Future contributors might treat this as validated production experience rather than theoretical analysis.

packages/hosting/src/adapters/nitro_cache_plugin_template.ts
```diff
@@ -124,8 +138,65 @@ export default defineNitroPlugin(() => {
...
+      warned = true;
+    }
+    storage.mount('cache', amplifyDriver);
+  };
+
+  // Nitro exposes \`hooks.hook\` on the runtime app instance. Probe for
+  // both the storage:mounts variant (newer Nitro) and a generic
+  // \`request\` hook fallback (so we re-check at the top of each
+  // request — cheap inline call, no I/O).
+  if (process.env.AMPLIFY_NITRO_CACHE_DISABLE !== 'true') {
```
adrianjoshua-strutt: Nit: Nobody has a custom cache plugin deployed that relies on displacing this mount (pre-launch). Consider just always re-binding and adding the escape hatch later when someone actually asks for it -- removes a conditional path and an undocumented config surface that'll be invisible to users who don't read source.

packages/hosting/src/adapters/nitro_cache_plugin_template.ts
```diff
@@ -124,8 +138,65 @@ export default defineNitroPlugin(() => {
...
+    if (nitroApp?.hooks?.hook) {
+      try {
+        nitroApp.hooks.hook('storage:mounts', remountIfDisplaced);
+        // eslint-disable-next-line @aws-amplify/amplify-backend-rules/no-empty-catch
+      } catch {
+        // hook name not supported by this Nitro version — fall back to
+        // the request-time check below.
+      }
+      try {
+        nitroApp.hooks.hook('request', remountIfDisplaced);
```
adrianjoshua-strutt: Nit: This runs on every single HTTP request. The function body re-checks `getMounts('cache')` each time. Would an early-exit at the top (`if (ours) return` is already there, but `getMounts` itself might be non-trivial) or a flag (`displaced = false; ... if (!displaced) return;`) help avoid the overhead on the hot path?

osama-rizk:
> customers. We are allowed to break stuff. We should not account for that and add unnecessary guards to account for customers that do not exist.

This gets added by mistake. PR description updated

packages/hosting/src/adapters/ipx_lambda_template.ts
osama-rizk: Fixed in 8500f5a8. Added segment-boundary check (`pathname === ptn || startsWith(ptn + '/')`) plus a userinfo guard. Extracted matchers to `ipx_allowlist.ts`; 23 unit tests in `ipx_allowlist.test.ts` cover the original bug, userinfo spoofing, protocol-relative URLs, encoded slashes, port/protocol constraints, and the wildcard hostname semantics.

packages/hosting/src/defaults.ts
osama-rizk: Dropped the basicAuth feature entirely in 8500f5a8. Discovered while investigating that Nitro 2.13 doesn't actually ship `routeRules.basicAuth` natively — it was an adapter-side extension we'd layered on. Pre-launch + 0 customers + not in upstream = removing was safer than re-architecting. Users do basic auth in their own SSR handler. We can re-add later with proper design (CloudFront KeyValueStore or Lambda@Edge with rotation) if customers actually ask. Removed ~120 lines across defaults / schema / types / nitro adapter / cdn_construct.

packages/hosting/src/constructs/hosting_construct.ts
osama-rizk: Flipped to default-ON in 8500f5a8. `monitoring?: { enabled?: boolean }` defaults to `true`; opt out with `{ enabled: false }`. Also rewrote the comment to drop the made-up production history.

packages/hosting/src/constructs/hosting_construct.ts
osama-rizk: Removed entirely in 8500f5a8. You're right — Node SnapStart isn't supported, the escape hatch would error at deploy. Rather than gate it behind a runtime check and ship a non-functional knob, removed the prop, the implementation block, the changeset entry, and the comment. Re-add when Node SnapStart GAs.

packages/hosting/src/constructs/hosting_construct.ts
osama-rizk: Fixed in 8500f5a8. Dropped `readonly` from `monitoringTopic` and removed the `(this as any)` cast. The field is logically immutable post-construction but is assigned mid-constructor (after the inner `MonitoringConstruct` is built), so making it non-readonly with a doc-comment about the convention was the honest fix.

packages/hosting/src/constructs/monitoring_construct.ts
osama-rizk: Added in 8500f5a8 — `monitoring_construct.test.ts` has 13 tests covering: disabled path (no topic, no alarms even with all resources), enabled topic provisioning (default + BYO + no-resources), each per-resource alarm branch (CloudFront 5xx with the `Region: Global` dimension, SSR errors+throttles pair, image-opt errors, DLQ depth), the full prop set (5 alarms, all wired to the same SNS action), and a threshold/period regression guard.

packages/hosting/src/constructs/cdn_construct.ts
osama-rizk: Reworded in 8500f5a8 — dropped the "customers hit this" framing and rewrote as design rationale ("this scales poorly under N routes because each prefixed behavior consumes a slot"). Will keep an eye on this pattern across the rest of the construct comments going forward.

packages/hosting/src/adapters/nitro_cache_plugin_template.ts
osama-rizk: Removed the `AMPLIFY_NITRO_CACHE_DISABLE` opt-out in 8500f5a8. Always re-binds now. Agree the env var was speculative — pre-launch with no users running custom cache plugins, the conditional path was carrying weight it didn't earn.

packages/hosting/src/adapters/nitro_cache_plugin_template.ts
osama-rizk: Fixed in 8500f5a8 — added a `settled` flag that flips on first successful remount, with an `if (settled) return` early-exit at the top of the hook. After the displacement (if any) is corrected once, every subsequent request becomes a single boolean check instead of re-running `getMounts('cache')` and the array scan.

adrianjoshua-strutt (commented):
A few more suggestions on top of the resolved comments. Great work addressing everything so quickly.

---

### 1. Tracing: OTEL env vars alone won't produce traces

The `compute.tracing` implementation sets `OTEL_EXPORTER_OTLP_ENDPOINT` and `OTEL_SERVICE_NAME` on the Lambdas, but without the ADOT Lambda Layer or a user-installed OTel SDK with `NODE_OPTIONS=--require ...`, these env vars are inert. A user who enables `tracing: { otelEndpoint: 'http://collector:4318' }` will expect traces to appear and get nothing.

Suggestion: Either auto-attach the ADOT layer when tracing is set, or add a clear note in the JSDoc that this is a "bring your own instrumentation" feature and the user needs to add the OTel SDK or Lambda layer independently.

---

### 2. Lambda warmup sends raw JSON to http-server Lambdas

The EventBridge warmup event is a raw JSON object (`{ source: 'amplify-hosting-warmup' }`). For `type: 'handler'` (OpenNext), the handler can detect and short-circuit. But for `type: 'http-server'` (Nitro/Astro via Lambda Web Adapter), LWA expects an HTTP-shaped event (Function URL or API GW format). Sending raw JSON will cause a 500 on every warmup invoke.

Suggestion: Either skip warmup wiring for http-server compute, or send a properly-shaped HTTP GET event that the adapter can handle (e.g., a health-check path like `/__warmup`).

---

### 3. Monitoring alarm threshold is absolute, not rate

The PR description says "SSR Lambda error rate > 1% over 5 min" but the implementation uses `threshold: 5` (absolute count). For a high-traffic site (10K RPM), 5 errors is noise. For a low-traffic site (2 RPM), 5 errors means complete outage.

Suggestion: Use a CloudWatch math expression alarm: `Errors / Invocations * 100 >= 1` for actual rate-based alerting. Or at minimum, document that this is an absolute threshold customers should tune.

---

### 4. Nitro cache `getKeys()` doesn't paginate

`ListObjectsV2Command` returns max 1000 keys per call. If the cache has more objects, `getKeys` silently returns only the first page. `clear()` then leaves orphaned objects. Also, `clear()` fires unbounded concurrent `DeleteObject` calls which could hit S3's 3,500 DELETE/s rate limit.

Suggestion: Add pagination via `IsTruncated` + `ContinuationToken`, and use `DeleteObjectsCommand` (batch delete, 1000 per call) instead of individual deletes.

---

### 5. Font deployment may overwrite Cache-Control headers

The font Content-Type BucketDeployment re-uploads font files with the correct `contentType` but no `cacheControl` set. Since `prune: false` still overwrites existing objects, fonts in `_next/static/media/` lose their `max-age=31536000, immutable` header from the earlier hashed-asset deployment.

Suggestion: Add `cacheControl: [CacheControl.fromString('public, max-age=31536000, immutable')]` to the font BucketDeployment, or restrict the glob to non-hashed paths only.

---

### 6. Version drift ranges are wider than tested

`VERIFIED_NITRO_RANGE = '>=2.10.0 <3.0.0'` covers all of Nitro 2.x, but if a breaking change ships in 2.14, the adapter won't warn. Same for Astro `<7.0.0` (6.x doesn't exist yet).

Suggestion: Consider tightening to the actual last-tested minor (e.g., `<2.14.0`) and bumping when you verify newer versions. Or add a comment explaining "VERIFIED" means "believed compatible" not "actively tested against every version in range."

osama-rizk:
Thanks for the second pass — all six addressed in `b6861f034a`.

**1. Tracing: OTEL env vars alone won't produce traces** ✅
Went with the docs option (kept the construct SDK-agnostic so it doesn't pin an OTel version or pull a layer into every deploy). The `compute.tracing` JSDoc now spells out "bring your own instrumentation" — `otelEndpoint` alone is inert; you must additionally attach the ADOT Lambda layer **or** bundle an OTel SDK via `NODE_OPTIONS=--require @opentelemetry/auto-instrumentations-node/register`.

**2. Warmup sends raw JSON to http-server Lambdas** ✅ — good catch, real 500.
Gated warmup to `type: 'handler'` (OpenNext) only. For `http-server` (Nitro/Astro via LWA) it's skipped with a stderr warning pointing at `provisionedConcurrency`. Chose skip-with-warning over synthesizing an HTTP event so we don't have to track LWA's expected event shape.

**3. Monitoring alarm threshold is absolute, not rate** ✅
Replaced `threshold: 5` with a CloudWatch math expression `(errors / invocations) * 100`, default **1%**, tunable via `monitoring.ssrErrorRatePercent`. `treatMissingData: NOT_BREACHING` + the ratio yielding 0 on no-traffic windows avoids false alarms on quiet periods.

**4. Nitro cache `getKeys()` doesn't paginate** ✅ — real data-loss bug.
`getKeys` now pages via `IsTruncated` + `NextContinuationToken`, and `clear()` batch-deletes with `DeleteObjectsCommand` (1000/call) instead of unbounded concurrent `DeleteObject`.

**5. Font deployment overwrites Cache-Control** ✅
Set the mutable tier (`s-maxage=…`) on the font pass. A single `BucketDeployment` can't intersect "font-ext AND immutable-path" (include globs are additive OR), so per-file re-tiering isn't expressible — documented that. Net: hashed fonts keep CloudFront's 1y edge cache via `s-maxage` (browser revalidates) instead of the no-header state they had before; non-hashed public fonts get exactly the mutable directive they want.

**6. Version drift ranges wider than tested** ✅
Tightened `VERIFIED_ASTRO_RANGE` `<7.0.0 → <6.0.0` and `VERIFIED_NITRO_RANGE` `<3.0.0 → <2.14.0` to the actual tested ceilings, and documented that "verified" = believed-compatible, not tested-against-every-release. Also corrected the comments to reference the real enforcement points (`assertAstroVersion` floor; the Nitro patcher's `UpstreamPatchPatternChangedError`) — there are no `warnIf*` functions for these two.

721/721 hosting unit tests pass (added a rate-threshold test).

adrianjoshua-strutt (approved):

adrianjoshua-strutt:
One more thought on M2 (prerendered pages from S3):

Since P1.8 already reads `prerender-manifest.json` to extract trailing-slash redirect routes, you have the full list of prerendered pages available at synth time. Would it be feasible to also emit dedicated static behaviors (or CloudFront Function routing entries) for those pages so they serve directly from S3 instead of going through the catch-all Lambda?

Right now, a page like `/about` that was prerendered by `generateStaticParams` still hits the SSR Lambda (catch-all `/*`), which re-renders it from the bundle. The HTML is already in the build output — serving it from S3 would save the Lambda invocation cost and reduce TTFB for those routes from ~200ms (cold Lambda) to ~30ms (S3 cache hit).

Not suggesting this for this PR, but wanted to flag it as a natural follow-up since you already have the data to do it.

osama-rizk:
> One more thought on M2 (prerendered pages from S3):
> 
> Since P1.8 already reads `prerender-manifest.json` to extract trailing-slash redirect routes, you have the full list of prerendered pages available at synth time. Would it be feasible to also emit dedicated static behaviors (or CloudFront Function routing entries) for those pages so they serve directly from S3 instead of going through the catch-all Lambda?
> 
> Right now, a page like `/about` that was prerendered by `generateStaticParams` still hits the SSR Lambda (catch-all `/*`), which re-renders it from the bundle. The HTML is already in the build output — serving it from S3 would save the Lambda invocation cost and reduce TTFB for those routes from ~200ms (cold Lambda) to ~30ms (S3 cache hit).
> 
> Not suggesting this for this PR, but wanted to flag it as a natural follow-up since you already have the data to do it.

Sure will address that in a follow up PR.

PR #3230, opened 2026-06-05
feat(hosting): SSR hosting batch — Astro S3 + ISR seeding, config surface, provisioned concurrency, multi-edge, fail-soft WS/cron/proxy

osama-rizk:
## What

A batch of Amplify self-managed hosting improvements across the framework adapters (Next/OpenNext, Nitro/Nuxt, Astro) and the L3 construct. Started as the Astro prerendered-from-S3 + OpenNext ISR cache-seeding work, then grew to cover a review-driven sweep of correctness, cost, configurability, and "fail-soft" fixes surfaced by a deep hosting review.

12 commits · 30 files · ~+1.6k/−0.3k. Each commit is atomic, has a changeset, and ships tests. Full hosting suite: **750 passing**.

> **Preview note:** `@aws-amplify/hosting` is Developer Preview and not yet released. Several commits change CloudFormation logical IDs / resource topology (e.g. the edge-function id, the SSR provisioned-concurrency alias). That's safe pre-release (no installed base to upgrade), but worth knowing if cherry-picking onto a deployed stack.

### Serving & caching
- **Serve prerendered (SSG) Astro pages from S3.** The adapter's prerendered-page → S3 routing was gated on the deprecated `output: 'hybrid'`, so modern `output: 'server'` + per-page `prerender = true` builds had every prerendered page fall through to the catch-all SSR Lambda — re-rendering frozen HTML and paying a Lambda invocation + cold TTFB (~200ms) instead of a ~30ms S3 hit. Now any non-`static` build walks `dist/client/**/*.html` and emits `target: 'static'` (S3) behaviors (bare `/about` + subtree `/about/*`) for each prerendered page, bringing Astro to parity with Nuxt/Nitro. *Verified live: `/about` and `/about/` return `server: AmazonS3` with no `x-amz-apigw-id`.* A `MAX_PRERENDERED_STATIC_PAGES` cap (8) keeps the per-page behaviors under CloudFront's 25-behavior limit; overflow pages still serve correctly via the SSR Lambda, with the dropped count logged.
- **Seed the OpenNext ISR cache at deploy time** (S3 incremental cache + DynamoDB tag table via `dynamodb-provider`). Prerendered ISR/SSG pages are now a cache HIT on first request instead of a cold on-demand render; tag-based revalidation works before a page's first hit. *(Verified live: `x-nextjs-cache` flips MISS→HIT; tag table seeded with 4776 rows.)*
- **Honor OpenNext `disableTagCache`** — no longer provisions an unused, billed DynamoDB table when the user opted out of the tag cache.
- **Emit proper `stale-while-revalidate` for Nitro SWR rules** — a `swr` route rule now emits `s-maxage=N, stale-while-revalidate=…` (no `max-age`) instead of hard `max-age=N, s-maxage=N`.

### Compute & routing
- **Wire `compute.provisionedConcurrency` end-to-end for SSR** — previously inert; now creates a `live` alias and points the REST API integration at it so warm instances actually serve traffic.
- **Support multiple edge routes outside us-east-1** — unique `EdgeFunction-<name>` construct id (the literal `'EdgeFunction'` collided in the shared cross-region `edge-lambda-stack`); also fixes a cross-region LogGroup synth failure on the edge function.
- **Stop Astro deploying a dead image-opt Lambda** — the adapter set `imageOptimization` with no route targeting it (a 1024 MB / reserved-concurrency-10 Lambda receiving zero traffic, mis-typed as `handler` over an HTTP listener). `/_image` now rides the SSR Lambda.

### Configurability (public `defineHosting` surface)
- **Surface `cdn.ssrDefaultTtl` / `cdn.responseHeadersPolicy` / `cdn.webAclArn`** on `HostingProps.cdn` (already plumbed downstream; were unreachable without a cast). Unblocks edge-cached SSR, sharing one RHP across stacks to dodge the account quota, and BYO WAF.
- **Surface `skewProtection`** in `HostingProps` + validate `maxAge <= storage.buildRetentionDays` (a cookie outliving the build prefix pins viewers to a deleted build → 403).
- **Widen `compute.timeout`** public type to `Duration | number` (matches the construct's existing coercion).

### Fail-soft policy (warn, don't block a working deploy)
- **WebSocket & cron warn across all adapters** instead of throwing. Neither is supported on this architecture (REST + request/response Lambda can't complete a WS upgrade; no general scheduler is wired), but the rest of the app deploys fine — so warn rather than fail the build. Also catches Nitro 3's renamed `features.websocket` flag (the 2.x check was `experimental.websocket` only).
- **Upstream-proxy rewrites (`routeRules.proxy`) warn instead of throwing** `RewritesNotYetSupportedError` — the proxy already works via the SSR Lambda's Nitro runtime (it's just not edge-optimized yet).
- **Accept OpenNext's native `aws-apigw-streaming` wrapper** in `validateUserOpenNextConfig` (was forcing users onto the community `aws-lambda-streaming` wrapper we monkeypatch).
- **nitropack version-range warning** (`warnIfNitroOutOfRange`) + corrected the Nitro patcher's documented semantics — zero patches is a legitimate Nitro v3 state (no patch needed), not an error.

## Tests
- 750 hosting unit tests pass; new tests for every change above, plus a shared `feature_warnings` module + tests.
- Live verification on the next-app-router stress app (444209914584): ISR MISS→HIT, tag-table seeded, hashed-asset `immutable`, security headers, POST body integrity, no regressions vs the VERIFY-2 matrix.

## Notes
- A few review findings were **corrected during implementation** rather than applied verbatim — e.g. the proposed torn-deploy fix would have created a CFN dependency cycle, and "throw on zero Nitro patches" would have broken the working Nitro v3 path. Origin-verify (`aws:Referer`) was live-verified to work correctly, so no fix was needed there.


osama-rizk:
## Live verification + scope correction

Redeployed the stress-test apps and probed response origin (`x-amz-apigw-id` = SSR Lambda; `server: AmazonS3` = S3). **Findings changed the scope of this PR — now Astro-only.**

### Astro — fixed ✅
| Route | Before | After |
|---|---|---|
| `/about` (server-mode `prerender = true`) | SSR Lambda (`x-amz-apigw-id`) | **`server: AmazonS3`** |
| `/about/` | SSR Lambda | **`server: AmazonS3`** |
| `/api-test`, `/` (SSR) | Lambda | Lambda (unchanged — correct control) |

### Next.js — reverted, not applicable ⚠️
The deployed probe revealed OpenNext stores prerendered output as **`.cache` blobs** in its incremental cache (`.open-next/cache/products/0.cache`, `dashboard.cache`), **not** flat `.html` in `assets/`. Those are read by the SSR Lambda's cache handler, which is also what drives ISR revalidation + cache-tag purging. Serving from S3 would bypass ISR entirely. `assets/` contains only `_next/static/*` + public files — no page HTML for a static route to target.

So the Next implementation I first wrote was effectively dead code (walked a dir that never has page HTML). **Reverted it.** Prerendered Next pages correctly continue through the Lambda (`x-nextjs-cache: HIT`).

### Nuxt — no regression ✅
`/about` still `server: AmazonS3` (Nuxt already did this; untouched).

### Regression sweep (redeployed Next + Astro)
Astro D1/D2/D3 + 404, Next B1/B2/B4/B18/B16 + POST `/api/echo` body integrity — all pass. No regressions.

### Net
Astro `server`-mode prerendered pages now serve from S3 (the actual gap). Next is architecturally not applicable (documented in the changeset). 724/724 unit tests pass (3 new Astro `server`-mode tests). Verification recorded in the VERIFY-2 matrix (new Section E).

adrianjoshua-strutt (approved):

Issue #3211, opened 2026-06-01
🚀 RFC: Self-Managed Hosting and CI/CD


adrianjoshua-strutt:
# RFC: Self-Managed Hosting and CI/CD for Amplify Gen2

**Status:** Developer Preview. This feature is under active development. APIs may change before GA. We are publishing this RFC to gather community feedback on the design and prioritize features.

Development branch: https://github.com/aws-amplify/amplify-backend/tree/snapshot/iac-hosting

## Motivation

Amplify Gen2 uses a code-first approach. You define backend resources in TypeScript using `define*()` functions (`defineAuth`, `defineData`, `defineStorage`, `defineFunction`). Each one wraps CDK constructs, giving you full control: version-controlled infrastructure, reviewable in PRs, and extensible via CDK escape hatches.

Amplify Hosting provides a managed solution for deploying and hosting frontend applications. It handles CI/CD, builds, and serves your app. And for many teams, that's exactly what they need.

But some teams want more control. They want to:

- **Version control their hosting infrastructure:** review CloudFront, S3, and Lambda configurations in pull requests, just like they do for backend resources.
- **Use CDK escape hatches:** customize cache behaviors, attach WAF rules, tune Lambda memory, or add resources that Hosting doesn't expose.
- **Own the infrastructure:** deploy into their own account with full visibility into every resource.
- **Use their own CI/CD:** integrate with GitHub Actions, GitLab CI, or Jenkins instead of a managed pipeline.
- **Deploy backend independently:** ship backend changes without a frontend build, or frontend changes without touching the backend.

This RFC introduces two new constructs that bring hosting and CI/CD into the same code-first model:

- **`defineHosting()`**: define your frontend hosting infrastructure as code
- **`definePipeline()`**: define your CI/CD pipeline as code

Both follow the same patterns you already know from Amplify Gen2, and both are fully optional. Amplify Hosting continues to work as it does today.

Each piece is independently adoptable:

- Want to deploy only your backend? Just use `ampx deploy --backend`.
- Want to add self-managed hosting? Add `amplify/hosting.ts`.
- Want a fully automated pipeline? Add `amplify/pipeline.ts`.
- Already have CI/CD (GitHub Actions, GitLab CI)? Skip `definePipeline` and call `ampx deploy` from your existing pipeline.

---

### Prerequisites

- An existing Amplify Gen2 project (or create one with `npm create amplify@latest`)
- AWS credentials configured
[...]

PR #3246, opened 2026-06-17
refactor(hosting): import hosting construct and adapters from @aws-blocks/hosting

osama-rizk:
## Summary

The Amplify self-hosting solution previously implemented its hosting infrastructure (the L3 CDK construct, framework adapters, deploy manifest, and build defaults) as a **fork** of [`@aws-blocks/hosting`](https://www.npmjs.com/package/@aws-blocks/core). This PR removes the fork and **imports the construct and adapters directly from `@aws-blocks/hosting`**, per the self-hosting RFC (#3211).

The public API of `@aws-amplify/hosting` is **unchanged**. All Amplify-specific glue is retained.

## What changed

**Now imported from `@aws-blocks/hosting`** (fork deleted, ~28k lines removed):
- `constructs/` — the L3 `HostingConstruct` (+ `cdn`, `compute`, `storage`, `dns`, `waf`, `monitoring`, `security_headers`, `skew_protection`)
- `adapters/` — Next.js, Nitro, Nuxt, Astro, SPA/static adapters + `detectFramework`/`getAdapter`
- `manifest/` — `DeployManifest` types + Zod schema
- `defaults.ts`, `constants.ts`

**Kept (Amplify-specific glue), rewired onto the imported construct:**
- `factory.ts` — `defineHosting()`: deploy lock, backend identifier, `StackMetadataBackendOutputStorageStrategy`, IPC synth
- `pipeline/` — `definePipeline()` / `AmplifyPipelineConstruct` (not part of aws-blocks)
- `build/runner.ts` — `runBuild`

**Re-export shims** preserve the public API and package `exports` subpaths (`.`, `./constructs`, `./adapters`, `./error`):
- `index.ts`, `types.ts`, `hosting_error.ts`, `adapters/index.ts`, `constructs/hosting_construct.ts`
- `AmplifyHostingConstruct` is re-exported as an alias of the upstream `HostingConstruct` (verified identical at runtime), so consumers — including the vanilla-CDK integration tests — keep working unchanged.

**Other:**
- `package.json`: add `@aws-blocks/hosting@^0.1.1` (the version published to the **public** npm registry — what CI installs and the e2e suite was validated against; `0.23.0` exists only on the internal mirror today); drop deps now provided upstream (`cross-spawn`, `fast-glob`, `jiti`, `local-pkg`, `parse-cache-control`, `semver`, `backend-platform-test-stubs`).
- Regenerated `API.md`; exported `DefinePipelineProps` (clears a pre-existing `ae-forgotten-export`); synced `tsconfig.json` references.

## Behavioral notes

The upstream construct + adapters move together, so two cosmetic deltas are internally consistent: staging dir `.hosting` (was `.amplify-hosting`) and the compute env-var prefix `HOSTING_` (was `AMPLIFY_`). No repo code outside this package reads the old names. The upstream package is also slightly *ahead* of the fork (e.g. it ships the `spaFallback` / default 404 page feature).

## Validation

- `tsc --build` for `@aws-amplify/hosting` ✅
- `tsc --build` for `@aws-amplify/integration-tests` (the cross-package consumer) ✅
- 106/106 unit tests pass, incl. `sub_path_exports` (all four subpaths resolve) ✅
- Runtime smoke test confirms `AmplifyHostingConstruct === @aws-blocks/hosting HostingConstruct` ✅
- ESLint + Prettier clean on changed files ✅
- `changeset status`, `check:package-lock` pass ✅

### Live deployments (real-world stress apps)

Deployed Amplify Gen2 stress-test apps against this branch's hosting construct (`ampx deploy --backend` + `--frontend`) and probed the live CloudFront URLs:

| App | Framework | Deployed URL | Status |
|---|---|---|---|
| amplify-next-app-router | Next.js App Router (OpenNext SSR) | https://d16f3ng9h9869u.cloudfront.net/ | ✅ live (200) |
| amplify-astro-stress | Astro (server + prerender) | https://d3mayu2k5zisdj.cloudfront.net/ | ✅ live (200) — SSR/prerender/api/redirect/404/headers verified |

E2E / integration tests will run on this PR.


packages/hosting/package.json
```diff
@@ -43,20 +43,7 @@
...
-    "semver": "^7.6.0"
-  },
-  "devDependencies": {
-    "@aws-amplify/backend-platform-test-stubs": "^0.3.10",
-    "@opennextjs/aws": "~3.10.0",
-    "@types/cross-spawn": "^6.0.6",
-    "@types/parse-cache-control": "^1.0.4",
-    "@types/semver": "^7.5.8",
-    "zod": "3.25.17"
+    "@aws-blocks/hosting": "^0.1.1"
```
adrianjoshua-strutt, line 46: This is a little confusing to me, could you help me understand the intent. The title and changeset reference 0.23.0, but the pin here is ^0.1.1, and the comment in index.ts notes 0.23.0 only exists on the internal mirror. Which version is this meant to ship against, and is it the one the tests were validated on? Mainly want to be sure a clean public install resolves what you actually tested.

packages/hosting/src/constructs/hosting_construct.test.ts
```diff
@@ -1,3037 +0,0 @@
-import { afterEach, describe, it } from 'node:test';
```
adrianjoshua-strutt, line 1: All the construct, adapter, and manifest tests come out here since the implementation moved upstream. Would it be sensible to keep some of them (or a slimmer parity set) in this repo so we still have local coverage that the wired-up version produces what we expect? Or is the plan to fully rely on the upstream package's suite for this? Genuinely asking, want to understand the coverage story.

package.json
```diff
@@ -107,6 +107,9 @@
   "workspaces": [
     "packages/*"
   ],
+  "overrides": {
+    "aws-cdk-lib": "2.257.0"
```
adrianjoshua-strutt: Could you say a bit more about why the aws-cdk-lib override is workspace-wide rather than scoped to the hosting package? Trying to understand whether every package in the monorepo needs to move to 2.257.0 for this, or whether it's specifically a hosting peer-dep requirement.

.github/dependency_review_config.yml
```diff
@@ -45,3 +45,10 @@ allow-ghsas:
   - 'GHSA-jmr7-xgp7-cmfj' # fast-xml-parser DoS through entity expansion in DOCTYPE (HIGH)
   - 'GHSA-gh4j-gqv2-49f6' # fast-xml-parser XMLBuilder XML Comment and CDATA injection (MODERATE)
   - 'GHSA-fj3w-jwp8-x2g3' # fast-xml-parser stack overflow in XMLBuilder with preserveOrder (LOW)
+  # Transitive vulnerabilities reachable via @opennextjs/aws -> next. @opennextjs/aws is an
+  # optional peer dependency of @aws-blocks/hosting (which @aws-amplify/hosting consumes), so npm
+  # installs it on a runtime path. These advisories pre-exist on the base branch (where the same
+  # packages were dev-scoped) and cannot be patched without upstream @opennextjs/aws / next releases.
+  - 'GHSA-gv7w-rqvm-qjhr' # esbuild dev-server enables any website to read responses (HIGH)
```
adrianjoshua-strutt, line 52: Could you walk through why these three advisories need allowlisting as part of this PR, including the esbuild HIGH? From what I can tell they come in because @opennextjs/aws lands on a runtime path now. Want to understand why that is necessary for this specific change, and whether allowlisting them is the only option here.

packages/hosting/package.json
osama-rizk: `^0.1.1` is correct — fixed the stale `0.23.0` in the PR description.

Public npm only has `0.1.x` (`latest = 0.1.1`); `0.23.0` is internal-mirror-only, so CI can't install it. `^0.1.1` is what CI installs and what the e2e suite passed against. APIs are compatible, so it's a one-line bump once `0.23.x` hits public npm.

packages/hosting/src/constructs/hosting_construct.test.ts
osama-rizk: Agreed — gap, not intentional. Behavioral tests belong upstream now, but we still want wiring coverage.

Added `shim_parity.test.ts` (12b30d2c): asserts `AmplifyHostingConstruct ===` upstream `HostingConstruct`, the subpaths (`./constructs`/`./adapters`/`./error`) re-export the expected symbols, and a minimal manifest synthesizes a CloudFront + S3 origin. Catches drift without duplicating upstream's suite. 112/112 pass.

package.json
osama-rizk: It's hosting's peer requirement, not a repo-wide move. `@aws-blocks/hosting` needs `aws-cdk-lib ^2.257.0`; `2.257.0` satisfies everyone else's `^2.234.1`, so it just pins the single hoisted copy. npm only honors `overrides` at the workspace root — there's no package-scoped option.

(The paired `cloud-assembly-schema` pin is the same fix: without it, synth emitted schema-54 manifests the pinned CDK CLI couldn't read — the original `vanilla_cdk` e2e failure.)

.github/dependency_review_config.yml
osama-rizk: Correct — all three come via `@opennextjs/aws → next` (an optional peer of `@aws-blocks/hosting`, on a runtime path). They already exist on base (dev-scoped there), aren't patchable by us (need upstream `next`/`esbuild` releases), and match the existing `fast-xml-parser` allowlist precedent here.

Only alternative is demoting `@opennextjs/aws` to a dev dep — happy to do that instead if you prefer. The esbuild HIGH is a dev-server SSRF that doesn't apply to our usage.

adrianjoshua-strutt (approved):

PR #3260, opened 2026-07-01
fix(backend-function): avoid serializing full env in malformed data-env error

ahmedhamouda78:
## Problem

When the data environment variables consumed by `getAmplifyDataClientConfig` are missing or malformed, the internal `extendEnv` validation throws an error that serializes the **entire** runtime environment into the message:

```ts
throw new Error(
  `The data environment variables are malformed. env=${JSON.stringify(env)}`,
);
```

In a Lambda runtime, `env` includes values unrelated to the data configuration — for example the standard `AWS_*` variables and any additional environment values configured on the function. Embedding the whole environment in an error message is more than is needed to diagnose the problem, and that message can be surfaced in logs. The error only needs to identify which of the expected data variables is missing or malformed.

## Changes

- `extendEnv` now computes which of the three expected data variables (`..._MODEL_INTROSPECTION_SCHEMA_BUCKET_NAME`, `..._MODEL_INTROSPECTION_SCHEMA_KEY`, `..._GRAPHQL_ENDPOINT`) are missing or not strings, and throws an error that lists only those **variable names** — it no longer serializes `env` values into the message.
- The new message stays actionable: it names exactly which variables need to be provided as strings.

Before:
```
The data environment variables are malformed. env={ ...entire environment serialized... }
```
After:
```
The data environment variables are malformed. Expected string values for: AMPLIFY_DATA_MODEL_INTROSPECTION_SCHEMA_BUCKET_NAME, AMPLIFY_DATA_MODEL_INTROSPECTION_SCHEMA_KEY, AMPLIFY_DATA_GRAPHQL_ENDPOINT.
```

## Validation

Added a unit test to `packages/backend-function/src/runtime/get_amplify_clients_configuration.test.ts` that calls `getAmplifyDataClientConfig` with a malformed data environment and asserts that:
- the thrown message names the three expected data variables, and
- none of the surrounding environment values (e.g. `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`, and an additional configured env value) appear in the message.

## Checklist

### DIFF
--- a/.changeset/redact-data-env-error-message.md
+++ b/.changeset/redact-data-env-error-message.md
@@ -1,0 +1,8 @@
+---
+'@aws-amplify/backend-function': patch
+'@aws-amplify/backend': patch
+---
+
+fix(backend-function): avoid serializing the full environment in the malformed data-env error
+
+When the data environment variables are missing or malformed, `getAmplifyDataClientConfig` no longer serializes the entire runtime environment into the thrown error. The message now lists only the names of the missing/malformed variables, which keeps it debuggable without including unrelated environment values.
--- a/packages/backend-function/src/runtime/get_amplify_clients_configuration.test.ts
+++ b/packages/backend-function/src/runtime/get_amplify_clients_configuration.test.ts
@@ -138,15 +138,6 @@ void describe('getAmplifyDataClientConfig', () => {
   });
 
   void describe('when the data environment variables are malformed', () => {
-    // The malformed-env error should name the missing/malformed variables
-    // without echoing the values of unrelated environment entries.
-    const unrelatedEnvValues = [
-      'TEST_VALUE for AWS_ACCESS_KEY_ID',
-      'TEST_VALUE for AWS_SECRET_ACCESS_KEY',
-      'TEST_VALUE for AWS_SESSION_TOKEN',
-      'TEST_VALUE for MY_OTHER_ENV_VAR',
-    ];
-
     const malformedEnv = {
       AWS_ACCESS_KEY_ID: 'TEST_VALUE for AWS_ACCESS_KEY_ID',
       AWS_SECRET_ACCESS_KEY: 'TEST_VALUE for AWS_SECRET_ACCESS_KEY',
@@ -157,6 +148,11 @@ void describe('getAmplifyDataClientConfig', () => {
       // The derived data env vars (bucket/key/endpoint) are intentionally
       // absent, which triggers the malformed-env validation.
     };
+
+    // None of the entries present in the fixture are the missing data
+    // variables, so none of their values should appear in the error. Deriving
+    // the list from the fixture keeps the two in sync if the fixture changes.
+    const unrelatedEnvValues = Object.values(malformedEnv);
 
     void it('names the missing variables without echoing unrelated env values', async () => {
       await assert.rejects(
@@ -180,6 +176,43 @@ void describe('getAmplifyDataClientConfig', () => {
               `error message must not contain the value "${value}"`,
             );
           }
+          return true;
+        },
+      );
+    });
+
+    void it('lists only the invalid variable names, not the ones present', async () => {
+      // Only the GraphQL endpoint is missing; the bucket and key are present
+      // as valid strings, so the error must name the endpoint alone.
+      const envMissingOnlyEndpoint = {
+        ...malformedEnv,
+        AMPLIFY_DATA_MODEL_INTROSPECTION_SCHEMA_BUCKET_NAME:
+          'TEST_VALUE for AMPLIFY_DATA_MODEL_INTROSPECTION_SCHEMA_BUCKET_NAME',
+        AMPLIFY_DATA_MODEL_INTROSPECTION_SCHEMA_KEY:
+          'TEST_VALUE for AMPLIFY_DATA_MODEL_INTROSPECTION_SCHEMA_KEY',
+        // AMPLIFY_DATA_GRAPHQL_ENDPOINT is still absent.
+      };
+
+      await assert.rejects(
+        async () =>
+          await getAmplifyDataClientConfig(
+            envMissingOnlyEndpoint,
+            mockS3Client,
+          ),
+        (error: Error) => {
+          assert.match(error.message, /AMPLIFY_DATA_GRAPHQL_ENDPOINT/);
+          assert.ok(
+            !error.message.includes(
+              'AMPLIFY_DATA_MODEL_INTROSPECTION_SCHEMA_BUCKET_NAME',
+            ),
+            'present bucket variable must not be listed as malformed',
+          );
+          assert.ok(
+            !error.message.includes(
+              'AMPLIFY_DATA_MODEL_INTROSPECTION_SCHEMA_KEY',
+            ),
+            'present key variable must not be listed as malformed',
+          );
           return true;
         },
       );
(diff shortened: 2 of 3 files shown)


packages/backend-function/src/runtime/get_amplify_clients_configuration.test.ts
```diff
@@ -136,4 +136,53 @@ void describe('getAmplifyDataClientConfig', () => {
       });
     });
   });
+
+  void describe('when the data environment variables are malformed', () => {
+    // The malformed-env error should name the missing/malformed variables
+    // without echoing the values of unrelated environment entries.
+    const unrelatedEnvValues = [
```
soberm: [minor] `unrelatedEnvValues` is hardcoded independently from the values in `malformedEnv`. If someone updates the fixture values without also updating this array, the negative assertions pass vacuously (the old strings won't appear in the error regardless) and the test silently loses its anti-regression value. Coupling them avoids this drift, e.g. derive the list from the fixture directly:

```ts
const malformedEnv = {
  AWS_ACCESS_KEY_ID: 'test-access-key',
  // ...
};
const unrelatedEnvValues = [
  malformedEnv.AWS_ACCESS_KEY_ID,
  malformedEnv.AWS_SECRET_ACCESS_KEY,
  malformedEnv.AWS_SESSION_TOKEN,
  malformedEnv.MY_OTHER_ENV_VAR,
];
```

packages/backend-function/src/runtime/get_amplify_clients_configuration.test.ts
```diff
@@ -136,4 +136,53 @@ void describe('getAmplifyDataClientConfig', () => {
...
+      AWS_SECRET_ACCESS_KEY: 'TEST_VALUE for AWS_SECRET_ACCESS_KEY',
+      AWS_SESSION_TOKEN: 'TEST_VALUE for AWS_SESSION_TOKEN',
+      AWS_REGION: 'TEST_VALUE for AWS_REGION',
+      AMPLIFY_DATA_DEFAULT_NAME: 'AmplifyData',
+      MY_OTHER_ENV_VAR: 'TEST_VALUE for MY_OTHER_ENV_VAR',
+      // The derived data env vars (bucket/key/endpoint) are intentionally
+      // absent, which triggers the malformed-env validation.
+    };
+
+    void it('names the missing variables without echoing unrelated env values', async () => {
```
soberm, line 157: [minor] This only tests the "all three variables absent" scenario. The key behavioral property introduced here is that the error lists *only* the invalid variable names, not all three unconditionally. That property goes untested. Worth adding a case where just one variable is missing (e.g. only `AMPLIFY_DATA_GRAPHQL_ENDPOINT`) and asserting that the other two names do not appear in the error message.

packages/backend-function/src/runtime/get_amplify_clients_configuration.test.ts
```diff
@@ -136,4 +136,53 @@ void describe('getAmplifyDataClientConfig', () => {
...
+      'TEST_VALUE for AWS_SECRET_ACCESS_KEY',
+      'TEST_VALUE for AWS_SESSION_TOKEN',
+      'TEST_VALUE for MY_OTHER_ENV_VAR',
+    ];
+
+    const malformedEnv = {
+      AWS_ACCESS_KEY_ID: 'TEST_VALUE for AWS_ACCESS_KEY_ID',
+      AWS_SECRET_ACCESS_KEY: 'TEST_VALUE for AWS_SECRET_ACCESS_KEY',
+      AWS_SESSION_TOKEN: 'TEST_VALUE for AWS_SESSION_TOKEN',
+      AWS_REGION: 'TEST_VALUE for AWS_REGION',
```
soberm, line 145: [nit] `malformedEnv` has `AWS_REGION` but its value isn't included in `unrelatedEnvValues`, so the negative check doesn't cover it. Looks like an oversight given all the other unrelated keys are represented.

packages/backend-function/src/runtime/get_amplify_clients_configuration.test.ts
```diff
@@ -136,4 +136,53 @@ void describe('getAmplifyDataClientConfig', () => {
       });
     });
   });
+
+  void describe('when the data environment variables are malformed', () => {
+    // The malformed-env error should name the missing/malformed variables
```
soberm: [nit] This comment just restates the surrounding `describe` label. Remove it.

.changeset/redact-data-env-error-message.md
```diff
@@ -0,0 +1,8 @@
+---
+'@aws-amplify/backend-function': patch
+'@aws-amplify/backend': patch
```
soberm, line 3: [question] Is the `@aws-amplify/backend` bump intentional here? The code change is only in `backend-function`. Does `@aws-amplify/backend` re-export `getAmplifyDataClientConfig` or have some other dependency relationship that warrants co-versioning?

packages/backend-function/src/runtime/get_amplify_clients_configuration.test.ts
ahmedhamouda78: Good catch — fixed in 1a2c92c. `unrelatedEnvValues` is now derived directly from the fixture (`Object.values(malformedEnv)`), so it can't drift out of sync when the fixture changes.

packages/backend-function/src/runtime/get_amplify_clients_configuration.test.ts
ahmedhamouda78: Agreed — this was the more important gap. Added a second test in 1a2c92c that leaves the bucket and key present as valid strings and omits only `AMPLIFY_DATA_GRAPHQL_ENDPOINT`, then asserts the endpoint is named while the two present variables are not listed. That locks in the "lists only the invalid names" property.

packages/backend-function/src/runtime/get_amplify_clients_configuration.test.ts
ahmedhamouda78: Fixed in 1a2c92c — resolved as a side effect of deriving `unrelatedEnvValues` from `Object.values(malformedEnv)`, which now includes `AWS_REGION`'s value in the negative assertions.

packages/backend-function/src/runtime/get_amplify_clients_configuration.test.ts
ahmedhamouda78: Removed in 1a2c92c.

.changeset/redact-data-env-error-message.md
ahmedhamouda78: Yes, intentional — and it's actually required by this repo's own changeset completeness check (`scripts/check_changeset_completeness.ts`, the `checkBackendDependenciesVersion` rule). `@aws-amplify/backend` depends on `@aws-amplify/backend-function`, so when `backend-function` gets a version bump the check requires `@aws-amplify/backend` to bump by at least the same kind — otherwise CI fails with a "changeset is missing packages" error. My first push had only the `backend-function` patch and `check_pr_changesets` failed for exactly this reason; adding the `@aws-amplify/backend` patch resolved it. It's consistent with how prior `backend-function` changes have been co-versioned.

soberm (approved):

sarayev:
Nice cleanup — dropping the full-env serialization is the right call, and I like that the message still names exactly which vars are off. One tiny thing (non-blocking): since the check also catches vars that are entirely absent, wording it as "missing or non-string values for:" might read a hair more accurately than "expected string values for:". Either way, happy to approve.

sarayev (approved):
Approving — clean, security-positive change with good test coverage. Left one non-blocking wording nit in a separate comment.

PR #3263, opened 2026-07-02
feat(sandbox): support CloudFormation/CDK Express mode via `ampx sandbox --express`

### LINKED ISSUE #3261: Support CloudFormation/CDK Express mode for faster sandbox deployments
### Environment information

```plain text
System:
  OS: macOS 26.5.1
  CPU: (10) arm64 Apple M5
  Memory: 159.33 MB / 32.00 GB
  Shell: /bin/zsh
Binaries:
  Node: 24.18.0 - /Users/fossamagna/.vite-plus/js_runtime/node/24.18.0/bin/node
  Yarn: 1.22.22 - /Users/fossamagna/.vite-plus/bin/yarn
  npm: 11.16.0 - /Users/fossamagna/.vite-plus/js_runtime/node/24.18.0/bin/npm
  pnpm: 10.33.4 - /Users/fossamagna/.vite-plus/bin/pnpm
NPM Packages:
  @aws-amplify/ai-constructs: Not Found
  @aws-amplify/auth-construct: Not Found
  @aws-amplify/backend: 1.22.0
  @aws-amplify/backend-ai: Not Found
  @aws-amplify/backend-auth: Not Found
  @aws-amplify/backend-cli: 1.8.2
  @aws-amplify/backend-data: Not Found
  @aws-amplify/backend-deployer: Not Found
  @aws-amplify/backend-function: Not Found
  @aws-amplify/backend-output-schemas: Not Found
  @aws-amplify/backend-output-storage: Not Found
  @aws-amplify/backend-secret: Not Found
  @aws-amplify/backend-storage: Not Found
  @aws-amplify/cli-core: Not Found
  @aws-amplify/client-config: Not Found
  @aws-amplify/data-construct: Not Found
  @aws-amplify/data-schema: Not Found
  @aws-amplify/deployed-backend-client: Not Found
  @aws-amplify/form-generator: Not Found
  @aws-amplify/model-generator: Not Found
  @aws-amplify/platform-core: Not Found
  @aws-amplify/plugin-types: Not Found
  @aws-amplify/sandbox: Not Found
  @aws-amplify/schema-generator: Not Found
  @aws-cdk/toolkit-lib: Not Found
  aws-amplify: 6.17.0
  aws-cdk-lib: 2.253.1
  typescript: 6.0.3
AWS environment variables:
  AWS_EC2_METADATA_DISABLED = true
No CDK environment variables
```

### Describe the feature

I'd like ampx sandbox to support the newly released [CloudFormation and CDK Express mode](https://aws.amazon.com/about-aws/whats-new/2026/06/aws-cloudformation-cdk/) as an opt-in deployment option.

Express mode lets stack operations complete once resource configuration is applied, instead of waiting for extended stabilization (e.g. traffic-readiness checks, region propagation). AWS reports up to ~4x faster deployments, and it disables rollback by default to enable fast fix-and-retry iteration — a good fit for the sandbox development loop.

Today the sandbox deploy path uses `@aws-cdk/toolkit-lib` programmatically (`Toolkit.deploy()`), and for sandbox it passes:

```
deploymentMethod: { method: 'hotswap', fallback: { method: 'direct' } }
```

The request is to wire Express mode into this path — e.g. an opt-in flag such as `ampx sandbox --express` (or a config option) that enables Express mode when the underlying @aws-cdk/toolkit-lib exposes it (the corresponding CLI/SDK surface is `cdk deploy --express` / `--deployment-config '{"mode":"EXPRESS"}'`). 

### Use case

When AI coding agents are iterating on amplify source code, the deployment feedback loop against the sandbox environment is a major bottleneck. Faster sandbox deployments via Express mode would let both human developers and AI agents receive deploy feedback significantly sooner, tightening the edit → deploy → verify cycle.

fossamagna:
## Problem

`ampx sandbox` has no way to opt into the newly released CloudFormation/CDK Express mode, which lets stack operations complete once resource configuration is applied instead of waiting for extended stabilization. For the sandbox edit -> deploy -> verify loop (used by both humans and AI coding agents), this is a meaningful speedup.

Close #3261

## Changes

- Add an opt-in `ampx sandbox --express` flag. The flag threads `express: true` through `SandboxOptions` -> the sandbox executor -> `BackendDeployer` -> `Toolkit.deploy()`. Express is only applied to sandbox deployments; the
  existing `hotswap` + `direct` fallback deployment method is preserved.
- Bump `@aws-cdk/toolkit-lib` `1.19.0` -> `1.32.0`, the version that exposes the `express` deploy option and sets `DeploymentConfig.Mode = 'EXPRESS'` on the CloudFormation request.
- Raise `@aws-sdk/client-cloudformation` to `^3.1078.0`. `toolkit-lib` only requires `^3`, so the tree resolved an older SDK (3.1017.0) that does not model the `DeploymentConfig` field and silently dropped it, leaving deployments in `STANDARD` mode. Support was added in `3.1077.0`.
- Surface the Express Mode stabilization warning (`CDK_TOOLKIT_W5902`) in the sandbox output, e.g. `Stack deployed using Express Mode. Resources still stabilizing: ...`.

_None yet — the `--express` flag help text is included; longer-form docs can follow._

## Validation

- Unit tests added/updated:
  - `backend-deployer`: asserts `express: true` is passed to `Toolkit.deploy()` only for sandbox deployments, and never for branch deployments.
  - `sandbox`: asserts `express` is forwarded through the executor and `file_watching_sandbox` to `BackendDeployer`.
  - `backend-cli`: asserts `--express` parses and reaches `sandbox.start`.
  - `cli-core`: asserts the `CDK_TOOLKIT_W5902` Express warning is printed.
- Manual verification: created a test project with `npm run setup:test-project <name>` and ran `ampx sandbox --express` against it, confirming the deployment runs in Express mode and the Express Mode stabilization warning is shown.
- Verified at the dependency-resolution level that `@aws-cdk/toolkit-lib@1.32.0` resolves `@aws-sdk/client-cloudformation@3.1078.0` (which serializes the `DeploymentConfig` Express field).
- Note: Express mode applies only to full CloudFormation deployments; changes handled by hotswap bypass CloudFormation and do not emit the Express warning.

## Checklist

### DIFF
--- a/.changeset/sandbox-express-mode.md
+++ b/.changeset/sandbox-express-mode.md
@@ -1,0 +1,7 @@
+---
+'@aws-amplify/backend-cli': minor
+'@aws-amplify/sandbox': minor
+'@aws-amplify/backend-deployer': minor
+---
+
+Add opt-in `ampx sandbox --express` flag that enables CloudFormation/CDK Express mode for faster sandbox deployments. Bumps `@aws-cdk/toolkit-lib` to `1.32.0`.
--- a/.changeset/sandbox-express-mode.md
+++ b/.changeset/sandbox-express-mode.md
@@ -3,6 +3,7 @@
 '@aws-amplify/sandbox': minor
 '@aws-amplify/backend-deployer': minor
 '@aws-amplify/cli-core': patch
+'@aws-amplify/plugin-types': patch
 ---
 
 Add opt-in `ampx sandbox --express` flag that enables CloudFormation/CDK Express mode for faster sandbox deployments. When a deployment completes with resources still stabilizing, the Express Mode warning (e.g. `Stack deployed using Express Mode. Resources still stabilizing: ...`) is surfaced in the sandbox output. Bumps `@aws-cdk/toolkit-lib` to `1.32.0` and `@aws-sdk/client-cloudformation` to `^3.1078.0` (the version that adds the `DeploymentConfig` Express mode field to the CloudFormation request; older SDK versions silently drop it).
--- a/.changeset/sandbox-express-mode.md
+++ b/.changeset/sandbox-express-mode.md
@@ -5,4 +5,4 @@
 '@aws-amplify/cli-core': patch
 ---
 
-Add opt-in `ampx sandbox --express` flag that enables CloudFormation/CDK Express mode for faster sandbox deployments. Bumps `@aws-cdk/toolkit-lib` to `1.32.0` and `@aws-sdk/client-cloudformation` to `^3.1078.0` (the version that adds the `DeploymentConfig` Express mode field to the CloudFormation request; older SDK versions silently drop it).
+Add opt-in `ampx sandbox --express` flag that enables CloudFormation/CDK Express mode for faster sandbox deployments. When a deployment completes with resources still stabilizing, the Express Mode warning (e.g. `Stack deployed using Express Mode. Resources still stabilizing: ...`) is surfaced in the sandbox output. Bumps `@aws-cdk/toolkit-lib` to `1.32.0` and `@aws-sdk/client-cloudformation` to `^3.1078.0` (the version that adds the `DeploymentConfig` Express mode field to the CloudFormation request; older SDK versions silently drop it).
--- a/.changeset/sandbox-express-mode.md
+++ b/.changeset/sandbox-express-mode.md
@@ -2,6 +2,7 @@
 '@aws-amplify/backend-cli': minor
 '@aws-amplify/sandbox': minor
 '@aws-amplify/backend-deployer': minor
+'@aws-amplify/cli-core': patch
 ---
 
-Add opt-in `ampx sandbox --express` flag that enables CloudFormation/CDK Express mode for faster sandbox deployments. Bumps `@aws-cdk/toolkit-lib` to `1.32.0`.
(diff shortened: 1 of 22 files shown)


sarayev:
Hi @fossamagna , 


Thanks for your contribution. 

The unit coverage here is solid, but it all mocks `Toolkit.deploy` — so we're asserting that we *ask* for `express: true`, not that a deploy actually runs in Express mode. The gap is that the feature could silently no-op (fall back to `STANDARD`) and every test would still pass. Would it be worth one integration test that asserts Express mode reaches the CFN request? That'd lock in the behavior the manual verification confirmed and catch a regression automatically.

sarayev (approved):

.changeset/sandbox-express-mode.md
```diff
@@ -0,0 +1,8 @@
+---
+'@aws-amplify/backend-cli': minor
+'@aws-amplify/sandbox': minor
+'@aws-amplify/backend-deployer': minor
+'@aws-amplify/cli-core': patch
```
sarayev, line 5: `@aws-amplify/plugin-types` needs to be added to changeset file as well.

fossamagna:
@sarayev 
Thanks for the suggestion. `express_mode_deployment_config.test.ts` sends `CreateStack`/`UpdateStack` through the real resolved `@aws-sdk/client-cloudformation` and asserts `DeploymentConfig.Mode=EXPRESS` reaches the serialized request. Runs in the normal in-memory suite (no `run-e2e`), and fails if the SDK is downgraded below 3.1077.0 (where the field is dropped) — the exact silent-no-op gap you flagged.

.changeset/sandbox-express-mode.md
fossamagna: Added `@aws-amplify/plugin-types: patch` to the changeset.

PR #3264, opened 2026-07-02
feat(hosting): make pipeline a thin shim over @aws-blocks/pipeline

osama-rizk:
## Summary

Makes the `@aws-amplify/hosting` **pipeline** module a thin shim over
[`@aws-blocks/pipeline`](https://github.com/aws-devtools-labs/aws-blocks/tree/main/packages/pipeline),
matching how the hosting construct, adapters, and error types already re-export
`@aws-blocks/*` (guarded by `shim_parity.test.ts`). The pipeline module was the
last piece still forked — ~1,000 lines duplicating generic CDK Pipelines logic
that already lives upstream.

While validating the shim end-to-end (build → link test apps → live deploy),
the branch also picked up the dependency and CI changes needed to make the
whole `snapshot/iac-hosting` feature branch releasable: a workspace-wide
`aws-cdk-lib` peer bump, a `@aws-blocks/hosting` upgrade, a couple of
pre-existing backend bugs surfaced by the newer CDK, and changeset/CI hygiene.

### Pipeline shim (`@aws-amplify/hosting`)
| File | Change |
|---|---|
| `src/pipeline/deploy_stage.ts` | Re-export `DeployStage` / `DeployStageProps` from `@aws-blocks/pipeline` |
| `src/pipeline/types.ts` | Re-export shared config types (`PipelineSourceConfig`, `PipelineSynthConfig`, `PipelineStageConfig`, `BranchConfig`) from blocks; keep only Amplify-owned types locally (`DefinePipelineProps`, and `PipelineProps` = blocks props + `stageFactory`/`_sourceOverride`/`_postStageHook`) |
| `src/pipeline/pipeline_construct.ts` | Replace the ~560-line fork with a thin `AmplifyPipelineConstruct extends Pipeline` (from blocks). Constructor keeps the 4-arg `_internal` async-init param for API compatibility and forwards it to the upstream `Pipeline` |
| `src/pipeline/pipeline_factory.ts` | **Unchanged** — `definePipeline()`, file discovery, ambient scope, and two-phase orchestration remain Amplify-owned glue |
| `src/pipeline/pipeline_construct.test.ts` | Replace the behavioral fork suite (now owned upstream) with shim-parity + glue + smoke tests |

### Dependencies & workspace
- Add `@aws-blocks/pipeline ^0.1.1`; upgrade `@aws-blocks/hosting ^0.1.1 → ^0.1.4`.
- Bump the `aws-cdk-lib` peer floor to `^2.257.0` (and `constructs → ^10.6.0`) across **all 16 workspace packages** so the whole version set agrees with the aws-blocks constructs' floor (avoids a baseline-deps `ERESOLVE`).

### Pre-existing backend fixes surfaced by CDK 2.257
- `packages/backend/src/engine/nodejs_function_project_root.ts` (new helper) + `branch_linker_construct.ts` + `backend_secret_fetcher_provider_factory.ts`: anchor each `NodejsFunction`'s `projectRoot`/`depsLockFilePath` to the handler's own directory. Without this, synth from a consumer/workspace-linked project throws `PathNotUnderRoot` (the entry ships inside `@aws-amplify/backend`, outside the cwd-derived root) — this was failing `e2e_backend_output` on the base branch too.

### Tooling / e2e
- `create-amplify`: bumped so it re-vends with the already-corrected `aws-cdk-lib@2.257.0` scaffold pin (otherwise `npm create amplify` fails `ERESOLVE` against the bumped peer).
- Integration test-projects `standalone-hosting-ssr` + `-monorepo`: bump Next `15.5.15 → ^16.2.9`. `@aws-blocks/hosting@0.1.4`'s OpenNext image-opt patch is only correct for Next ≥16 (see below).
- `scripts/check_pr_size.ts`: exclude `.changeset/` from the size check (release metadata, like the already-excluded `package-lock.json`/`API.md`).
- Consolidated the feature branch's ~60 changesets into one.

## The one piece that isn't a pure re-export: the two-phase deploy

`definePipeline()` needs the Amplify **two-phase deployment** (deploy backend →
`ampx generate outputs` → build frontend → `cdk deploy` hosting), expressed via
the internal `_postStageHook`, which `@aws-blocks/pipeline` does not model. The
shim's thin subclass re-adds only this: it wraps `stageFactory` so
`_postStageHook` runs inline as each stage synthesizes (capturing per-stage
backend `CfnOutput`s from `globalThis`), then injects the returned steps onto
each built `StageDeployment` via `addPost()` (a `CodePipeline` stays mutable
until `buildPipeline()`). Everything else — source, synth, branches, approvals,
bake time, trigger filters, cross-account, sync/async construction — delegates
to blocks unchanged.

## Validation

- **Pipeline shim verified end-to-end**: built the full workspace, linked the
  `test-projects/pipeline` app to the local build, and synthesized via the
  `ampx deploy --pipeline` path → valid CloudFormation with the expected stages
  (S3 source override, Synth, Assets, backend deploy stage). Construct tests
  5/5, factory (two-phase `definePipeline`) tests 32/32.
- **CI green** on the changed surface: `build`, `lint`, `check_pr_size`,
  `check_pr_changesets`, `check_api_extract`, `test_with_baseline_dependencies`,
  `e2e_backend_output`, `e2e_create_amplify`, `e2e_package_manager`,
  `e2e_sandbox`, and the hosting `e2e_deployment` suite.

## Known upstream issue (not fixed here)

`@aws-blocks/hosting@0.1.4`'s `patchImageOptimizerForNext155` has an inverted
version assumption: it believes Next **15.5** added the `fetchInternalImage`
`maximumResponseBody` arg, but Next **16** added it (15.5 is still 4-arg). On
Next 15.5 the patch corrupts an already-correct call → `/_next/image` 500s. This
PR sidesteps it by pinning the SSR e2e apps to Next 16 (the arity the patch
handles). The real fix belongs upstream in `@aws-blocks/hosting` (gate the patch
on Next ≥16).


packages/hosting/src/pipeline/pipeline_construct.ts
```diff
@@ -72,492 +72,187 @@ const DEFAULT_SYNTH_COMMANDS = [
...
+ * aws-blocks names the branch CodePipeline `${id}-${safeBranch}` and the stage
+ * `${id}-${safeBranch}-Stage-${name}`, so the stage's construct id begins with
+ * the owning pipeline's id followed by `-Stage-`. We match on that to pick the
+ * correct branch pipeline, then read its synth ShellStep's primary input (the
+ * source file set).
+ *
+ * Falls back to `undefined` (the hook tolerates a missing source by producing
+ * no steps) if the source cannot be resolved.
+ */
+const resolveSource = (stage: cdk.Stage): IFileSetProducer => {
```
bobbor: **`resolveSource()` — silent-failure risk from id-prefix matching**

`resolveSource()` associates a `DeployStage` with its owning branch `CodePipeline` purely by string prefix (`stage.node.id.startsWith(`${child.node.id}-Stage-`)`) and returns `undefined as unknown as IFileSetProducer` on no match. This hard-codes `@aws-blocks/pipeline`'s internal construct-naming convention in three coupled places: this prefix match, `wrapProps`'s `postSteps.set(stage.stageName, …)`, and `applyPostStageHook`'s keying on `stageDeployment.stageName`.

If a future `@aws-blocks/pipeline` release renames its stage / child-pipeline constructs, the prefix match silently fails → `_postStageHook` returns `[]` → the hosting deploy step vanishes from the pipeline with **no error and no CI signal**.

There's also an undocumented second assumption: for the `startsWith` to hold, `child.node.id` must already equal `${id}-${safeBranch}`, which isn't stated alongside the `${id}-${safeBranch}-Stage-${name}` convention noted in `wrapProps`.

Suggestions:
- Throw a descriptive error instead of the silent `undefined` cast when no source resolves (fail loud, not silent).
- Prefer resolving via construct type / metadata rather than an id-string prefix.
- Add a unit test asserting `resolveSource` finds the branch pipeline's file set for a multi-branch config, so a future upstream rename trips CI instead of silently dropping the deploy step.

packages/hosting/package.json
```diff
@@ -43,12 +43,13 @@
     "@aws-amplify/backend-output-storage": "^1.3.4",
     "@aws-amplify/platform-core": "^1.11.0",
     "@aws-amplify/plugin-types": "^1.12.0",
-    "@aws-blocks/hosting": "^0.1.1"
+    "@aws-blocks/hosting": "^0.1.4",
```
bobbor, line 46: **`@aws-blocks/hosting@0.1.4` image-optimizer patch — customer-facing risk CI can't catch**

Bumping to `@aws-blocks/hosting@0.1.4` pulls in `patchImageOptimizerForNext155`, which is applied unconditionally at build for any customer app. Worth calling out a runtime risk the e2e suite structurally can't surface:

The patch has **no Next-version detection** — it's a structural regex over the OpenNext-bundled `image-optimization-function` bundle that injects `void 0` as a 4th arg to shift `handleRequest` into the 5-arg `fetchInternalImage(href, _req, _res, maximumResponseBody, handleRequest)` slot.

That 5th arg (`maximumResponseBody`) is **not** an original 15.5 or 16.0 feature — it's the CVE-2026-44577 security backport (GHSA-h64f-5h5j-jqjh) that landed in **Next 15.5.16 and 16.2.5** on 2026-05-06. So the patch is only correct when the customer's Next has the 5-arg signature:

| Customer Next | `fetchInternalImage` arity | Patch result |
|---|---|---|
| ≥ 15.5.16 / ≥ 16.2.5 | 5-arg | ✅ correct |
| 15.5.0 – 15.5.15 | 4-arg | ❌ injects `void 0` → handler shifted off the end → 500 on `/_next/image` |
| 16.0.0 – 16.2.4 | 4-arg | ❌ same |

This PR pins the e2e apps to `next@^16.2.9` (≥ 16.2.5), so CI always exercises the correct branch and never hits the broken path. But `@aws-blocks/hosting@0.1.4` ships as a **runtime dependency**, so a customer on Next 15.5.0–15.5.15 or 16.0.0–16.2.4 would get broken image optimization.

Suggestions (best → minimum):
1. Fix upstream in `@aws-blocks/hosting` to gate the patch on the resolved Next version (only inject when ≥ 15.5.16 / ≥ 16.2.5), then bump the dep.
2. Add a build-time guard in the Amplify Next adapter that reads the customer's resolved `next` version and errors/skips the patch below the threshold.
3. At minimum, document a minimum Next version (≥ 15.5.16 / ≥ 16.2.5) in the hosting release notes — silent image breakage is a poor customer experience.

bobbor:
**Three failing checks — all introduced by this PR — and what's needed to green them**

1. **`check_api_changes`** — `packages/hosting/API.md` appears hand-edited. The `create()` shim builds a base `Pipeline` via `Pipeline.create()` and re-tags it with `Object.setPrototypeOf`, so api-extractor may emit a return type of `Pipeline` rather than `AmplifyPipelineConstruct`. Re-run the API-report generation (`npm run update:api` / the repo's api-extractor task) on a clean full install and commit the regenerated `API.md` instead of editing it by hand; adjust the `create()` return-type annotation if the emitted `.d.ts` disagrees.

2. **`e2e_iam_access_drift`** — introduced by the `aws-cdk-lib ^2.234.1 → ^2.257.0` bump (newer CDK emits different default IAM policy structures), compounded by the new `nodejs_function_project_root` helper changing `projectRoot`/`depsLockFilePath` → asset hashes → Lambda logical IDs → IAM refs. Regenerate the IAM baseline/allow-list snapshot against 2.257 and sanity-diff to confirm no unintended permission changes slipped in.

3. **`e2e_amplify_outputs_backwards_compatibility`** — same CDK-bump root cause: synthesized template hashes, output keys, and logical IDs shift with 2.257. Regenerate the backwards-compatibility snapshot against the 2.257 baseline.

Checks 2 and 3 are expected consequences of the intentional CDK bump — they just need baseline regeneration plus a sanity diff. Check 1 is a mechanical API-report regen.

packages/hosting/src/pipeline/pipeline_construct.ts
```diff
@@ -72,492 +72,187 @@ const DEFAULT_SYNTH_COMMANDS = [
...
-
+    const postSteps = new Map<string, Array<ShellStep | CodeBuildStep>>();
+    const instance = (await Pipeline.create<TConfig>(
+      scope,
+      id,
+      wrapProps(props, postSteps),
+    )) as AmplifyPipelineConstruct<TConfig>;
+    // Re-assign the prototype so callers get an AmplifyPipelineConstruct
+    // instance (Pipeline.create constructs a base Pipeline internally).
+    Object.setPrototypeOf(instance, AmplifyPipelineConstruct.prototype);
```
soberm, line 126: [major] `Object.setPrototypeOf` is used here to make the `Pipeline.create()` return value appear as an `AmplifyPipelineConstruct` instance. This carries real costs: (1) `Object.setPrototypeOf` is a well-known V8 deoptimization that degrades performance for the object's lifetime; (2) the object was never constructed via `new AmplifyPipelineConstruct()`, so if the constructor gains validation or instance-initialization logic the async path silently bypasses it; (3) the test at line 82 only asserts `instanceof AmplifyPipelineConstruct`, which passes because of the prototype swap, not because the instance was properly constructed.

The root cause is that `Pipeline.create()` creates a `new Pipeline(...)` internally with no subclass-factory hook, and the private `_ASYNC_INIT` symbol is inaccessible from this shim. A clean fix requires an upstream change to `@aws-blocks/pipeline` (e.g., `Pipeline.create()` accepting a constructor factory, or an async-init mechanism that works from a subclass). Worth filing upstream so this can be removed.

packages/hosting/src/pipeline/pipeline_construct.ts
```diff
@@ -72,492 +72,187 @@ const DEFAULT_SYNTH_COMMANDS = [
...
-  const triggerOnPush = hasTriggerFilters
-    ? false
-    : (branchConfig.triggerOnPush ?? props.source.triggerOnPush ?? true);
-
-  const source =
-    props._sourceOverride ??
-    CodePipelineSource.connection(props.source.repo, branchConfig.branch, {
-      connectionArn: props.source.connectionArn,
-      triggerOnPush,
+  // NOTE: `wrappedStageFactory` must NOT be an async function. aws-blocks'
```
soberm, line 148: [minor] This comment documents a dependency on `@aws-blocks/pipeline`'s internal async detection (`stageFactory.constructor.name === 'AsyncFunction'`). That check is not part of the upstream public API: if it is removed, replaced with a different mechanism, or if a bundler renames `AsyncFunction`, the guard stops working and callers get an unexpected or missing error with no clear cause.

Worth filing upstream to replace the name-based check with an explicit `isAsync?: boolean` flag on `PipelineProps`, so the shim no longer has to work around an implementation detail.

packages/hosting/src/pipeline/pipeline_construct.ts
soberm: [major] `resolveSource` depends on two internal naming conventions of `@aws-blocks/pipeline`: branch `CodePipeline` IDs follow `${id}-${safeBranch}` and stage construct IDs follow `${id}-${safeBranch}-Stage-${name}`. These are implementation details, not public API contracts. A cosmetic rename upstream (for example, `-Stage-` becoming `-Deploy-`) silently breaks source resolution for every pipeline using `_postStageHook`, including the production two-phase deploy path. The failure is invisible: `resolveSource` hits the fallback at line 230, the hook receives `undefined` typed as `IFileSetProducer`, and synthesis produces broken CloudFormation with no error.

At minimum, add a multi-branch test confirming each stage receives its own branch's source rather than just a truthy value. Longer term, ask upstream to expose the per-branch source file set through a public API so this fragile ID-pattern matching can be removed.

packages/hosting/src/pipeline/pipeline_construct.ts
```diff
@@ -72,492 +72,187 @@ const DEFAULT_SYNTH_COMMANDS = [
...
-      stage,
-      additionalPostSteps,
-    );
   }
-
-  addTriggerFilters(codePipeline, branchConfig, props);
-
-  return codePipeline;
+  // Should not happen in practice; the hook returns [] when source is falsy.
+  return undefined as unknown as IFileSetProducer;
```
soberm: [major] The declared return type is `IFileSetProducer` but this path returns `undefined as unknown as IFileSetProducer`, a type lie. The comment says "the hook tolerates a missing source by producing no steps," but that is not enforced anywhere: nothing prevents a `_postStageHook` implementation from using `source` as a `ShellStep` `input` (the existing test at line 101 does exactly that). If `resolveSource` reaches this branch, the hook receives `undefined`, passes it to `ShellStep({ input: undefined })`, and synthesis produces invalid CloudFormation with no compile-time or early-runtime error.

Suggested fix: change the return type to `IFileSetProducer | undefined`. In `runHook`, guard the hook call with a null-check and emit a `CDK Annotations.addWarning` on the stage when source cannot be resolved, so the failure is visible at synth time rather than silently producing a broken deploy.

packages/hosting/src/pipeline/pipeline_construct.ts
```diff
@@ -72,492 +72,187 @@ const DEFAULT_SYNTH_COMMANDS = [
...
-  codePipeline: CodePipeline,
-  branchConfig: BranchConfig<TConfig>,
-  props: PipelineProps<TConfig>,
+/**
+ * Append the captured `_postStageHook` steps to the built stage deployments.
+ *
+ * aws-blocks builds each branch's CodePipeline (adding stages via `addStage`,
+ * which internally creates a Wave named after the stage) inside its
+ * constructor. The pipeline remains mutable until `buildPipeline()`, so we can
+ * still attach post steps here via `StageDeployment.addPost()`.
```
soberm, line 254: [minor] `applyPostStageHook` calls `stageDeployment.addPost()` after `super()` returns, relying on the assumption that `@aws-blocks/pipeline` never calls `buildPipeline()` inside its constructor. `addPost()` is silently ignored (or may throw) after `buildPipeline()` has been called. The comment documents the assumption, but there is no guard. An upstream change that eagerly builds the pipeline would silently drop all `_postStageHook` steps from the production hosting deploy without any error at synth or deploy time.

Consider adding an assertion in the construct test that the injected step actually appears in the synthesized CloudFormation, so the test would fail loudly if this ordering assumption is ever violated.

packages/hosting/src/pipeline/pipeline_construct.test.ts
```diff
@@ -1,1218 +1,132 @@
...
-          a.ActionTypeId?.Category === 'Deploy' &&
-          a.ActionTypeId?.Provider === 'CloudFormation',
-      );
-      assert.ok(
-        deployActions.length >= 2,
-        `Expected at least 2 deploy actions for 2 stacks, got ${deployActions.length}`,
-      );
-    });
+/**
+ * Shim + glue coverage for `AmplifyPipelineConstruct`.
```
soberm, line 16: [major] This suite shrinks from 1,218 lines to 132 (roughly 90% reduction). The rationale, that generic behavior is owned upstream, is sound in principle, but leaves a concrete gap. All validation tests that were here (invalid ARN, empty branches, duplicate stage and branch names, bakeTime limit, cross-account key guard, empty-stack guard, branch-name collision after sanitization, ComputeType.SMALL annotation) are now trusted entirely to `@aws-blocks/pipeline`. If upstream tests regress or behavior changes, nothing in this repo catches it during dependency upgrades.

Two additions would close the most important gaps:
1. Contract tests asserting upstream still throws (or does not throw) on the public-API validation cases. These can be lightweight and are worth the maintenance cost.
2. Additional `_postStageHook` tests for: (a) multi-branch behavior, confirming each stage's hook receives its own branch's source rather than the other branch's; (b) the async `create()` path with a real async `stageFactory`; (c) behavior when `resolveSource` cannot find the matching pipeline.

packages/backend/src/engine/nodejs_function_project_root.ts
```diff
@@ -0,0 +1,68 @@
...
+ * keeps `projectRoot`, `depsLockFilePath`, and `entry` mutually contained in
+ * every layout — the monorepo root in local/CI dev, and the consumer's project
+ * root once the package is installed under `node_modules`. Both values must be
+ * passed to `NodejsFunction` so it does not fall back to cwd-based discovery for
+ * the lock file.
+ * @param handlerDir - Absolute directory containing the bundled handler entry.
+ * @param packageRootFallback - Absolute package root to use when no lock file is
+ *   found above the handler (still contains the entry, so containment passes).
+ */
+export const resolveNodejsFunctionBundlingRoot = (
```
soberm, line 45: [minor] `resolveNodejsFunctionBundlingRoot` has several distinct code paths (lock file found immediately, found at a higher directory level, no lock file anywhere, multiple lock file formats present at the same level) but no unit tests. Because the result is captured as a module-level constant in both callers, a bug here causes a `PathNotUnderRoot` error at synth time with no clear link back to this utility. The function is straightforward to test with a temporary-directory fixture.

packages/backend/src/engine/backend-secret/backend_secret_fetcher_provider_factory.ts
```diff
@@ -17,6 +18,14 @@ const backendSecretLambdaFilePath = path.join(
   'backend_secret_fetcher.js',
 );
 
+// Anchor bundling to the handler's own project so the entry AND lock file stay
+// contained even when synth runs from a consumer/test project with a different
+// cwd (avoids CDK's `PathNotUnderRoot`). See resolveNodejsFunctionBundlingRoot.
+const secretLambdaBundlingRoot = resolveNodejsFunctionBundlingRoot(
+  resourcesRoot,
+  path.join(dirname, '..', '..', '..'),
```
soberm: [nit] `path.join(dirname, '..', '..', '...')` resolves to the `@aws-amplify/backend` package root (three levels up from `engine/backend-secret/`), but a reader has to count separators and know the directory layout to confirm this. A brief inline comment or a named constant like `packageRoot` would make the intent self-documenting. The same three-level pattern appears in `branch_linker_construct.ts` as well.

packages/hosting/src/pipeline/types.ts
```diff
@@ -319,70 +121,42 @@ export type DefinePipelineProps<TConfig = Record<string, unknown>> = {
...
+    /**
+     * Internal escape hatch: override the GitHub source with a custom source.
+     *
+     * Used in tests to provide an S3 source instead of requiring a real GitHub
+     * connection. `@aws-blocks/pipeline` consumes this at runtime but (as of
+     * 0.1.1) omits it from its public `PipelineProps` type, so the shim
+     * re-declares it here to keep it type-visible. Not part of the public API.
+     * @internal
+     */
+    readonly _sourceOverride?: CodePipelineSource;
```
soberm, line 146: [minor] `_sourceOverride` is re-declared here because, per the comment, `@aws-blocks/pipeline` accepts it at runtime but omits it from its public `PipelineProps` type. This creates a latent coupling: if upstream renames or removes the internal prop, TypeScript does not catch it at the call site, and the integration silently stops working (the source override is ignored and any test or e2e that depends on it falls back to a live CodeConnections ARN, causing a non-obvious CI failure). Track this in the upstream backlog to get `_sourceOverride` officially declared so the re-declaration here can be removed.

packages/platform-core/package.json
```diff
@@ -46,7 +46,7 @@
     "zod": "3.25.17"
   },
   "peerDependencies": {
-    "aws-cdk-lib": "^2.234.1",
+    "aws-cdk-lib": "^2.257.0",
     "constructs": "^10.0.0"
```
soberm: [minor] The PR description says "`constructs -> ^10.6.0` across all 16 workspace packages," but only `@aws-amplify/hosting` received the `constructs` peer bump. All other packages that declare `constructs` as a peer dependency (`ai-constructs`, `auth-construct`, `backend`, `backend-ai`, `backend-auth`, `backend-data`, `backend-function`, `backend-storage`, `platform-core`, `plugin-types`) still declare `^10.0.0`. Since `aws-cdk-lib@^2.257.0` itself requires `constructs@^10.3.0`, the `^10.0.0` declaration is misleading: it implies compatibility with `constructs@10.0.x` that does not actually exist once `aws-cdk-lib@2.257.0` is in the dependency graph. The peer ranges should be consistent across all packages that received the CDK bump.

scripts/check_pr_size.ts
```diff
@@ -3,10 +3,17 @@ import { GitClient } from './components/git_client.js';
...
+const EXCLUDE = [
+  'package-lock.json',
+  'API.md',
+  'expected-cdk-out',
+  '.changeset/',
+];
 
 const MAX_LINES_ADDED = 3500;
-const MAX_LINES_REMOVED = 1000;
+const MAX_LINES_REMOVED = 2500;
```
soberm, line 20: [nit] `MAX_LINES_REMOVED` is permanently bumped from 1000 to 2500. The bump was genuinely needed: even after the new `.changeset/` exclusion, the real code deletions in this PR (`pipeline_construct.ts`, `types.ts`, `pipeline_construct.test.ts`) total roughly 2,000 lines. That said, 2500 is now the permanent guard for all future PRs. A short comment here explaining the rationale would help a future reviewer understand whether the limit is intentionally loose or was widened to accommodate a specific refactoring PR.

packages/hosting/src/pipeline/pipeline_construct.ts
osama-rizk: Good catch — fixed in 179208e.

- **Fail loud**: `resolveSource()` now throws a descriptive error instead of the `undefined as unknown as IFileSetProducer` cast. (The downstream hook feeds `source` straight into `CodeBuildStep.input`, so a silent miss wouldn't even produce `[]` — it would drop or corrupt the hosting deploy step. So throwing is the right call.)
- **Documented the second assumption**: added a comment noting the id-prefix relies on `child.node.id === ${id}-${safeBranch}`.
- **CI guard**: added a multi-branch unit test (`resolves each stage source from its own branch pipeline`) that asserts a real source file set is resolved for every stage across branches. A future `@aws-blocks/pipeline` construct rename now trips this test (and the throw) instead of silently dropping the step.

On the "resolve via construct type / metadata rather than id-string prefix" suggestion — I kept the id-prefix match: aws-blocks exposes no public stage→owning-pipeline API, so metadata resolution would just trade one undocumented coupling for another. The fail-loud + test combination gives us the CI signal without adding a second coupling. If upstream later exposes a first-class accessor, we can switch to it.

packages/hosting/package.json
osama-rizk: Thanks — agreed, and nice detail on the CVE-2026-44577 backport being the actual source of the 5-arg signature (that's a more precise root cause than the "Next 16 added the arg" framing in my earlier commit message).

We're aware of this and will fix it upstream in `@aws-blocks/hosting` — gating `patchImageOptimizerForNext155` on the resolved Next version (only inject for the ≥ 15.5.16 / ≥ 16.2.5 5-arg signature) rather than applying it structurally/unconditionally. That's the correct home for the fix since the patch code and the adapter both live in blocks; doing it there fixes it for every consumer, not just this shim.

This PR only re-exports the blocks adapters, so no code change here. Pinning the e2e apps to `next@^16.2.9` keeps CI on the correct branch until the upstream gate ships; once it does we'll bump the dep.

osama-rizk:
Thanks for the triage. I dug into the three failing runs, and the actual failure modes are transient infra issues rather than snapshot drift — so the "regenerate baselines" remediation doesn't quite apply. Details:

**1. `check_api_changes`** — not an `API.md` mismatch. The job failed on a flaky `ENOTEMPTY: directory not empty, rmdir '.../api-validation-projects/.../node_modules/...graphql-api-construct/...'` during the install/cleanup step (retried 3×, same infra error each attempt; no API-report assertion ran). I still validated the concern directly: regenerated `packages/hosting/API.md` on a clean full install via `npm run update:api` → **zero diff**. api-extractor emits `AmplifyPipelineConstruct` (not `Pipeline`) for `create()`, so the committed report is correct and not mis-hand-edited. → **re-run**.

**2. `e2e_iam_access_drift`** — not an IAM assertion failure. All 5 attempts hit the **25-minute deploy timeout** (`Script timed out after 25 minutes`) on the Node 18 runner. Also worth noting this test is a *live differential* test: it deploys the PR version and the last-published npm version and diffs them at runtime via IAM Access Analyzer `CheckNoNewAccess` — there's **no committed IAM baseline snapshot** to regenerate. → **re-run** (and unrelated to the CDK bump).

**3. `e2e_amplify_outputs_backwards_compatibility`** — deploy timeouts (attempts 1–2) then `[CredentialsError] Unable to get backend outputs due to invalid credentials` from `ampx generate outputs` (attempts 3–5, expired e2e creds mid-run). Same live-differential design as #2, no committed snapshot. → **re-run**.

Net: all three are re-run candidates (slow-deploy timeouts / expired creds / flaky npm cleanup), not baseline regen. I've kicked off re-runs. Separately, I pushed 179208e addressing the `resolveSource` review comment.

packages/hosting/src/pipeline/pipeline_construct.ts
osama-rizk: Agreed on all three costs, and you've pinpointed the root cause exactly: `Pipeline.create()` does `new Pipeline(...)` internally and `_ASYNC_INIT` is `private static`, so there's no subclass-factory hook the shim can use — I confirmed this against the published 0.1.4 dist. So there's no clean in-shim fix today; `setPrototypeOf` is the least-bad option to preserve the historical `AmplifyPipelineConstruct` identity on the async path.

I'll file an upstream request for `@aws-blocks/pipeline` to support a subclass factory (e.g. `Pipeline.create()` accepting a constructor, or a protected async-init hook), and remove the prototype swap once it lands. Leaving as-is here with that tracked.

packages/hosting/src/pipeline/pipeline_construct.ts
osama-rizk: Right — this is the shim working around an upstream implementation detail (`stageFactory.constructor.name === 'AsyncFunction'`), which a bundler rename or an upstream refactor could break. I'll file the upstream ask for an explicit `isAsync?: boolean` on `PipelineProps` and drop the name-based workaround once it's available. No shim change for now; the comment stays as the record of the coupling.

packages/hosting/src/pipeline/pipeline_construct.ts
osama-rizk: Fixed in 179208e (before this second round). `resolveSource` now **throws** a descriptive error instead of returning `undefined as unknown as IFileSetProducer`, so a cosmetic upstream rename (e.g. `-Stage-` → `-Deploy-`) trips loudly at synth time rather than producing broken CloudFormation. I also added the multi-branch test you asked for (`resolves each stage source from its own branch pipeline`), which asserts each stage receives *its own* branch's source (not just a truthy value). Longer term I agree the real fix is a public upstream API for the per-branch source file set — I'll file that so this ID-pattern matching can be removed.

packages/hosting/src/pipeline/pipeline_construct.ts
osama-rizk: Fixed in 179208e — the `undefined as unknown as IFileSetProducer` cast is gone; `resolveSource` now throws when it can't resolve a source. On the specific suggestion (return `IFileSetProducer | undefined` + `Annotations.addWarning`): I went with throwing instead. A synth-time warning can be lost in output noise and would still let a broken pipeline synthesize; since the downstream hook feeds `source` straight into `ShellStep`/`CodeBuildStep` `input`, hard-failing is the safer choice — no invalid CloudFormation is ever produced.

packages/hosting/src/pipeline/pipeline_construct.ts
osama-rizk: Good call — addressed in aa39f6a. `applyPostStageHook` now tracks which `postSteps` entries actually attached to a built `StageDeployment` and **throws** if any didn't, so an upstream change that eagerly calls `buildPipeline()` (making `addPost()` a silent no-op) or renames stages fails loud instead of dropping the hosting deploy. The existing construct tests already assert the injected step appears in the synthesized CloudFormation (both the single-stage and the new multi-branch case check the CodeBuild BuildSpec), so that guard is covered in CI too.

packages/hosting/src/pipeline/pipeline_construct.test.ts
osama-rizk: Addressed the concrete gaps; pushed back on re-forking the rest.

Your (a)/(b)/(c) are now covered:
- **(a) multi-branch source isolation** — `resolves each stage source from its own branch pipeline` asserts each stage gets *its own* branch's source.
- **(b) async `create()` with a real async `stageFactory`** — `supports the async create() path` already exercises this.
- **(c) `resolveSource` can't find the pipeline** — now throws (179208e), and `applyPostStageHook` also throws on an unattached step (aa39f6a).

On the contract tests for the upstream validation cases (invalid ARN, empty branches, duplicate names, bakeTime, cross-account, ComputeType): I added **one** canary — `delegates upstream prop validation (invalid connection ARN throws)` — as an upgrade tripwire. I deliberately didn't re-add all nine: they're owned and tested by `@aws-blocks/pipeline`, and duplicating them here re-forks exactly the surface this PR exists to delete. The single canary confirms delegation is live so a dependency upgrade that regresses the public validation contract still trips CI here.

packages/backend/src/engine/nodejs_function_project_root.ts
osama-rizk: Done in aa39f6a — added `nodejs_function_project_root.test.ts` with a temp-dir fixture covering all four paths: lock file alongside the handler, lock file at an ancestor, nearest-ancestor-wins when locks exist at two levels, a non-npm format (`pnpm-lock.yaml`), and the no-lock-anywhere fallback to the package root with `depsLockFilePath: undefined`. 5/5 passing.

packages/backend/src/engine/backend-secret/backend_secret_fetcher_provider_factory.ts
osama-rizk: Done in aa39f6a — extracted the three-levels-up path into a named `packageRoot` constant with a comment spelling out the directory walk, in both `backend_secret_fetcher_provider_factory.ts` and `branch_linker_construct.ts`.

packages/hosting/src/pipeline/types.ts
osama-rizk: Agreed — this is the same class of coupling as the async-detection one: `@aws-blocks/pipeline` consumes `_sourceOverride` at runtime but omits it from its public `PipelineProps` type, so we re-declare it and TS can't catch an upstream rename at the call site. I'll track an upstream ask to declare `_sourceOverride` officially (or expose a supported source-override option) so the re-declaration here can be removed. Keeping it for now since the e2e source override depends on it; the `@internal` comment documents the coupling.

packages/platform-core/package.json
osama-rizk: Good catch — you're right, the description overstated it: only `@aws-amplify/hosting` had actually gotten the `constructs` bump. Fixed in aa39f6a: bumped the `constructs` peer floor to `^10.6.0` across all 10 remaining packages that declare it (`ai-constructs`, `auth-construct`, `backend`, `backend-ai`, `backend-auth`, `backend-data`, `backend-function`, `backend-storage`, `platform-core`, `plugin-types`), so the peer range is consistent and no longer implies a `10.0.x` compatibility that can't exist once `aws-cdk-lib@2.257.0` (which requires `constructs@^10.5.0`) is in the graph. Changeset note updated too.

scripts/check_pr_size.ts
osama-rizk: Done in aa39f6a — added a comment at `MAX_LINES_REMOVED` explaining the 1000→2500 bump (accommodates de-fork/deletion PRs like this one, which removes the ~2,000-line forked pipeline implementation) and noting it's intentional headroom that can be tightened if it proves loose.

packages/platform-core/package.json
osama-rizk: Follow-up: I had to walk this back. Bumping `constructs` to `^10.6.0` **workspace-wide** caused a real e2e regression — `e2e_iam_access_drift` and `e2e_amplify_outputs_backwards_compatibility` started timing out. Root cause: those tests do a clean `npm install` of a `create-amplify` app from the local proxy, and a `^10.6.0` floor that's *stricter than create-amplify's scaffolded `constructs@^10.0.0`* sent npm's resolver into ~21 min of backtracking (baseline install 5m → 21m), blowing the 25-min per-attempt guard → retries → CI creds expired mid-run.

So in 6ae97f6 I scoped it: only `@aws-amplify/hosting` (the one package that consumes the `@aws-blocks` constructs directly) keeps `^10.6.0`; the other 10 go back to `^10.0.0`.

On the "misleading range" point — I'd argue `^10.0.0` is actually fine: `aws-cdk-lib@2.257.0` declares `constructs@^10.5.0` as its own peer, so any install that pulls in 2.257 transitively resolves a `>=10.5` `constructs` regardless of what these packages declare. The peer floor here doesn't gate the effective version; it just avoids over-constraining the resolver. Net: `aws-cdk-lib` stays `^2.257.0` workspace-wide (required — `test_with_baseline_dependencies` pins only `aws-cdk-lib` to minVersion and would `ERESOLVE` on a split floor), and `constructs` is raised only where it's genuinely needed.

soberm (approved):

packages/hosting/package.json
```diff
@@ -43,12 +43,13 @@
...
+    "@aws-blocks/pipeline": "^0.1.1"
   },
   "peerDependencies": {
     "@opennextjs/aws": ">=3.10.0",
-    "aws-cdk-lib": "^2.234.1",
-    "constructs": "^10.0.0"
+    "aws-cdk-lib": "^2.257.0",
+    "constructs": "^10.6.0"
```
bobbor, line 52: **Lockfile not regenerated after the `constructs ^10.6.0` restore — likely the real cause of `check_api_changes`, not an API break**

The final commit (`290caef`, “restore constructs ^10.6.0 peer floor across all packages”) bumps the `constructs` **peer** floor to `^10.6.0` across the workspace `package.json` manifests, but `package-lock.json` still records `constructs: ^10.0.0` for those same workspace nodes — it wasn’t regenerated after the restore. `npm ci` (and the clean install that `check_api_changes` runs before api-extractor) requires manifest/lock parity, so this desync is a strong candidate for the red `check_api_changes`, compounded by the `ENOTEMPTY` teardown flake the retry addresses.

I independently diffed the committed `API.md` at head (`290caef`) vs the base tip (`a281f3a`): only `packages/hosting/API.md` changed (+12/−42), and every delta is fully explained by the intended shim refactor — `AmplifyPipelineConstruct` now `extends Pipeline<TConfig>`, the forked config types (`BranchConfig`, `PipelineSourceConfig`/`StageConfig`/`SynthConfig`) are re-exported from `@aws-blocks/pipeline`, and `PipelineProps` collapses into `PipelineProps_2<TConfig> & { hosting-specific fields }`. The one removed member, `codePipelines`, is inherited from the base `Pipeline` (declared `public readonly codePipelines: ReadonlyMap<string, CodePipeline>` upstream), so its absence from the subclass report is an api-extractor inheritance artifact — not a breaking removal.

So the API surface looks correct and non-breaking; this reads as a tooling/install issue rather than real API drift. Regenerating `package-lock.json` so it matches the `^10.6.0` peer floors should restore clean-install parity and, with it, `check_api_changes`.

packages/hosting/src/pipeline/pipeline_construct.test.ts
```diff
@@ -1,1218 +1,207 @@
...
-          defaultPipelineProps({
-            source: {
-              repo: 'my-org/my-app',
-              connectionArn: VALID_CONNECTION_ARN,
-              triggerFilters: ['packages/**'],
-            },
-          }),
-        ),
-    );
+  void it('supports the async create() path', async () => {
```
bobbor, line 95: **Test gap: the async `create()` path never exercises `_postStageHook`**

This async `create()` test asserts `instanceof AmplifyPipelineConstruct` — good, it does exercise the `Object.setPrototypeOf` re-tag — but it supplies no `_postStageHook`. So the most fragile new interaction is uncovered: the async factory building its own `postSteps` map, driving the wrapped `stageFactory`, and draining it via `applyPostStageHook()` *after* the prototype swap. The sync constructor path has multi-branch `_postStageHook` coverage, but the async path (the one that depends on the prototype hack) doesn’t. Could you add an async `create()` case with a `_postStageHook` and assert the post-steps land on the correct StageDeployment(s)?

packages/hosting/src/pipeline/pipeline_construct.ts
```diff
@@ -72,492 +72,221 @@ const DEFAULT_SYNTH_COMMANDS = [
...
+ * `${id}-${safeBranch}-Stage-${name}`, so a stage's construct id is exactly its
+ * owning pipeline's id followed by `-Stage-${name}`. We match on that prefix to
+ * pick the correct branch pipeline, then read its synth ShellStep's primary
+ * input (the source file set).
+ *
+ * If no owning pipeline can be resolved we throw rather than return an empty
+ * source: the downstream hook feeds this straight into `CodeBuildStep.input`,
+ * so a silent miss would either drop the hosting deploy step or synthesize a
+ * broken one with NO CI signal. Throwing means a future upstream construct
+ * rename trips loudly here (and in the `resolveSource` unit test) instead.
```
bobbor, line 216: **Test gap: `resolveSource()` throw path is untested, and this comment references a test that doesn’t exist**

The new fail-loud behavior (throwing instead of the old `undefined as unknown as IFileSetProducer` cast) is a solid improvement, but only the happy path is asserted (the multi-branch test checks resolved sources). This comment references a `resolveSource` unit test for the rename/no-match case, but no such negative test exists. Could you either add a test asserting `resolveSource()` (and the `applyPostStageHook` unattached-steps guard) throws the descriptive error when the upstream stage-naming assumption breaks, or drop the comment? A negative test here is exactly what would trip CI if a future `@aws-blocks/pipeline` release renames its stage constructs — otherwise that regression stays silent.

bobbor:
**Status on the currently red / in-progress checks**

Triage of the non-green checks for reviewers:

- **`check_api_changes` (red)** — looks like a tooling/install issue, not an API break. `package-lock.json` wasn’t regenerated after the `constructs ^10.6.0` restore in `290caef`, so it’s out of sync with the manifests and the clean-install step fails (compounded by the `ENOTEMPTY` teardown flake the retry addresses). An independent base-vs-head diff of the committed `API.md` shows only the intended, non-breaking shim changes (details in the inline comment on `packages/hosting/package.json`). Regenerating the lockfile should clear this.
- **`e2e_notices (macos-14)` (red)** — nothing in the 95-file diff touches notices, and the same test passes on ubuntu, so this is the known macos-14 flakiness rather than a regression. A re-run (or a few) should settle it.
- **`e2e_iam_access_drift` / `e2e_amplify_outputs_backwards_compatibility` (re-running)** — no baseline snapshot files changed in the diff, and the CI commits target exactly the `ExpiredToken`/timeout failures (per-attempt role re-assume + 25→40m timeouts), so “infra flakiness, not snapshot drift” is credible. Pending green.

Net: no evidence of a real functional or API regression from the diff — the remaining red is lockfile parity + known macos flakiness. Regenerating `package-lock.json` is the one concrete action that should move the needle.

packages/hosting/package.json
osama-rizk: Confirmed and fixed in 298898d — you're exactly right. `290caef` bumped the 10 workspace `package.json` peer floors to `^10.6.0` but didn't regenerate the lock; all 10 lock nodes still read `^10.0.0`. Regenerated `package-lock.json` (diff is precisely 10× `^10.0.0`→`^10.6.0`, no version/integrity churn), restoring manifest/lock parity for the clean install. And thanks for independently confirming the `API.md` diff is the intended, non-breaking shim change — matches what I saw regenerating it locally (zero diff).

packages/hosting/src/pipeline/pipeline_construct.test.ts
osama-rizk: Added in 298898d — new test `injects _postStageHook steps via the async create() path`: async `stageFactory` + `_postStageHook`, asserting the injected step lands as a CodeBuild action after the `setPrototypeOf` re-tag + `applyPostStageHook()` drain. Covers the async postSteps path the sync tests didn't.

packages/hosting/src/pipeline/pipeline_construct.ts
osama-rizk: Fixed in 298898d. Exported `resolveSource` (`@internal`) and added a negative test `throws a descriptive error when no owning branch pipeline matches` — drives a Stage whose scope has no matching CodePipeline (the post-rename world) and asserts the fail-loud throw. So the comment now references a real test, and an upstream stage-construct rename trips CI here instead of silently dropping the deploy step. (`@internal` keeps it out of the public API report.)

osama-rizk:
Thanks for the triage — agreed on all three, and acted on the concrete one:

- **`check_api_changes`**: fixed the root cause you identified — `package-lock.json` was out of sync with the `constructs ^10.6.0` manifests. Regenerated it in `298898d` (parity restored; `API.md` confirmed non-breaking). The ENOTEMPTY retry (`98ece5c1`) covers the teardown flake.
- **`e2e_notices (macos-14)`**: agreed, pre-existing macOS-only flake (passes on ubuntu/windows; diff touches nothing notices-related). Re-run candidate.
- **`e2e_iam_access_drift` / `amplify_outputs_backwards_compatibility`**: per-attempt role re-assume (`47135b0`) is confirmed working in the latest run — logs show `Refreshed … credentials for attempt N` and `✔ should not drift iam policies` (test passes). Remaining issue is purely the job hitting the 120-min cap because the create-amplify install takes ~26–30 min in CI (measured), so a passing attempt lands at ~1h52m. No functional/API/snapshot regression — it's install-time vs. the job budget.

---

**Update: these two tests are pre-existing flakes — they fail on `main` too, not just this PR.**

Latest `main` run ([28665436967](https://github.com/aws-amplify/amplify-backend/actions/runs/28665436967), `main` @ `4849fadca`, 2026-07-03) shows **both** jobs FAILED with none of this PR's changes present:
- `run_health_checks / e2e_iam_access_drift` → **failure**
- `run_health_checks / e2e_amplify_outputs_backwards_compatibility` → **failure**

So this is a pre-existing flakiness in these two baseline-diff e2e jobs (heavy `create-amplify` install vs. the per-attempt/120-min budget + 1h role session), independent of the pipeline-shim work. This PR's CI commits (per-attempt role re-assume, 25→40m attempt timeout) actually *improve* their odds versus `main` — the latest run here shows the test assertion passing (`✔ should not drift iam policies`) before the job-level cap. Not a regression introduced by this PR.

bobbor (approved):
## Approving ✅

All review feedback from the earlier rounds has been addressed and every non-e2e check is green (including `check_api_changes`, which now completes in ~35 min after the npm-install hardening — skip audit/funding + fetch-retry for transient `ECONNRESET` — that resolved the earlier 60-min timeout).

- **Lockfile parity** — `package-lock.json` was regenerated (`298898d`) so the `constructs ^10.6.0` peer floors match the manifests, clearing the clean-install path. An independent base-vs-head diff confirmed the only public-API change is the intended shim (`AmplifyPipelineConstruct extends Pipeline`, config types re-exported from `@aws-blocks/pipeline`, `codePipelines` inherited from the public base class) — no unintended API drift.
- **`resolveSource()` fail-loud** — now throws a descriptive error instead of the `undefined as unknown as IFileSetProducer` cast, backed by a real negative test, plus the `applyPostStageHook` unattached-steps guard.
- **Async `create()` coverage** — new test exercises the `create()` + `_postStageHook` path (prototype re-tag via `Object.setPrototypeOf` + post-steps landing on the correct StageDeployment).

The design — a thin shim over `@aws-blocks/pipeline` replacing ~1k lines of forked pipeline logic — is sound, and the cdk/constructs floor bump is coherent.

The only remaining non-green checks are `e2e_*` (the two baseline-diff jobs `e2e_iam_access_drift` / `e2e_amplify_outputs_backwards_compatibility`, plus a cancelled Windows `e2e_create_amplify` leg) — known/pre-existing infra flakes that also fail on `main`, not regressions from this PR. `e2e_notices` (incl. macos-14) is now passing.

One tracked, non-blocking follow-up: the `@aws-blocks/hosting` `patchImageOptimizerForNext155` patch should be version-gated upstream (Next ≥15.5.16 / ≥16.2.5) so customers on older Next patch versions aren't affected by the `fetchInternalImage` signature shift — to be handled in `@aws-blocks/hosting`, not here.

Nice cleanup. 🎉

PR #3265, opened 2026-07-02
feat(backend-notifications): add defineNotifications factory for Customer Profiles

soberm:
## Summary
Adds `@aws-amplify/backend-notifications`, a new `defineNotifications` backend construct that provisions push notifications backed by **Amazon Connect Customer Profiles**. It gives an Amplify app a SigV4-authenticated API to identify users and register/remove device tokens, an authoritative DynamoDB device store with strongly-consistent single-owner semantics, and a Connect-journey push-delivery path that cannot leak notifications across users.

## What it provisions
- **Customer Profiles**: a single object type, `AmplifyProfile`, keyed on `principalId` — the Cognito Identity Pool `identityId`, always derived server-side from the SigV4-verified caller (never read from the request body).
- **DynamoDB device store**: the authoritative device→owner record (PK `deviceId`, GSI on `principalId`, native TTL), with strongly-consistent single-owner semantics.
- **HTTP API (API Gateway, payload format 1.0)** with three IAM/SigV4 routes served by one write Lambda:
  - `POST /identify-user` — upserts the caller's profile.
  - `POST /register-device` — registers or re-homes a device token to the caller.
  - `POST /remove-device` — removes a device the caller owns (idempotent no-op otherwise).
  All three are callable by both authenticated and guest Cognito Identity Pool identities via SigV4/IAM.
- **Push-delivery Lambda** (Connect journey custom action) that gates delivery on the DynamoDB ownership record.

## Modes
- **Attach** (`domainName`): reuse an existing Customer Profiles domain.
- **Create-from-scratch**: provision a new Amazon Connect instance and Customer Profiles domain.

## Client configuration
Surfaces `notifications.amazon_connect { endpoint, aws_region }` in the generated client config (client-config schema v1.5). The amplify-js Customer Profiles push provider consumes this key.

## Security properties
- `principalId` is server-derived from the SigV4 context; reserved profile attribute keys are rejected and stripped, so a caller cannot bind another user's identity.
- Delivery ownership gate (GSI enumerate → strongly-consistent point-read → ownership check) ensures a re-homed device never receives a previous owner's pushes.
- PII-safe logging: raw SDK error messages are never logged.
- Least-privilege IAM; API Gateway throttling on the guest-writable stage; 90-day CloudWatch log retention on the Lambdas; APNS/FCM credentials rendered as CFN dynamic references (never plaintext in the template).

## Testing
- Unit tests across the construct, factory, API handlers, device store, and the push-delivery ownership gate.
- A create-mode integration test that deploys real infrastructure and exercises the three SigV4 routes (authenticated and guest), profile field mapping, device registration / re-home / removal, server-side validation, and the `principalId`-injection regression.

## Deferred (tracked for post-preview)
- Customer-managed KMS keys on the Customer Profiles domain and the DynamoDB table.
- A request body-size cap.
- ARN partition parameterization (Amazon Connect / Customer Profiles / Outbound Campaigns are commercial-partition-only today).

## Checklist


packages/backend-notifications/src/factory.ts
```diff
@@ -0,0 +1,197 @@
...
+    notifications: AmplifyNotifications,
+  ): void => {
+    outputStorageStrategy.addBackendOutputEntry(customOutputKey, {
+      version: '1',
+      payload: {
+        // Serialized `Partial<ClientConfig>` — surfaced under `custom` in
+        // `amplify_outputs.json`, matching `backend.addOutput({ custom })`.
+        customOutputs: JSON.stringify({
+          custom: {
+            [OUTPUT_KEY]: {
```
ekjotmultani, line 196: Should this be under `analytics.amazon_connect_customer_profiles` (with `aws_region`) rather than `custom.CustomerProfiles`? That's the key the JS provider in #14864 reads, so the construct and the clients would line up on the same shape.

packages/backend-notifications/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@aws-amplify/backend-output-schemas": "^1.8.0",
+    "@aws-amplify/platform-core": "^1.11.1",
+    "@aws-amplify/plugin-types": "^1.12.1"
+  },
+  "devDependencies": {
+    "@aws-amplify/backend-output-storage": "^1.3.5",
+    "@aws-amplify/backend-platform-test-stubs": "^0.3.10",
+    "@aws-sdk/client-connect": "^3.1079.0",
```
ekjotmultani, line 37: These `@aws-sdk/*` deps are pinned to `^3.1079.0`, which doesn't resolve on npm. Is that version right? A clean install fails on it.

packages/backend-notifications/package.json
```diff
@@ -0,0 +1,49 @@
...
+  },
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@aws-amplify/backend-output-schemas": "^1.8.0",
+    "@aws-amplify/platform-core": "^1.11.1",
+    "@aws-amplify/plugin-types": "^1.12.1"
+  },
+  "devDependencies": {
+    "@aws-amplify/backend-output-storage": "^1.3.5",
+    "@aws-amplify/backend-platform-test-stubs": "^0.3.10",
```
ekjotmultani, line 36: `@aws-amplify/backend-platform-test-stubs@^0.3.10` doesn't look published on npm, so installing this package outside the monorepo fails. Is it meant to be internal only?

packages/backend-notifications/src/factory.ts
soberm: I am not sure yet were we should put the config for Customer Profiles since its not an Analytics solution. The most sense probably makes `notifications.amazon_connect_customer_profiles`.

packages/backend-notifications/package.json
soberm: These exist on npm, e.g., [@aws-sdk/client-connect@3.1079](https://www.npmjs.com/package/@aws-sdk/client-connect/v/3.1079.0), [@aws-sdk/client-connectcampaignsv2@3.1079](https://www.npmjs.com/package/@aws-sdk/client-connectcampaignsv2/v/3.1079.0), [@aws-sdk/client-customer-profiles@3.936](https://www.npmjs.com/package/@aws-sdk/client-customer-profiles/v/3.936.0).

packages/backend-notifications/package.json
soberm: This is a private package for testing which is supposed to be only available within the workspace.

.changeset/backend-notifications.md
```diff
@@ -0,0 +1,12 @@
...
+
+- Two modes: zero-config create-from-scratch (provisions a new Connect instance + Customer Profiles domain, wires Outbound Campaigns v2, the CTR store, and a message-templates knowledge base so Journeys work with no console setup) and attach-to-existing (`domainName` provided, additive registration only).
+- Authenticated and guest (unauthenticated, IAM/SigV4 `POST /identify-user-guest`) support. Profiles are kept separate (no `MergeProfiles`). Device ownership is authoritative in a dedicated DynamoDB Devices table (PK `deviceId`, GSI on `profileId`, native TTL): registering/re-homing a device is a strongly-consistent last-writer-wins `UpdateItem` on the `deviceId` (overwriting the record IS the eviction), and push delivery enumerates a profile's devices via the GSI then gates each send on a strongly-consistent point read of the `deviceId` — so a physical device routes to exactly one profile at any instant with no eventual-consistency window (this eliminates the immediate-switch cross-user push leak). Guest profiles are reaped by a shorter TTL. The request option is `guestIdentityId`.
```
sarayev: **[MAJOR]** This changeset (which becomes the published changelog) describes behavior that isn't in the code: there's no `POST /identify-user-guest` route, no `guestIdentityId` request option, the GSI is on `principalId` (not `profileId`), and there's no separate guest profile / shorter guest TTL. Could we rewrite this to match what shipped — three SigV4 routes (`/identify-user`, `/register-device`, `/remove-device`), a single `principalId`-keyed `AmplifyProfile`, and the DynamoDB device store?

packages/backend-notifications/README.md
```diff
@@ -0,0 +1,253 @@
...
+  - `AmplifyGuestProfile` — a guest profile, keyed by the unauthenticated
+    Cognito Identity Pool `identityId`.
+  - `AmplifyDevice` — a device object (keyed by a stable `deviceId`).
+- An HTTP API + `identify-user` Lambda that find-or-creates the caller's profile
+  and registers their device, exposing two routes:
+  - `POST /identify-user` — authenticated, authorized by a **Cognito user-pool
+    JWT** authorizer.
+  - `POST /identify-user-guest` — guest, authorized by **IAM/SigV4** (callable
+    with unauthenticated Cognito Identity Pool credentials).
+- A push-delivery Lambda (a Connect **Journey Custom-action** target) plus a
```
sarayev: **[MAJOR]** This "What it provisions" section looks like it's from an earlier design revision. The code has **one** object type (`AmplifyProfile`, keyed on `principalId`), not three — there's no `AmplifyGuestProfile` or `AmplifyDevice` (devices live in DynamoDB). And there are **three** routes, all IAM/SigV4 — no `/identify-user-guest` and no Cognito user-pool JWT authorizer (line 27's "JWT authorizer is bound to your app's Cognito user pool" is also inaccurate). Could we refresh the README to match the shipped design so adopters aren't misled?

packages/backend-notifications/src/construct.ts
```diff
@@ -0,0 +1,1197 @@
...
+  };
+};
+
+/**
+ * Amazon Connect Customer Profiles backend for the identifyUser contract,
+ * packaged as an AWS CDK construct. Provisions:
+ *   - AmplifyProfile object type: searchable `cognitoUserKey` (PROFILE + UNIQUE)
+ *     identity key + person / targeting attribute schema.
+ *   - DynamoDB Devices table: authoritative device store (PK `deviceId`, GSI on
+ *     `profileId`, native TTL) with strongly-consistent single-owner semantics.
```
sarayev: **[NIT]** Small doc drift: these comments still say `cognitoUserKey`, "AmplifyProfile / AmplifyGuestProfile object types", and "GSI on `profileId`", but the code uses a single object type keyed on `principalId` with the GSI on `principalId` (same in the class JSDoc at ~92 and ~449). Mind scrubbing these so the comments match the implementation?

packages/backend-notifications/src/construct.ts
```diff
@@ -0,0 +1,1197 @@
...
+        identityManagementType: 'CONNECT_MANAGED',
+        instanceAlias: this.sanitizeInstanceAlias(
+          props.instanceAlias ?? baseName,
+        ),
+      });
+
+      profilesDomain = new CfnDomain(this, 'ProfilesDomain', {
+        domainName,
+        defaultExpirationDays: expirationDays,
+      });
```
sarayev, line 403: **[MINOR]** The Customer Profiles domain is created without a `defaultEncryptionKey`, so it falls back to default encryption. For regulated customers it'd be nice to optionally thread a customer-managed KMS key through here (pairs with the DynamoDB table at ~452). Fine to defer for preview, but worth a tracking note.

packages/backend-notifications/src/construct.ts
```diff
@@ -0,0 +1,1197 @@
...
+        type: dynamodb.AttributeType.STRING,
+      },
+      billingMode: dynamodb.BillingMode.PAY_PER_REQUEST,
+      timeToLiveAttribute: 'ttl',
+      pointInTimeRecovery: true,
+      // Authoritative device store: branch/pipeline deploys RETAIN so a stack
+      // teardown never silently drops live device registrations; sandbox deploys
+      // DESTROY so ephemeral dev / E2E data is cleaned up on delete.
+      removalPolicy: isSandbox ? RemovalPolicy.DESTROY : RemovalPolicy.RETAIN,
+    });
```
sarayev, line 501: **[MINOR]** The Devices table relies on DynamoDB's default AWS-owned encryption — no `encryption`/`encryptionKey` option is exposed. Since this stores device tokens + principal linkage, should we consider offering a CMK prop (same story for the Profiles domain at ~366)?

packages/backend-notifications/src/construct.ts
```diff
@@ -0,0 +1,1197 @@
...
+      );
+    }
+
+    // ---- Lambda handler ----------------------------------------------------
+    // Pre-bundled at build time (esbuild) into a self-contained asset, so the
+    // construct works regardless of the consuming project's root (Amplify's
+    // NodejsFunction guardrail forbids an entry outside the app root).
+    const codePath = props.lambdaCodePath ?? defaultLambdaCodePath;
+
+    const apiFn = new lambda.Function(this, 'ApiFn', {
```
sarayev, line 540: **[MINOR]** These `lambda.Function`s don't set `logRetention`/`logGroup`, so they get never-expiring CloudWatch log groups with AWS-managed encryption (same for `PushHandlerFn` ~694 and `CampaignAssociationFn` ~931). Could we attach a log group with an explicit retention (and optionally a KMS key) to avoid unbounded log accumulation?

packages/backend-notifications/src/construct.ts
```diff
@@ -0,0 +1,1197 @@
...
+    // unauthenticated Identity Pool caller signing the same endpoints.
+    const iamAuthorizer = new HttpIamAuthorizer();
+    const integration = new HttpLambdaIntegration('WriteIntegration', apiFn, {
+      payloadFormatVersion: apigwv2.PayloadFormatVersion.VERSION_1_0,
+    });
+
+    const httpApi = new apigwv2.HttpApi(this, 'IdentifyUserApi', {
+      apiName: 'notifications-write-api',
+      createDefaultStage: true,
+    });
```
sarayev, line 632: **[MINOR]** The HTTP API's default stage has no throttling configured. Since guest (unauthenticated) callers can hit `identify-user` and write PII, an unthrottled endpoint is a bit of an abuse/cost risk — could we add per-route `throttlingBurstLimit`/`throttlingRateLimit` on the default stage?

packages/backend-notifications/src/construct.ts
```diff
@@ -0,0 +1,1197 @@
...
+    // isn't known at synth time, so account scoping is the tightest guard here.
+    for (const servicePrincipal of CONNECT_INVOKE_SERVICE_PRINCIPALS) {
+      pushFn.addPermission(
+        `Invoke-${servicePrincipal.replace(/[^a-zA-Z0-9]/g, '-')}`,
+        {
+          principal: new iam.ServicePrincipal(servicePrincipal),
+          action: 'lambda:InvokeFunction',
+          sourceAccount: stack.account,
+        },
+      );
```
sarayev, line 967: **[MINOR]** Nice confused-deputy guard with `sourceAccount`. In create-from-scratch mode the instance ARN is actually known (`connectInstance.attrArn`), so we could tighten this further with `sourceArn` to pin invocation to the specific instance rather than any Connect instance in the account. The "not known at synth time" note only really holds for attach mode — worth conditionally adding `sourceArn` on the create path?

packages/backend-notifications/src/lambda/push/delivery.ts
```diff
@@ -0,0 +1,366 @@
...
+        JSON.stringify({ present: Boolean(owner) }),
+      );
+      continue;
+    }
+
+    const channelType = normalizeChannelType(owner.channelType);
+    if (!channelType) {
+      results.push({
+        deviceToken: owner.token,
+        channelType: 'GCM',
```
sarayev: **[NIT]** Tiny thing: `channelType: 'GCM'` here is a placeholder on a SKIPPED/unsupported-channel result just to satisfy the type (delivery is `false` anyway). A dedicated sentinel or making `channelType` optional on this result shape would read a little more honestly.

packages/backend-notifications/src/lambda/api/validation.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  for (const key of ['email', 'name', 'phone'] as const) {
+    const value = (userProfile as Record<string, unknown>)[key];
+    if (value !== undefined && !isBoundedString(value)) {
+      return {
+        ok: false,
+        error: `userProfile.${key} must be a string <= ${MAX_ATTRIBUTE_LENGTH} chars`,
+      };
+    }
+  }
```
sarayev, line 70: **[NIT]** `isBoundedString` accepts empty strings for `email`/`name`/`phone`, which then get silently dropped by `compact()` in `mapping.ts`. End state is correct, just mildly redundant — no change needed unless you'd prefer to reject empties explicitly here.

sarayev (commented):
Really solid implementation — the security-critical paths all check out: `principalId` is server-derived from the SigV4-verified `cognitoIdentityId` (never the body), payload format 1.0 is explicitly set, the DynamoDB ownership gate (GSI enumerate → strongly-consistent point-read → conditional delete) correctly prevents cross-user push leaks, logging is PII-safe, and IAM is genuinely least-privilege with well-documented exceptions. No blocking correctness or security issues found.

My main ask before merge is **documentation fidelity**: the README and changeset both describe a superseded design (JWT auth, a `/identify-user-guest` route, and three Customer Profiles object types) that the code no longer implements — these will actively mislead users, so they should be rewritten to match the shipped single-object-type / three-SigV4-route / DynamoDB-device-store design. Beyond that, a few preview-acceptable hardening gaps worth addressing or explicitly deferring: no API Gateway throttling on guest-writable routes, no explicit Lambda log-group retention, and no customer-managed KMS option on the table/domain. The inline notes cover the smaller items. Nice work overall!

---

- **[MINOR] PR description output-key mismatch:** The PR description says the endpoint is surfaced under `notifications.amazon_connect_customer_profiles`, but the code/schema use `notifications.amazon_connect` (`src/constants.ts:149`, `src/factory.ts:196-199`, `client-config/src/client-config-schema/schema_v1.5.json:372`). The code is internally consistent — just the description needs a tweak. (Filed as a body comment since it doesn't anchor to a diff line.)

packages/backend-notifications/src/lambda/campaign-association/integration.ts
```diff
@@ -0,0 +1,210 @@
...
+  domainName: string,
+): string => `arn:aws:profile:${region}:${account}:domains/${domainName}`;
+
+/** connect-campaigns instance ARN used as the Customer Profiles integration Uri. */
+export const campaignsInstanceArnOf = (
+  account: string,
+  region: string,
+  connectInstanceId: string,
+): string =>
+  `arn:aws:connect-campaigns:${region}:${account}:instance/${connectInstanceId}`;
```
sarayev, line 45: **[MINOR]** These ARN builders hardcode the `arn:aws:` partition, whereas the construct uses `stack.partition`. That'll break in `aws-cn`/`aws-us-gov`. Since `account`/`region` are already threaded through, could we pass the partition too and interpolate it here?

packages/backend-notifications/src/lambda/campaign-association/integration.ts
```diff
@@ -0,0 +1,210 @@
...
+    new ListRolesCommand({ PathPrefix: CONNECT_CAMPAIGNS_SLR_PATH_PREFIX }),
+  );
+  const arn = roles?.[roles.length - 1]?.Arn;
+  if (!arn) {
+    throw new Error(
+      'Outbound Campaigns service-linked role not found after onboarding',
+    );
+  }
+  return arn;
+};
```
sarayev, line 100: **[MINOR]** `roles[roles.length - 1]` assumes `ListRoles` returns the newest SLR last, but the ordering isn't contractual. Usually there's only one `connect-campaigns` SLR so this is low-risk, but matching on the expected role-name pattern would be more robust than relying on list position. What do you think?

packages/backend-notifications/src/factory.ts
soberm: Settled on `notifications.amazon_connect` for the client-config output key — this is a notifications feature (not analytics), and the amplify-js Customer Profiles provider (#14866) reads that same key, so the construct and clients are aligned. Updated the PR description to match.

packages/backend-notifications/package.json
soberm: Confirmed — client-connect and client-connectcampaignsv2 are pinned ^3.1079.0 and client-customer-profiles ^3.936.0; all three resolve on npm (currently 3.1093.0). Clean install succeeds.

packages/backend-notifications/package.json
soberm: Yes — @aws-amplify/backend-platform-test-stubs is an internal workspace-only package under devDependencies, so it doesn't affect installs of the published package.

packages/backend-notifications/src/construct.ts
```diff
@@ -0,0 +1,1163 @@
...
+   * lowercase alphanumeric characters + single hyphens, no leading/trailing hyphen, not
+   * starting with the reserved `d-` prefix, and within Connect's 1–62 char
+   * bound.
+   */
+  private sanitizeInstanceAlias(raw: string): string {
+    let alias = raw
+      .toLowerCase()
+      .replace(/[^a-z0-9-]/g, '-')
+      .replace(/-+/g, '-')
+      .replace(/^-+|-+$/g, '');
```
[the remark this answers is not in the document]
soberm: Addressed in e407c266 — the instance-alias sanitizer now bounds raw input length (<=256 chars) before any regex, and the transform is non-backtracking (split('-').filter(Boolean).join('-')), so there's no polynomial blow-up on untrusted input.

packages/backend-notifications/src/construct.ts
soberm: Fixed in e851e11c — scrubbed the stale comments (cognitoUserKey, AmplifyGuestProfile, GSI on profileId) to match the shipped design: a single principalId-keyed AmplifyProfile with the GSI on principalId.

packages/backend-notifications/src/construct.ts
soberm: Deferred for preview with a tracking TODO at the domain — will thread an optional customer-managed defaultEncryptionKey post-preview.

packages/backend-notifications/src/construct.ts
soberm: Same — deferred for preview with a tracking TODO; will add an optional CMK encryptionKey post-preview (pairs with the Profiles domain).

packages/backend-notifications/src/construct.ts
soberm: Done in e851e11c — attached explicit LogGroups with ONE_MONTH retention to all three Lambdas (ApiFn, PushHandlerFn, CampaignAssociationFn), removalPolicy gated sandbox=DESTROY / branch=RETAIN. Log-group KMS deferred.

packages/backend-notifications/src/construct.ts
soberm: Added in e851e11c — the $default stage now sets DefaultRouteSettings throttling (burst 50 / rate 100 rps) to bound abuse/cost on the guest-writable routes.

packages/backend-notifications/src/construct.ts
soberm: Done — added sourceArn: connectInstance.attrArn on the create-from-scratch path (where the instance ARN is known at synth), keeping sourceAccount; attach mode stays sourceAccount-only since the ARN isn't known then.

packages/backend-notifications/src/lambda/push/delivery.ts
soberm: Fixed in 90569d15 — channelType is now optional on the delivery result and omitted on the skipped/unsupported-channel path instead of the misleading 'GCM'.

packages/backend-notifications/src/lambda/api/validation.ts
soberm: Leaving as-is — empty values are dropped by compact() in mapping.ts so the end state is correct; agree it's not worth an explicit reject.

packages/backend-notifications/src/lambda/campaign-association/integration.ts
soberm: Added a note at the ARN builders — deferring partition parameterization since Amazon Connect / Customer Profiles / Outbound Campaigns are commercial-partition-only today (not in aws-cn/aws-us-gov). Will thread partition through if/when the services expand.

packages/backend-notifications/src/lambda/campaign-association/integration.ts
soberm: Fixed in 84523d07 — now selects the connect-campaigns SLR by RoleName pattern (AWSServiceRoleForConnectCampaigns...) with a safe fallback, instead of relying on ListRoles order.

.changeset/backend-notifications.md
soberm: Rewritten in af161d29 to match the shipped design — three IAM/SigV4 routes (/identify-user, /register-device, /remove-device), a single principalId-keyed AmplifyProfile, and the DynamoDB device store; removed the /identify-user-guest, guestIdentityId, separate guest profile/TTL, and profileId-GSI references.

packages/backend-notifications/README.md
soberm: Rewritten in af161d29 — 'What it provisions' now reflects one object type (AmplifyProfile, principalId-keyed), devices in DynamoDB, and three IAM/SigV4 routes; removed the /identify-user-guest route and corrected the inaccurate JWT-authorizer line.

packages/integration-tests/src/test-project-setup/notifications_project.ts
```diff
@@ -0,0 +1,1591 @@
...
+    return (search.Items ?? []).filter((item) => !!item.ProfileId);
+  };
+
+  /** Send an UNSIGNED POST (no SigV4 headers) — used to prove the IAM gate. */
+  private unsignedPost = async (
+    endpoint: string,
+    path: string,
+    body: unknown,
+  ): Promise<SignedResponse> => {
+    const res = await fetch(`${endpoint}${path}`, {
```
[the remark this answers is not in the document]
soberm: Resolved earlier in 52267a2a — the test now validates the endpoint (https + *.execute-api.<region>.amazonaws.com) before the fetch, sanitizing the file-derived value; the CodeQL job is green on the latest run.

packages/backend-notifications/src/construct.ts
soberm: Bumped the log-group retention to 90 days (RetentionDays.THREE_MONTHS) — one month was on the low side for post-incident debugging.

.changeset/backend-notifications.md
soberm: Tightened the changeset further in fb783a39 — rewrote it as a plain feature description and removed the remaining design-diff phrasing (e.g. the guest-object-type / MergeProfiles / one-principal-maps-to-one-profile wording) so it reads as what the package adds rather than a diff against an earlier design.

sarayev (approved):

ekjotmultani (approved):

PR #3174, opened 2026-04-20
feat: standalone SSR hosting & CI/CD for Amplify Gen 2 (defineHosting / definePipeline / self-managed secrets)

adrianjoshua-strutt:
RFC / discussion: https://github.com/aws-amplify/amplify-backend/issues/3211

## What & why

Amplify Gen 2 is code-first: you define backend resources in TypeScript with `define*()` functions. This PR extends that model to **frontend hosting and CI/CD**, for teams that want to own their hosting infrastructure (version-controlled, PR-reviewable, CDK-extensible) and/or run it from their own account and pipeline — without giving up the managed Amplify Hosting option.

Everything here is **optional and independently adoptable**:
- Backend only → `ampx deploy --backend`
- Add self-managed hosting → add `amplify/hosting.ts`
- Add a managed pipeline → add `amplify/pipeline.ts`
- Bring your own CI/CD → call `ampx deploy` from it

## What's included

- **`defineHosting()`** (`@aws-amplify/hosting`) — a framework-agnostic L3 hosting construct (Next.js, Nuxt/Nitro, Astro, SPA/static) that deploys SSR/SSG on CloudFront + Lambda via an OpenNext build: KVS edge routing, ISR cache seeding, image optimization, multi-domain + WAF, configurable cache/headers, skew protection. Built on `@aws-blocks/hosting` 0.2.0.
- **`definePipeline()`** (`@aws-amplify/hosting/pipeline`) — a self-mutating CodePipeline (one per branch) with a two-phase backend-then-hosting deploy and typed per-stage config. Built on `@aws-blocks/pipeline` 0.2.0.
- **Self-managed secrets & config** — `secret()` (AWS Secrets Manager) and `config()` (SSM Parameter Store), read at runtime with `getSecret`/`getConfig` from the CDK-free `@aws-amplify/hosting/runtime` entry. `byoSecret()`/`byoConfig()` reference existing entries with no user CDK. Values are set out of band and never enter source or the template.
- **CLI** — `ampx deploy` gains `--backend`/`--frontend` and defaults `--identifier` to the sanitized `package.json` name; new `ampx secret` / `ampx config` (`set`/`get`/`list`/`remove`) manage self-managed hosting values.
- **Dependency alignment** — `aws-cdk-lib` peer floor → `^2.257.0`, `constructs` → `^10.6.0`; `@aws-cdk/toolkit-lib` → 1.40.0 (a hotswap fallback stays in STANDARD mode).
[...]

PR #3283, opened 2026-07-15
ci(e2e): harden E2E CI reliability + dependency updates

osama-rizk:
> **Note:** this PR now also contains the Node modernization from #3284 (matrix `[22,24,26]`), a `c8` bump to `^12` so coverage runs on Node 26, and a Node-22 pin for `e2e_notices` to avoid a Node-24 Windows libuv crash (nodejs/node#56645). Description updated to match.

## Purpose

Extract the **E2E CI reliability + dependency** work from the `snapshot/iac-hosting` feature branch (PR #3174) into a standalone PR against `main`. This lets the CI-hardening land on mainline independently of the hosting feature, de-risking that large merge. **This PR carries none of the hosting feature code** (no `defineHosting`/`definePipeline`, no `@aws-blocks/*` deps, no hosting fixtures).

### 1. Credential lifetime (fixes `ExpiredTokenException` on long e2e jobs)
- `run_with_e2e_account`: new `credential_duration_seconds` input + `assume_with_fallback()` that re-assumes both roles per attempt and **falls back to 3600s** if the role's `MaxSessionDuration` is lower (never regresses).
- `backwards_compatibility` wired to `credential_duration_seconds: 5400` + `attempt_timeout_minutes: 75`; `iam_access_drift` to `55`.
- `setup_profile`: documents the `MaxSessionDuration` cap.

### 2. Install speed
- New `warm_verdaccio_cache` action — cross-run `actions/cache` over the shared verdaccio storage so third-party packages aren't re-downloaded each run (wired into `check_api`, `iam_drift`, `backwards_compat`, `create_amplify`).
- New `disable_windows_defender_for_installs` action (Windows-only, best-effort) — excludes install paths + `node.exe`/`npm.exe` from real-time scanning.
- `npm_proxy_controller.ts`: shared-cache mode (`setUpWithSharedCache`) preserving the third-party cache, re-vending only workspace packages.
- **Node version modernization** (merged from #3284): dropped **EOL Node 18 and 20**; the health-checks matrix is now `["22", "24", "26"]` (22 & 24 Active LTS, 26 released 2026-05-05 → LTS Oct 2026). Node 26 is confirmed resolvable on setup-node and green across the full matrix in CI. Single-node jobs were re-mapped 18→22 / 20→24 / 22→26; the `engines.node` support contract in package manifests is intentionally left unchanged.

### 3. Runner OOM
- `check_api_changes`: bound validation concurrency to 4 (was unbounded 27-way parallel installs OOM-killing the 2-core/7GB runner).
- `api_changes_validator.ts`: retry/backoff + `--prefer-dedupe`.

### 4. Misc script robustness
- `check_pr_size.ts`: guard against the empty-diff pathspec error; exclude `.changeset/` (release metadata) from the size check. **Caps left at main's 1000/1000** (the hosting-merge cap bump was intentionally not brought over).
- New `clean_workspace_packages_from_verdaccio_storage.ts` helper; `publish_runner.ts` tweak; `.gitignore`/`.prettierignore` `API.md.api.md` entries; `.eslint_dictionary.json` words for the new comments.

### 5. Auto-run e2e for internal PRs targeting `main`
`do_include_e2e.ts`: any PR from this repo (non-fork) targeting `main` now runs e2e automatically, so mainline regressions surface before merge without a maintainer having to add the `run-e2e` label. **Fork-safe:** auto-run is gated on the PR head being `aws-amplify/amplify-backend`; fork PRs still use the label path and never hit the OIDC-credentials error (auto-running on forks would otherwise make `do_include_e2e` throw on every external contribution).

## Verification
- ✅ `tsc --build packages/integration-tests scripts` — clean
- ✅ `eslint --max-warnings 0` on all changed files — clean (dictionary updated)
- ✅ `prettier --check` — clean

## Scope guarantees
- `git diff --stat main...HEAD` touches only `.github/`, `scripts/`, root config, and the private `@aws-amplify/integration-tests` package (test/util files only).
- Full-diff grep for `hosting` / `@aws-blocks` / `standalone_hosting` / `vanilla_cdk` / `defineHosting` / `snapshot/iac-hosting` → **0 matches**.
- Empty changeset (CI/infra-only, no package release).

## Deliberately NOT included (kept with the hosting feature)
- `dependency_review_config.yml` allowlist (next/astro advisories for hosting demo fixtures)
- `.c8rc.json` coverage tweaks, `integration-tests/tsconfig.json` (hosting fixture refs)


.github/actions/run_with_e2e_account/action.yml
```diff
@@ -102,6 +118,84 @@ runs:
...
+              [ "$dur" = "3600" ] && break
+              echo "WARN: re-assume at ${dur}s failed (MaxSessionDuration too low?); retrying at 3600s" >&2
+            done
+            return 1
+          }
+
+          # Execution role -> default credential env (used by cdk/ampx deploy).
+          local exec_creds
+          if exec_creds=$(assume_with_fallback "$E2E_EXECUTION_ROLE_ARN" "gha-e2e-exec-${ATTEMPT}"); then
+            export AWS_ACCESS_KEY_ID=$(echo "$exec_creds" | jq -r '.AccessKeyId')
```
sarayev: **[major]** The re-assumed credentials here aren't masked in the logs.

When `setup_profile` fetches credentials it goes through `configure-aws-credentials`, which calls `core.setSecret()` so the access key / secret / session token show up as `***` in the workflow logs. This new path assumes the roles directly with `aws sts assume-role-with-web-identity` and exports the results without that masking.

Since this repo is public, if any later step ends up echoing the environment (a tool running under `set -x`, npm debug output, a crash dump), we'd be printing live ~1h deploy-capable credentials into public CI logs.

Could we mask them right after pulling them out with `jq`, before exporting? Something like:

```bash
local ak sk st
ak=$(echo "$exec_creds" | jq -r '.AccessKeyId')
sk=$(echo "$exec_creds" | jq -r '.SecretAccessKey')
st=$(echo "$exec_creds" | jq -r '.SessionToken')
echo "::add-mask::$sk"
echo "::add-mask::$st"
export AWS_ACCESS_KEY_ID=$ak AWS_SECRET_ACCESS_KEY=$sk AWS_SESSION_TOKEN=$st
```

Same for the tooling-role set before the `aws configure set` calls.

.github/actions/run_with_e2e_account/action.yml
```diff
@@ -30,7 +30,11 @@ inputs:
   attempt_timeout_minutes:
     description: Timeout for each attempt in minutes
     required: false
-    default: '25'
+    # 40 (was 25): a cold-cache create-amplify install from the local verdaccio
+    # proxy can legitimately exceed the old 25-min cap and time out; retries then
+    # outlive the 1h OIDC session token (ExpiredTokenException). A 40-min attempt
+    # lets a cold install finish in one pass within the credential window.
+    default: '40'
```
sarayev, line 37: **[minor]** Bumping the default `attempt_timeout_minutes` from 25 → 40 affects every job that relies on the default, not just the ones tuned in this PR. Could we double-check that for those jobs `retry_attempts × attempt_timeout` still fits inside their job-level `timeout-minutes`? The explicitly-tuned jobs (iam_drift, backwards_compat, create_amplify) look fine — it's the default-relying ones I'd want to confirm.

.github/workflows/health_checks.yml
```diff
@@ -43,7 +43,7 @@ on:
         description: 'Node versions list (as JSON array).'
         required: false
         type: string
-        default: '["18", "20", "22"]'
+        default: '["22", "24", "26"]'
```
sarayev, line 46: **[minor]** Small mismatch between the description and the change: the PR body says "drop EOL Node 18 → `["20","22"]`", but the matrix here becomes `["22","24","26"]` (so Node 20 is dropped too, and 24/26 added). Given the timing that's probably the right call — just want to confirm it's intentional and update the description so the two line up. Also worth a quick check that Node 26 resolves on the setup-node toolcache, since a version that isn't available yet would fail every matrix leg.

.github/workflows/health_checks.yml
```diff
@@ -658,6 +747,16 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cdk-lib-version: ${{ needs.resolve_inputs.outputs.cdk_lib_version }}
       - run: cd packages/cli && npm link
+      - name: Reset git tree after npm link
```
sarayev, line 750: **[nit]** `git checkout .` after `npm link` resets the whole working tree rather than just what `npm link` touched. It's safe on a clean CI checkout, but if the intent is to undo the package.json/lockfile churn from linking, scoping it (e.g. `git checkout -- packages/cli/package.json package-lock.json`) or a one-line comment naming what it's reverting would make it clearer.

.github/actions/run_with_e2e_account/action.yml
osama-rizk: Good catch — fixed in 1b4724b812. We now register the secret access key and session token with the runner's log masker via `::add-mask::` for **both** the execution and tooling roles before exporting/writing them, mirroring what `configure-aws-credentials` does with `core.setSecret()`. (AccessKeyId is left unmasked to match the SDK action's behavior; only the secret + session token are sensitive.)

.github/workflows/health_checks.yml
osama-rizk: Added a comment in 1b4724b812 explaining the intent: the full-tree reset is deliberate because the verdaccio warm-cache + local-publish steps right after are sensitive to *any* working-tree diff (not just package.json/lockfile), so we want a guaranteed-pristine checkout rather than a scoped revert that might miss a stray file `npm link` touched.

.github/workflows/health_checks.yml
osama-rizk: Confirmed intentional, and I've updated the PR description to match. The `["22","24","26"]` matrix is correct: this branch also absorbed the Node-modernization work (originally #3284), which drops **both** EOL Node 18 and 20 and adds 24 + 26. Node 26 is confirmed resolvable on setup-node — the latest CI run is green across all `test_with_coverage`/e2e legs on 26 (26.5.0 from the toolcache). Description now reflects the full scope (matrix, the `c8@^12` bump for Node-26 coverage, and the Node-22 pin for `e2e_notices`).

.github/actions/run_with_e2e_account/action.yml
osama-rizk: Good check — I audited every caller and no default-relying job is adversely affected, so I'd propose leaving the default bump as-is:

- **All 7 `run_with_e2e_account` usages in `health_checks.yml` set `attempt_timeout_minutes` explicitly** (55, 75, 120, 25, 35/25, 40/25) — none rely on the default.
- The **only** default-relying caller is `canary_checks.yml` → `live_dependency_health_checks`, which has `timeout-minutes: 20`. Since `20 < 25 < 40`, that job's own 20-min cap is the binding limit under **both** the old (25) and new (40) default — so the bump changes nothing for it.

So there's no job where `retry_attempts × attempt_timeout` newly overruns its `timeout-minutes`. Happy to also set `attempt_timeout_minutes: 20` explicitly on the canary job as a follow-up if you'd prefer it not depend on the default at all, but it's not needed for correctness here.

sarayev (approved):
All four review comments are addressed and verified:

- **Credential masking (major)** — re-assumed execution + tooling secrets and session tokens are now registered with `::add-mask::` before export/`aws configure set`, mirroring `configure-aws-credentials`' `core.setSecret()`. AccessKeyId intentionally left unmasked, which matches the SDK action's behavior.
- **Default `attempt_timeout_minutes` 25→40** — confirmed safe: all 7 `run_with_e2e_account` usages in `health_checks.yml` set it explicitly; the only default-relying caller (`canary_checks` → `live_dependency_health_checks`) is bound by its own `timeout-minutes: 20`, so the bump changes nothing for it.
- **Node matrix vs. description** — description updated; `["22","24","26"]` confirmed intentional (drops EOL 18 & 20), Node 26 resolvable on setup-node, no stray 18/20 refs left in any workflow.
- **`git checkout .` after `npm link`** — explanatory comment added clarifying the full-tree reset is deliberate (the verdaccio warm-cache + local-publish steps are sensitive to any working-tree diff).

Also verified the description's claims against the code (credential wiring, warm-cache job coverage, concurrency cap, PR-size guards, empty changeset, scope guarantees) and the fork-safe e2e auto-run logic. Nice, well-documented hardening work.

sarayev:
Thanks for addressing the comments! 🙌 Everything looks great now — appreciate the thorough responses and the extra verification (Node 26 availability, timeout audit). Approved.

Simone319 (approved):

PR #3285, opened 2026-07-16
fix(model-generator): skip empty graphql documents to prevent Unexpected <EOF> with @function resolvers

### LINKED ISSUE #3280: generate graphql-client-code fails with SyntaxError: Unexpected <EOF> when schema uses @function resolvers
### Environment information

```plain text
- Amplify CLI: `@aws-amplify/backend-cli` (tested 1.8.0 and latest)
- Node: v22.12.0 (also reproducible on v20.19.4 in CI)
- OS: macOS (also fails in Amplify Console build environment)
- Framework: Next.js 15, Amplify Gen 2
```

### Describe the bug

`ampx generate graphql-client-code --format graphql-codegen` fails with a `SyntaxError: Unexpected <EOF>` when the schema contains mutations or queries backed by `@function` resolvers.

### Reproduction steps

1. Create an Amplify Gen 2 project with a schema that includes `@function` resolvers:

```graphql
type Mutation {
  myCustomMutation(input: MyInput!): MyResult
    @function(name: "myLambdaFunction")
    @aws_cognito_user_pools
}
```

2. Deploy the backend (so outputs exist)

3. Run:

```bash
npx ampx generate graphql-client-code --format graphql-codegen --branch dev --app-id <your-app-id>
```

### Expected behavior

Codegen generates TypeScript types and operation documents for all schema types, skipping `@function`-backed operations gracefully (or generating stub operations for them).

### Actual behavior

```
[SyntaxError] Syntax Error: Unexpected <EOF> (4:1)
  2 | 
  3 | 
> 4 | 
    | ^
```

The codegen appears to generate an empty `.graphql` operations file (because `@function` resolvers don't produce auto-generated client operations), and then the GraphQL parser attempts to parse the empty file and crashes on EOF.

### Workaround

Using `@graphql-codegen/cli` (community package) with the `typescript` plugin reading `schema.graphql` directly generates schema types successfully. Frontend per-operation types (`MutationVariables`, `QueryResult`) are maintained manually.

### Impact

Any Gen 2 project using `@function` resolvers cannot use the native `ampx generate graphql-client-code` command. This forces manual maintenance of frontend type files, which is error-prone and increases development overhead.

### Additional context

- This has persisted across multiple CLI versions over the past year
- The project has 50+ mutations/queries, most backed by `@function` resolvers
- The `@graphql-codegen/cli` workaround works for Lambda types but cannot generate the per-operation types that `ampx` would normally produce

ahmedhamouda78:
## Problem

`ampx generate graphql-client-code` crashes with `SyntaxError: Unexpected <EOF>` when an Amplify Data schema uses only `@function`-backed custom operations (custom queries/mutations bound to a Lambda) and no `@model` types.

Fixes #3280

## Root cause

When a schema exposes only `@function`-backed custom operations and no `@model` types, AppSync generates **no subscriptions**. The code-generation formatter still emits a **comment-only** document (just the auto-generated header) for the empty `subscriptions` operation type:

```
# this is an auto generated file. This will be overwritten


```

That comment-only string is then handed to `graphql.parse()` (via `AppSyncGraphqlTypesGenerator` → `Source` → `@aws-amplify/graphql-generator`'s `generateTypes` → `parseAndMergeQueryDocuments`). Because a GraphQL document must contain at least one definition, `parse()` throws `Syntax Error: Unexpected <EOF>`, aborting the entire client-code generation. The reporter's `(4:1)` error position corresponds exactly to a 3-line header + blank line, then EOF at line 4.

## Fix

Add an `isEmptyGraphqlDocument()` helper (`packages/model-generator/src/empty_graphql_document.ts`) that treats empty / whitespace-only / comment-only documents as "empty", while still reporting genuinely malformed **non-empty** documents as non-empty so their real parse error surfaces downstream instead of being silently dropped.

Filter these empty documents out before they reach the parser in both:
- `AppSyncGraphqlTypesGenerator` — before wrapping each statement in a `graphql` `Source`
- `AppSyncGraphqlDocumentGenerator` — before `resultBuilder`

The fix lives at the `model-generator` boundary, so it is independent of the installed `@aws-amplify/graphql-generator` version.

## Test coverage

All tests exercise real code paths (real generators, real schemas, the real `@aws-amplify/graphql-generator`, and the real `graphql.parse` — no mocks of the code under test):

- `empty_graphql_document.test.ts` — unit tests for the helper: the comment-only header document, empty string, whitespace-only, comment-only, valid single/multi-definition docs, and a malformed-but-non-empty doc (which must remain non-empty).
- `graphql_types_generator.test.ts` — an end-to-end test with an `@function`-representative schema (Query + Mutation, no Subscription) that previously crashed, plus a regression test asserting the **unfiltered** path rejects with `Could not parse graphql operations` / `Unexpected <EOF>` while the **filtered** path succeeds.
- `graphql_document_generator.test.ts` — asserts no empty documents are emitted for an `@function`-representative schema.

## Verification

```
tsc --build packages/model-generator        → exit 0
eslint --max-warnings 0 (all changed files)  → exit 0
node:test (packages/model-generator)         → 48 passed, 0 failed
changeset status --since=main                → @aws-amplify/model-generator: patch
```


packages/model-generator/src/graphql_document_generator.ts
```diff
@@ -39,6 +40,12 @@ export class AppSyncGraphqlDocumentGenerator implements GraphqlDocumentGenerator
       relativeTypesPath,
     });
 
-    return this.resultBuilder(generatedStatements);
+    const nonEmptyStatements = Object.fromEntries(
+      Object.entries(generatedStatements).filter(
+        ([, contents]) => !isEmptyGraphqlDocument(contents),
+      ),
+    );
```
Simone319, line 47: Would there be a case that all the generatedStatements have empty content? Will we throw unexpected error for that case?

packages/model-generator/src/graphql_types_generator.ts
```diff
@@ -43,9 +44,9 @@ export class AppSyncGraphqlTypesGenerator implements GraphqlTypesGenerator {
       target: 'graphql',
     });
 
-    const queries = Object.entries(generatedStatements).map(
-      ([filename, contents]) => new Source(contents, filename),
-    );
+    const queries = Object.entries(generatedStatements)
+      .filter(([, contents]) => !isEmptyGraphqlDocument(contents))
+      .map(([filename, contents]) => new Source(contents, filename));
```
Simone319, line 49: Same here, are we handling the case when queries is empty list?

packages/model-generator/src/graphql_document_generator.ts
ahmedhamouda78: Good question. Two parts:

**Is it reachable?** Not from a real schema. `generateStatements` always emits at least one non-empty document for any valid schema — a deployed AppSync schema must have a `Query` root, and I confirmed that Query-only, Mutation-only, and Subscription-only schemas each still produce a non-empty `queries/mutations/subscriptions.graphql`. So `nonEmptyStatements` becoming `{}` (every statement filtered) isn't reachable in practice.

**Would it throw if it were `{}`?** No. Here the `resultBuilder` is `(fileMap) => new AppsyncGraphqlGenerationResult(fileMap)` (`create_graphql_document_generator.ts`). `AppsyncGraphqlGenerationResult` just stores the map: `getResults()` returns `{}`, and `writeToDirectory()` maps over `Object.entries({})` → `Promise.all([])` → `{ filesWritten: [] }`, writing nothing. There's no `parse()` on this path, so no `Unexpected <EOF>` risk — an empty map is a no-op, not an error.

I added a regression test locking this in: **`appsync_graphql_generation_result.test.ts` → "handles an empty file map without throwing (all operation documents filtered out) (#3285)"** — asserts `getResults()` → `{}`, `writeToDirectory()` → `{ filesWritten: [] }`, and zero `fs` writes. (commit `77ab9c3a`)

packages/model-generator/src/graphql_types_generator.ts
ahmedhamouda78: Yes, the empty-list case is handled. If every statement is filtered out, `queries` is `[]` and flows into `generateTypes` → `generateFromString` → `parseAndMergeQueryDocuments([])` (`@aws-amplify/graphql-types-generator@3.6.0`). That function does `sources.map(source => parse(source))` — with an empty array the callback never runs, so **`graphql.parse()` is never called** (i.e. no `Unexpected <EOF>`) — and it returns `concatAST([])`, which is an empty-but-valid `DocumentNode` (`{ kind: 'Document', definitions: [] }`). `validateQueryDocument` then finds no operations and reports no errors, so codegen succeeds and emits only the boilerplate `API.ts`.

Verified empirically: `generateTypes({ schema, target: 'typescript', queries: [] })` returns `{ 'API.ts': <header-only> }` without throwing. (As on the other thread, this all-empty case isn't reachable from a real schema since `generateStatements` always emits ≥1 non-empty document — but it degrades gracefully regardless.)

Regression test added: **`graphql_types_generator.test.ts` → "does not throw when every operation document is filtered out and `queries` is an empty list (#3285)"** — asserts `generateTypes({ …, queries: [] })` does not reject and yields `['API.ts']`. (commit `77ab9c3a`)

Simone319 (approved):

Issue #825, opened 2023-12-15
`ctrl+c` on the `npm create amplify@latest` on path prompt crashes with stack trace


ykethan:
### Environment information

```plain text
System:
    OS: macOS 13.6.1
    CPU: (8) arm64 Apple M1
    Memory: 115.36 MB / 16.00 GB
    Shell: 3.6.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 18.18.0 - ~/.local/share/nvm/v18.18.0/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 9.8.1 - ~/.local/share/nvm/v18.18.0/bin/npm
    pnpm: 7.26.2 - /opt/homebrew/bin/pnpm
    bun: Not Found
    Watchman: Not Found
  npmPackages:
    @aws-amplify/backend: Not Found
    @aws-amplify/backend-cli: Not Found
    aws-amplify: Not Found
    aws-cdk: Not Found
    aws-cdk-lib: Not Found
    typescript: ^5.3.2 => 5.3.3
```

### Description

1. `npm create amplify@latest`
2. `ctrl+c` on first prompt
3. observe error

![Screenshot 2023-12-14 at 10 36 50 PM](https://github.com/aws-amplify/amplify-backend/assets/87995712/0e9c9d6f-6afd-4103-82cb-636e3a4e1d98)

similar to https://github.com/aws-amplify/amplify-backend/issues/321

mehuljariwala:
I verified this is still present on current `main`. `getProjectRoot()` runs before the existing top-level `try`, so Inquirer rejecting with `User force closed the prompt` bypasses the error boundary and produces the stack trace.

I am preparing a focused fix that keeps the bin entry thin, moves the create workflow into a testable runner, treats only the known prompt-force-close error as a clean cancellation, and preserves current logging plus exit code 1 for all other failures. I will add unit coverage for cancellation, ordinary errors, and the successful path, plus the required patch changeset.

PR #3291, opened 2026-07-21
fix(create-amplify): exit quietly when a prompt is cancelled with ctrl+c

### LINKED ISSUE #321: ctrl+c on the `npx amplify sandbox` exit prompt crashes with stack trace
1. `npx amplify sandbox`
2. ctrl+c
3. ctrl+c
4. observe error

```
[Sandbox] Watching for file changes...
^C[Sandbox] Shutting down
? Would you like to delete all the resources in your sandbox environment (This cannot be undone)? 
(y/N)
file:///Users/josefai/Documents/projects/aws-amplify/tmp-samsara-cli/examples/next.js/node_modules/@inquirer/core/dist/esm/lib/create-prompt.mjs:36
                    reject(new Error(`User force closed the prompt with ${code} ${signal}`));
                           ^

Error: User force closed the prompt with 0 null
    at file:///Users/josefai/Documents/projects/aws-amplify/tmp-samsara-cli/examples/next.js/node_modules/@inquirer/core/dist/esm/lib/create-prompt.mjs:36:28
    at Emitter.emit (file:///Users/josefai/Documents/projects/aws-amplify/tmp-samsara-cli/examples/next.js/node_modules/@inquirer/core/node_modules/signal-exit/dist/mjs/index.js:67:19)
    at #processEmit (file:///Users/josefai/Documents/projects/aws-amplify/tmp-samsara-cli/examples/next.js/node_modules/@inquirer/core/node_modules/signal-exit/dist/mjs/index.js:236:27)
    at #process.emit (file:///Users/josefai/Documents/projects/aws-amplify/tmp-samsara-cli/examples/next.js/node_modules/@inquirer/core/node_modules/signal-exit/dist/mjs/index.js:187:37)
    at process.callbackTrampoline (node:internal/async_hooks:130:17)
```

ihopenre-eng:
## Problem

`npm create amplify@latest` prints a stack trace when you press ctrl+c at the first prompt.

`getProjectRoot()` is awaited at module top level in `packages/create-amplify/src/create_amplify.ts`, outside the `try` that guards the rest of the flow. A cancelled prompt rejects with `ExitPromptError` from `@inquirer/core`, nothing catches it, and the process dies on an unhandled rejection instead of just stopping.

**Issue number, if available:** #825 (same class as the already-fixed #321)

## Changes

- Move the prompt, and the objects built from its result, inside the existing `try`.
- Treat a cancelled prompt as an explicit intent to exit: nothing is printed for it. The detection matches on the message rather than the error type, exactly as `packages/cli/src/error_handler.ts` already does, so this package takes no new dependency on the prompt library.
- Changeset added (`create-amplify`, patch).

The file's header comment asks to keep logic in this package to a minimum, so the change is a scope move plus a four-line guard.

## Validation

**Reproduced against the built binary** (`packages/create-amplify/lib/create_amplify.js`, `npm_config_user_agent` set so the CLI's package-manager detection is satisfied, prompt cancelled):

before

```
ExitPromptError: User force closed the prompt with 13 null
    at .../node_modules/@inquirer/core/dist/esm/lib/create-prompt.js:57:20
    at Emitter.emit (.../node_modules/signal-exit/dist/mjs/index.js:67:19)
    ...
Node.js v22.17.1
exit 13
```

after

```
? Where should we create your project? (.)
exit 1
```

**Cancellation delivered as a rejection while the event loop is alive** (what ctrl+c does, as opposed to a closed stdin, which exits the process before the rejection can be observed). Reduced to the two shapes of the module:

| await placement | result |
| --- | --- |
| outside the `try`, as on `main` | `ExitPromptError: User force closed the prompt with SIGINT null` plus stack trace |
| inside the `try`, as in this PR | no output, exit 1 |

**Repo checks:** `npm run build` passes, `eslint --max-warnings 0` is clean on the changed file, `prettier --check` passes, and the pre-commit hook (lint-staged, tsconfig-refs, create-amplify-deps, package-lock checks) ran clean.

No behaviour changes for successful runs or for any other error, which still print through `printer.log(format.error(err))` and exit 1.


packages/create-amplify/src/create_amplify.ts
```diff
@@ -19,26 +19,38 @@ import { getProjectRoot } from './get_project_root.js';
...
+  );
 
-try {
   await amplifyProjectCreator.create();
 } catch (err) {
-  printer.log(format.error(err), LogLevel.ERROR);
+  if (!isUserForceClosePromptError(err)) {
+    printer.log(format.error(err), LogLevel.ERROR);
+  }
   process.exitCode = 1;
```
sharonyajain: The fix adds no automated regression test. The new guard (moving the prompt into the try plus isUserForceClosePromptError) is verified only by the manual before/after in the description. Being a top-level-await bin module with a non-exported helper, nothing pins the behavior.

```ts
} catch (err) {
  if (!isUserForceClosePromptError(err)) {
    printer.log(format.error(err), LogLevel.ERROR);
  }
  process.exitCode = 1;
}
```

**Why it matters:** no test for the recurred class -> a future refactor can again move an awaited prompt outside the try, or drift the message match, with a green build -> the Ctrl+C stack-trace bug regresses a third time and users again see a crash on cancel.

**Suggestion:** Extract the flow (or at least isUserForceClosePromptError + the catch branch) into a small exported, testable function and add a unit test that drives it with an Error whose message includes 'User force closed the prompt' (asserting nothing is logged and exitCode is 1) and with a generic Error (asserting it is logged). This locks the exact class that has now recurred twice.

packages/create-amplify/src/create_amplify.ts
```diff
@@ -19,26 +19,38 @@ import { getProjectRoot } from './get_project_root.js';
...
 
-const projectRoot = await getProjectRoot();
+/**
+ * Ctrl+C during a prompt rejects with an ExitPromptError from the prompt
+ * library. The intent to exit is explicit, so there is nothing to report.
+ * Matched on the message rather than the type so that this package does not
+ * take a direct dependency on the prompt library, the same way the CLI's
+ * error handler does it.
+ */
+const isUserForceClosePromptError = (err: unknown): boolean =>
```
sharonyajain: This helper and the literal 'User force closed the prompt' now exist in two places: here and in packages/cli/src/error_handler.ts (isUserForceClosePromptError, line ~136). The comment explains the string match avoids a direct prompt-library dependency, which is sound, but the string itself is duplicated across packages rather than shared. Both packages already depend on @aws-amplify/cli-core (imported at the top of this file), which is a natural home for a shared predicate.

```ts
const isUserForceClosePromptError = (err: unknown): boolean =>
  err instanceof Error && err.message.includes('User force closed the prompt');
```

**Why it matters:** duplicated magic string across two packages -> when the upstream @inquirer message text changes, one site can be updated and the other missed -> the un-updated package silently reverts to printing a stack trace on Ctrl+C, reintroducing exactly this bug in one entry point only.

**Suggestion:** Move isUserForceClosePromptError (and the literal) into the shared @aws-amplify/cli-core and consume it from both create-amplify and the CLI's error_handler, so the message contract lives in one place. This keeps the no-prompt-library-dependency property while removing the drift risk.

packages/create-amplify/src/create_amplify.ts
sharonyajain: The fix adds no automated regression test. The new guard (moving the prompt into the try plus isUserForceClosePromptError) is verified only by the manual before/after in the description. Being a top-level-await bin module with a non-exported helper, nothing pins the behavior.

```ts
} catch (err) {
  if (!isUserForceClosePromptError(err)) {
    printer.log(format.error(err), LogLevel.ERROR);
  }
  process.exitCode = 1;
}
```

**Why it matters:** no test for the recurred class -> a future refactor can again move an awaited prompt outside the try, or drift the message match, with a green build -> the Ctrl+C stack-trace bug regresses a third time and users again see a crash on cancel.

**Suggestion:** Extract the flow (or at least isUserForceClosePromptError + the catch branch) into a small exported, testable function and add a unit test that drives it with an Error whose message includes 'User force closed the prompt' (asserting nothing is logged and exitCode is 1) and with a generic Error (asserting it is logged). This locks the exact class that has now recurred twice.

sharonyajain:
Hi @ihopenre-eng,
The core change looks right: moving the getProjectRoot() await inside the try and treating the force-close as a clean cancel is exactly the correct approach, and it mirrors how packages/cli/src/error_handler.ts already handles it.

Before it can go in, there are two things still open:

- A unit test for the cancellation path — covering that a force-closed prompt exits quietly (no stack trace) while an ordinary error still logs and sets a non-zero exit code. This is the main blocker to merging.

- The exit code on cancel — the current change suppresses the error output but still sets process.exitCode = 1. Worth a deliberate decision on whether an intentional Ctrl+C should exit 0 or 1; happy to weigh in.

Are you still interested in picking these up? No worries either way, just let us know. Thanks again for the contribution!

packages/create-amplify/src/create_amplify.ts
ihopenre-eng: Added in eae0a4d. The catch body is now an exported `handleError` in `packages/create-amplify/src/handle_error.ts`, and `create_amplify.ts` only calls it, so the bin module keeps the "as little logic as possible" property the header comment asks for.

`handle_error.test.ts` drives it both ways, in the same style as the force-close case in `packages/cli/src/error_handler.test.ts`:

- an error whose message contains `User force closed the prompt`: `printer.log` is not called, `process.exitCode` is 1
- any other error: `printer.log` is called once with the formatted message, `process.exitCode` is 1

`process.exitCode` is restored in `afterEach` so the suite itself does not exit non-zero.

Verified locally: 21/21 tests pass in the package, and the new test fails as expected when the guard is removed from the compiled handler (`pass 1, fail 1`), so it does pin the behaviour rather than passing vacuously.

packages/create-amplify/src/create_amplify.ts
ihopenre-eng: Agreed that the string is the contract and that drift is the risk. I left it in this package for now, and here is the reasoning so you can overrule it if you prefer the move.

Moving the predicate into `@aws-amplify/cli-core` adds a symbol to that package's public API, which means an API.md regeneration and a changeset for cli-core and the CLI on top of a four line Ctrl+C fix. That is a bigger surface to review than the duplication it removes, and cli-core is consumed well beyond these two call sites.

What the duplication actually costs is covered now: both occurrences are pinned by tests, `packages/cli/src/error_handler.test.ts` for the CLI and the new `packages/create-amplify/src/handle_error.test.ts` here. If the upstream message text changes, both tests go red rather than one entry point silently regressing.

Happy to do the cli-core extraction as a separate PR if you would rather have the single source of truth, since it touches two more packages and reads better on its own.

ihopenre-eng:
Thanks for the review, still interested. Both points are addressed in eae0a4d.

**1. Unit test.** The catch body moved into an exported `handleError` in `packages/create-amplify/src/handle_error.ts`, and `create_amplify.ts` now only calls it, so the bin module stays as thin as its header comment asks. `handle_error.test.ts` covers both branches: a force-closed prompt logs nothing and exits 1, any other error is logged once and exits 1. Ran the package suite locally, 21/21 pass, and confirmed the test fails when the guard is removed, so it pins the behaviour rather than passing vacuously.

**2. Exit code on cancel.** I kept 1, for consistency with what the CLI already does. In `packages/cli/src/error_handler.ts` the force-close case only suppresses the output, while the `unhandledRejection` listener has already set `process.exitCode = 1`, so `ampx` already exits non-zero when a prompt is cancelled. Independently of that, 0 would tell any script or CI job wrapping `npm create amplify` that a project was created when none was, so a non-zero code is the more accurate signal. If you would rather have cancellation exit 0, that reads to me as a deliberate cross-package decision for both entry points, and I am happy to do it in a follow-up so this fix does not change CLI behaviour as a side effect.

I also replied on the shared-predicate thread with why I left it local for now, along with an offer to do the cli-core extraction separately.

sharonyajain (approved):

PR #3303, opened 2026-08-11
fix(e2e-cleanup): tear down all hosting resources instead of orphaning them

sarayev:
Consolidates the three previously-stacked PRs (#3304, #3305 closed in favour of this one).

## What

The hourly `e2e-resource-cleanup` job deletes resources directly while the stacks that own them are still deleting. That poisons those stacks' delete path — a deleted custom-resource execution role leaves CloudFormation waiting on a Lambda it can no longer assume — so stacks land in `DELETE_FAILED` and every later run fails identically. **The job creates the backlog it is meant to drain.** It also deleted S3 origin buckets out from under CloudFront distributions (orphaned distributions, subdomain-takeover risk), and never fully emptied versioned buckets.

## Fix

Under `scripts/components/e2e-cleanup/`, wired into `cleanup_e2e_resources.ts`:

- **`StackDeleter`** — indexes the resources of every not-yet-deleted stack **before** any delete is issued; all six direct sweeps skip anything a live stack still owns. Fails closed (skips everything) and loud (`process.exitCode = 1`) when the index can't be built. Stacks are inspected with bounded concurrency so an account with hundreds of stacks is indexed well within the job timeout.
- **`S3BucketEmptier`** — paginates `ListObjectVersions` on both `KeyMarker` and `VersionIdMarker`, batches `DeleteObjects` at 1000, throws on `Errors[]`, re-empties once on `BucketNotEmpty`.
- **`CloudFrontDistributionCleaner`** — maps buckets to distributions via their S3 origin domains and converges across hourly runs (enabled → disable → wait for `Deployed` → delete). A bucket is only deleted once its distributions are gone, so the origin never outlives the distribution.

### Draining the existing backlog

Verified read-only against a test account: **every** stuck root stack blocks on exactly one resource, its nested hosting stack, and that nested stack in turn blocks only on `AWS::S3::Bucket` — a versioned bucket CloudFormation can't delete because it isn't empty (`409 ... You must delete all versions in the bucket`). The bucket is invisible from the root, which is why the drain never saw it.

`findBucketsBlockingStackDeletion` therefore walks nested stacks to find the blocking buckets and empties them **before** the stack delete is retried. The bucket itself is left to CloudFormation, so its record of the resource stays intact and its own delete succeeds on the retry.

Retaining the nested stack instead — the one-line allowlist change — would *detach* it from its root: the root would reach `DELETE_COMPLETE` while the nested stack silently kept its bucket and its distribution. That converts a visible failure into an invisible orphan, so `AWS::CloudFormation::Stack` is deliberately **not** in the retain allowlist. A nested-stack failure that emptying can't fix (e.g. a distribution) is logged for a human, since `deleteStack` only ever sees the nested stack resource itself.

### Region scoping

S3 buckets and IAM roles are listed **account wide**, so each per-region job sees the buckets and roles of all four regions while the ownership index only covered one — a role owned by a live stack in another region looked free. The index is now built across every region in the workflow matrix before any sweep. Region-scoped types (Cognito, DynamoDB, SSM, log groups) stay scoped to the current region, because their names are only unique within a region and the equally-named resource of another region would otherwise keep a genuinely stale one from ever being cleaned up. `e2e_test_regions.test.ts` parses the workflow and fails if the region list drifts from the matrix.

## Tests

94/94 (`npm run test:scripts`), 45 new. `tsc`, `eslint --max-warnings 0`, `prettier --check` clean.

## Notes

- **IAM prerequisite (out of repo):** the cleanup role needs `cloudfront:ListDistributions/GetDistribution/GetDistributionConfig/UpdateDistribution/DeleteDistribution/TagResource` and `cloudformation:DescribeStacks/DescribeStackResources/ListStackResources`. Both degrade gracefully — CloudFront reaping is inert and the sweeps fail closed until it lands — so merge order doesn't matter.
- **`check_pr_size` is red and advisory** — not a required check on `main` (precedent: #3265 merged at +15,994). ~660 of the added lines are `package-lock.json` and roughly two thirds of the rest are tests. Happy to re-split if you'd rather review it in pieces.
- `e2e_notices macos-14` and `e2e_iam_access_drift` are pre-existing flakes that also fail on `main`.
- No `packages/*` changes, so no changeset.

### DIFF
--- a/scripts/cleanup_e2e_resources.ts
+++ b/scripts/cleanup_e2e_resources.ts
@@ -1,9 +1,5 @@
 import {
   CloudFormationClient,
-  DeleteStackCommand,
-  ListStacksCommand,
-  ListStacksCommandOutput,
-  StackStatus,
   StackSummary,
 } from '@aws-sdk/client-cloudformation';
 import {
@@ -72,6 +68,7 @@ import {
   ListTablesCommandOutput,
   TableDescription,
 } from '@aws-sdk/client-dynamodb';
+import { StackDeleter } from './components/e2e-cleanup/stack_deleter.js';
 
 const amplifyClient = new AmplifyClient({
   maxAttempts: 5,
@@ -145,44 +142,54 @@ const isStale = (creationDate: Date | undefined): boolean | undefined => {
   return now.getTime() - creationDate.getTime() > staleDurationInMilliseconds;
 };
 
-const listAllStaleTestStacks = async (): Promise<Array<StackSummary>> => {
-  let nextToken: string | undefined = undefined;
-  const stackSummaries: Array<StackSummary> = [];
-  do {
-    const listStacksResponse: ListStacksCommandOutput = await cfnClient.send(
-      new ListStacksCommand({
-        NextToken: nextToken,
-        StackStatusFilter: Object.keys(StackStatus).filter(
-          (status) => status != StackStatus.DELETE_COMPLETE,
-        ) as Array<StackStatus>,
-      }),
-    );
-    nextToken = listStacksResponse.NextToken;
-    listStacksResponse.StackSummaries?.filter(
-      (stackSummary) =>
-        stackSummary.StackName?.startsWith(TEST_AMPLIFY_RESOURCE_PREFIX) &&
-        isStackStale(stackSummary),
-    ).forEach((item) => {
-      stackSummaries.push(item);
-    });
-  } while (nextToken);
-  return stackSummaries;
+const stackDeleter = new StackDeleter(
+  cfnClient,
+  TEST_AMPLIFY_RESOURCE_PREFIX,
+  isStackStale,
+);
+
+/**
+ * Deleting a resource that a stack still owns poisons the delete path of that stack, which is
+ * how stacks end up stuck in DELETE_FAILED forever. Deleting the execution role of a custom
+ * resource for example leaves CloudFormation waiting for a Lambda function that can no longer be
+ * assumed. The ownership index must be captured before any deletion is requested, otherwise the
+ * resources of the stacks that are being deleted look free while CloudFormation still uses them.
+ */
+const liveStackResources = await stackDeleter.getResourcesOwnedByLiveStacks();
+if (!liveStackResources.isComplete) {
+  console.log(
+    'Could not determine which resources still belong to stacks. Only stack deletion will run',
+  );
+}
+
+const isOwnedByLiveStack = (
+  resourceType: string,
+  physicalResourceId: string | undefined,
+): boolean => {
+  if (
+    !liveStackResources.isOwnedByLiveStack(resourceType, physicalResourceId)
+  ) {
+    return false;
+  }
+  console.log(
+    `Skipping direct deletion of ${physicalResourceId} ${resourceType}. It belongs to a stack that has not finished deleting`,
+  );
+  return true;
 };
 
-const allStaleStacks = await listAllStaleTestStacks();
+const allStaleStacks = await stackDeleter.listStaleTopLevelStacks();
 
 for (const staleStack of allStaleStacks) {
-  if (staleStack.StackName) {
-    const stackName = staleStack.StackName;
-    try {
-      await cfnClient.send(new DeleteStackCommand({ StackName: stackName }));
-      console.log(`Successfully kicked off ${stackName} stack deletion`);
-    } catch (e) {
-      const errorMessage = e instanceof Error ? e.message : '';
-      console.log(
-        `Failed to kick off ${stackName} stack deletion. ${errorMessage}`,
-      );
-    }
+  try {
+    await stackDeleter.deleteStack(staleStack);
+    console.log(
+      `Successfully kicked off ${staleStack.StackName} stack deletion`,
+    );
+  } catch (e) {
+    const errorMessage = e instanceof Error ? e.message : '';
+    console.log(
+      `Failed to kick off ${staleStack.StackName} stack deletion. ${errorMessage}`,
+    );
   }
 }
 
@@ -262,6 +269,9 @@ const emptyAndDeleteS3Bucket = async (bucketName: string): Promise<void> => {
 for (const staleBucket of staleBuckets) {
   if (staleBucket.Name) {
     const bucketName = staleBucket.Name;
+    if (isOwnedByLiveStack('AWS::S3::Bucket', bucketName)) {
+      continue;
+    }
     try {
       await emptyAndDeleteS3Bucket(bucketName);
       console.log(`Successfully deleted ${bucketName} bucket`);
@@ -297,6 +307,9 @@ const staleUserPools = await listStaleCognitoUserPools();
 
 for (const staleUserPool of staleUserPools) {
   if (staleUserPool.Name) {
+    if (isOwnedByLiveStack('AWS::Cognito::UserPool', staleUserPool.Id)) {
+      continue;
+    }
     try {
       const describeUserPoolResponse = await cognitoClient.send(
         new DescribeUserPoolCommand({
@@ -446,6 +459,9 @@ const listAllStaleRoles = async (): Promise<Array<Role>> => {
 
 const allStaleRoles = await listAllStaleRoles();
 for (const staleRole of allStaleRoles) {
+  if (isOwnedByLiveStack('AWS::IAM::Role', staleRole.RoleName)) {
+    continue;
+  }
   try {
     // delete inline policies
     const inlinePolicies: ListRolePoliciesCommandOutput = await iamClient.send(
@@ -519,6 +535,9 @@ const listAllStaleSSMParameters = async (): Promise<
 
 const allStaleSSMParameters = await listAllStaleSSMParameters();
 for (const staleSSMParameter of allStaleSSMParameters) {
+  if (isOwnedByLiveStack('AWS::SSM::Parameter', staleSSMParameter.Name)) {
+    continue;
+  }
   try {
     await ssmClient.send(
       new DeleteParameterCommand({
@@ -568,6 +587,11 @@ const listAllStaleDynamoDBTables = async (): Promise<
 
 const allStaleDynamoDBTables = await listAllStaleDynamoDBTables();
 for (const staleDynamoDBTable of allStaleDynamoDBTables) {
+  if (
+    isOwnedByLiveStack('AWS::DynamoDB::Table', staleDynamoDBTable.TableName)
+  ) {
+    continue;
+  }
   try {
     await ddbClient.send(
       new DeleteTableCommand({
@@ -614,6 +638,9 @@ const listAllStaleTestLogGroups = async (): Promise<Array<LogGroup>> => {
 
 const allStaleLogGroups = await listAllStaleTestLogGroups();
 for (const logGroup of allStaleLogGroups) {
+  if (isOwnedByLiveStack('AWS::Logs::LogGroup', logGroup.logGroupName)) {
+    continue;
+  }
   try {
     await cloudWatchClient.send(
       new DeleteLogGroupCommand({
--- a/scripts/cleanup_e2e_resources.ts
+++ b/scripts/cleanup_e2e_resources.ts
@@ -2,6 +2,7 @@ import {
   CloudFormationClient,
   StackSummary,
 } from '@aws-sdk/client-cloudformation';
+import { CloudFrontClient } from '@aws-sdk/client-cloudfront';
 import {
   CloudWatchLogsClient,
   DeleteLogGroupCommand,
@@ -58,6 +59,7 @@ import {
   ListTablesCommandOutput,
   TableDescription,
 } from '@aws-sdk/client-dynamodb';
+import { CloudFrontDistributionCleaner } from './components/e2e-cleanup/cloudfront_distribution_cleaner.js';
 import { S3BucketEmptier } from './components/e2e-cleanup/s3_bucket_emptier.js';
 import {
   GuardedResourceType,
@@ -70,6 +72,9 @@ const amplifyClient = new AmplifyClient({
 const cfnClient = new CloudFormationClient({
   maxAttempts: 5,
   retryMode: 'adaptive',
+});
+const cloudFrontClient = new CloudFrontClient({
+  maxAttempts: 5,
 });
 const cloudWatchClient = new CloudWatchLogsClient({
   maxAttempts: 5,
@@ -143,6 +148,10 @@ const stackDeleter = new StackDeleter(
   isStackStale,
 );
 const s3BucketEmptier = new S3BucketEmptier(s3Client);
+const cloudFrontDistributionCleaner = new CloudFrontDistributionCleaner(
+  cloudFrontClient,
+  TEST_AMPLIFY_RESOURCE_PREFIX,
+);
 
 /**
  * Deleting a resource that a stack still owns poisons the delete path of that stack, which is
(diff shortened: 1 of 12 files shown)


scripts/components/e2e-cleanup/stack_deleter.ts
```diff
@@ -0,0 +1,284 @@
...
+        new DeleteStackCommand({ StackName: stackName }),
+      );
+      return [];
+    }
+    const failedResources = (
+      await this.listStackResources(stackSummary.StackId ?? stackName)
+    ).filter(
+      (stackResource) =>
+        stackResource.ResourceStatus === ResourceStatus.DELETE_FAILED,
+    );
```
Simone319: **Blocking:** Line 167's `await this.listStackResources(...)` is unguarded (unlike the index-build call wrapped at L128–141), so a throttle or a missing `ListStackResources` grant throws; the caller (`cleanup_e2e_resources.ts:188`) then catches and moves on without ever sending `DeleteStack`, giving that `DELETE_FAILED` stack zero delete attempts — the exact backlog this PR exists to drain. Please guard this call so an inspection failure still falls back to attempting the deletion (the "retry as-is" path at L180–183 already does this and needs no extra IAM).

scripts/components/e2e-cleanup/stack_deleter.ts
```diff
@@ -0,0 +1,284 @@
...
+ * Deletes stale e2e test CloudFormation stacks and tells which resources are still owned by stacks.
+ */
+export class StackDeleter {
+  private activeStacks: Array<StackSummary> | undefined = undefined;
+
+  /**
+   * Creates stack deleter.
+   */
+  constructor(
+    private readonly cfnClient: CloudFormationClient,
```
Simone319, line 175: **Important:** The ownership index is built region-scoped (`cfnClient` resolves `AWS_REGION`), but the IAM role sweep uses account-global `ListRoles`, and the job is a 4-region matrix in one account — so a role owned by a live `DELETE_FAILED` stack in another region isn't in this job's index and gets deleted, poisoning that stack's delete path (see also `cleanup_e2e_resources.ts:436–458`). This is a pre-existing hole the PR shrinks (4/4 → 3/4 jobs), not a regression; the durable fix is to run the global sweeps (IAM, S3) in one designated region whose job builds the index across all four regions. The S3 bucket sweep has the same gap, saved only accidentally by S3's cross-region `PermanentRedirect`.

scripts/components/e2e-cleanup/stack_deleter.ts
```diff
@@ -0,0 +1,284 @@
...
+ *
+ * The custom resources are the ones that time out on delete, and their bucket is deleted by the
+ * bucket sweep of a subsequent run. Any other resource type, in particular
+ * `AWS::CloudFront::Distribution`, must never be abandoned.
+ */
+const RESOURCE_TYPES_SAFE_TO_RETAIN = [
+  'Custom::S3AutoDeleteObjects',
+  'Custom::CDKBucketDeployment',
+  'AWS::S3::Bucket',
+];
```
Simone319, line 28: **Important:** Amplify Gen2 backends are nested-stack apps and CDK places the `S3AutoDeleteObjects`/`CDKBucketDeployment` custom resources inside nested stacks, so a stuck root's failed resource is type `AWS::CloudFormation::Stack` — absent from this allowlist → it falls to the "manual attention" plain-retry branch (L166–186) and the new retain/unblock path never engages, so for the common Amplify case the drain adds nothing over the pre-existing hourly re-delete. Please confirm against the actual stuck-stack list whether the backlog is mostly nested-stack roots, and if so, consider how to unblock them.

scripts/components/e2e-cleanup/stack_deleter.ts
```diff
@@ -0,0 +1,284 @@
...
+        }
+      } catch (error) {
+        if (this.isStackNotFoundError(error)) {
+          continue;
+        }
+        isComplete = false;
+        this.log(
+          `Unable to list resources of ${stackSummary.StackName} stack, skipping direct resource deletion. ${this.getErrorMessage(error)}`,
+        );
+      }
```
Simone319: **Important:** When any `listStackResources` fails, `isComplete` goes false and all six guarded sweeps skip (the unguarded branch sweep at `cleanup_e2e_resources.ts:416–434` still runs, so the "Only stack deletion will run" log at L160 is itself inaccurate), yet the script exits 0 so the workflow's `nick-fields/retry` never fires — an invisible multi-region cleanup outage. Set `process.exitCode = 1` after the delete phase (non-fatal — explicitly not `process.exit(1)`/rethrow, so cleanup that did run still completes), accepting the job then goes red hourly until the cause clears. Separately, the `1+N` index calls are fully serial against a 15-min timeout — add bounded concurrency (5–10) and `retryMode: 'adaptive'`; if you do, add a `deletionRequested` guard so the index-before-delete ordering survives the interleaving.

scripts/cleanup_e2e_resources.ts
```diff
@@ -262,6 +269,9 @@ const emptyAndDeleteS3Bucket = async (bucketName: string): Promise<void> => {
 for (const staleBucket of staleBuckets) {
   if (staleBucket.Name) {
     const bucketName = staleBucket.Name;
+    if (isOwnedByLiveStack('AWS::S3::Bucket', bucketName)) {
```
Simone319, line 293: **Important:** All six type strings passed to `isOwnedByLiveStack` (here and at L310, L462, L538, L591, L641) are currently correct, but a typo (e.g. `'AWS::Logs::Loggroup'`) would hit `?? false` (`stack_deleter.ts:64–68`) and silently fail open, invisible to tests — latent maintainability risk. Export a typed union of the type strings from `stack_deleter.ts` and type the `resourceType` param to it, so a typo becomes a compile error (preferable to matching physical-ID-only, which risks collisions since CDK derives bucket/role/param names from shared prefixes).

scripts/components/e2e-cleanup/stack_deleter.ts
```diff
@@ -0,0 +1,284 @@
...
+          new ListStacksCommand({
+            NextToken: nextToken,
+            StackStatusFilter: Object.keys(StackStatus).filter(
+              (status) => status != StackStatus.DELETE_COMPLETE,
+            ) as Array<StackStatus>,
+          }),
+        );
+      nextToken = listStacksResponse.NextToken;
+      stackSummaries.push(...(listStacksResponse.StackSummaries ?? []));
+    } while (nextToken);
```
Simone319, line 462: **Important (test):** The mocks never return a `NextToken`, so the pagination loops here and at L232–244 never take a second iteration — a truncated `ListStacks`/`ListStackResources` page would silently drop resources from the ownership index, leaving those at-risk resources unprotected, and no test would catch it. Please add a test where the paginated call returns a `NextToken` on the first page so the second-page merge is exercised.

scripts/components/e2e-cleanup/stack_deleter.ts
sarayev: Fixed. The `listStackResources` call in the `DELETE_FAILED` branch is now wrapped in try/catch and falls through to a plain `DeleteStack` on throw, so a stack is never silently skipped. Covered by `still requests the deletion when the resources of a DELETE_FAILED stack cannot be listed`.

scripts/cleanup_e2e_resources.ts
sarayev: Fixed. `GuardedResourceType` is now an exported union, `isOwnedByLiveStack`'s `resourceType` param is typed to it, and all six call sites updated — a typo'd resource type is a compile error instead of a silent no-match.

scripts/components/e2e-cleanup/stack_deleter.ts
sarayev: Fixed. Added `NextToken` pagination tests for both `ListStacks` and `ListStackResources`, each asserting the second page's contents reach the index.

scripts/components/e2e-cleanup/stack_deleter.ts
sarayev: Fixed in 25fff12847. Since S3 buckets and IAM roles are listed account wide, the ownership index is now built across every region in the workflow matrix before any sweep runs. Region-scoped types (Cognito, DynamoDB, SSM, log groups) stay scoped to the current region on purpose — their names are only unique within a region, so checking them against another region would keep a genuinely stale resource from ever being cleaned up. `e2e_test_regions.test.ts` parses the workflow and fails if the region list drifts from the matrix.

scripts/components/e2e-cleanup/stack_deleter.ts
sarayev: Added in 25fff12847 — `getResourcesOwnedByLiveStacks` and the new nested-stack walk both inspect stacks 8 at a time, which matters more now that the index covers four regions. `retryMode: 'adaptive'` and the `deletionRequested` guard are unchanged. Covered by `indexes every stack even when many of them are inspected at once`.

scripts/components/e2e-cleanup/stack_deleter.ts
sarayev: Fixed in 25fff12847, and the read-only investigation went a level deeper than the root: every stuck root blocks on its nested hosting stack, and that nested stack blocks **only** on `AWS::S3::Bucket` — a versioned bucket CloudFormation cannot delete because it is not empty (`409 ... You must delete all versions in the bucket`). Invisible from the root, which is why the drain never saw it.

So instead of two-step root-last, the fix is "unblock, do not detach": `findBucketsBlockingStackDeletion` walks nested stacks, empties the blocking buckets, and only then retries the plain `DeleteStack` on the root. CloudFormation still owns the bucket, so its own delete succeeds and both stacks reach `DELETE_COMPLETE` — no detached nested stack, no orphan, and `AWS::CloudFormation::Stack` stays out of the retain allowlist. A nested failure that emptying cannot fix (a distribution, say) is logged for a human, since `deleteStack` only ever sees the nested stack resource itself.

Simone319 (approved):

sarayev:
### `test_with_baseline_dependencies` failure is a pre-existing flake, not from this PR

Flagging this so the red check isn't attributed to this change.

**The failure** ([job](https://github.com/aws-amplify/amplify-backend/actions/runs/31602579252/job/94134506903), step _Run unit and integration tests_) is a single test out of 2013:

```
tests 2013 | pass 2012 | fail 1

packages/backend-function/lib/function_env_type_generator.test.js:47
  "generated type definition file has valid syntax"
  Error: ENOENT: no such file or directory,
    open '<root>/.amplify/generated/env/testFunction.ts'
```

**Root cause: a cross-file test-isolation race on a shared directory.**

That test writes the shim with `generateTypedProcessEnvShim()` and then `await import(filePath)`. Five test files each wipe the *same repo-root* directory in an `after()` hook:

```ts
await fsp.rm(path.join(process.cwd(), '.amplify'), { recursive: true, force: true });
```

- `packages/backend-function/src/function_env_type_generator.test.ts`
- `packages/backend-function/src/function_env_translator.test.ts`
- `packages/backend-function/src/factory.test.ts`
- `packages/backend-function/src/layer_parser.test.ts`
- `packages/integration-tests/src/test-in-memory/data_storage_auth_with_triggers.test.ts`

`tsx --test` runs test *files* in parallel from the repo root, so a sibling file's cleanup can delete `<root>/.amplify` in the window between the write and the import — hence the intermittent ENOENT.

**Why this PR can't be the cause:**

1. This PR only touches `scripts/**` plus one root `devDependencies` entry. It changes no package source and no test in `packages/`.
2. `scripts/set_baseline_dependency_versions.ts` only pins `aws-cdk-lib` (`targetDependencyNames = new Set(['aws-cdk-lib'])`) and only iterates `glob('packages/*')` — it never reads the root `package.json`, so the dependency added here isn't involved in baseline pinning at all.
3. Most conclusively: the **same package test suite passed in all 9 `test_with_coverage` matrix jobs on this exact commit** (`25fff12`, ubuntu/macos/windows × node 22/24/26). Only the baseline-deps run, which happened to lose the race, went red.

I'll re-run the job once the overall workflow finishes. Happy to send the isolation fix (give each of those suites its own temp dir instead of sharing `<root>/.amplify`) as a separate PR — it's unrelated to this one, so I've kept it out of scope here.

sarayev:
### Status of the two remaining red/stuck checks — both reproduce on `main` without this PR

Following up on my previous comment, here's the full accounting. **Every** non-passing check on this PR is either advisory or independently reproducible on `main` at this branch's base commit.

---

#### 1. `test_scripts` — hanging (not failing), pre-existing

This job has been `in_progress` for ~2h and will eventually be `cancelled` when it hits the 6h job cap (the job has no `timeout-minutes`).

**It is not this PR.** The same job hangs on `main` at `8a0c918` — this branch's base commit, which contains none of this PR's code:

| `main` run | `test_scripts` |
|---|---|
| [31595921662](https://github.com/aws-amplify/amplify-backend/actions/runs/31595921662) | `in_progress` (stuck) |
| [31587240316](https://github.com/aws-amplify/amplify-backend/actions/runs/31587240316) | `in_progress` (stuck) |

In both runs every other job reached a terminal state — `test_scripts` is the only one stuck. The 2-hourly `main` cron also shows four consecutive `conclusion: cancelled` runs (`31578738758`, `31570374008`, `31563762352`, `31558402053`), which is the signature of jobs burning the 6h cap rather than failing.

Locally the full suite is green in 40s, so nothing in this PR loops:

```
ℹ tests 94   ℹ pass 94   ℹ fail 0   ℹ duration_ms 39355
```

**Likely external trigger.** `@changesets/cli@3.0.0` became the `latest` dist-tag on **2026-08-11T12:09:29.967Z**, shortly before `main` started hanging. Two script tests shell out to changesets through *unpinned* `npx` inside a fresh `mkdtemp` directory:

- `scripts/components/release_lifecycle.test.ts:108`
- `scripts/components/dependabot_version_update_handler.test.ts:91`

```ts
await $`npx --package @changesets/cli -- changeset init`;
```

Because the cwd is a temp dir with no `node_modules`, the root `devDependencies` range `^2.26.1` is bypassed entirely and `npx` resolves the registry's `latest` (now `3.0.0`).

One caveat so nobody burns time on the obvious patch: **pinning that invocation to `@changesets/cli@^2.26.1` alone does not fix it.** I verified both ranges in a clean temp dir and they fail identically, because the ESM-only `human-id` is pulled transitively either way:

```
Error [ERR_REQUIRE_ESM]: require() of ES Module .../human-id/dist/index.js
  from .../@changesets/write/dist/changesets-write.cjs.js not supported.
```

On Node 18 that throws immediately; on Node 22 (this job's version) `require(ESM)` is permitted, so it proceeds past that point and then blocks — consistent with a hang instead of a crash. I couldn't read the CI log to confirm the exact blocking line because the run is still in progress, so treat the precise hang site as inferred rather than proven. Either way it's in code this PR doesn't touch, and it needs a repo-level fix (pin the transitive resolution and/or add `timeout-minutes` so this surfaces as a failure in minutes instead of a 6h silent stall). Happy to open that separately.

---

#### 2. `e2e_package_manager (windows-2025, npm, 22)` — pre-existing

Also fails on `main` run [31587240316](https://github.com/aws-amplify/amplify-backend/actions/runs/31587240316), alongside `(windows-2025, yarn-classic, 22)`. Not related to this change.

---

#### 3. `check_pr_size` — advisory

Not a required check (confirmed via the branch-protection API); expected for this changeset.

---

**Summary:** `lint`, `check_api_extract`, `check_api_changes`, all 9 `test_with_coverage` jobs, and effectively all `e2e_*` jobs are green on `25fff12`. I'm deliberately not force-pushing or amending, since none of the above is caused by this branch and a re-push would just restart a ~2h queue. Ready for review.

PR #3304, opened 2026-08-11
fix(e2e-cleanup): delete every object version when emptying a test bucket

sarayev:
> Stacked on #3303 — review that first. This PR targets `fix/e2e-cleanup-live-stack-ownership`, so its diff only shows the S3 changes.

## Problem

Test buckets are versioned, and the previous emptying logic could not reliably empty one. `DeleteBucket` then failed with `BucketNotEmpty` and the bucket leaked. Three separate defects:

**1. Version pagination was broken.**

A single `nextToken` variable was shared across two sequential `do/while` loops:

```ts
let nextToken: string | undefined = undefined;
do { /* ListObjectsV2 ... */ nextToken = listObjectsResponse.NextContinuationToken; } while (nextToken);

do {
  const listVersionsResponse = await s3Client.send(
    new ListObjectVersionsCommand({ Bucket: bucketName, KeyMarker: nextToken }),
  );
  ...
  nextToken = listVersionsResponse.NextKeyMarker;
} while (nextToken);
```

The versions loop started from the leftover value of the objects loop (always `undefined` at that point — the first loop only exits when it is falsy), and it only ever advanced `NextKeyMarker`. `ListObjectVersions` pages on **both** `KeyMarker` and `VersionIdMarker`, so a key with more versions than fit in one page had the rest of its versions silently skipped.

Pagination was also driven by "is there a marker" rather than by `IsTruncated`, which is the documented termination signal.

**2. `DeleteObjects` failures were discarded.**

`DeleteObjects` is a partial-success API — it reports per-object failures in `Errors[]` in a `200` response instead of throwing. The response was never inspected, so a bucket that could not be emptied looked like a clean success and the real reason for the subsequent `BucketNotEmpty` was lost.

**3. Requests were not batched.**

`DeleteObjects` accepts at most 1000 keys. A `ListObjectVersions` page holding more than that (versions + delete markers combined) was rejected outright.

## Fix

`S3BucketEmptier` (new, `scripts/components/e2e-cleanup/s3_bucket_emptier.ts`):

- Paginates on **both** `NextKeyMarker` and `NextVersionIdMarker`, terminating on `IsTruncated`.
- Deletes versions **and** delete markers, batched at 1000.
- Throws when the response reports `Errors[]`, including the first few object keys and codes so the log says *why*.
- Retries `DeleteBucket` once after emptying again on `BucketNotEmpty`, because bucket emptiness is eventually consistent and a concurrent test may still be writing. Any other failure propagates unchanged.

The separate `ListObjectsV2` pass is gone: `ListObjectVersions` already returns the current version of every object, so it was duplicated work and doubled the request count on every bucket.

## Tests

6 new unit tests in `s3_bucket_emptier.test.ts`: full version + delete-marker removal, pagination across both markers, 1000-item batching, throwing on reported `Errors[]`, the `BucketNotEmpty` re-empty-and-retry path, and that other failures are *not* retried.

```
ℹ tests 75
ℹ pass 75
ℹ fail 0
```

`tsc --noEmit -p scripts/tsconfig.json`, `eslint --max-warnings 0`, and `prettier --check` are all clean.

## Notes for the reviewer

- No new IAM permissions. The same `s3:ListBucketVersions` / `s3:DeleteObject` / `s3:DeleteBucket` actions are used; only the call pattern changed.
- No `packages/*` code is touched, so there is no changeset.

## Validation

- [x] Unit tests pass
- [x] `tsc`, `eslint`, `prettier` clean
- [x] PR size check passes

---

### Review stack

This fix is split into three stacked PRs to stay under the repo's 1000-line PR size limit. Each is independently correct and reviewable:

1. **#3303 — stop deleting resources that live stacks still own** (base, targets `main`) — the primary bug, plus draining the existing `DELETE_FAILED` backlog.
2. **#3304 — delete every object version when emptying a test bucket** (targets #3303's branch).
3. **#3305 — delete the CloudFront distributions of test buckets** (targets #3304's branch) — the subdomain-takeover risk.

Merge in order 3303 → 3304 → 3305.

> **Why "no checks reported" on this PR:** `health_checks.yml` only triggers on PRs targeting `main`, `hotfix`, or `feature/**`. This PR targets its parent's `fix/...` branch, so CI is idle by design. GitHub auto-retargets this PR to `main` when its parent merges, and the full suite runs at that point. #3303 (the base of the stack, targeting `main`) is fully green.
>
> Verified locally on this exact branch in the meantime:
> ```
> tsc --noEmit -p scripts/tsconfig.json   clean
> eslint --max-warnings 0                 clean
> prettier --check                        clean
> npm run test:scripts                    75 pass / 0 fail
> npm run diff:check <parent>             within size limits
> ```

### DIFF
--- a/scripts/cleanup_e2e_resources.ts
+++ b/scripts/cleanup_e2e_resources.ts
@@ -9,17 +9,7 @@ import {
   DescribeLogGroupsCommandOutput,
   LogGroup,
 } from '@aws-sdk/client-cloudwatch-logs';
-import {
-  Bucket,
-  DeleteBucketCommand,
-  DeleteObjectsCommand,
-  ListBucketsCommand,
-  ListObjectVersionsCommand,
-  ListObjectsV2Command,
-  ListObjectsV2CommandOutput,
-  ObjectIdentifier,
-  S3Client,
-} from '@aws-sdk/client-s3';
+import { Bucket, ListBucketsCommand, S3Client } from '@aws-sdk/client-s3';
 import {
   CognitoIdentityProviderClient,
   DeleteUserPoolCommand,
@@ -68,6 +58,7 @@ import {
   ListTablesCommandOutput,
   TableDescription,
 } from '@aws-sdk/client-dynamodb';
+import { S3BucketEmptier } from './components/e2e-cleanup/s3_bucket_emptier.js';
 import {
   GuardedResourceType,
   StackDeleter,
@@ -151,6 +142,7 @@ const stackDeleter = new StackDeleter(
   TEST_AMPLIFY_RESOURCE_PREFIX,
   isStackStale,
 );
+const s3BucketEmptier = new S3BucketEmptier(s3Client);
 
 /**
  * Deleting a resource that a stack still owns poisons the delete path of that stack, which is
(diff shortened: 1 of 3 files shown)


Simone319:
Have we triggered the e2e tests to ensure it actually works? Plus, look like we're not merging to main.

sarayev:
Closing in favour of **#3303**, which now contains this change.

`check_pr_size` turned out **not** to be a required status check on `main` — it is advisory and does not block merge (precedent: #3265 merged at 15,994 additions with the same check red). The 1000-line limit was the only reason this was split into a three-PR stack, so I consolidated all three components back into #3303 to keep one reviewable narrative instead of three partial ones.

No work is lost: #3303's branch was fast-forwarded to this stack's tip, so every commit here is already in it, unchanged. Review continues on #3303.

PR #3305, opened 2026-08-11
fix(e2e-cleanup): delete the CloudFront distributions of test buckets

sarayev:
> Stacked on #3304 (which is stacked on #3303) — review those first. This PR targets `fix/e2e-cleanup-s3-object-versions`, so its diff only shows the CloudFront changes.

## Problem

**Nothing in the cleanup job ever deleted CloudFront distributions.**

Hosting e2e tests create them. When a stack fails to delete, the distribution is left behind. The bucket sweep then deletes the origin bucket **out from under it**.

That leaves a running distribution whose S3 origin points at a bucket name that no longer exists. S3 bucket names are globally claimable, so **anyone can create that bucket and serve their own content from the distribution's `*.cloudfront.net` domain** — a subdomain-takeover primitive on an AWS-owned domain. The distributions also keep costing money indefinitely and are never reclaimed.

The ordering is exactly backwards: today the bucket goes first and the distribution is orphaned.

## Fix

`CloudFrontDistributionCleaner` (new, `scripts/components/e2e-cleanup/cloudfront_distribution_cleaner.ts`):

**1. Index distributions by their origin bucket**

`buildBucketToDistributionsIndex()` pages `ListDistributions` once per run and maps `bucketName -> DistributionSummary[]`, parsing the three S3 origin domain formats CloudFront reports:

```
bucket.s3.<region>.amazonaws.com
bucket.s3.amazonaws.com
bucket.s3-website-<region>.amazonaws.com
```

Only buckets matching the `amplify-` test prefix are indexed, so a non-test distribution is never touched.

**2. Converge across runs, not within one**

A distribution cannot be deleted while it is enabled, and disabling one takes tens of minutes to propagate. Blocking the hourly job on that is not viable, so reaping is idempotent and spread across runs:

| Distribution state | Action | Result |
|---|---|---|
| `Enabled` | `UpdateDistribution` with `Enabled: false` | `disable-requested` |
| disabled, `Status != Deployed` | wait | `disable-requested` |
| disabled, `Status == Deployed` | `DeleteDistribution` | `deleted` |
| no distributions for bucket | — | `none` |

**3. Reverse the ordering — bucket last**

The bucket sweep now reaps distributions *first* and **retains the bucket** unless every distribution using it is gone. This closes the takeover window: the origin bucket outlives the distribution, not the other way round.

## Graceful degradation (important)

CloudFront reaping is **inert until the IAM role is updated**, and is designed so that landing this ahead of the role change is safe:

- `ListDistributions` failure (e.g. `AccessDenied`) is logged and returns an **empty index** rather than throwing. Every bucket then reaps `none` and the rest of the cleanup behaves exactly as before.
- A per-distribution failure is treated as **still present** (`disable-requested`), so the caller retains the bucket rather than orphaning the distribution. Failing to reap never escalates into the very problem this fixes.

### IAM prerequisite (out of repo)

The `e2e-resource-cleanup` role needs:

```
cloudfront:ListDistributions
cloudfront:GetDistribution
cloudfront:GetDistributionConfig
cloudfront:UpdateDistribution
cloudfront:DeleteDistribution
cloudfront:TagResource
```

(plus, from #3303: `cloudformation:DescribeStacks`, `cloudformation:DescribeStackResources`, `cloudformation:ListStackResources`)

Until that lands this PR is a no-op beyond one extra logged `ListDistributions` attempt per run.

## Tests

9 new unit tests in `cloudfront_distribution_cleaner.test.ts`: all three origin domain formats, non-test origins ignored, `ListDistributions` pagination, the `AccessDenied` empty-index path, and each of the four reap outcomes including "one of two distributions still present ⇒ retain bucket" and "reap failed ⇒ retain bucket".

```
ℹ tests 75
ℹ pass 75
ℹ fail 0
```

`tsc --noEmit -p scripts/tsconfig.json`, `eslint --max-warnings 0`, and `prettier --check` are all clean.

## Notes for the reviewer

- Adds `@aws-sdk/client-cloudfront` to root `devDependencies`, pinned to `^3.936.0` to match every sibling `@aws-sdk/client-*` declaration.
- No `packages/*` code is touched, so there is no changeset.
- Draining the existing `DELETE_FAILED` backlog happens in #3303; this PR stops new orphans from being created.

## Validation

- [x] Unit tests pass
- [x] `tsc`, `eslint`, `prettier` clean
- [x] PR size check passes

---

### Review stack

This fix is split into three stacked PRs to stay under the repo's 1000-line PR size limit. Each is independently correct and reviewable:

1. **#3303 — stop deleting resources that live stacks still own** (base, targets `main`) — the primary bug, plus draining the existing `DELETE_FAILED` backlog.
2. **#3304 — delete every object version when emptying a test bucket** (targets #3303's branch).
3. **#3305 — delete the CloudFront distributions of test buckets** (targets #3304's branch) — the subdomain-takeover risk.

Merge in order 3303 → 3304 → 3305.

> **Why "no checks reported" on this PR:** `health_checks.yml` only triggers on PRs targeting `main`, `hotfix`, or `feature/**`. This PR targets its parent's `fix/...` branch, so CI is idle by design. GitHub auto-retargets this PR to `main` when its parent merges, and the full suite runs at that point. #3303 (the base of the stack, targeting `main`) is fully green.
>
> Verified locally on this exact branch in the meantime:
> ```
> tsc --noEmit -p scripts/tsconfig.json   clean
> eslint --max-warnings 0                 clean
> prettier --check                        clean
> npm run test:scripts                    75 pass / 0 fail
> npm run diff:check <parent>             within size limits
> ```

### DIFF
--- a/.eslint_dictionary.json
+++ b/.eslint_dictionary.json
@@ -48,6 +48,7 @@
   "chown",
   "claude",
   "cloudformation",
+  "cloudfront",
   "codebase",
   "codegen",
   "cognito",
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,6 +17,7 @@
         "@aws-amplify/eslint-plugin-amplify-backend-rules": "^0.0.2",
         "@aws-sdk/client-amplify": "^3.936.0",
         "@aws-sdk/client-cloudformation": "^3.936.0",
+        "@aws-sdk/client-cloudfront": "^3.936.0",
         "@aws-sdk/client-cloudwatch-logs": "^3.936.0",
         "@aws-sdk/client-cognito-identity-provider": "^3.936.0",
         "@aws-sdk/client-dynamodb": "^3.936.0",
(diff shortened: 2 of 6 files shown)


sarayev:
Closing in favour of **#3303**, which now contains this change.

`check_pr_size` turned out **not** to be a required status check on `main` — it is advisory and does not block merge (precedent: #3265 merged at 15,994 additions with the same check red). The 1000-line limit was the only reason this was split into a three-PR stack, so I consolidated all three components back into #3303 to keep one reviewable narrative instead of three partial ones.

No work is lost: #3303's branch was fast-forwarded to this stack's tip, so every commit here is already in it, unchanged. Review continues on #3303.

PR #3307, opened 2026-08-13
fix(ai-constructs): strip extra Bedrock tool use fields before sending to AppSync

### LINKED ISSUE #3302: responseComponents tool use fails with "field not defined for AmplifyAIToolUseBlockInput"
### Environment information

```plain text
System:
  OS: macOS 26.6
  CPU: (12) arm64 Apple M3 Pro
  Memory: 195.22 MB / 36.00 GB
  Shell: /bin/zsh
Binaries:
  Node: 22.22.2 - /opt/homebrew/bin/node
  Yarn: undefined - undefined
  npm: 10.9.7 - /opt/homebrew/bin/npm
  pnpm: undefined - undefined
NPM Packages:
  @aws-amplify/ai-constructs: 1.6.2
  @aws-amplify/auth-construct: 1.9.0
  @aws-amplify/backend: 1.18.0
  @aws-amplify/backend-ai: 1.5.4
  @aws-amplify/backend-auth: 1.8.0
  @aws-amplify/backend-cli: 1.8.0
  @aws-amplify/backend-data: 1.6.2
  @aws-amplify/backend-deployer: 2.1.4
  @aws-amplify/backend-function: 1.15.1
  @aws-amplify/backend-output-schemas: 1.8.0
  @aws-amplify/backend-output-storage: 1.3.5
  @aws-amplify/backend-secret: 1.4.1
  @aws-amplify/backend-storage: 1.4.2
  @aws-amplify/cli-core: 2.2.2
  @aws-amplify/client-config: 1.9.0
  @aws-amplify/data-construct: 1.16.3
  @aws-amplify/data-schema: 1.20.5
  @aws-amplify/deployed-backend-client: 1.8.1
  @aws-amplify/form-generator: 1.2.5
  @aws-amplify/model-generator: 1.2.1
  @aws-amplify/platform-core: 1.11.1
  @aws-amplify/plugin-types: 1.12.2
  @aws-amplify/sandbox: 2.1.3
  @aws-amplify/schema-generator: 1.4.0
  @aws-cdk/toolkit-lib: 1.6.1
  aws-amplify: 6.14.4
  aws-cdk-lib: 2.263.0
  typescript: 5.8.3
```

## Bug description

When using `responseComponents` on `AIConversation`, any LLM tool call causes the conversation handler Lambda to fail with an AppSync validation error. No response is delivered to the client and the streaming turn silently errors out.

## Error

```
Variable '$input' got invalid value ... at 'input.accumulatedTurnContent[0].toolUse';
Field 'type' is not defined by type 'AmplifyAIToolUseBlockInput'
```

## Steps to reproduce

1. Define an `a.conversation()` and use `responseComponents` on `AIConversation` in the frontend
2. Prompt the LLM to invoke the component tool
3. Observe the Lambda error in CloudWatch — the streaming turn never completes

## Root cause

`@aws-sdk/client-bedrock-runtime` defines `ToolUseBlockStart` with an optional `type?: ToolUseType` field. In `ConversationTurnResponseSender.serializeContent`, the block is spread verbatim before being sent to AppSync:

```ts
// packages/ai-constructs/src/conversation/runtime/conversation_turn_response_sender.ts
const input = JSON.stringify(block.toolUse.input);
return { toolUse: { ...block.toolUse, input } };  // spreads `type` into the mutation
```

The AppSync-generated `AmplifyAIToolUseBlockInput` only accepts `{ toolUseId, name, input }` — there is no `type` field. AppSync rejects the mutation.

## Proposed fix

In `packages/ai-constructs/src/conversation/runtime/conversation_turn_response_sender.ts`, explicitly pick only the fields `AmplifyAIToolUseBlockInput` accepts instead of spreading the full Bedrock SDK object:

```ts
const input = JSON.stringify(block.toolUse.input);
const { toolUseId, name } = block.toolUse;
return { toolUse: { toolUseId, name, input } };
```

## Affected versions

| Package | Version |
|---|---|
| `@aws-amplify/ai-constructs` | 1.6.2 |
| `@aws-sdk/client-bedrock-runtime` | 3.1104.0 |

sharonyajain:
## Problem

Fixes #3302.

When using `responseComponents` on `AIConversation`, any LLM tool call makes the conversation handler Lambda fail with an AppSync validation error and the streaming turn silently errors out with no response delivered to the client:

```
Variable '$input' got invalid value ... at 'input.accumulatedTurnContent[0].toolUse';
Field 'type' is not defined by type 'AmplifyAIToolUseBlockInput'
```

## Root cause

`ConversationTurnResponseSender.serializeContent` spread the entire Bedrock SDK `ToolUseBlock` into the AppSync mutation:

```ts
const input = JSON.stringify(block.toolUse.input);
return { toolUse: { ...block.toolUse, input } };
```

Newer `@aws-sdk/client-bedrock-runtime` versions add extra fields to the tool use block (e.g. `type`), which flow through the streaming accumulator in `bedrock_converse_adapter.ts` and this spread into the mutation. The AppSync-generated `AmplifyAIToolUseBlockInput` only accepts `{ toolUseId, name, input }`, so AppSync rejects the mutation. `serializeContent` is the single choke point for both the streaming and non-streaming response paths.

## Fix

Explicitly pick only the fields `AmplifyAIToolUseBlockInput` accepts instead of spreading the whole SDK object. This is also defensive against future SDK field additions.

```ts
const input = JSON.stringify(block.toolUse.input);
const { toolUseId, name } = block.toolUse;
return { toolUse: { toolUseId, name, input } };
```

## Tests

Added two regression tests (non-streaming + streaming) in `conversation_turn_response_sender.test.ts` that attach an extra `type` field to the tool use block and assert it is stripped from the mutation. Verified they **fail** against the old spread implementation and **pass** with the fix. All 7 tests in the file pass; `tsc --build` and `prettier` are clean.

## Checklist

- [x] Changeset added (`@aws-amplify/ai-constructs`, patch)
- [x] Unit tests added and passing

### DIFF
--- a/packages/ai-constructs/src/conversation/runtime/conversation_turn_response_sender.test.ts
+++ b/packages/ai-constructs/src/conversation/runtime/conversation_turn_response_sender.test.ts
@@ -300,6 +300,118 @@ void describe('Conversation turn response sender', () => {
     });
   });
 
+  void it('does not forward extra tool use fields not accepted by AppSync', async () => {
+    const userAgentProvider = new UserAgentProvider(
+      {} as unknown as ConversationTurnEvent,
+    );
+    mock.method(userAgentProvider, 'getUserAgent', () => '');
+    const graphqlRequestExecutor = new GraphqlRequestExecutor(
+      '',
+      '',
+      userAgentProvider,
+    );
+    const executeGraphqlMock = mock.method(
+      graphqlRequestExecutor,
+      'executeGraphql',
+      () =>
+        // Mock successful Appsync response
+        Promise.resolve(),
+    );
+    const sender = new ConversationTurnResponseSender(
+      event,
+      userAgentProvider,
+      graphqlRequestExecutor,
+    );
+    // Newer `@aws-sdk/client-bedrock-runtime` versions add extra fields (e.g. `type`)
+    // to the tool use block. These are not part of the AppSync-generated
+    // `AmplifyAIToolUseBlockInput` and must not be forwarded, otherwise the mutation
+    // is rejected with "Field 'type' is not defined by type 'AmplifyAIToolUseBlockInput'".
+    const toolUseBlock = {
+      toolUse: {
+        name: 'testTool',
+        toolUseId: 'testToolUseId',
+        input: {
+          testPropertyKey: 'testPropertyValue',
+        },
+        type: 'tool_use',
+      },
+    } as unknown as ContentBlock.ToolUseMember;
+    const response: Array<ContentBlock> = [toolUseBlock];
+    await sender.sendResponse(response);
+
+    assert.strictEqual(executeGraphqlMock.mock.calls.length, 1);
+    const request = executeGraphqlMock.mock.calls[0]
+      .arguments[0] as GraphqlRequest<MutationResponseInput>;
+    assert.deepStrictEqual(request.variables.input.content, [
+      {
+        toolUse: {
+          input: JSON.stringify(toolUseBlock.toolUse.input),
+          name: toolUseBlock.toolUse.name,
+          toolUseId: toolUseBlock.toolUse.toolUseId,
+        },
+      },
+    ]);
+  });
+
+  void it('does not forward extra tool use fields not accepted by AppSync when streaming', async () => {
+    const userAgentProvider = new UserAgentProvider(
+      {} as unknown as ConversationTurnEvent,
+    );
+    mock.method(userAgentProvider, 'getUserAgent', () => '');
+    const graphqlRequestExecutor = new GraphqlRequestExecutor(
+      '',
+      '',
+      userAgentProvider,
+    );
+    const executeGraphqlMock = mock.method(
+      graphqlRequestExecutor,
+      'executeGraphql',
+      () =>
+        // Mock successful Appsync response
+        Promise.resolve(),
+    );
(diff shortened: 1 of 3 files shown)


sarayev (approved):

packages/ai-constructs/src/conversation/runtime/conversation_turn_response_sender.test.ts
```diff
@@ -300,6 +300,118 @@ void describe('Conversation turn response sender', () => {
...
+    const sender = new ConversationTurnResponseSender(
+      event,
+      userAgentProvider,
+      graphqlRequestExecutor,
+    );
+    // Newer `@aws-sdk/client-bedrock-runtime` versions add extra fields (e.g. `type`)
+    // to the tool use block. These are not part of the AppSync-generated
+    // `AmplifyAIToolUseBlockInput` and must not be forwarded, otherwise the mutation
+    // is rejected with "Field 'type' is not defined by type 'AmplifyAIToolUseBlockInput'".
+    const toolUseBlock = {
```
sarayev, line 329: nit (non-blocking): this setup — the mock/executor wiring, sender construction, and the `toolUseBlock` literal — is duplicated almost verbatim with the streaming test just below. A small shared helper or a `beforeEach` would keep them tighter. Fine to merge as-is.

PR #3309, opened 2026-08-17
fix(backend-notifications): require Identity Resolution to be disabled on the Customer Profiles domain used by defineNotifications

soberm:
## Summary

`defineNotifications` relies on a one-to-one mapping between the caller principal (the Cognito Identity Pool `principalId`) and its Amazon Connect Customer Profiles profile. Amazon Connect Identity Resolution (profile merging) combines profiles across principals, which does not preserve that mapping. This change ensures `defineNotifications` only operates against a Customer Profiles domain with Identity Resolution disabled.

## Behavior

- **Create-from-scratch mode:** the Customer Profiles domain is provisioned with automatic (`Matching`) and rule-based (`RuleBasedMatching`) matching disabled.
- **Attach mode (deploy time):** the target domain is validated during deployment; the deployment fails with an actionable message if Identity Resolution is enabled.
- **Runtime:** the `identify-user` API verifies that the attached domain has Identity Resolution disabled and returns an error when it is enabled, using a short-lived per-domain cache and failing closed when the domain's configuration cannot be confirmed.

## Impact (breaking)

A deployment that attaches to — or a running app whose attached domain has — Identity Resolution enabled will be rejected. Attach `defineNotifications` to a dedicated Customer Profiles domain with Identity Resolution disabled, or disable Identity Resolution on the domain.

## Testing

- Unit tests (319 passing): create-mode matching disabled; deploy-time validation allow/deny (including rule-based `PENDING`); runtime allow/deny, cache hit/miss/expiry, fail-closed and serve-stale-within-grace; least-privilege `profile:GetDomain` IAM.
- Manual deploy validation: attach to a non-merging domain succeeds; attach to a merging domain fails the deployment; enabling Identity Resolution after deploy causes `identify-user` writes to be rejected without a redeploy; disabling restores writes.


packages/backend-notifications/src/lambda/identity-resolution/check.ts
```diff
@@ -0,0 +1,118 @@
...
+ * being false (or the whole block being absent) is a pass.
+ *
+ * Pure and side-effect free: the deploy-time guard handler calls this on a live
+ * `GetDomain` response, and the unit tests call it on literal fixtures.
+ * @param domain The `GetDomain` response (or any object with the same shape).
+ * @returns `{ merging: false }` when the domain is safe to attach to, otherwise
+ * the enabled mechanism and its status.
+ */
+export const evaluateMerging = (domain: DomainMatchingView): MergingVerdict => {
+  if (domain.Matching?.Enabled === true) {
```
sarayev, line 61: **[major]** Two edge cases worth covering here since this is the security-critical predicate:

- It only inspects `Matching.Enabled === true`, but the API can also carry `Matching.AutoMerging.Enabled`. It's likely safe because `AutoMerging` requires `Matching.Enabled: true`, but a defensive `if (domain.Matching?.AutoMerging?.Enabled === true) return { merging: true, mechanism: 'AutoMerging' };` (or a comment citing that invariant) would save the next reader from re-deriving it.
- When `RuleBasedMatching.Enabled === true` with `Status: 'PENDING_DISABLE'`, we reject unconditionally. That's the correct fail-closed behavior, but it transiently blocks while the service finishes disabling. Surfacing `Status` in the customer-facing message ("merging is being disabled, wait and retry") would save a confusing debugging session.

packages/backend-notifications/src/lambda/api/handler.ts
```diff
@@ -72,6 +79,58 @@ const response = (
...
+ * The device routes are deliberately EXCLUDED. Both address the authoritative
+ * DynamoDB Devices table by `deviceId` and gate ownership on the SigV4-derived
+ * `principalId`; neither reads or writes Customer Profiles, so no profile merge
+ * can create, redirect, or be triggered by them. `register-device` is a
+ * principalId-keyed last-writer-wins claim, and `remove-device` must stay
+ * available: blocking de-registration would strand devices registered against a
+ * profile precisely when merging is on.
+ */
+const MERGING_GATED_ROUTES: ReadonlySet<WriteRoute> = new Set<WriteRoute>([
+  'identify-user',
```
sarayev, line 99: **[nit]** A one-liner here explaining why only `identify-user` is gated (register/remove-device don't write to Customer Profiles) would help the next reader who might wonder whether the other routes were left out by mistake.

packages/backend-notifications/src/lambda/api/handler.ts
```diff
@@ -72,6 +79,58 @@ const response = (
...
+      `[write] ${route}.refused`,
+      JSON.stringify({
+        freshness: decision.freshness,
+        mechanism: decision.verdict.mechanism,
+        status: decision.verdict.status,
+      }),
+      mergingEnabledMessage(domainName, decision.verdict),
+    );
+    return response(409, { message: MERGING_REJECTED_MESSAGE });
+  }
```
sarayev, line 134: **[minor]** 409 reads as "retry with a different payload" to most clients, but here the request itself is fine, it's the domain config that's wrong. A 403 ("server refuses to authorize this action") may signal the intent more clearly. Non-blocking, your call.

packages/backend-notifications/src/lambda/shared/merging_guard.ts
```diff
@@ -0,0 +1,252 @@
...
+  domainName: string,
+  opts: MergingGateOptions = {},
+): Promise<MergingGateDecision> => {
+  const now = opts.nowMs ?? Date.now();
+  const ttlMs = opts.ttlMs ?? resolveTtlMs();
+  const staleGraceMs = opts.staleGraceMs ?? MERGING_CHECK_STALE_GRACE_MS;
+
+  const cached = cache.get(domainName);
+  const ageMs = cached ? now - cached.storedAtMs : Number.POSITIVE_INFINITY;
+  if (cached && ageMs >= 0 && ageMs < ttlMs) {
```
sarayev, line 216: **[nit]** Nice that `ageMs >= 0` handles the clock going backwards. Could we add a test that passes `nowMs < storedAtMs` and asserts a refetch happens? That locks the behavior in so a future refactor doesn't silently drop it.

sarayev (commented):
Took a careful pass focused on the security behavior, and the core fix looks solid. `identify-user` resolves the profile entirely from the server-derived `cognitoIdentityId` (via the `PROFILE, UNIQUE` key on `principalId`), and the request body only carries attribute values, so disabling Identity Resolution cleanly closes the merge-based profile takeover. Left a few inline notes below, mostly minor.

**[major]** One question on the routing side that I couldn't anchor inline because the file isn't in this diff: `upsertDeviceOwner` in `packages/backend-notifications/src/lambda/shared/device_store.ts`. It's an unconditional last-writer-wins `UpdateItem` on the client-supplied `deviceId` key (no `ConditionExpression`), whereas `remove-device` is ownership-gated on `principalId`. So a caller who knows a victim's `deviceId` can overwrite/evict their registration, which is a denial-of-delivery on that device's routing (not interception, since `principalId` is server-derived). Since the report calls out notification routing, was device re-homing part of the reported scenario, or is this an intentional "device belongs to whoever last registered it on this install" design that's out of scope here? If it's in scope, an `attribute_not_exists(principalId) OR principalId = :caller` condition would block cross-owner overwrites, happy to talk through the tradeoff with the legit device-handoff flow.

soberm:
Confirmed intentional. This is pre-existing last-writer-wins re-homing: a device belongs to whoever most recently registers it on that install, and overwriting the item is the eviction (the docblock calls it register-or-re-home). device_store.ts is not part of this change, so I have left the semantics as-is here. It is likely to be revisited and hardened separately in the future; happy to pick up the conditional-write against the legitimate device-handoff flow at that time.

packages/backend-notifications/src/lambda/identity-resolution/check.ts
soberm: Took the defensive branch rather than just the comment. `Matching.Enabled` and `Matching.AutoMerging.Enabled` are independently optional in the response schema, so the "AutoMerging implies Matching" invariant is a service contract the type system does not enforce. It is now checked explicitly and reported as its own `AutoMerging` mechanism, with the invariant documented. Added three cases: `AutoMerging` alone, `Matching` taking precedence when both are on, and a present-but-disabled `AutoMerging` block still passing.

On the status point: `RuleBasedMatchingStatus` is `PENDING | IN_PROGRESS | ACTIVE`, there is no `PENDING_DISABLE`, so there is no transient-disable state to special-case. All three already mean "merging requested" and are refused. The status is surfaced in the deploy-time message and in the runtime log, just not in the caller-facing body (that one withholds domain and mechanism deliberately). I noted the complete status domain in the docblock.

packages/backend-notifications/src/lambda/api/handler.ts
soberm: That reasoning is in the `MERGING_GATED_ROUTES` docblock just above the set: "Only `identify-user` qualifies: it is the sole route that writes profile state", then a paragraph on why the device routes are deliberately excluded (they address DynamoDB by `deviceId` and gate on `principalId`, and blocking de-registration would strand devices). I think the diff hunk cut off mid-docblock and hid it. Left as-is to avoid duplicating it, happy to reword if it is not landing where you would look for it.

packages/backend-notifications/src/lambda/api/handler.ts
soberm: Thought about this one and kept 409, but documented why. 403 is already used a few lines up for the genuinely authorization-related case (no verified caller identity), so reusing it here would collapse two conditions needing different fixes, "fix your credentials" versus "fix the domain configuration", into one status a client cannot tell apart. 409's "conflict with the current state of the target resource" describes the domain-config conflict fairly precisely, and the 409/503 split already carries the retryability signal. Added a comment so this does not get re-litigated.

packages/backend-notifications/src/lambda/shared/merging_guard.ts
soberm: Added, and covered both guards rather than just the cache-hit one. `REFETCHES rather than trusting an entry dated in the future` primes the cache at `storedAtMs: 10_000`, calls with `nowMs: 0`, and asserts `freshness: 'fresh'` with exactly one `GetDomain`. A second test covers the `ageMs >= 0` check on the stale-grace path: same skew but with revalidation also failing, asserting `reject-unverified`, so a future-dated entry cannot be resurrected through the grace window either.

sarayev (approved):

PR #3319, opened 2026-08-31
fix(deps): bump aws-cdk-lib to ^2.267.0 to fix Windows/Node 26.8 asset fingerprint EINVAL

soberm:
## Problem

On Windows, CDK asset fingerprinting fails with `EINVAL` on Node.js >= 26.8.0. Any Amplify command that hashes assets (`ampx sandbox`, `ampx pipeline-deploy`, anything that synthesizes a stack containing an asset) throws before it can produce an asset hash, so the backend cannot be deployed at all from a Windows machine on a current Node runtime.

Root cause is in `aws-cdk-lib`, not in this repo. In `aws-cdk-lib` 2.244.0 (the version pinned here), `core/lib/fs/fingerprint.js` opens each file for hashing with the flag combination `O_DSYNC | O_RDONLY | O_SYNC`. Historically those sync flags were silently ignored on Windows. Node.js 26.8.0 started passing them through for real, and libuv rejects that combination on Windows with `EINVAL`.

The upstream fix shipped in `aws-cdk-lib` 2.254.0: `fingerprint.js` now opens with `O_RDONLY` only. Per the upstream maintainer on aws/aws-cdk#38692 the resolution for downstream consumers is purely a dependency bump — there is no source change to make here.

**Issue number, if available:** upstream aws/aws-cdk#38692

## Changes

Bump `aws-cdk-lib` from `^2.234.1` to `^2.267.0` (latest at time of writing; the fix landed in 2.254.0) across all 16 workspace packages that declare it, plus `templates/construct`. The caret-range convention already used in the repo is preserved.

- `package-lock.json` regenerated — `aws-cdk-lib` now resolves to **2.267.0** (was 2.244.0).
- `packages/create-amplify/src/default_packages.json` — regenerated via `npm run update:create-amplify-deps` so the pinned `aws-cdk-lib@2.267.0` that `create-amplify` scaffolds stays in sync with the library (required by the `check:create-amplify-deps` CI guard).
- Changeset added with a `patch` bump for the affected published packages.

Lockfile churn is small and confined to `aws-cdk-lib` and its bundled dependency set (`@aws-cdk/cloud-assembly-schema`, `aws-cdk-lib/node_modules/*`); it is a net reduction because 2.267.0 declares fewer bundled deps. No unrelated dependencies were touched.

**Corresponding docs PR, if applicable:** n/a

## Validation

This is a dependency-only change to fix a defect inside a third-party package, so there is no new runtime behavior in this repo to cover with tests. Validation was therefore aimed at (a) proving the upstream defect is actually gone in the resolved version and (b) proving the bump is non-breaking.

1. **Confirmed the upstream fix is present in the installed tree.** `node_modules/aws-cdk-lib/core/lib/fs/fingerprint.js`, `contentFingerprintMiss`:

   ```js
   const fd = fs().openSync(file, fs().constants.O_RDONLY);
   ```

   `O_RDONLY` only — the `O_SYNC`/`O_DSYNC` combination that libuv rejects on Windows is gone. Installed version confirmed as `2.267.0`.

2. **Full monorepo build is clean.** `npm run build` (`tsc --build packages/* scripts` + post-compile bundling) completes with zero TypeScript errors, so the bump introduces no type or peer-dependency breakage anywhere in the workspace.

3. **CI guard scripts pass:** `check:package-lock`, `check:package-versions`, `check:create-amplify-deps`, and `check:api` all exit 0. `check:api` reports no API surface change, i.e. no public type shifted between 2.244.0 and 2.267.0.

4. Prettier reports all changed files conform to the repo code style.

## Checklist

> Note on the last item: this PR does not modify E2E tests or SDK call sites, but it does change the CDK version used for synthesis and asset hashing, so a maintainer may want to run the checks with `run-e2e` set as a precaution.

### DIFF
--- a/.changeset/bump-aws-cdk-lib-windows-fingerprint-einval.md
+++ b/.changeset/bump-aws-cdk-lib-windows-fingerprint-einval.md
@@ -1,0 +1,18 @@
+---
+'@aws-amplify/backend-output-storage': patch
+'@aws-amplify/backend-notifications': patch
+'@aws-amplify/backend-function': patch
+'@aws-amplify/backend-deployer': patch
+'@aws-amplify/backend-storage': patch
+'@aws-amplify/backend-data': patch
+'@aws-amplify/backend-auth': patch
+'@aws-amplify/auth-construct': patch
+'@aws-amplify/ai-constructs': patch
+'@aws-amplify/platform-core': patch
+'@aws-amplify/plugin-types': patch
+'@aws-amplify/backend-cli': patch
+'@aws-amplify/backend-ai': patch
+'@aws-amplify/backend': patch
+---
+
+Bump aws-cdk-lib to ^2.267.0 to fix Windows asset-fingerprint EINVAL under Node.js >= 26.8.0 (aws/aws-cdk#38692)
--- a/.changeset/bump-aws-cdk-lib-windows-fingerprint-einval.md
+++ b/.changeset/bump-aws-cdk-lib-windows-fingerprint-einval.md
@@ -13,6 +13,7 @@
 '@aws-amplify/backend-cli': patch
 '@aws-amplify/backend-ai': patch
 '@aws-amplify/backend': patch
+'create-amplify': patch
 ---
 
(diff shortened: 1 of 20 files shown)


soberm:
Superseded by #3320 (opened from an internal branch so the E2E suite can run — fork PRs are gated out of E2E). See #3320.

PR #3326, opened 2026-09-02
fix(e2e-cleanup): fail-closed CloudFront origin guard + broaden S3 origin matching to prevent dangling distributions

sarayev:
## Problem

The e2e resource janitor (`scripts/cleanup_e2e_resources.ts`) can delete the **origin S3 bucket** of a CloudFront distribution that still exists. The result is a *dangling distribution*: a live distribution whose origin bucket name is now unclaimed, so anyone who creates a bucket with that name controls what the distribution serves.

This is the mechanism behind AWS AppSec finding `palisade.udd.cloudfront.distribution.dangling` (SIM **V2338459152**) observed in the e2e test accounts.

`CloudFrontDistributionCleaner` (added in #3303) already lists distributions first, guards origin buckets, and does a two-phase `disable → wait → delete`. That closed the common case, but two residual gaps still let the janitor delete an origin bucket:

### Gap 1 — the origin guard failed OPEN on a `ListDistributions` error

`buildBucketToDistributionsIndex` (`cloudfront_distribution_cleaner.ts` L53-85 before this change) caught any failure of `ListDistributions` and returned an **empty `Map`** (L78-84). An empty index is indistinguishable from "no bucket is used as an origin", so the bucket delete pass (`cleanup_e2e_resources.ts` L302-336 before this change) ran completely **unguarded**: every stale bucket looked origin free and was deleted, turning every distribution in the account into a dangling distribution.

Any transient CloudFront hiccup was enough to trigger it — throttling (`Rate exceeded`), a permission gap on `cloudfront:ListDistributions`, or a failure while paginating after the first page had already succeeded.

### Gap 2 — S3 origin matching missed several origin forms

`S3_ORIGIN_DOMAIN_PATTERN` (L17-18) and `getTestBucketOrigins` (L180-193) only matched virtual-hosted `*.amazonaws.com` REST / website / dualstack forms. Any other S3 origin form silently fell through, the bucket then looked origin free, and it was deleted — the same dangling outcome, just via a different path. Missed forms: legacy path-style origins (bucket in the *path*, not the host), non-standard partitions (`amazonaws.com.cn`), and any endpoint form the regex had never seen.

**Issue number, if available:** n/a (internal AWS AppSec finding `palisade.udd.cloudfront.distribution.dangling`, SIM V2338459152)

### Gap 1: fail CLOSED on an incomplete distribution index

Mirrors the existing fail-closed pattern of `LiveStackResourceIndex` in `stack_deleter.ts` (L96-130), which already protects *every* resource when the stack-ownership index could not be fully built.

- **`scripts/components/e2e-cleanup/cloudfront_distribution_cleaner.ts`**
  - New exported `BucketToDistributionsIndex` class carrying an `isComplete` flag plus `getDistributions(bucketName)` / `getBucketNames()`. `buildBucketToDistributionsIndex` now returns this instead of a bare `Map`.
  - A `ListDistributions` failure — on the first page or on any later page — yields `isComplete === false` instead of an empty `Map`, and logs that stale buckets must be retained because their origin usage is unknown.
  - `DistributionReapResult` gains `'index-incomplete'`. `reapDistributionsForBucket` returns it immediately, without making a single CloudFront call, when the index is incomplete. This is defense in depth: no caller can delete an origin bucket on the strength of an index that knows nothing.

- **`scripts/components/e2e-cleanup/stale_bucket_sweeper.ts`** (new) — the bucket delete loop, extracted from `cleanup_e2e_resources.ts` so it is unit testable (the surrounding script is a top-level-await entry point and cannot be imported). When the index is **not** complete it skips bucket deletion **entirely** for the run, retains every stale bucket, logs the reason, and sets `process.exitCode = 1` — the same "fail the job so the run is retried and the cause is visible" signal the incomplete stack-ownership index already uses (`cleanup_e2e_resources.ts` L721-732). Retaining buckets is safe and self-healing: a later healthy run deletes them.

- **`scripts/cleanup_e2e_resources.ts`** — constructs `StaleBucketSweeper` (with the existing `isOwnedByLiveStack` guard bound to `AWS::S3::Bucket`) and replaces the inline delete loop with a single `sweep(...)` call. Behavior and log strings for the healthy path are unchanged; the existing stack-index `process.exitCode` block is untouched.

### Gap 2: broaden S3 origin matching

In `cloudfront_distribution_cleaner.ts`, `getOriginBucketName` now recognizes, in order:

1. **Virtual-hosted** (regression-preserved): `bucket.s3.us-west-2.amazonaws.com`, `bucket.s3.amazonaws.com`, `bucket.s3-website-us-west-2.amazonaws.com`, `bucket.s3.dualstack.us-west-2.amazonaws.com` — plus optional `fips` and a **non-standard partition suffix** such as `bucket.s3.cn-north-1.amazonaws.com.cn`.
2. **Legacy path-style**, where the host names no bucket at all (`s3.amazonaws.com`, `s3.<region>.amazonaws.com`) and the bucket is the leading path segment. Taken from the domain name when it carries a path, otherwise from the origin's `OriginPath`.
3. **`S3OriginConfig` present** → the origin *is* an S3 origin even if its suffix is one this script has never seen, so the bucket name is read from the leading label of the host.

Genuinely non-S3 origins (`example.com`, `*.execute-api.*`, `*.elb.*`, and a bare `s3.<region>.amazonaws.com` with no bucket anywhere) still yield no bucket, so nothing new is treated as an origin bucket.

**No changeset**: this PR only touches `scripts/` and `.eslint_dictionary.json`, no released package. Verified with `changeset status --since origin/main` ("NO packages to be bumped") and `scripts/check_changeset_completeness.ts` (both exit 0).

**Corresponding docs PR, if applicable:** n/a

## Validation

All commands run from the repo root on this branch.

| Command | Result |
| --- | --- |
| `npx tsc --build scripts` | exit 0, no diagnostics |
| `npm run test:dir scripts/components/e2e-cleanup/cloudfront_distribution_cleaner.test.ts` | **12/12 pass**, 0 fail |
| `npm run test:dir scripts/components/e2e-cleanup/stale_bucket_sweeper.test.ts` | **6/6 pass**, 0 fail |
| `npm run test:scripts` (full script suite) | **109 pass, 0 fail** (24 suites) |
| `npx eslint --max-warnings 0 scripts/` | exit 0 |
| `npx tsx scripts/check_package_json.ts` | exit 0 |
| `npx prettier --check .` | "All matched files use Prettier code style!" |

`npm run lint` over the whole repo also reports one pre-existing error in `packages/backend-auth/src/reference_construct.test.ts:115` (`@typescript-eslint/restrict-template-expressions`), which reproduces on unmodified `main` and is unrelated to this PR.

New/updated unit tests (`node:test`, SDK clients mocked exactly as the existing tests do):

**Gap 1**
- `ListDistributions` throws → `index.isComplete === false`, no bucket names, reason logged.
- A failure on a *later* page (first page already succeeded) → `isComplete === false`.
- `reapDistributionsForBucket` on an incomplete index → `'index-incomplete'` with **zero** CloudFront calls.
- `StaleBucketSweeper.sweep` with an incomplete index → **retains all** buckets, `deletedBucketNames: []`, **zero** S3 and zero CloudFront calls, and the failure signal fires.
- The default signal sets `process.exitCode = 1` (asserted with save/restore).
- Healthy-path regressions: an unused bucket is emptied and deleted; a bucket with a live distribution is retained (no `DeleteBucket`); a bucket a live stack still owns is skipped without touching CloudFront; a delete failure is logged and the sweep continues to the remaining buckets.

**Gap 2**
- Bucket extracted for: REST regional, REST global, website, dualstack (regressions) + `amazonaws.com.cn`, path-style via `OriginPath`, path-style via a domain name carrying the path, and `S3OriginConfig` with an unknown suffix.
- No bucket returned for genuinely non-S3 origins (`example.com`, `execute-api`, `elb` with `CustomOriginConfig`, and a bucket-less `s3.<region>.amazonaws.com`).

`StaleBucketSweeper` tests exercise the **real** `S3BucketEmptier` and the **real** `CloudFrontDistributionCleaner`; only the AWS SDK clients are mocked.

## Follow-up / NOT addressed in this PR

These are architectural and belong to the Amplify CLI team; they are out of scope here. This PR stops the janitor from *creating* dangling distributions, but it does not reap the ones already wedged behind a stuck stack.

**GAP 3 — the janitor cannot un-wedge a `DELETE_FAILED` hosting-standalone stack, and never independently reaps its distribution**

1. **`deleteStack` retries uselessly forever.** `stack_deleter.ts` (L24-28, L300-356) can only make progress on a `DELETE_FAILED` stack by retaining resources whose type is in `RESOURCE_TYPES_SAFE_TO_RETAIN`. When the blocker is the stack's **custom-resource Lambda function**, that type is not in the list, so no `RetainResources` set can unblock the delete: every hourly run re-issues the same `DeleteStack` and fails the same way, indefinitely.
2. **The distribution is never reaped independently.** `CloudFrontDistributionCleaner` only acts on distributions matched to a stale bucket that is already **ownership-free**. While the parent stack is not `DELETE_COMPLETE` (i.e. `DELETE_FAILED`), its bucket is still owned by a live stack and is correctly skipped — so the distribution is never even considered, no matter how long the stack stays wedged. There is no independent sweep for "distribution parented to a stack that will never finish deleting".

**Root cause of the wedge** (confirmed in the affected account): the custom-resource Lambdas' **IAM execution roles were deleted** — `aws iam get-role` on the role returns `NoSuchEntity`. Without an assumable execution role the Lambda can never run, so it never sends a response to CloudFormation, and the stack delete eventually fails with *"did not receive a response from your Custom Resource"*. The e2e janitor's own IAM-role sweep is the likely deleter of those roles, which is exactly why the stack-ownership index in `stack_deleter.ts` exists and must stay fail-closed.

A durable fix needs some combination of: an independent distribution sweep keyed on the distribution's own staleness/tags rather than on an ownership-free bucket; and a way to un-wedge a custom-resource-blocked stack (e.g. recreating the execution role, or responding to the pending CloudFormation callback directly).

## Checklist

- [x] If this PR requires a change to the [Project Architecture README](../PROJECT_ARCHITECTURE.md), I have included that update in this PR. — not required, no architectural change
- [x] If this PR requires a docs update, I have linked to that docs PR above. — not required
- [x] If this PR modifies E2E tests, makes changes to resource provisioning, or makes SDK calls, I have run the PR checks with the `run-e2e` label set. — this changes the e2e *cleanup* script (not the e2e tests themselves); it is covered by unit tests and I cannot apply labels on this repo, so please add `run-e2e` if you would like a full run.

### DIFF
--- a/scripts/components/e2e-cleanup/cloudfront_distribution_cleaner.ts
+++ b/scripts/components/e2e-cleanup/cloudfront_distribution_cleaner.ts
@@ -6,16 +6,28 @@ import {
   GetDistributionConfigCommand,
   ListDistributionsCommand,
   ListDistributionsCommandOutput,
+  Origin,
   UpdateDistributionCommand,
 } from '@aws-sdk/client-cloudfront';
 
 /**
- * Matches the S3 origin domain name formats CloudFront reports, for example
- * `bucket.s3.us-west-2.amazonaws.com`, `bucket.s3.amazonaws.com` and
- * `bucket.s3-website-us-west-2.amazonaws.com`. The first group is the bucket name.
+ * Matches the virtual hosted style S3 origin domain names CloudFront reports, for example
+ * `bucket.s3.us-west-2.amazonaws.com`, `bucket.s3.amazonaws.com`,
+ * `bucket.s3-website-us-west-2.amazonaws.com`, `bucket.s3.dualstack.us-west-2.amazonaws.com`
+ * and the China partition form `bucket.s3.cn-north-1.amazonaws.com.cn`. The first group is the
+ * bucket name.
  */
-const S3_ORIGIN_DOMAIN_PATTERN =
-  /^(.+?)\.s3(?:[.-]website)?(?:\.dualstack)?(?:[.-][a-z0-9-]+)?\.amazonaws\.com$/;
+const S3_VIRTUAL_HOSTED_ORIGIN_DOMAIN_PATTERN =
+  /^(.+?)\.s3(?:[.-]website)?(?:[.-]fips)?(?:\.dualstack)?(?:[.-][a-z0-9-]+)?\.amazonaws\.com(?:\.[a-z]{2})?$/;
+
+/**
+ * Matches the legacy path style S3 origin hosts, where the host carries no bucket name at all and
+ * the bucket is the leading segment of the path, for example `s3.amazonaws.com` or
+ * `s3.us-west-2.amazonaws.com` with a `/bucket` path. The path is taken from the domain name when
+ * it carries one and from the origin path otherwise.
+ */
+const S3_PATH_STYLE_ORIGIN_HOST_PATTERN =
+  /^s3(?:[.-]website)?(?:[.-]fips)?(?:\.dualstack)?(?:[.-][a-z0-9-]+)?\.amazonaws\.com(?:\.[a-z]{2})?$/;
 
 const DEPLOYED_STATUS = 'Deployed';
 
@@ -23,9 +35,49 @@ const DEPLOYED_STATUS = 'Deployed';
  * Outcome of an attempt to reap the distributions of a single bucket.
  *
  * `disable-requested` means at least one distribution still exists, therefore the origin
- * bucket must be retained. `deleted` means every distribution of the bucket is gone.
+ * bucket must be retained. `index-incomplete` means the distributions of the bucket are unknown,
+ * therefore the origin bucket must be retained as well. `deleted` means every distribution of the
+ * bucket is gone.
  */
-export type DistributionReapResult = 'none' | 'disable-requested' | 'deleted';
+export type DistributionReapResult =
+  | 'none'
+  | 'index-incomplete'
+  | 'disable-requested'
+  | 'deleted';
+
+/**
+ * Index of the CloudFront distributions that use each test bucket as an origin.
+ *
+ * `isComplete` is false when the distributions could not be listed. A caller must not delete any
+ * bucket while the index is incomplete: an incomplete index makes every bucket look origin free,
+ * so deleting buckets would leave dangling distributions behind that serve a bucket name anybody
+ * can claim. Retaining the buckets instead is always safe because a later healthy run deletes them.
+ */
+export class BucketToDistributionsIndex {
+  /**
+   * Creates an index of distributions by the name of the test bucket they use as an origin.
+   */
+  constructor(
+    private readonly distributionsByBucketName: Map<
+      string,
+      Array<DistributionSummary>
+    >,
+    readonly isComplete: boolean,
+  ) {}
+
+  /**
+   * Returns the distributions that use the bucket as an origin, or an empty list if none do.
+   */
+  getDistributions = (bucketName: string): Array<DistributionSummary> =>
+    this.distributionsByBucketName.get(bucketName) ?? [];
+
+  /**
+   * Returns the names of the test buckets that are used as a distribution origin.
+   */
+  getBucketNames = (): Array<string> => [
+    ...this.distributionsByBucketName.keys(),
+  ];
+}
 
 /**
  * Deletes CloudFront distributions that are left behind by failed e2e test stack deletions.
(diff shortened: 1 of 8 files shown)


scripts/components/e2e-cleanup/cloudfront_distribution_cleaner.ts
```diff
@@ -47,42 +104,71 @@ export class CloudFrontDistributionCleaner {
...
+          const listDistributionsResponse: ListDistributionsCommandOutput =
+            await this.cloudFrontClient.send(
+              new ListDistributionsCommand({ Marker: marker }),
+            );
+          const distributionList: DistributionList | undefined =
+            listDistributionsResponse.DistributionList;
+          for (const distribution of distributionList?.Items ?? []) {
+            const { bucketNames, hasUnrecognizedS3Origin } =
+              this.getTestBucketOrigins(distribution);
+            if (hasUnrecognizedS3Origin) {
```
Simone319, line 133: `hasUnrecognizedS3Origin` is evaluated across every distribution in the account, so a single unrelated distribution with an S3 origin form these patterns don't recognize marks the whole index incomplete — retaining every stale bucket and failing the run on each pass until that unrelated distribution is fixed. This is safe, but it can wedge the cleanup job red indefinitely with a cause that isn't obvious from a generic failed run. Would it be worth narrowing the fail-closed trigger to origins that can't be excluded as non-test buckets, or emitting a distinct/alarmable signal so the blocked cause is actionable?

scripts/components/e2e-cleanup/cloudfront_distribution_cleaner.ts
```diff
@@ -113,7 +208,10 @@ export class CloudFrontDistributionCleaner {
   ): Promise<DistributionReapResult> => {
     const distributionId = distribution.Id;
     if (!distributionId) {
```
Simone319, line 225: A distribution summary missing its `Id` returns `index-incomplete`, which correctly retains the bucket — but unlike the global incomplete index it never calls `signalIncompleteRun()`, so a persistently unreapable distribution silently retains its bucket forever with a green run. It also overloads `index-incomplete` to mean both 'the whole index is untrustworthy' and 'this one distribution can't be reaped'. Consider a distinct result (e.g. `unreapable`) and/or signalling an incomplete run here so the condition stays visible.

scripts/components/e2e-cleanup/cloudfront_distribution_cleaner.ts
sarayev: Addressed in 22ef2b4d. I kept the account-wide guard fail-closed because an unrecognized S3 origin cannot be safely proven unrelated to an e2e bucket. The index now carries a concrete reason for each incomplete path (including the distribution ID for unrecognized origins), and the sweeper emits the stable `CloudFrontDistributionIndexIncomplete` signal with that reason before failing the run. Added coverage for both the unrecognized-origin and API-error reasons.

scripts/components/e2e-cleanup/cloudfront_distribution_cleaner.ts
sarayev: Addressed in 35c7daa4. A distribution summary without an ID now returns a distinct `unreapable` result with highest aggregate precedence. The sweeper retains the bucket, emits `CloudFrontDistributionUnreapable`, and calls `signalIncompleteRun`, so this condition can no longer remain silently green. Added focused cleaner and sweeper coverage for the result, retention, log signal, and failed-run signal.

Simone319 (approved):

Issue #3316, opened 2026-08-26
sandbox/pipeline-deploy crashes with "Cannot convert undefined or null to object" from toolkit-lib's countAssemblyResults


Millroy094:
### Environment information

```plain text
System:
  OS: macOS 26.6.2
  CPU: (10) arm64 Apple M1 Max
  Memory: 233.44 MB / 32.00 GB
  Shell: /bin/zsh
Binaries:
  Node: 24.12.0 - /Users/millroy.fernandes/.nvm/versions/node/v24.12.0/bin/node
  Yarn: undefined - undefined
  npm: 11.6.2 - /Users/millroy.fernandes/.nvm/versions/node/v24.12.0/bin/npm
  pnpm: 11.1.3 - /Users/millroy.fernandes/.nvm/versions/node/v24.12.0/bin/pnpm
NPM Packages:
  @aws-amplify/auth-construct: 1.11.2
  @aws-amplify/backend: 1.24.0
  @aws-amplify/backend-ai: Not Found
  @aws-amplify/backend-auth: 1.9.4
  @aws-amplify/backend-cli: 1.9.0
  @aws-amplify/backend-data: 1.8.0
  @aws-amplify/backend-deployer: 2.2.0
  @aws-amplify/backend-function: 1.18.2
  @aws-amplify/backend-output-storage: 1.3.5
  @aws-amplify/backend-secret: 1.4.2
  @aws-amplify/backend-storage: 1.5.0
  @aws-amplify/cli-core: 2.2.6
  @aws-amplify/client-config: 1.11.0
  @aws-amplify/data-construct: 1.17.7
  @aws-amplify/data-schema: 1.23.0
  @aws-amplify/deployed-backend-client: 1.8.2
  @aws-amplify/form-generator: 1.2.7
  @aws-amplify/model-generator: 1.2.4
  @aws-amplify/platform-core: 1.11.1
  @aws-amplify/plugin-types: 1.12.2
  @aws-amplify/sandbox: 2.3.0
  @aws-amplify/schema-generator: 1.4.1
  @aws-cdk/toolkit-lib: 1.32.0
  aws-amplify: 6.20.0
  aws-cdk-lib: 2.237.1
  typescript: 5.9.3
npm warn exec The following package was not found and will be installed: cdk@2.1138.0
CDK CLI Version: 2.1138.0 (build a4c0282)
normal verbosity, no debugging
AWS environment variables:
  AWS_EC2_METADATA_DISABLED = true
[...]

PR #3329, opened 2026-09-08
fix(backend-deployer): map toolkit-lib countAssemblyResults crash to a fault

sharonyajain:
## Problem

`npx ampx sandbox` and `ampx pipeline-deploy` crash during synth with a generic, misleading error that blames the customer's backend code:

```
[SyntaxError] Unable to build the Amplify backend definition.
 ∟ Caused by: [TypeError] Cannot convert undefined or null to object
Resolution: Check the Caused by error and fix any issues in your backend code
```

Running with `--debug` shows the crash is not in the customer's `amplify/backend.ts` at all — it originates inside `@aws-cdk/toolkit-lib`:

```
TypeError: Cannot convert undefined or null to object
    at Object.values (<anonymous>)
    at .../@aws-cdk/toolkit-lib/lib/toolkit/private/count-assembly-results.js
    at countAssemblyResults (.../count-assembly-results.js)
    at synthAndMeasure (.../toolkit.js)
    at async Toolkit.synth (.../toolkit.js)
```

`countAssemblyResults` calls `Object.values(stack.metadata)` with no null guard, and at least one nested stack that a Gen 2 backend (auth + multi-model data + storage) generates has `metadata` undefined — so `Object.values(undefined)` throws. This is a library bug, not a customer error, but `backend-deployer` re-wraps the `TypeError` via the generic `TypeError → SyntaxError` mapping and tells the customer to fix their own backend.

**Issue number, if available:** #3316

## Changes

- `cdk_error_mapper.ts`: add a specific branch — placed **before** the generic `TypeError` branch — that detects this crash signature (`TypeError` + `Cannot convert undefined or null to object` + `countAssemblyResults` in the stack) and maps it to a **fault** (`CDKSynthAssemblyMetadataFault`) with an accurate, non-blaming message and a `resolution` that points at the upstream bug, instead of a customer-facing `SyntaxError`. The original error is preserved as the cause.

**Note on scope:** this PR only reclassifies the error so customers are no longer told to debug backend code that is not at fault. It does **not** fix the crash itself — the missing null guard lives in `@aws-cdk/toolkit-lib` and the actual fix belongs upstream in `aws/aws-cdk-cli` (`Object.values(s.metadata ?? {})`, tracked in aws/aws-cdk-cli#1952). A `@aws-cdk/toolkit-lib` version bump was considered but dropped: no published release contains the guard yet, so bumping would not fix the crash, and `main` deliberately pins `1.32.0`.

## Validation

- Added 3 unit tests in `cdk_error_mapper.test.ts`:
  - the crash signature maps to `CDKSynthAssemblyMetadataFault` and is classified as a `FAULT` (not a customer `ERROR`);
  - the original `TypeError` is preserved as the `cause`;
  - an unrelated `TypeError` still maps to the existing backend `SyntaxError` (no regression / over-broad match).
- `backend-deployer` error-mapper suite: **65/65 pass**. `tsc --build` on `backend-deployer` clean, prettier + eslint clean.
- Manual verification: N/A — the change is error classification only; unit coverage over `getAmplifyError` exercises the exact mapping path.

## Checklist

no linked-issue close: partially addresses #3316 (improves the error message); the underlying crash fix is upstream in aws/aws-cdk-cli, so this PR intentionally does not close the issue.


packages/backend-deployer/src/cdk_error_mapper.ts
```diff
@@ -167,6 +167,24 @@ export class CdkErrorMapper {
         },
         error,
       );
+    } else if (
+      error.name === 'TypeError' &&
+      /Cannot convert undefined or null to object/.test(error.message) &&
+      /countAssemblyResults/.test(error.stack ?? '')
```
soberm, line 173: [minor] Name-based matching is the brittle part of this guard. `error.name === 'TypeError'` holds for the raw throw, but if any intermediate layer re-wraps the error before it reaches this mapper, the `name` check silently fails and the customer sees the old misleading `SyntaxError` again. The `countAssemblyResults` stack regex is the real, stable anchor here. Not fixable without an upstream change, but a one-line note that this match depends on the error arriving un-rewrapped would help future maintainers reason about why it might stop firing.

packages/backend-deployer/src/cdk_error_mapper.ts
```diff
@@ -167,6 +167,24 @@ export class CdkErrorMapper {
...
+      // @aws-cdk/toolkit-lib bug: countAssemblyResults calls
+      // Object.values(stack.metadata) with no null guard and throws when a
+      // synthesized stack has no metadata. This is a library fault, not a
+      // customer backend error, so it must not fall through to the generic
+      // TypeError branch below. Refs aws-amplify/amplify-backend#3316.
+      return new AmplifyFault(
+        'CDKSynthAssemblyMetadataFault',
+        {
+          message:
+            'Unable to synthesize the Amplify backend due to a bug in the CDK toolkit library (@aws-cdk/toolkit-lib) that fails when a synthesized stack has no metadata.',
```
soberm, line 192: [nit] Every other user-facing branch supplies a `resolution`; this fault omits one. Faults are internal-signal by nature, but a resolution would improve the CLI output and give the customer a next step, e.g. "This is a known upstream bug in aws/aws-cdk-cli; a retry is unlikely to help — track the linked issue for a fix." Optional.

packages/backend-deployer/src/cdk_error_mapper.test.ts
```diff
@@ -545,3 +545,51 @@ void describe('Node version error per deployment type', () => {
...
+  void it('classifies the crash as a fault, not a user error', () => {
+    const error = cdkErrorMapper.getAmplifyError(
+      buildCountAssemblyResultsError(),
+    );
+    assert.equal(error.name, 'CDKSynthAssemblyMetadataFault');
+    // Must NOT tell the customer to fix their backend code.
+    assert.ok(
+      !error.resolution?.includes('fix any issues in your backend code'),
+      'the fault must not blame the customer backend code',
+    );
```
soberm: [minor] This assertion is effectively a no-op. The new `AmplifyFault` sets no `resolution`, so `error.resolution` is `undefined`, and `!undefined?.includes(...)` -> `!undefined` -> `true` — it passes regardless of behavior (it would even pass if the resolution *did* contain that string, as long as it were undefined). The meaningful checks in this test are already the `error.name` equality (line 571) and the `assert.match(error.message, /@aws-cdk\/toolkit-lib/)` on line 577. Suggest either dropping this assertion or replacing it with a positive discriminator, e.g. asserting the classification is a FAULT rather than a USER error.

packages/backend-deployer/src/cdk_error_mapper.ts
sharonyajain: Good call. Fixed in `1a133ce0`. Added a comment above the branch noting the match depends on the `TypeError` reaching the mapper without being wrapped by an intermediate layer, and that the `countAssemblyResults` stack frame is the stable anchor while `error.name` is the fragile part.

packages/backend-deployer/src/cdk_error_mapper.ts
sharonyajain: Fixed in `1a133ce0`. Added a `resolution`: "This is a known upstream bug in @aws-cdk/toolkit-lib, not an issue in your backend code; retrying is unlikely to help. Track aws-amplify/amplify-backend#3316 for a fix." — so the CLI gives the customer a real next step and steers them away from a pointless retry.

packages/backend-deployer/src/cdk_error_mapper.test.ts
sharonyajain: Fixed in `1a133ce0` by replacing it with the positive discriminator you suggested: `assert.equal(error.classification, 'FAULT')`, which actually asserts the crash is classified as a FAULT rather than a user `ERROR`

sharonyajain:
This PR is merged: https://github.com/aws/aws-cdk-cli/pull/1952

packages/backend-deployer/src/cdk_error_mapper.ts
```diff
@@ -167,6 +167,30 @@ export class CdkErrorMapper {
...
+      // Object.values(stack.metadata) with no null guard and throws when a
+      // synthesized stack has no metadata. This is a library fault, not a
+      // customer backend error, so it must not fall through to the generic
+      // TypeError branch below. Refs aws-amplify/amplify-backend#3316.
+      // This match relies on the TypeError reaching the mapper without being
+      // wrapped by an intermediate layer: the countAssemblyResults stack frame
+      // is the stable anchor, while the error.name check would silently miss
+      // if a layer wraps the error into a new type before it gets here.
+      return new AmplifyFault(
+        'CDKSynthAssemblyMetadataFault',
```
Simone319, line 189: Since the upstream guard (aws/aws-cdk-cli#1952) is already merged, once the next @aws-cdk/toolkit-lib release ships this branch becomes dead code — it will keep matching a countAssemblyResults frame that no longer throws. Consider adding a TODO linking #1952 and a follow-up to bump @aws-cdk/toolkit-lib to the fixed release and remove this branch (and its tests) then, so the stopgap doesn't linger indefinitely once the real fix is available.

Simone319 (commented):
One non-blocking note that can't be line-anchored here because the CDKDeploymentError union is unchanged by this PR: CDKSynthAssemblyMetadataFault is passed as a bare string to AmplifyFault but isn't a member of the CDKDeploymentError union (~line 643), unlike every other named error. It compiles (the constructor accepts string), but callers that exhaustively switch on CDKDeploymentError won't see this fault and nothing keeps the literal and the union in sync — worth adding it to the union.

sharonyajain:
@Simone319 on the CDKDeploymentError union note: fixed in `3bc92bfc` — added `CDKSynthAssemblyMetadataFault` to the `CDKDeploymentError` union so the named fault is consistent with every other error there (previously it compiled only because `AmplifyFault`'s first param accepts a bare `string`). Both of your notes are addressed in that commit.

sharonyajain:
@Simone319 on the dead-code note (`cdk_error_mapper.ts` `CDKSynthAssemblyMetadataFault` branch): fixed in `3bc92bfc` — added a TODO above the branch linking aws/aws-cdk-cli#1952 and noting that once a @aws-cdk/toolkit-lib release ships the guard and this package bumps its pin, `countAssemblyResults` stops throwing and this branch becomes dead code to remove as part of that bump. Left the branch in place for now since no fixed toolkit-lib release exists yet (main still pins 1.32.0). (Replying here rather than in-thread since the rebase orphaned the original inline anchor.)

PR #3144, opened 2026-03-22
fix: resolve high and critical Dependabot security vulnerabilities

Simone319:
## Description

Fix all high and critical Dependabot vulnerabilities by upgrading direct dependencies. No npm overrides are used — all fixes propagate to downstream consumers.

### Vulnerability Fixes

| Vulnerability | Package | Fix |
|--------------|---------|-----|
| fast-xml-parser CRITICAL (entity expansion) | `@aws-sdk/client-bedrock-runtime` in `packages/ai-constructs` | Upgraded to `^3.936.0` (SDK 3.936+ uses `@aws-sdk/xml-builder` instead of fast-xml-parser) |
| minimatch HIGH (ReDoS) | `@microsoft/api-extractor` | Upgraded `7.43.8` → `7.57.7` (uses minimatch 10.x natively) |
| minimatch HIGH (ReDoS) | `minimatch` in `packages/backend-deployer` | Already at `10.2.3` (direct dep) |

### Remaining Vulnerabilities (unfixable from this repo)

- **fast-xml-parser 4.x** in `@aws-amplify/data-construct` and `@aws-amplify/graphql-api-construct` — these are bundled packages from `amplify-category-api` repo, already at latest version. No patched 4.x line exists (fix requires 5.x). Allowlisted in dependency review config.
- **fast-xml-parser 4.x** transitively via `aws-amplify@6.14.4` → `@aws-amplify/storage` → older AWS SDK. Cannot upgrade aws-amplify beyond 6.14.4 due to Node 18 Lambda incompatibility (`File is not defined`).

### Other Changes

- **`@aws-cdk/toolkit-lib`**: `1.16.0` → `1.19.0` in `packages/plugin-types` and `packages/backend-deployer`
- **`aws-amplify`**: Pinned to exact `6.14.4` in `packages/integration-tests` and `packages/seed` (prevents `@aws-amplify/storage@6.13.x+` which has `data instanceof File` crash on Node 18 Lambda)
- **`@zip.js/zip.js`**: Pinned to exact `2.7.52` in `packages/integration-tests` (2.8.x has WASM malloc crash)
- **Conversation handler test**: Made streaming padding assertion conditional (aws-amplify 6.14.4 changed response format)

### CI Fixes

- **ESLint OOM**: Narrowed `parserOptions.project` glob from `**/tsconfig.json` to explicit paths; increased `NODE_OPTIONS` heap size
[...]

PR #3337, opened 2026-09-09
fix(seed): restore caret range for aws-amplify to allow dedup (#3323)

### LINKED ISSUE #3323: @aws-amplify/seed 1.1.3 pins aws-amplify to an exact version, creating a duplicate unconfigured Amplify singleton
### Environment information

```
System:
  OS: Windows 11 (10.0.26200)
  Node: v24.15.0
  npm: 11.12.1

Packages:
  @aws-amplify/seed: 1.1.3        <-- the problem
  aws-amplify: 6.20.0             (declared as ^6.20.0 across the monorepo)
  @aws-amplify/backend: 1.24.0
  @aws-amplify/backend-cli: 1.9.0
```

npm workspaces monorepo, `aws-amplify` declared as `^6.20.0` in every workspace.

### Describe the bug

`@aws-amplify/seed@1.1.3` declares its `aws-amplify` dependency as an **exact version**:

```json
"dependencies": { "aws-amplify": "6.14.4" }
```

Every prior release used a range:

| version | `dependencies.aws-amplify` |
| --- | --- |
| 1.1.0 | `^6.0.16` |
| 1.1.1 | `^6.0.16` |
| 1.1.2 | `^6.0.16` |
| **1.1.3** | **`6.14.4`** |

`aws-amplify` exports `Amplify` as a **module-level singleton**. When an app depends on any `aws-amplify` other than exactly `6.14.4`, npm can no longer dedupe, so it installs a second copy at `node_modules/@aws-amplify/seed/node_modules/aws-amplify`. The seed script then configures one instance while `@aws-amplify/seed` reads the other:

```
from @aws-amplify/seed : node_modules\@aws-amplify\seed\node_modules\aws-amplify\dist\cjs\index.js   (6.14.4)
from repo root         : node_modules\aws-amplify\dist\cjs\index.js                                  (6.20.0)
SAME MODULE FILE?      : false
```

So a completely ordinary seed script:

```ts
const outputs = JSON.parse(await readFile("./amplify_outputs.json", "utf8"));
Amplify.configure(outputs);              // configures the hoisted 6.20.0
await createAndSignUpUser({ ... });      // reads the nested, unconfigured 6.14.4
```

fails immediately with:

```
[SeedingFailedError] Seed failed to complete
  ∟ Caused by: [AuthUserPoolException] Auth UserPool not configured.
```

Two things make this expensive to diagnose:

1. **The error points at the wrong thing.** It reads as a malformed or missing `amplify_outputs.json`. The outputs are fine — configuring the hoisted copy by hand yields a fully populated `Auth.Cognito` with the correct `userPoolId` and `userPoolClientId`. Nothing in the message hints at duplicate module instances.
2. **`ampx sandbox seed` still exits 0** after printing `[SeedingFailedError]`, so the failure is invisible to CI or any caller that checks the exit code.

Since 1.1.3 is currently the latest release, there is no forward fix — the only options are downgrading to 1.1.2 or forcing the transitive dependency with an npm `overrides` entry.

### Reproduction steps

1. In a project that declares `aws-amplify` at any version other than exactly `6.14.4` (e.g. `^6.20.0`), install `@aws-amplify/seed@1.1.3`.
2. Confirm the duplicate: `node_modules/@aws-amplify/seed/node_modules/aws-amplify` exists alongside the hoisted `node_modules/aws-amplify`.
3. Write a seed script that calls `Amplify.configure(outputs)` and then any `@aws-amplify/seed` helper, e.g. `createAndSignUpUser`.
4. Run `ampx sandbox seed`.
5. It fails with `[AuthUserPoolException] Auth UserPool not configured` — and exits 0.

Downgrading to `@aws-amplify/seed@1.1.2` restores `^6.0.16`, npm dedupes to a single `aws-amplify`, and the identical seed script succeeds.

### Suggested fix

Restore a range (`^6.0.16`, or `^6.14.4` if newer APIs are needed) so npm can dedupe. Given that `aws-amplify` exports a singleton that the consuming app is expected to configure, it arguably belongs in `peerDependencies` rather than `dependencies` — an exact `dependencies` pin makes a second, unconfigurable instance the default outcome for anyone not on that exact version.

Separately, `ampx sandbox seed` exiting 0 on `SeedingFailedError` looks like a bug worth fixing on its own.

Simone319:
Fixes #3323

`@aws-amplify/seed@1.1.3` pinned `aws-amplify` to an exact `6.14.4` in `dependencies` (releases 1.1.0-1.1.2 used `^6.0.16`). Because `aws-amplify` is a module-level singleton (`Amplify.configure()` mutates shared state), the exact pin defeats npm deduplication whenever the consumer uses any other 6.x version. npm then installs a nested copy under `node_modules/@aws-amplify/seed/node_modules/aws-amplify`; the app configures the hoisted copy while seed's helpers read the nested, unconfigured one, producing `AuthUserPoolException` wrapped as `SeedingFailedError`.

This restores the caret range so npm can dedup against any `^6.x` app.

Regression introduced in 88c4759 ("resolve Dependabot vulnerabilities #3144").

## Testing
- Install `@aws-amplify/seed` with the change in a project pinning `aws-amplify@^6.20.0`
- Confirm no `@aws-amplify/seed/node_modules/aws-amplify` duplicate is created
- `ampx sandbox seed` completes without `Auth UserPool not configured`

### DIFF
--- a/.changeset/fix-seed-aws-amplify-pin.md
+++ b/.changeset/fix-seed-aws-amplify-pin.md
@@ -1,0 +1,5 @@
+---
+'@aws-amplify/seed': patch
+---
+
+Loosen `aws-amplify` from an exact version pin (`6.14.4`) back to a caret range (`^6.14.4`) so npm can deduplicate it against the consuming app's `aws-amplify` version. The exact pin (introduced in #3144, shipped in 1.1.3) forced a second, nested `aws-amplify` copy whenever the app used any other 6.x version, producing an unconfigured Amplify singleton and `Auth UserPool not configured` / `SeedingFailedError` during `ampx sandbox seed`.
--- a/packages/seed/package.json
+++ b/packages/seed/package.json
@@ -29,6 +29,6 @@
     "@aws-amplify/platform-core": "^1.11.1",
     "@aws-amplify/plugin-types": "^1.12.1",
     "@aws-sdk/client-cognito-identity-provider": "^3.936.0",
-    "aws-amplify": "6.14.4"
+    "aws-amplify": "^6.14.4"
   }
 }


Simone319:
Superseded by https://github.com/aws-amplify/amplify-backend/pull/3339 — recreated from an upstream branch so e2e checks can run (fork PRs skip e2e).

PR #3338, opened 2026-09-09
fix(scripts): handle symbol-brand keys in check_api_changes

osama-rizk:
## Problem

The release PR ([#3330](https://github.com/aws-amplify/amplify-backend/pull/3330)) and every PR to `main` fail the `check_api_changes` job:

```
Validation of @aws-amplify/hosting failed:
index.ts: error TS2304: Cannot find name 'BYO_BRAND'.
```

`@aws-amplify/hosting`'s `ByoValue` is branded with a module-local `unique symbol` (`BYO_BRAND`) used as a computed property key. API Extractor's report references the key (`[BYO_BRAND]`) but never declares the private symbol, so the usage the validator generates from the report cannot compile. It was skipped on the feature PR (#3174) — a new package has no baseline `API.md` — and began failing once `main` carried the baseline.

## Why not just exclude the type

Excluding `ByoValue` (via `excludedTypesByPackageName`) cascades: dropping it removes its import, so every member that references it (`byoSecret`, `byoConfig`, `isByoValue`, `HostingProps.environment`) then fails with `TS2304: Cannot find name 'ByoValue'`. Excluding the whole BYO surface (and `HostingProps`) is unacceptable.

A source-only fix is also impossible in a single PR: the validator reconstructs the **baseline** type from `main`'s committed `API.md`, so no change on this branch alters the generated code (circular — fixing the report requires merging a PR that must first pass this check).

## Fix

Make the usage generator tolerant of symbol brands generally: when reconstructing a type, drop members keyed by a bare-identifier computed property (a `unique symbol` brand like `[BYO_BRAND]`). Such a brand is a nominal marker that can't be exercised in generated usage anyway. `ByoValue`'s brand is made **optional** so the reconstruction (brand dropped) stays assignable to the published type.

- `scripts/.../api_usage_statements_generators.ts` — strip symbol-brand members (recurses through intersections/unions/parentheses); verbatim output otherwise, so all other types are unaffected.
- `packages/hosting/src/byo.ts` — make the `BYO_BRAND` key optional (set at runtime by `byoSecret`/`byoConfig`; `API.md` is unchanged — API Extractor renders symbol keys the same).
- Added a generator unit test for symbol-brand stripping (10/10 pass).

Published types (tsc-emitted `lib/*.d.ts`) are unaffected. Fixes all current and future branded types, no per-type exclusion list.

## Notes
- Tooling + internal-type change; empty-frontmatter changeset (no published-package bump; the hosting type change rides on the existing first-release changeset).
- After merge, the release PR ([#3330](https://github.com/aws-amplify/amplify-backend/pull/3330)) regenerates and `check_api_changes` passes.


scripts/components/api-changes-validator/api_usage_statements_generators.ts
```diff
@@ -5,6 +5,75 @@ import {
...
+ * i.e. a `unique symbol` brand such as `readonly [BYO_BRAND]: true`.
+ *
+ * (`[Symbol.iterator]` is a PropertyAccessExpression, not an Identifier, so
+ * well-known symbols are intentionally NOT matched; nor are string/number
+ * computed keys like `['foo']`.)
+ */
+const isSymbolBrandMember = (member: ts.TypeElement): boolean =>
+  !!member.name &&
+  ts.isComputedPropertyName(member.name) &&
+  ts.isIdentifier(member.name.expression);
```
Simone319: This matches on key shape only — any `[Identifier]`-keyed member is stripped regardless of its value type — so an exported unique-symbol-keyed member that carries real API would be silently dropped from the baseline, and a breaking change to it would pass this check undetected. Since this generator runs on every PR, a silent false-negative here is the worst failure mode. Consider narrowing the match to the actual brand pattern (only strip when the member's value type is the literal `true`); a symbol-keyed member with any other value then falls through to verbatim and fails loud (TS2304) instead of vanishing.

scripts/components/api-changes-validator/api_usage_statements_generators.ts
```diff
@@ -136,7 +205,9 @@ export class TypeUsageStatementsGenerator implements UsageStatementsGenerator {
     const baselineTypeName = `${typeName}Baseline`;
     const functionParameterName = `${constName}FunctionParameter`;
     // declare type with same content under different name.
-    let usageStatement = `type ${baselineTypeName}${genericTypeParametersDeclaration} = ${this.typeAliasDeclaration.type.getText()}${EOL}`;
+    let usageStatement = `type ${baselineTypeName}${genericTypeParametersDeclaration} = ${renderTypeWithoutSymbolBrands(
```
Simone319, line 250: Because the baseline drops the brand, this generated assignment only compiles when the brand is optional in the current type — so a future type with a required symbol brand hard-fails with a confusing 'property is missing', and 'fixes all current and future branded types' really only covers optional brands. A non-exported unique symbol can't be reconstructed or imported, so required brands genuinely can't pass while keeping a real assignability check; better to make that explicit than implicit. Proposal: (1) when a matched brand member is NOT optional, fail with an actionable message (e.g. '<Type> has a required unique-symbol brand <name>; declare it optional [BRAND]?: true or exclude the type') instead of the cryptic missing-property error; (2) reword the claim/changeset to 'optional symbol brands' and note the convention in a comment by the helper; (3) add a required-brand regression test. (If required brands must pass transparently, the only sound route is to export the brand symbol and import it in generated usage so both sides reference the same nominal symbol — at the cost of exposing the brand in the public API.)

scripts/components/api-changes-validator/api_usage_statements_generators.ts
```diff
@@ -5,6 +5,75 @@ import {
...
+const isSymbolBrandMember = (member: ts.TypeElement): boolean =>
+  !!member.name &&
+  ts.isComputedPropertyName(member.name) &&
+  ts.isIdentifier(member.name.expression);
+
+/**
+ * Whether a type node (descending through intersections, unions and
+ * parentheses) declares a symbol-brand member.
+ */
+const containsSymbolBrandMember = (typeNode: ts.TypeNode): boolean => {
```
Simone319, line 44: Stripping only descends through intersections, unions, parentheses and top-level type-literal members, so a brand nested inside a member's value type (foo: { [SYM]: true }) or wrapped in an array/generic/mapped type is still emitted verbatim and fails with TS2304. That's a fail-loud gap rather than a silent one, so it's a completeness limit rather than a safety hole — worth a comment noting the boundary and a test documenting it.

scripts/components/api-changes-validator/api_usage_statements_generators.ts
```diff
@@ -5,6 +5,75 @@ import {
...
+const renderTypeWithoutSymbolBrands = (typeNode: ts.TypeNode): string => {
+  if (!containsSymbolBrandMember(typeNode)) {
+    return typeNode.getText();
+  }
+  if (ts.isTypeLiteralNode(typeNode)) {
+    const members = typeNode.members
+      .filter((member) => !isSymbolBrandMember(member))
+      // `getText()` includes each member's terminating `;`; strip it so the
+      // rejoined members are not doubly-separated.
+      .map((member) => member.getText().replace(/;\s*$/, ''));
```
Simone319: Rejoining members after stripping the trailing ';' with /;\s*$/ then '; ' assumes API Extractor's ';'-separated output; it would be brittle if a member's getText() ever ended in ',' or a trailing inline comment. Low risk given the fixed input format — a one-line comment noting the assumption would be enough.

packages/hosting/src/byo.ts
```diff
@@ -14,7 +14,11 @@ const BYO_BRAND = Symbol.for('@aws-amplify/hosting.byo');
 
 /** Inert marker describing an existing store entry to wire into `environment`. */
 export type ByoValue = {
-  readonly [BYO_BRAND]: true;
+  // Optional so the API-change validator (which reconstructs this type from the
+  // API report, where the private BYO_BRAND symbol is not declared) can drop the
+  // brand key and still have the reconstruction assignable to this type. The
+  // brand is always set at runtime by byoSecret()/byoConfig().
+  readonly [BYO_BRAND]?: true;
```
Simone319, line 28: Making the brand optional weakens the nominal guarantee: a plain { kind, ref } object now structurally satisfies ByoValue at compile time, so callers are no longer forced through byoSecret()/byoConfig() (only the runtime isByoValue still enforces it). Also, [BYO_BRAND]: true becoming ?: true does change the emitted .d.ts, so 'published types unaffected' isn't quite accurate. If the nominal guarantee matters, consider fixing this generator-side (e.g. a cast or a synthesized local symbol in generated usage) rather than loosening the published type.

scripts/components/api-changes-validator/api_usage_statements_generators.ts
osama-rizk: Agreed — fixed in d19dbf045a. `isSymbolBrandMember` now also requires the value type to be the literal `true`, so only the inert brand shape is stripped. A symbol-keyed member with any other value type falls through to verbatim and fails loud (TS2304) instead of vanishing. Added a regression test: `keeps a symbol-keyed member whose value type is not `true``.

scripts/components/api-changes-validator/api_usage_statements_generators.ts
osama-rizk: Agreed on all three, done in d19dbf045a:
1. A matched brand that is **required** now throws an actionable error (`declare it optional ([BRAND]?: true) or add the type to excludedTypesByPackageName`) instead of the cryptic TS2741.
2. Reworded to 'optional symbol brands' in the changeset and the helper comment.
3. Added a required-brand regression test (`throws an actionable error for a required symbol brand`).

On the exported-symbol alternative: noted it as the future route to a required brand, but it can't unblock this PR — the validator reconstructs the baseline from main's committed API.md, which references the un-exported symbol, so nothing on this branch changes the generated code.

scripts/components/api-changes-validator/api_usage_statements_generators.ts
osama-rizk: Agreed — it's a fail-loud completeness limit, not a safety hole. Documented the boundary in the `containsSymbolBrandMember` doc comment and added a test (`does not strip a brand nested in a member value type`) showing a nested brand is emitted verbatim. (d19dbf045a)

scripts/components/api-changes-validator/api_usage_statements_generators.ts
osama-rizk: Agreed — added a comment noting the assumption (API Extractor emits `;`-separated members) next to the strip/rejoin. (d19dbf045a)

packages/hosting/src/byo.ts
osama-rizk: You're right on the accuracy point — corrected: the emitted `.d.ts` **does** change (`API.md` is what's unchanged, because API Extractor renders the symbol key identically). Fixed in the changeset and the byo.ts comment now spells out the nominal trade-off.

Pushing back on the generator-side fix though — neither alternative holds for this constraint:
- A **synthesized local symbol** in generated usage is a *different* unique symbol than the package's `BYO_BRAND`; `[local]: true` is not assignable to `[BYO_BRAND]: true`, so it still fails TS2741 (unique symbols are nominal by identity).
- A **cast** compiles but removes the `kind`/`ref` assignability check — the exact silent false-negative flagged in the other thread.

The deciding constraint is that the baseline is reconstructed from main's committed report (symbol-branded) and gets stripped to `{ kind, ref }`, so the current type must not *require* any brand the stripped baseline lacks → optional is forced while the baseline stays symbol-branded (which it does — API Extractor renders it the same post-merge). I've documented the trade-off and a follow-up: migrate `ByoValue` to a declarable brand (exported symbol or reserved string-literal prop) and update the baseline in the same PR, which would let the brand be required again. (d19dbf045a)

Simone319 (approved):

PR #3341, opened 2026-09-10
fix(scripts): remove infeasible required-brand throw in check_api_changes + refresh hosting README

osama-rizk:
## Problem

#3338 merged the symbol-brand handling for `check_api_changes`, but it included a throw that regressed the check: `check_api_changes` is now **failing on `main` and on the release PR ([#3330](https://github.com/aws-amplify/amplify-backend/pull/3330))** for `@aws-amplify/hosting`:

```
Error: A required unique-symbol brand '[BYO_BRAND]' cannot be validated. ...
```

## Root cause

The throw fires when a matched brand member is not marked optional. But `check_api_changes` reconstructs types from the **baseline API report**, and **API Extractor renders symbol-keyed properties without the `?`** — a source `readonly [BRAND]?: true` appears in the report as `readonly [BRAND]: true`. So the `questionToken` check always sees the brand as *required* and throws on the current baseline. Optional-vs-required simply isn't recoverable from the report.

## Fix

Remove the required-brand throw and strip the inert (`: true`) symbol-brand shape **unconditionally** — the behavior that originally passed `check_api_changes` before the throw was added. The narrow value-type match (only `: true`) is kept, so a symbol-keyed member with real API still fails loud (`TS2304`) rather than being silently dropped.

The convention (brands must be declared optional in source) is documented in the helper and changeset; a genuinely required brand would surface downstream as `TS2741`, which can't be pre-empted from the report.

- Remove the throw + the required-brand throw test.
- Keep narrow match, nested-brand boundary doc/test, and `;`-join handling.
- Generator tests: 12/12 pass.

Unblocks the release PR #3330.

### DIFF
--- a/.changeset/api-check-remove-required-brand-throw.md
+++ b/.changeset/api-check-remove-required-brand-throw.md
@@ -1,0 +1,18 @@
+---
+---
+
+fix(scripts): strip symbol brands unconditionally in check_api_changes (build script only, no published package change)
+
+The symbol-brand handling merged in #3338 threw an error when a matched brand
+member was not marked optional. But `check_api_changes` reconstructs types from
+the baseline API report, and API Extractor renders symbol-keyed properties
+WITHOUT the `?` (a source `[BRAND]?: true` shows as `[BRAND]: true`) — so the
+check always saw the brand as "required" and threw on the current baseline,
+failing `check_api_changes` for `@aws-amplify/hosting` on every PR and on the
+release PR.
+
+Remove the infeasible required-brand throw and strip the inert (`: true`) brand
+shape unconditionally (the behavior that originally passed). Optional-vs-required
+is not recoverable from the report, so the convention is documented instead: such
+brands must be declared optional in the source type (a genuinely required brand
+would surface downstream as `TS2741`).
--- a/.changeset/hosting-readme-refresh.md
+++ b/.changeset/hosting-readme-refresh.md
@@ -1,0 +1,11 @@
+---
+'@aws-amplify/hosting': patch
+---
+
+docs: refresh README for the 1.0 API
+
+- Add SvelteKit to the built-in adapters table and exports, and stop listing it as "not built in".
+- Mention Nuxt/Astro/SvelteKit SSR in the intro and complete the `framework` union.
+- Document secrets & environment variables (`secret()`/`config()`/`byoSecret()`/`byoConfig()`, `ampx secret|config set`, `getSecret`/`getConfig` from `@aws-amplify/hosting/runtime`, store namespaces).
+- Document self-managed CI/CD via `definePipeline()` (source, branches/stages, `bakeTime`, `synth`, `crossAccountKeys`, `selfMutation`, `getStageConfig()`), and reframe the `ampx pipeline-deploy` note accordingly.
+- Add the default-on `monitoring` and `skewProtection` options and `buildOutputDir` to the configuration table.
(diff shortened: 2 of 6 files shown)


osama-rizk:
Folded the hosting README refresh into this PR (commit 92555ad5b8) so the docs publish with the next version.

**README changes (`@aws-amplify/hosting`: patch → 1.0.1):**
- SvelteKit added to the built-in adapters table + exports; removed from the "not built in" lists.
- Intro and the `framework` union now cover Nuxt/Nitro/Astro/SvelteKit SSR.
- New **Secrets & Environment Variables** section (`secret()`/`config()`/`byoSecret()`/`byoConfig()`, `ampx secret|config set`, `getSecret`/`getConfig` from `@aws-amplify/hosting/runtime`, store namespaces).
- New **CI/CD with definePipeline()** section; reframed the `ampx pipeline-deploy` note to point at it.
- Config table: default-on `monitoring` & `skewProtection`, plus `buildOutputDir`.

This bumps hosting to 1.0.1, which is what makes the refreshed README appear on the npm package page (npm only re-renders on a new publish).

Simone319 (approved):

Simone319:
**Follow-up: the optional `[BYO_BRAND]` is a tooling workaround — let's not let it become the permanent design**

**Concern with making the brand optional.** To get past `check_api_changes`, this PR keeps `[BYO_BRAND]?: true` **optional** — and that's a real weakening of the type's nominal guarantee. A plain `{ kind, ref }` object now type-checks as a `ByoValue`, so callers are no longer *forced* through `byoSecret()`/`byoConfig()` at compile time; only the runtime `isByoValue` still enforces the brand. The optionality isn't something `ByoValue` wants — it's a concession to the validator, which drops the symbol brand from the reconstructed baseline and can't even distinguish optional from required (API Extractor renders symbol keys without the `?`).

**The real decision — don't let "optional by accident" become the permanent design.** Optional is a fine short-term unblock, but it shouldn't silently harden into the contract. The thing worth deciding *explicitly* is how much a compile-time forgery guarantee matters for `ByoValue`, and then choosing a brand strategy on purpose — rather than inheriting "symbol brands must be optional" as an implicit rule for every future branded type.

**Preferred option: a *required* string-literal brand** (drop the `unique symbol`):
```ts
export type ByoValue = {
  readonly __brand: 'amplify.hosting.byo'; // required
  readonly kind: 'secret' | 'config';
  readonly ref: string;
};
```
- API Extractor renders a string-literal member normally — no undeclared symbol (`TS2304`), no stripped `?` — so the brand can stay **required** and `check_api_changes` passes with **zero special-casing**.
- Restores the compile-time guarantee the optional symbol brand gives up.
- No public symbol export, no generator complexity; if `ByoValue` is the only symbol-branded type, the `isSymbolBrandMember`/`renderTypeWithoutSymbolBrands` machinery from #3338/#3341 could eventually be retired.
- Tradeoff: the key is nameable, so *deliberate* forgery is possible (same as the export-symbol route) — but accidental structural forgery (`{ kind, ref }`) is prevented, which is the case that matters.

For the record: **keep optional + runtime `isByoValue`** (acceptable if `ByoValue` is only produced by the factories and always consumed through runtime-validated APIs — then document it as the deliberate choice); or **export the symbol, restore required** (strongest symbol-based typing, but widens the public API with a brand symbol and needs generator work, for the same forgeability as the string-literal option).

**Recommendation:** merge this PR to unblock #3330, then adopt the required string-literal brand as a follow-up and record the chosen convention in a short decision note — so the optional brand doesn't become permanent by default.

PR #3347, opened 2026-09-14
fix(backend-deployer): decouple internal type-check from project TypeScript version

### LINKED ISSUE #3274: backend-deployer: incompatible with TypeScript 7.0 (Go rewrite) — ts.sys is undefined
### Environment information

```plain text
System:
OS: Linux 4.14 Amazon Linux 2023
CPU: (8) x64 Intel(R) Xeon(R) Platinum 8275CL CPU @ 3.00GHz
Memory: 13.56 GB / 15.19 GB
Shell: /bin/bash
Binaries:
Node: 24.18.0 - /root/.nvm/versions/node/v24.18.0/bin/node
Yarn: undefined - undefined
npm: 11.16.0 - /root/.nvm/versions/node/v24.18.0/bin/npm
pnpm: 11.10.0 - /root/.nvm/versions/node/v24.18.0/bin/pnpm
NPM Packages:
@aws-amplify/ai-constructs: Not Found
@aws-amplify/auth-construct: Not Found
@aws-amplify/backend: Not Found
@aws-amplify/backend-ai: Not Found
@aws-amplify/backend-auth: Not Found
@aws-amplify/backend-cli: Not Found
@aws-amplify/backend-data: Not Found
@aws-amplify/backend-deployer: Not Found
@aws-amplify/backend-function: Not Found
@aws-amplify/backend-output-schemas: Not Found
@aws-amplify/backend-output-storage: Not Found
@aws-amplify/backend-secret: Not Found
@aws-amplify/backend-storage: Not Found
@aws-amplify/cli-core: Not Found
@aws-amplify/client-config: Not Found
@aws-amplify/data-construct: Not Found
@aws-amplify/data-schema: 1.26.0
@aws-amplify/deployed-backend-client: Not Found
@aws-amplify/form-generator: Not Found
@aws-amplify/model-generator: Not Found
@aws-amplify/platform-core: Not Found
@aws-amplify/plugin-types: Not Found
@aws-amplify/sandbox: Not Found
@aws-amplify/schema-generator: Not Found
@aws-cdk/toolkit-lib: Not Found
aws-amplify: 6.18.0
aws-cdk-lib: Not Found
typescript: 6.0.3
npm warn exec The following package was not found and will be installed: cdk@2.1129.0
CDK CLI Version: 2.1129.0 (build 629ca49)
normal verbosity, no debugging
AWS environment variables:
AWS_AMPLIFY_WEB_DYNAMIC_RUNTIME = nodejs16.x
AWS_EXECUTION_ENV = AWS_ECS_EC2
AWS_CONTAINER_CREDENTIALS_RELATIVE_URI = /v2/credentials/REDACTED
AWS_AMPLIFY_AL2023_DEFAULT_BUILD_IMAGE = true
AWS_STEP_NAME = BUILD
AWS_BRANCH = test
AWS_JOB_ID = REDACTED
AWS_DEFAULT_REGION = us-east-1
AWS_GITHUB_SSH_PUBLIC_KEYS = REDACTED
AWS_REGION = us-east-1
AWS_AMPLIFY_SKEW_PROTECTION_ENABLED = false
AWS_APP_ID = REDACTED
AWS_PREVIOUS_COMMIT_ID = REDACTED
AWS_PLATFORM = WEB
AWS_AMPLIFY_BUILD_IMAGE_VERSION = 2025-09
AWS_BRANCH_ARN = arn:aws:amplify:us-east-1:REDACTED:apps/REDACTED/branches/test
AWS_AMPLIFY_INTERNAL_IMAGE_OPTIMIZATION = true
AWS_BUILDSPEC_CONTENT = REDACTED
AWS_COMMIT_ID = REDACTED
AWS_CLONE_URL = REDACTED
AWS_AMPLIFY_DEPLOYMENT_ID = REDACTED
AWS_REQUEST_ID = REDACTED
AWS_AMPLIFY_BUILD_COMPUTE_TYPE = LARGE_16GB
No CDK environment variables
```

### Describe the bug

I bumped my Typescript version to 7.0.3 (released yesterday) and got an error during the amplify deploy stage.

TypeScript 7.0 (Project Corsa) is a complete rewrite of the TypeScript compiler in Go.
The npm package no longer ships the JavaScript-based Compiler API (`ts.sys`, `ts.createProgram`, `ts.readConfigFile`, etc.). These are all `undefined`.

`@aws-amplify/backend-deployer@2.1.7` uses the TypeScript Compiler API directly in
`lib/ts_compiler.js`:

### Reproduction steps

node [repro.js](https://github.com/user-attachments/files/29843589/repro.js)

Expected output with typescript@6.0.3: "ts.sys.readFile is a function: true"
Expected output with typescript@7.0.2: "TypeError: Cannot read properties of undefined (reading 'readFile')"

sharonyajain:
## Problem

`@aws-amplify/backend-deployer` type-checks the backend definition using the TypeScript **JavaScript Compiler API** (`ts.sys`, `ts.readConfigFile`, `ts.parseJsonConfigFileContent`, `ts.createIncrementalProgram`, `ts.getPreEmitDiagnostics`) in `packages/backend-deployer/src/ts_compiler.ts`. `typescript` was declared as a **peerDependency** (`^5.0.0`), so `import ts from 'typescript'` in the type-check worker resolved to the **project's** installed `typescript`.

TypeScript 7.0 (Project Corsa, the Go rewrite) ships **without** the JavaScript Compiler API until a later release, so `ts.sys` and friends are `undefined`. The first call — `ts.readConfigFile(configPath, ts.sys.readFile)` — throws `Cannot read properties of undefined (reading 'readFile')`, the worker exits non-zero, and the deploy fails with a `TSCompilerWorkerFault`.

The type-check runs in **both** hosted/pipeline deploys **and** local `ampx sandbox`, so a project cannot move any backend to TypeScript 7 without breaking local dev as well as CI — there is no partial-adoption path. Confirmed by three reporters on the issue.

> Supersedes #3346, which GitHub auto-closed (and locked against reopen) after a force-push transiently left the branch with no common history with `main`. The branch history is now clean; this is the same commit, re-opened.

**Issue number, if available:** #3274

## Changes

Decouple the deployer's internal type-check from the project's `typescript` version:

- **`packages/backend-deployer/package.json`** — move `typescript` from `peerDependencies` (`^5.0.0`) to a **direct dependency** pinned `~5.8.2` (matching the repo's own root pin). The deployer now ships a 5.x compiler that exposes the JavaScript Compiler API, so the type-check runs against it regardless of the version the app pins.
- **`packages/backend-deployer/src/ts_compiler.ts`** — add an explicit guard before the first `ts.sys` use: if the resolved compiler lacks the JavaScript Compiler API (`!ts.sys || typeof ts.readConfigFile !== 'function'`), throw a clear `UnsupportedTypeScriptVersionError` with an actionable resolution, instead of the opaque undefined-read crash. This also covers the residual case where an app forces a single hoisted `typescript@7` via `overrides`/`resolutions`.

This unblocks TypeScript 7 adoption **in the app** while the deployer type-checks with 5.x. It does not migrate the deployer itself to the TS7 compiler — that awaits the return of the JavaScript Compiler API in a later TypeScript release.

## Validation

- Added a unit regression test in `packages/backend-deployer/src/ts_compiler.test.ts` that simulates the TypeScript 7 API gap (`ts.sys` / `ts.readConfigFile` undefined) and asserts `compileProject` throws `UnsupportedTypeScriptVersionError`. The test stores and fully restores the original property descriptors (deleting the mock when a property had no own descriptor) so it cannot leak state into later tests.
- Verified the fix in an isolated harness against a **real `typescript@5.8.3`** (happy path type-checks clean, no false positive) and a **TS7-shaped stub** (`sys`/`readConfigFile` undefined → throws `UnsupportedTypeScriptVersionError`). The repo's own `node --test` lane could not be run locally because the workspace install requires an authenticated registry (public-registry `npm install` returns E401); it should be exercised in CI.

## Checklist

Fixes #3274


sharonyajain:
**Review disposition — test cleanup (`ts_compiler.test.ts`)**

`fixed` in `1594614c80ba35c85d1f647c6341e52060452adc`.

The finding: the `finally` block in the "resolved typescript lacks the JS Compiler API" test only restored `ts.sys` / `ts.readConfigFile` when `getOwnPropertyDescriptor` returned an own descriptor; if a property had no own descriptor, the `value: undefined` mock would leak into later tests. Legitimate and in-scope (a self-contained hardening of the test added by this PR), so I applied it.

Change: added an `else` branch to each restore that deletes the mocked property when there was no original own descriptor, so the module object is fully restored either way:

```ts
} finally {
  if (sysDescriptor) {
    Object.defineProperty(ts, 'sys', sysDescriptor);
  } else {
    delete (ts as { sys?: unknown }).sys;
  }
  if (readConfigFileDescriptor) {
    Object.defineProperty(ts, 'readConfigFile', readConfigFileDescriptor);
  } else {
    delete (ts as { readConfigFile?: unknown }).readConfigFile;
  }
}
```

I used the `delete (ts as { sys?: unknown }).sys` cast form rather than a bare `delete ts.sys` because this repo runs `eslint --max-warnings 0` with strict TS, and `delete` on a non-optional namespace member trips `@typescript-eslint`. Behaviour is identical to the suggested fix.

(Note: this supersedes #3346, which GitHub auto-closed and locked against reopen after a force-push transiently left the branch without common history with `main`. Same commit, re-opened here.)

sharonyajain:
**Review dispositions** — head `3450e8574838ed69df141218ebace85235121dd4` (previous reviewed head `1594614c`).

**1. Test cleanup `else` branch (Critical) — `rebutted`.** The premise doesn't hold: `Object.getOwnPropertyDescriptor` returns a descriptor for *any* own property, `configurable` or not — it returns `undefined` **only** when there is no own property at all. So the `else` (`delete`) branch runs exactly when there was nothing to restore, which is the correct restore for that case. A non-configurable own property still yields a truthy descriptor and takes the `Object.defineProperty` branch. No inconsistent state; leaving as-is.

**2. Error message should name the missing API(s) (Minor) — `fixed` in `3450e857`.** The guard now collects the specific missing members and names them: `... (missing: ts.sys, ts.readConfigFile)` (or just the one that's absent).

**3. Missing coverage: `ts.sys` present but `ts.readConfigFile` undefined (Missing Coverage) — `fixed` in `3450e857`.** Added a second test that leaves `ts.sys` intact and nulls only `ts.readConfigFile`, asserting the guard still throws. Verified in an isolated harness that all three shapes throw and the message lists exactly the missing member(s), while a real `typescript@5.8.3` compiles clean (no false positive).

**4. Pin `~5.8.2` vs `^5.8.2` (Architecture) — `rebutted`.** This is deliberate. `typescript` here is a **direct, bundled dependency of backend-deployer's own type-checker**, not a peer the consumer project shares — the whole point of the fix is that the deployer resolves *its own* compiler regardless of the project's `typescript`. So there's no cross-version conflict with a project's newer TS 5.x; they're separate installs. The tilde matches this repo's own root pin (`~5.8.2`), keeping the deployer on the exact line it's tested against; widening to `^5` would let it silently pull an untested 5.x minor. If maintainers prefer `^5.8.2` I'm happy to change it, but the conservative pin is intentional, not an oversight.

**5. Rename `UnsupportedTypeScriptVersionError` (Minor) — `fixed` in `3450e857`.** Agreed the old name over-claimed a version judgment the check doesn't make. Renamed to `TypeScriptCompilerApiUnavailableError`, which describes what's actually detected (the JS Compiler API is absent). Updated the assertion in both tests.

sharonyajain:
**CI fix** — head `71f7ba96aff815617d78bb775963ce22a8901db3`.

The `test_with_coverage` / `test_with_baseline_dependencies` lanes were red on the previous head (`3450e857`): the two guard tests threw `TypeError: Cannot redefine property: sys`. Root cause: the real installed `typescript` defines `ts.sys` / `ts.readConfigFile` as **non-configurable** own properties, so `Object.defineProperty(ts, 'sys', …)` in the test throws. (My earlier isolated harness used a stub `typescript` with configurable members, which masked this — my mistake; the real package test lane is the correct check.)

Fix: extracted the guard into an exported, injectable helper `assertTypeScriptCompilerApi(tsModule = ts)`. `compileProject` calls it with the real `ts`; the tests call it with a plain fake object, so they never mutate the non-configurable real module. Three tests now cover: both members missing (message lists both), `ts.sys` present + `ts.readConfigFile` missing (message lists only the one), and a well-formed API passing. The production call path is unchanged.

Verified by compiling the actual edited source + test against a real `typescript@5.8.3` and running `node --test`: all three guard tests pass (the `Cannot redefine property` failure no longer reproduces).

sarayev (approved):

